PR #13, opened 2018-07-01
Update README for macOS

nicolai86:
Document necessary environment variables for macOS for `go get` to succeed.

Without these, `go get` will fail, even if `openssl` is installed:

```
$ go get -u github.com/pions/webrtc
# github.com/pions/webrtc/internal/dtls
In file included from ../../pions/webrtc/internal/dtls/dtls.go:10:
./dtls.h:4:10: fatal error: 'openssl/bio.h' file not found
#include <openssl/bio.h>
         ^~~~~~~~~~~~~~~
1 error generated.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,13 @@ your platform before using pion-WebRTC.
 `sudo apt-get install libssl-dev`
 
 #### OSX
-`brew install openssl`
+
+```
+brew install openssl
+export CPATH=`brew --prefix`/opt/openssl/include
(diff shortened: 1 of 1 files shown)


Sean-Der:
Nice! Thank you for fixing this @nicolai86

Sean-Der:
Looking forward to replace OpenSSL, that's gonna be a fun one :)

nicolai86:
it will. also super useful for cross-compilation users like me, as I develop on macOS and deploy on ARM (raspberry pi zero w)

PR #14, opened 2018-07-02
add RTCConfiguration and RTCPeerConnection constructor

nicolai86:
This PR lays the foundation to get a `RTCConfiguration` object into the library without compromising local/ remote functionality.

### DIFF
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -53,12 +53,52 @@ func (t TrackType) String() string {
 	}
 }
 
+// RTCCredentialType specifies the type of credentials provided
+type RTCCredentialType string
+
+var (
+	// RTCCredentialTypePassword describes username+pasword based credentials
+	RTCCredentialTypePassword RTCCredentialType = "password"
+	// RTCCredentialTypeToken describes token based credentials
+	RTCCredentialTypeToken RTCCredentialType = "token"
+)
+
+// RTCICEServer describes a single ICE server, as well as required credentials
+type RTCICEServer struct {
+	CredentialType RTCCredentialType
+	URLs           []string
+	Username       string
+	Credential     string
+}
+
+func (c RTCICEServer) isSTUN() bool {
+	for _, url := range c.URLs {
+		if strings.HasPrefix(url, "stun:") {
+			return true
+		}
+	}
+	return false
+}
+
+// RTCConfiguration contains RTCPeerConfiguration options
+type RTCConfiguration struct {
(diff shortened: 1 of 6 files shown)


rtcpeerconnection.go
```diff
@@ -38,17 +37,40 @@ func (t TrackType) String() string {
...
 // RTCPeerConnection represents a WebRTC connection between itself and a remote peer
 type RTCPeerConnection struct {
 	Ontrack                    func(mediaType TrackType, buffers <-chan *rtp.Packet)
 	LocalDescription           *sdp.SessionDescription
 	OnICEConnectionStateChange func(iceConnectionState ice.ConnectionState)
 
+	config *RTCConfiguration
 	tlscfg *dtls.TLSCfg
 
-	iceUsername string
```
Sean-Der: This will need to stay (but I gave it an awful name, so renaming it would be great). This is the `ice-ufrag` and `ice-pwd` and there is only one per session. This isn't the username/password for a TURN server.

This is used for STUN binding requests, and all your candidates will share the same one.

rtcpeerconnection.go
```diff
@@ -77,26 +99,24 @@ func (r *RTCPeerConnection) CreateOffer() error {
...
 	r.portsLock.Lock()
 	defer r.portsLock.Unlock()
 
 	candidates := []string{}
 	basePriority := uint16(rand.Uint32() & (1<<16 - 1))
-	for id, c := range ice.HostInterfaces() {
-		port, err := network.NewPort(c+":0", []byte(r.icePassword), r.tlscfg, r.generateChannel, r.iceStateChange)
-		if err != nil {
-			return err
+	for id, server := range r.config.ICEServers {
```
Sean-Der: I think r.ports will end up being `portsAndTURN`(bad name) but you still want to bind to the IP+Port that ice.HostInterfaces() return.

So for STUN you won't create any new connections (the hole punching will happen and you won't notice anything different, just be listening on local ports like if you were in the same LAN) but for TURN you will need dedicated connections.

nicolai86: I see. I guess I'll need to add an identification if something is TURN or STUN; STUN servers are ignored effectivly ignored, as the local connections should suffice?

rtcpeerconnection.go
nicolai86: I've renamed it to `iceUfrag` and `icePwd` to be close to the RFC

nicolai86:
@Sean-Der  if you're fine with that I'll move the actual STUN implementation into another PR to get the configuration into master.

PR #30, opened 2018-07-05
Add H264 Support (Send, Receive and Packetization)

kc5nra:


pkg/rtp/packetizer.go
```diff
@@ -28,7 +30,7 @@ func NewPacketizer(mtu int, pt uint8, ssrc uint32, payloader Payloader, sequence
 		ssrc,
 		payloader,
 		sequencer,
-		0,
```
Sean-Der: Is this seeded properly? In RTCPeerConnection.go I put one in `func init()` I don't know if that was the right thing to do though.

Sean-Der: Just want to make sure on startup we don't get the same timestamp start every time.

Sean-Der:
LGTM!

PR #33, opened 2018-07-07
Move JSfiddle snippets to git

### LINKED ISSUE #32: Read example jsFiddles from GitHub
According to [this documentation](http://doc.jsfiddle.net/use/github_read.html) jsFiddles can be backed by GitHub. This would allow for easier contributions.

Sean-Der:
jsfiddle provides a way to create snippets from Github via a URL. This
way we can still provide easy demos, but get all the nice things from
having them in Git

Closes #32


Sean-Der:
Here is the jsfiddle bug about not being able to use branches https://github.com/jsfiddle/jsfiddle-issues/issues/1280

PR #36, opened 2018-07-09
basic STUN support

### LINKED ISSUE #18: Implement STUN
Currently pion-WebRTC only generates host candidates. Adding STUN shouldn't be too difficult and mostly just involves modifying SDP generation and a basic STUN client.

* Add a way for a user to pass a STUN server to `pion-WebRTC` add a new argument to RTCPeerConnection to match Javascript.

* Communicate with the STUN servers to get our public IP see https://github.com/pions/turn/tree/master/internal/client and how it communicates with a STUN server and parses the response to the binding request https://github.com/pions/turn/blob/master/internal/server/stun.go#L7-L17

* Start listening on a local port, but return the public IP in the candidate https://github.com/pions/webrtc/blob/master/rtcpeerconnection.go#L107 instead of a host address

nicolai86:
this PR adds support for modern STUN, as specified in [RFC 5389](https://tools.ietf.org/html/rfc5389). Once merged this will close #18.

Missing is stuns support, as well as automatic end 2 end tests to ensure that this functionality does not break. I'll be working on adding those afterwards.

### DIFF
--- a/examples/stun/main.go
+++ b/examples/stun/main.go
@@ -1,13 +1,29 @@
 package main
 
 import (
+	"bufio"
 	"encoding/base64"
 	"fmt"
+	"os"
 
 	"github.com/pions/webrtc"
+	"github.com/pions/webrtc/pkg/ice"
+	"github.com/pions/webrtc/pkg/rtp"
 )
 
 func main() {
+	reader := bufio.NewReader(os.Stdin)
+	rawSd, err := reader.ReadString('\n')
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println("")
+	sd, err := base64.StdEncoding.DecodeString(rawSd)
+	if err != nil {
+		panic(err)
+	}
+
 	peerConnection, err := webrtc.New(&webrtc.RTCConfiguration{
 		ICEServers: []webrtc.RTCICEServer{
 			{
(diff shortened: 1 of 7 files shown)


examples/stun/main.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	if err := peerConnection.CreateAnswer(); err != nil {
+		panic(err)
+	}
+
+	log.Printf("Looping...")
+	for {
```
Sean-Der: I don't think it makes a difference but I just do `select{}` I don't know if conceptually is better/worse!

examples/stun/main.go
```diff
@@ -0,0 +1,30 @@
...
+			{
+				URLs: []string{"stun:stun.l.google.com:19302"},
+			},
+		},
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	if err := peerConnection.CreateAnswer(); err != nil {
```
Sean-Der: Love this example! I would dump the session description maybe? The user can see their public IP might be helpful to see the difference.

rtcpeerconnection.go
```diff
@@ -127,9 +128,28 @@ func (r *RTCPeerConnection) CreateAnswer() error {
 			if server.serverType() != RTCServerTypeSTUN {
 				continue
 			}
-			// TODO connect to STUN server
-			_ = id
-			_ = server
+
+			for _, url := range server.URLs {
+				// TODO if one of the URLs does not work we should just ignore it.
+				client, err := stun.NewClient("udp", url, time.Second*5)
```
Sean-Der: This is so close to being able to do STUN, nice work!

You should be able to just do `network.NewPort("0.0.0.0:" + addr.Port, ....`

All you need to is update the typ of the candidate, and that should be it! You can get the examples from [here](https://webrtchacks.com/sdp-anatomy/)

examples/stun/main.go
nicolai86: gotcha.

examples/stun/main.go
nicolai86: gotcha

PR #56, opened 2018-08-01
Implement full ICE

Sean-Der:
The is the overall architecture.

`ICEAgent` is a member of `network/manager`. (All of the state in `network/manager` *could* go directly on RTCPeerConnection. I think we should keep the code in the top-level as terse as possible. If it isn't relevant for a user to browse/read it should be in internal or pkg.)

Each port push/pulls from `ICEAgent` the STUN traffic for send/recv. The ICEAgent itself has all the timers/selected-pair.

When data comes from the RTCPeerConnection it queries the ICEAgent for the selected pair, and if it has one it will admin the proper src/dest. Otherwise it will return nothing and the traffic will be dropped/queued.

----
All tests and lints pass, and the examples still work! After this lands IMO this is what we need to knock out.

* Integration tests (can Pion talk to itself?) that is gonna go a long way (send a packet, and the test passes when we recv)
* FireFox support
* Pion can open datachannels itself.
* Update CONTRIBUTOR doc/steer people toward what issues we care about most. We need to create more Go generic tasks, not deep stuff
* Public website


pkg/ice/candidate.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// CandidateHost is a Candidate of typ Host
+type CandidateHost struct {
+	CandidateBase
+}
+
+// String for CandidateHost
+func (c *CandidateHost) String(component int) string {
+	return fmt.Sprintf("udpcandidate %d udp %d %s %d typ host generation 0",
```
backkem: I have mixed feelings about this approach. It seems rather clean but this way we are leaking SDP details into the ICE package. E.g.: If we ever want to support the [ORTC API](https://draft.ortc.org/#rtcicecandidate*) this string format doesn't apply anymore.

Sean-Der: I agree!

Are you ok with the SDP package importing ICE? I can have it case off the Candidate type in there.

Then it can generate the string exactly the same way (and then in the future we can add ORTC support)

backkem: That does seem more appropriate. ICE does seem like a pretty big dependency for the SDP package but we can worry about that another time.

Sean-Der: Done!

internal/network/manager.go
```diff
@@ -43,86 +47,72 @@ type Manager struct {
...
+		return errors.Errorf("%s is not implemented", url.Type.String())
 	}
 
-	m.ports = append(m.ports, p)
-	return p.listeningAddr, nil
+	return nil
+}
+
+// Start allocates DTLS/ICE state that is dependent on if we are offering or answering
+func (m *Manager) Start(isOffer bool, remoteUfrag, remotePwd string) error {
```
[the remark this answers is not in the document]
Sean-Der: Yep! I am going to cut a ticket for support `ice-lite` and going to be testing against mediasoup. So will changed to `isControllingICE` and `dtlsRole` maybe?

DTLS does work right now using just offer/answer. The setup attribute just allows someone to fast start (which we should do in the future) but wasn't going to do yet.

Issue #55, opened 2018-07-31
API: Implement the Peer-to-peer Data API


backkem:
Implement the Peer-to-peer Data API as documented in the [spec](https://w3c.github.io/webrtc-pc/#peer-to-peer-data-api)

backkem:
A minimal API has meanwhile been built. We'll create separate issues when working on the remaining missing parts.

PR #57, opened 2018-08-02
[WIP] API: Data Channel API

backkem:
This WIP PR is intended to complete the WebRTC Peer-to-peer Data API (#55).

**TODO**
- [ ] Create DataChannel
- [ ] Complete RTCDataChannel object attributes and event handlers.
- [ ] Base SDP answer on actual data channels.
- [ ] Supply RTCSctpTransport with the correct information.


internal/log/zap.go
```diff
@@ -0,0 +1,25 @@
+package log
```
Sean-Der: thank you SO much for starting on this :) 

This might not be possible, but what do you think of us providing an Interface that people have to satisfy (and not including zap)

I just don't want to make people import/vendor another library (zap) if they don't have to

backkem: The interface is in log.go. We can move the zap implementation to another package so it doesn't have to be imported all the time.

Sean-Der: awesome!

My only concern is just if someone vendors/builds pion. I don't want us to be broken/have our size balloon from Zap (I am fine with this dependency though, logging is a solved problem and just happy with what is popular)

Sean-Der:
Really nice work on this PR, this is great to see :)

I can't wait to start helping with this stuff, this is gonna be huge when it lands.

internal/log/zap.go
backkem: Yea, that's what I meant. We can move the zap.go file to another package so it's no longer a dependency. A user of the lib can optionally use it if they want or roll their own implementation of the log.Logger interface.

examples/pion-to-pion/answer/main.go
```diff
@@ -0,0 +1,105 @@
...
+				fmt.Println("faild to send openchannel", err)
+			}
+		}
+	}
+
+	return peerConnection
+}
+
+func main() {
+	ln, err := net.Listen("unix", "../pion.sock")
```
backkem: I can't run this since I'm on Windows. We should migrate to something more cross-platform for the final example.

Bogdan-D:
Does it work for you, guys? I'm trying to use datachannels from api-datachannels branch, but got no any datachannel open events, not on js nor golang side

PR #59, opened 2018-08-02
Fix typos

brendan-rius:


Sean-Der:
Thank you @brendan-rius mind adding yourself to the contributors in the README! That is the only thing failing in CI

PR #67, opened 2018-08-08
Add extra documentations as well as random re-org to sync with RFC order

trivigy:
Basically this pull request is not going to contain any programming changes to the library with the exception of changed in the orders of variables in different structures. For the most part everything is really synced up but as I am going through the RFC that are still things that are in-flux and are not organized. It just makes it really hard to follow when I constantly have to jump from one place to another between the RFC and the code.

The bottom line goal I am trying to achieve here is so that if someone is reading our documentations they are not necessarily need to go to the RFC in order to understand the more in depth stuff. I am going at it this way because I am also learning the whole library. Once I am up to speed I'll be able to add more documentations and tutorials for a wider range of audience and not just in-depth information.

### DIFF
--- a/examples/gstreamer-receive/README.md
+++ b/examples/gstreamer-receive/README.md
@@ -14,7 +14,7 @@ go get github.com/pions/webrtc/examples/gstreamer-receive
 ```
 
 ### Open gstreamer-receive example page
-[jsfiddle.net](https://jsfiddle.net/usd3xmtz/2/) you should see your Webcam, two text-areas and a 'Start Session' button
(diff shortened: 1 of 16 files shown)


trivigy:
I ended up merging full-ice on top of this. Once full-ice is completely merged into master this branch is going to be only docs changes and not show all the other code changes.

trivigy:
I am actually killing this PR because of all the conflicts that are there due to full-ice changes. I am going to re-create this PR on top of the latest master once full-ice is merged.

PR #71, opened 2018-08-12
Complete rework of sdp parser to comply with rfc4566#section-5

Sean-Der:


Sean-Der:
I like where this is going!

One thing I would like to see is an Attribute struct. Would be cool if we could easily get/set attributes that are just value or key/value and do it properly instead of text munging every time. Stuff like https://github.com/pions/webrtc/blob/master/signaling.go#L123-L137 could be done better

internal/sdp/descriptions.go
```diff
@@ -82,21 +146,44 @@ type SessionDescription struct {
...
-	s.RepeatTimes = nil
-	s.TimeZones = nil
-	s.EncryptionKeys = nil
-	s.Attributes = nil
-	s.MediaDescriptions = nil
+// MediaDescription represents a media type.  Currently defined media are "audio",
+// "video", "text", "application", and "message", although this list
+// may be extended in the future
+// https://tools.ietf.org/html/rfc4566#section-5.14
+type MediaDescription struct {
```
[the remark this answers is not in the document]
trivigy: No good reason. This is just temporary while I am adding a bunch of new structs and print functions. Wanted to have it all in one place. I will break them all down once I finish.

trivigy:
You got it. I am going to add that too. I was planning to break it down as completely as I can. I was also planning to add all of the ice event somehow. In the rfc there is a lot of things that happen with the ice server settings depending on the set attributes. I will need to figure out a design for how to put it even handling or just pass in the network manager inside via the lexer object. If you have a preference to how you think it should ultimately be feel free to letting me know.

Sean-Der:
Nice work, excited to see text blobs becoming real structures :) a couple asks.

* keep the file names singular `media_descriptions` -> `media_description`
* Add tests for Marshal/Unmarshal of Chrome and FireFox SDP nothing fancy just want to increase the coverage and make sure we don't regress, if those pass for Offer+Answer I can't ask for anything better!
* Why changing from `Address -> IP` an address is valid in all those cases. You can have `foo.bar` for example, most of those were copied straight from the RFC

internal/sdp/jsep.go
```diff
@@ -34,17 +34,13 @@ const (
...
-		Attributes: []string{
-			//	"ice-options:trickle", // TODO: implement trickle ICE
-			"fingerprint:sha-256 " + fingerprint,
+		ProtocolVersion:  0,
+		Origin:           Origin{"-", newSessionID(), uint64(time.Now().Unix()), "IN", "IP4", "0.0.0.0"},
+		SessionName:      "-",
+		TimeDescriptions: []TimeDescription{{Timing{0, 0}, nil}},
+		Attributes: []Attribute{
+			// 	"Attribute(ice-options:trickle)", // TODO: implement trickle ICE
+			Attribute("fingerprint:sha-256 " + fingerprint),
```
Sean-Der: What do you think of making Attribute an interface, and having `AttributeKey` `AttributeKeyValue` would be nice if this just looked like `AttributeKeyValue("fingerprint", sha-256 " + fingerprint)` that is my biggest concern in the SDP code.

I don't want any string munging outside the SDP code (like the stuff in signaling.go)

trivigy: Yeah. I was thinking about exactly the same thing. Lets wrap up this run over the code first and then I will look into how to actually make what you are talking about. But generally speaking I am in agreement.

trivigy:
@Sean-Der 
> Why changing from Address -> IP an address is valid in all those cases. You can have foo.bar for example, most of those were copied straight from the RFC
A: The reason is because the address structure according to the RFC is actually more complicated. You can have an address that is describing multicast with a range of IPs. I wanted to actually parse all of that into a structure and keep it broken down. I also added IP format verification.

trivigy:
As for all your other comments I agree with and working on fixing them all up. I think that by end of tonight the first working version is going to be available.

PR #72, opened 2018-08-12
.gitignore improvement

trivigy:
Super tiny change. Just wanted to add just a tiny changes to gitignore so that people like me that work with jetbrains IDE do not have to suffer. :smile:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,13 @@
+### JetBrains IDE ###
+#####################
+.idea/
+
+### Folders ###
+#############
+bin/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,13 @@
-*.ivf
-tags
+### JetBrains IDE ###
+#####################
+.idea/
+
+### Folders ###
+#############
+bin/
+tags/
```
Sean-Der: just `tags` it is just a file

Issue #53, opened 2018-07-27
Firefox Support


dhannyz:
Firefox is using sdparta_N in their SDP offers as a bundle & mid value instead of the hardcoded 'data' value.
Additionally Firefox only supports a max 62bit value as a sessionID, currently the 64bit value in the answer causes Firefox to error during SDP answer parsing.

Sean-Der:
I have started work on this, will open a PR when I make some progress.

Sean-Der:
Hey @dhannyz can you try the PR out, it has full FireFox support (audio, video and datachannels) I hope to merge it soon.

Would love to get you back on pion-WebRTC

thanks!

Sean-Der:
This landed with #73 @dhannyz

PR #73, opened 2018-08-13
[WIP] FireFox support

Sean-Der:
This is for #53 

This is the status of examples
* [x] data-channels 
* [x] gstreamer-receive 
* [x] gstreamer-send  
* [x] save-to-disk

### DIFF
--- a/internal/sdp/util.go
+++ b/internal/sdp/util.go
@@ -44,7 +44,7 @@ func (t ConnectionRole) String() string {
 }
 
 func newSessionID() uint64 {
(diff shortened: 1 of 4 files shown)


Sean-Der:
Hey @backkem I believe this is good to go.

I think we can do a lot better with how we do SDP stuff, maybe we can land this and get your opinion on where we take this next.

Also once this lands we can chat about what is next `pion <-> pion` and making ipfs happen, feel free to grab me on Slack :)

PR #82, opened 2018-08-19
Allows sending raw RTP packets to track 

Sean-Der:
Before, we could only send RTCSample to outgoing tracks.

Now we can also send directly RTP packets. This allows for more efficient processing when you just want to forward media from a connection to another (if both connections use the same codec).

I have renamed the `Packets` member of `RTCTrack` which was only used for pulling packets from incoming tracks to `IncomingPackets`. Now, you can also push packets for outgoing tracks by queung them to the `OutgoingPackets` channel.

Pleasen note that one you've started sending RTCSample to a track, you can only send RTCSample and likewise, if you start sending RTP packets you can only send RTP packets.
Mixing both would mess up with the sequence number of the RTP packets.
You can see if the track accepts RTP packers or RTCSamples by inspecting the new `Accepts` member of a track which can either be `RCTTrackAcceptsOnlySamples`, `RCTTrackAcceptsOnlyRTPPackets` or `UndefinedYet`

You can now do something like:

```golang
track1.OutgoingPackets <- <-track2.IncomingPackets
```
to forward RTP packets from a track (`track1`) to another (`track2`) as long as both tracks share the same codec


Sean-Der:
Hey @brendan-rius @kc5nra I am having a hard time with this one conceptually. Feel free to push back, my only concern is that pion-WebRTC is easy/enjoyable to use! I think this change may make pion-WebRTC harder to use, but I don't know all the things people are trying to build. I also have lots biases.

If we ask users to craft RTP packets I believe people will spend more time modifying things/wrangle state. Things like ssrc/seqnum etc... will be exposed to users, and when proxying we don't even want this information to be passed between RTCPeerConnections.

I think in the long term embedders will be much happier if they just get audio/video and don't have to think about things above them. When jitterbuffer etc.. lands we will give hooks into that so users can decide what behavior they want.

These are the applications I can think of and how they would handle data.
----
## Sending raw media
Users just need to craft samples, we handle the rest currently. 

If we ask people to craft RTP packets it is more work on them (they need to break stuff up for MTU etc..) and I don't believe the extra flexibility will give much of a reward.

# Receiving raw media
Users just get audio/video data, they don't care about the contents of the RTP packets. I don't believe exposing the RTP packets will offer any additional info. We want to get to the point where jitter/rtx is all handled before the user even notices.

# Proxying
I see the use case here, you just want to copy + push RTP packets between RTCPeerConnections. All you need to do is update the ssrc/seqnum.

We need to de-jitter/rtx before proxying, we don't want to just send out packets as they come in otherwise it will just makes things messier. While we do this we should have enough data to take things to samples anyway. 

Does anyone think this is too much boilerplate/would make users unhappy (vs copying an RTP packet and change seqnum/ssrc)
````
destinations := []chan<- webrtc.RTCSample

peerConnection.Ontrack = func(track *webrtc.RTCTrack) {
  sampleBuilder := sampleCollector.New(sampleCollector.vp8)
  for {
    if sample := sampleBuilder.Push(<-track.IncomingPackets), sample != nil {
      for d := range destinations {
        d <- sample
      }
    }
  }
}
```

brendan-rius:
@Sean-Der I think you're right. So in the proxy example, sampleBuilder would rebuild samples from the incoming RTP packets right?

Sean-Der:
@brendan-rius yep exactly!

I will build the proxy example, and include you on the PR and we can see if it works for everyone (if not we can scrap it and try another way)

I am happy to try a bunch of different ways! Just want to make sure you think pion-WebRTC is pleasant to use (and if not call me out!)

thanks for the feedback :)

notedit:
I think pions-webrtc can just provide raw rtp packet in and out api.  pions-webrtc is a base lib.
we could provide some other lib/module to do rtp pack and unpack.

PR #88, opened 2018-08-27
Config upgrade

trivigy:
This pull request includes a spec compliant RTCConfiguration with all of the things associated with it like RTCCertificate, RTCIceServer, etc...


trivigy:
Just FYI that right now stuff is disconnected from the way old ice works and obviously because of that library is not going to work. I am finally done with the unittests and going to hook everything up back according to the old way. But note this is going to be temporarily until `RTCIceTransport` starts working at which point I will switch everything through that.

examples/data-channels-create/main.go
```diff
@@ -16,7 +16,7 @@ import (
 func main() {
 	// Prepare the configuration
 	config := webrtc.RTCConfiguration{
-		ICEServers: []webrtc.RTCICEServer{
```
Sean-Der: Can we keep `ICE` instead of `Ice` 

Since it is an acronym I think it makes sense (and we won't break existing users)

trivigy: +1

rtcpeerconnection.go
```diff
@@ -415,7 +415,7 @@ func (pc *RTCPeerConnection) SetRemoteDescription(desc RTCSessionDescription) er
 		for _, a := range m.Attributes {
 			if strings.HasPrefix(*a.String(), "candidate") {
 				if c := sdp.ICECandidateUnmarshal(*a.String()); c != nil {
-					// pc.networkManager.IceAgent.AddRemoteCandidate(c)
+					pc.networkManager.IceAgent.AddRemoteCandidate(c)
```
trivigy: Sorry for missing this stuff. Thanks for catching it.

PR #94, opened 2018-08-28
Move DOM wrapper errors into their own package

trivigy:


trivigy:
Lack of interest with the rest of the team this change has to be reworked to be more golang / pions-webrtc specific and less JS API DOMException kinda deal.

PR #101, opened 2018-09-03
Implement RTCP and Proxying

Sean-Der:
This PR adds code (and an example) on how to proxy audio/video traffic, this will be useful in the future when building an SFU.


Sean-Der:
Killing this PR it is too large to review, breaking into smaller ones.

PR #103, opened 2018-09-04
Transports

trivigy:
This is a fairly small PR. I realized that I need to go through the SCTP and DTLS protocols we have built so far in order to make sure they can support the necessary parts for the RTCDtlsTransport and RTCIceTransport and RTCSctpTransport structs. As I started going through those sections I started thinking the PR might end up being way too big so instead I broke it down into this one, which really includes the following, and later I will come with SCTP improvements PR, DTLS improvements PR, and finally will come back with the transports implemented.

PR includes:
- some re-organization
- createDataChannel with more compliance and all the relevant structs improved.
- more documentation
- minor code and style fixes here and there.


trivigy:
Btw, guys sorry there are tons of files touched. In practice not much real useful code was actually added. Mostly it is just re-organization of stuff all over the place.

trivigy:
Also, 1.8% coverage taking us over 40%, hell yeah baby!!!

backkem:
Have you tested the examples yet?

trivigy:
@backkem I have addressed the question of being too big on slack. The idea for now to keep everything public in one place until stuff is a little built up and then export the internals into their own places. So totally with agree with your unpointed comment and will address it as stuff becomes more spec compliant.

In terms of the examples not yet. I am going to test the data stream stuff before I merge but I don't know how to do the video stream testing since I am not using it in my own project or know how it works. That's why I really want the e2e testing.

trivigy:
@backkem But I would greatly appreciate comments regarding the organization and where I am placing public interfaces.

rtcdatachannel.go
```diff
@@ -147,7 +116,7 @@ func (d *RTCDataChannel) SendOpenChannelMessage() error {
 
 // Send sends the passed message to the DataChannel peer
 func (d *RTCDataChannel) Send(p datachannel.Payload) error {
-	if err := d.rtcPeerConnection.networkManager.SendDataChannelMessage(p, d.ID); err != nil {
+	if err := d.rtcPeerConnection.networkManager.SendDataChannelMessage(p, *d.ID); err != nil {
```
Sean-Der: To be consistent can/should we follow the NullBool pattern here? 

There are a lot of places through out this PR that do the pointers to primitives, should we do the NullBool pattern instead? I am happy to help with that also if you want to start a branch (we can do the same for the SDP library)

rtcpeerconnection.go
```diff
@@ -23,47 +25,85 @@ import (
...
+	SignalingState RTCSignalingState
+
+	// IceGatheringState attribute returns the ICE gathering state of the
+	// RTCPeerConnection instance.
+	// IceGatheringState RTCIceGatheringState // FIXME NOT-USED
+
+	// IceConnectionState attribute returns the ICE connection state of the
+	// RTCPeerConnection instance.
+	// IceConnectionState RTCIceConnectionState  // FIXME SWAP-FOR-THIS
+	IceConnectionState ice.ConnectionState // FIXME REMOVE
```
Sean-Der: What is wrong with having ice.ConnectionState be public?

We have to maintain mappings and boilerplate otherwise. I think conceptually we want `github.com/pions/webrtc` to be tiny, it is just the shim overtop while everything complicated lives inside pkg/internal

trivigy:
In terms of the empty implementations I just created those to know what else needs to be built. I put `panic()` inside of them but really they are commented out because the current lint doesn't allow "unused" functions. Lets keep them this way for now. If someone reads the code, they might create the issue or just help and contribute. I am hopefully will fill up most of them in the next few PRs. In terms of the TODO/FIXME lets keep it this way. Personally I like seeing those inside of libraries because they teach me what else is missing or not structured. Sometimes I read just to understand other's code and have to spend so much time just learning if something is missing/not-used. This way people can actually just see it and again, help contribute if they find it interesting. Again, hopefully most of it will start disappearing in the new few PRs.

Apart of why this looks bad is because I am flushing it out with respect to the specs. So you see all the things that are missing. I would recommend not to stress over them too much now. Our obvious goal is to fill everything up. But at least lets make the "shit" visible for everyone to improve. Hehehe... I am saying it in the most positive way possible. Making shit visible is really great for all of us. We can then improve on it.

examples/data-channels-create/main.go
```diff
@@ -46,9 +46,9 @@ func main() {
 		}
 	}
 
-	// Register the Onmessage to handle incoming messages
+	// Register the OnMessage to handle incoming messages
 	dataChannel.Lock()
-	dataChannel.Onmessage = func(payload datachannel.Payload) {
+	dataChannel.OnMessage = func(payload datachannel.Payload) {
```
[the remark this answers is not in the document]
trivigy: Hmmmmmmmm.... Not sure. Not really doing this for any reason other than being consistent with all of the other golang style syntax. So I am happy to change it back. Purely golang syntax aesthetics.

examples/data-channels-create/main.go
```diff
@@ -33,7 +33,7 @@ func main() {
 
 	// Set the handler for ICE connection state
 	// This will notify you when the peer has connected/disconnected
-	peerConnection.OnICEConnectionStateChange = func(connectionState ice.ConnectionState) {
+	peerConnection.OnIceConnectionStateChange = func(connectionState ice.ConnectionState) {
```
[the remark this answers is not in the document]
trivigy: This stuff we talked about in terms of the "ICE" vs "Ice" to be consistent. I guess again I am just being consistent. Generally speaking though I am really not as scared of breaking things for users much right now because we are not at 80% coverage and high spec compliance. I think when there are loads of dependent project is when we start worrying but as it stands right now, only recently people are started to adopt us according to https://godoc.org/github.com/pions/webrtc?importers. I really do believe that as long as we are improving in the direction of better readability people will make the right adjustments.

rtcdatachannel.go
trivigy: Yeah I did this but haven't gotten into this PR yet. Remember I decided to break the PR into smaller PRs because stuff started getting a little too big. :D

In this specific case those it is a little different. This is not because of the NULLBool thingy, it is because the specs, not the config struct but the actual struct, talks about the sctp getting configured through an out-bound channel establishment. This means that after `CreateDataChannel()` the ID may actually be still nil. It is an actual feature.

rtcpeerconnection.go
trivigy: Ohhh.... There is nothing wrong with this stuff. The actual spec keeps it separate. The `RTCIceConnectionState` is part of the WebRTC interface and the `IceAgent` has its own state management. I started working towards it but needed to fix up sctp a little to be able to finish `RTCIceTransport` which is the abstraction layer for the `IceAgent`. They literally do it that way in order to not have what we ended up doing which is sharing that enum in two separate packages. The webrtc and the ice. So I am just following the spec but cannot swap it right away in order to keep the PRs still functional and not break the library. Sorry.

PR #104, opened 2018-09-05
Tiny PR that fixes backward compatible data channels

trivigy:


trivigy:
Sorry about the hit on the coverage. I had to remove one test in order to not have to re-write this test right now.

trivigy:
Next proper PR this test will be completely rewritten with a whole bunch of new goodies and a spike in coverage.

rtcdatachannel.go
```diff
@@ -88,27 +92,35 @@ type RTCDataChannel struct {
...
-//
-// 	d.ReadyState = RTCDataChannelStateClosing
-//
-// 	// if err := d.rtcPeerConnection.networkManager.SendOpenChannelMessage(d.ID, d.Label); err != nil {
-// 	// 	return &rtcerr.UnknownError{Err: err}
-// 	// }
-// 	return nil
-//
-// }
+func (d *RTCDataChannel) generateID() error {
```
Sean-Der: IMO it is confusing to have this one the RTCDataChannel itself. It doesn't depend on anything state of the DataChannel we are dispatching off of, and would cause weird behavior if called multiple times.

rtcdatachannel_test.go
```diff
@@ -1,35 +1,31 @@
 package webrtc
 
-import (
```
Sean-Der: We can keep this test if we move ID stuff back

rtcicetransport.go
```diff
@@ -3,4 +3,28 @@ package webrtc
 // RTCIceTransport allows an application access to information about the ICE
 // transport over which packets are sent and received.
 type RTCIceTransport struct {
```
Sean-Der: Can we bring these back?

rtcdatachannel.go
trivigy: I know... The only reason it is like that is because I was in the middle of transitioning it into the `DataChannel` when I realized that I actually broke master. I will move it for this PR but ultimately it does not belong to `RTCPeerConnection`. It is purely an internal mechanism for leasing the `datachannels` which is being moved into the `sctpTransport` as well.

rtcpeerconnection.go
```diff
@@ -75,60 +75,72 @@ type RTCPeerConnection struct {
...
+	// Deprecated: Internal mechanism which will be removed.
+	networkManager *network.Manager
+}
 
 // New creates a new RTCPeerConfiguration with the provided configuration
 func New(configuration RTCConfiguration) (*RTCPeerConnection, error) {
+	// https://w3c.github.io/webrtc-pc/#constructor (Step #2)
 	// Some variables defined explicitly despite their implicit zero values to
 	// allow better readability to understand what is happening.
-	pc := RTCPeerConnection{
```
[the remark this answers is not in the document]
trivigy: No problem.

rtcicetransport.go
trivigy: Oh these never existed!!! This is all the new stuff that I am building. Again, it the PR just because I pushed stuff in order to fix master. This will be created in the next few PRs.

rtcdatachannel_test.go
trivigy: Yeah no problem. I will bring this back for now. But I am ultimately moving that function out of `RTCPeerConnection` since it does not belong there.

PR #116, opened 2018-09-10
Move IVFWriter to `pkg` and implement `SampleBuilder` with tests

Sean-Der:
Full description in commits

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -1,0 +1,96 @@
+package samplebuilder
+
+import (
+	"github.com/pions/webrtc/pkg/media"
+	"github.com/pions/webrtc/pkg/rtp"
+)
+
+// SampleBuilder contains all packets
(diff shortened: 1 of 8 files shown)


pkg/media/samplebuilder/samplebuilder.go
```diff
@@ -0,0 +1,95 @@
...
+type SampleBuilder struct {
+	maxLate uint16
+	buffer  [65536]*rtp.Packet
+
+	// Last seqnum that has been added to buffer
+	lastPush uint16
+
+	// Last seqnum that has been successfully popped
+	// -1 means if no pop has occurred
+	lastPopSeq       int32
```
kc5nra: I would probably keep lastPopSeq and all datatypes that operate on seqnums in uint16 rather than casting, and encode -1 somewhere else

Sean-Der: @kc5nra updated PR! thanks for the review

PR #121, opened 2018-09-11
Start RTCP package: add header parsing

maxhawkins:
More to come. Let me know if there's anything I need to change otherwise I'll proceed with the rest of the packet types.

### DIFF
--- a/pkg/rtcp/packet.go
+++ b/pkg/rtcp/packet.go
@@ -8,11 +8,11 @@ import (
 
 // RTCP packet types registered with IANA. See: https://www.iana.org/assignments/rtp-parameters/rtp-parameters.xhtml#rtp-parameters-4
 const (
-	TypeSR   = 200 // sender report        RFC 3550, 6.4.1
-	TypeRR   = 201 // receiver report      RFC 3550, 6.4.2
-	TypeSDES = 202 // source description   RFC 3550, 6.5
-	TypeBYE  = 203 // goodbye              RFC 3550, 6.6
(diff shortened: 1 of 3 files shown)


Sean-Der:
Hey @maxhawkins fantastic work!

Mind branching off of master instead (rtcp branch has C code so won't build in CI) so we don't get the tests.

pkg/rtcp/packet.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"encoding/binary"
+
+	"github.com/pkg/errors"
+)
+
+// RTCP packet types registered with IANA. See: https://www.iana.org/assignments/rtp-parameters/rtp-parameters.xhtml#rtp-parameters-4
+const (
+	TypeSR   = 200 // sender report        RFC 3550, 6.4.1
```
Sean-Der: If these stay public can you make them the full name `TypeSenderReport` if they stay private it is fine using a less descriptive name (I just want to ensure the GoDoc isn't confusing)

I think they will end up being private if you do an API like.
```
pkt := &rtcp.Nack{SequenceNumber: 5000,  SSRC: 1234}
raw, err := pkt.Marshal()
```

maxhawkins: Not sure what the API is going to look like yet. I'll make them public and unabbreviated for now.

maxhawkins:
OK, PTAL

maxhawkins:
OK I think this is ready to merge.

PR #131, opened 2018-09-14
Add nullvalue to express significant zero values

### LINKED ISSUE #105: Remove `reftypereturn` instead follow NullString pattern
Using a pointer to represent primary data types is prone to error/not self documenting.

Instead replace reftypereturn with `util.NullString util.NullUint16` etc.. they can't be in the WebRTC package since SDP needs to use them as well.

Sean-Der:
A common pattern in ECMAScript is that you can pass a
value OR null, however with Go this isn't possible. This adopts
the sql.NullString pattern so that we can as closely as possible
mirror the browser API for WebRTC

Resolves #105


Sean-Der:
@trivigy I know you had opinions on this so added as reviewer.

I want to use this for the SRTP package (SSRC may be null if we are handling RTCP package) and if that all works out will working on moving SDP to it.

trivigy:
@Sean-Der if you really care about it do it. I went back and forth on this and currently think to leave this be for now. The current thinking behind the opinion is that if we add this now, this will add an extra public interface that is pretty much embedded everywhere. So I think we should try to shy away from that until we hit like 70%/80% coverage and library stability. Then we could introduce this. I believe you hold a different opinion, and at the moment I care more about getting sctp and transports complete, so feel free to make a call on this.

Sean-Der:
@trivigy I don't want to merge if you think it is a bad idea!

But we won't be able to do this if we don't do it now (I believe) if our public API uses pointers to values, it will need to stay that way.

```
Also the sql has a bunch of functions that we are missing for setting the actual objects. Remember that they have this Valid member which means that no direct manipulation of these objects should be allowed to maintain state.
```

SQL packages has them like
```
type NullInt64 struct {
        Int64 int64
        Valid bool // Valid is true if Int64 is not NULL
}
```

We can add additional methods, but I don't they they are relevant (they are used by SQL driver right?) when I have done SQL code I just work with the members directly.

PR #134, opened 2018-09-14
[WIP] Generate payload types

### LINKED ISSUE #43: media engine: generate payload types
When creating a codec, dynamically generate a payload type in the range 96-127 if one wasn't provided. The generation is described in [RFC3551](https://tools.ietf.org/html/rfc3551#section-3)

porjo:
Resolves #43

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -41,26 +41,30 @@ type MediaEngine struct {
 
 // RegisterCodec registers a codec to a media engine
 func (m *MediaEngine) RegisterCodec(codec *RTCRtpCodec) uint8 {
+	if codec.PayloadType == nil {
+		// check for existing payload type
+		for _, c := range m.codecs {
+			if c.Equals(codec) {
+				codec.PayloadType = new(uint8)
+				*codec.PayloadType = *c.PayloadType
+				break
+			}
+		}
 
-	var payloadType uint8
-
-	// check for existing payload type
-	for _, c := range m.codecs {
-		if c.Name == codec.Name {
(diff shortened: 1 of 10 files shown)


mediaengine.go
```diff
@@ -14,51 +14,52 @@ func RegisterCodec(codec *RTCRtpCodec) {
...
 
-func (m *MediaEngine) getCodec(payloadType uint8) (*RTCRtpCodec, error) {
-	for _, codec := range m.codecs {
-		if codec.PayloadType == payloadType {
-			return codec, nil
+	var payloadType uint8
+
+	// check for existing payload type
+	for _, c := range m.codecs {
+		if c.Name == codec.Name {
```
porjo: Is it sufficient to identify the codec by name only?

backkem: The payload type defines a specific codec, including all its properties. Looking up a codec based on its properties can be done like in ``getCodecSDP``.

porjo:
> Those applications which need to define more than 32 dynamic payload types

That seems like an extreme use case to need more than 32 payload types per peer connection. Do we need to cater for that?

Sean-Der:
Hey @porjo really sorry this took so long, when I get home from work will get you the feedback needed to get merged.

thanks!

Sean-Der:
Hey @porjo I am really sorry this got ignored. That was really bad on our part, I passed this off and didn't realize you were waiting on feedback from us.

Hopefully this didn't push you away from using Pion! I am going to close this for now, I am going to go back and fix how painful it is to use the Media APIs in the next release.

PR #136, opened 2018-09-14
Examples: host examples locally

### LINKED ISSUE #120: Examples: Serve fiddles locally
We should run a local HTTPS server so people don't even need to open a jsfiddle. This will also allow people to prototype faster using tools they are already comfortable with.

backkem:
We can make it look fabulous down the line. I'm going for integration testing first.

Resolves #120

### DIFF
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -1,0 +1,3 @@
+---
+exclude_paths:
+  - examples/examples.json
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,13 @@
+# Examples
+This directory contains some example use-cases of Pions WebRTC.
+
+## Instructions
+You can host the examples locally on your machine as follows:
+``` sh
+go get github.com/pions/webrtc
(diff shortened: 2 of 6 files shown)


Sean-Der:
This is amazing! Really nice work.

Could you add SSL also, just generate a self signed cert? getUserMedia requires it. After that I think we should be good.

backkem:
Actually, I think localhost is exempted from this rule for testing purposes. This makes me lean toward using http so you don't have to constantly accept the self signed cert. Thoughts?

Sean-Der:
Great work! The only thing I would do is have people run the examples directly, since it reads off disk. I like having people `go install` only because jsfiddle sorta lowered the friction.

The examples could be as simple as this in one block
```
go get github.com/pions/webrtc
cd $GOPATH/src/github.com/pions/webrtc/examples
go run example.go
```

I also think after this we should maybe re-think the README, this probably should be the 'Official' way. We can leave up the jsfiddles with links (and if they break/regress we kill them if they cause trouble)

Issue #83, opened 2018-08-19
Design proposal for RTCIceTransport


trivigy:
For a little white I have been brainstorming on how to best enhance the ICE protocol and finally after conversations with @Sean-Der decided to put down the information I found in this proposal. This change is primarily only for enhancing internal readability and does not change a single thing in terms of the public API. In this proposal I am not going to cover in depth how the current design of the IceAgent is structures In fact for the most part how it operates is really not negotiable since it needs to comply with [RFC5245] (https://tools.ietf.org/html/rfc5245). 
>If you are interested please follow the code [HERE](https://github.com/pions/webrtc/blob/master/pkg/ice/agent.go#L44). Current implementation uses this object in two places. During initialization [HERE](https://github.com/pions/webrtc/blob/master/internal/network/manager.go#L66-L81) and when starting up the gathering process [HERE](https://github.com/pions/webrtc/blob/master/signaling.go#L145).

What I will mention however is why in my opinion this design is hard to read and how to package as its own module:
1. The code is initialized during application initialization but, in fact nothing within the module is actually used during that intialization nor is there any initialization specific information covered in the ICE RFC. This fact seems to have been actually addressed and extended using [RTCPeerConnection Interface Extensions] (https://www.w3.org/TR/webrtc/#rtcpeerconnection-interface-extensions) which talks about an object called `RTCIceTransport` whose suppose is to be an interface into the ICE process.
2. The whole process of gathering is kicked off with a `start()` function which does not convey the needed information that this is a gathering of candidates process.
3. Local candidate gathering is happening outside of the `IceAgent`. This makes things hard for isolation purposes.
4. `IceAgent` shares objects with the rest of the webrtc. This one makes it especially hard to make ICE its own module. In fact after reading the entire ICE RFC there is absolutely no mentioning of webrtc state management which raises the question of why ICE shares objects with webrtc at all.

**Solution**:
> Note this is not my solution an I am simply adopting the extended variation of the webrtc spec.

[...]

PR #138, opened 2018-09-14
Add getIceServers to support easy server parsing

trivigy:
Tiny PR that only adds one function that is used in later PRs for ease of ice server parsing.

Relates to #83

### DIFF
--- a/rtcconfiguration_test.go
+++ b/rtcconfiguration_test.go
@@ -1,0 +1,38 @@
+package webrtc
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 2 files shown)


rtcconfiguration_test.go
```diff
@@ -0,0 +1,39 @@
...
+				{
+					URLs: []string{expectedServerStr},
+				},
+			},
+		}
+
+		_, err := cfg.getIceServers()
+		assert.NotNil(t, err)
+	})
+
```
kc5nra: nit extra ws

trivigy: @kc5nra what? :)

PR #145, opened 2018-09-16
SRTP Context can now handle multiple SSRCes

### LINKED ISSUE #143: Move SRTP SSRC handling into SRTP package
Right now we put too much logic in `networkManager`

While working on SRTCP implementation issues have come up around this. It would be better to just have one inbound and one outbound SRTP Context.

This would make our network code cleaner, and makes SRTCP easier.

Sean-Der:
Improve the SRTP context so it can be sent multiple SSRCes.
This refactor simplifies our network code, and also enables us to
implement RTCP before we would only negotiate a context on RTP traffic
and RTCP breaks this pattern

Resolves #143


internal/network/port-receive.go
```diff
@@ -105,7 +90,25 @@ func (p *port) handleDTLS(raw []byte, srcAddr string) {
 
 	p.m.certPairLock.Lock()
 	if certPair := p.m.dtlsState.GetCertPair(); certPair != nil && p.m.certPair == nil {
+		var err error
 		p.m.certPair = certPair
+
+		p.m.srtpInboundContextLock.Lock()
```
kc5nra: Dunno if too nitty, but lock held  longer than needed -- doubt it matters

PR #162, opened 2018-09-22
Implement SFU example

### LINKED ISSUE #63: Can we have an example for attach current stream to other webRTC
Hi Pions

could you create an example show how to attach a stream track (track *webrtc.RTCTrack) from webRTC call A to another WebRTC call B and send it to B browser ? 

* Plus could we have sample to push stream track to another RTMP server :) 

Best regards,
Toan

Sean-Der:
`sfu` example shows how to build a simple broadcast only SFU

Resolves #63


internal/network/port-send.go
```diff
@@ -40,3 +41,22 @@ func (p *port) sendSCTP(buf []byte, dst fmt.Stringer) {
 		fmt.Println(err)
 	}
 }
+
+func (p *port) sendRTCP(buf []byte, dst net.Addr) {
+	p.m.srtpOutboundContextLock.Lock()
+	defer p.m.srtpOutboundContextLock.Unlock()
+	if p.m.srtpOutboundContext == nil {
+		fmt.Printf("Tried to send RTP packet but no SRTP Context to handle it \n")
```
kc5nra: Comment probably should say RTCP packet etc.

internal/network/manager.go
```diff
@@ -159,6 +159,22 @@ func (m *Manager) SendRTP(packet *rtp.Packet) {
 	}
 }
 
+// SendRTCP finds a connected port and sends the passed RTCP packet
+func (m *Manager) SendRTCP(pkt []byte) {
+	m.portsLock.Lock()
```
kc5nra: I noticed in several places we lock around things other than ports -- should we rename portsLock to something else? Or reduce the scope of the locks?

Sean-Der: It still is just for the ports (iceAgent interaction is thread safe)

I moved the locks down, should read a bit nicer!

Issue #159, opened 2018-09-19
DataChannel open message missing


Bogdan-D:
In current webrtc state dataChannel working only if they are opened from javascript.
Datachannel open from golang does nothing (even when I call `SendOpenChannelMessage` manually), also manual negoitaion is not possible

backkem:
Yes, you are correct. We are actively working on filling these gaps in our implementation.

In the mean time: The ``data-channels-create`` example has a workaround for this that works with Chrome. (In my testing it doesn't work when connecting to NodeJS thought). On [main.go#L42](https://github.com/pions/webrtc/blob/master/examples/data-channels-create/main.go#L42) you can see we manually call ``SendOpenChannelMessage`` when the ``ICEConnectionState`` changes to ``Connected``. Again, this is obviously a temporary workaround until we can complete the underlying implementation.

Bogdan-D:
Yeah, I've tried that workaround but without success

backkem:
Hi @Bogdan-D, can you please test again with the last master branch. We're now sending the ChannelOpen messages when creating data channels in Pions. These changes also make Pions to Pions and Pions to NodeJS data channels work in my testing.

Sean-Der:
Thanks for implementing this @backkem  :)

Closing for now, feel free to re-open @Bogdan-D if it is still an issue, thanks!

PR #169, opened 2018-09-22
ICE: Cleanup agent notifier

backkem:
Relates to #159

### DIFF
--- a/pkg/ice/agent.go
+++ b/pkg/ice/agent.go
@@ -59,7 +59,7 @@ type Agent struct {
 	sync.RWMutex
 
 	outboundCallback OutboundCallback
-	iceNotifier      func(ConnectionState)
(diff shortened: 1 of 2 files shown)


pkg/ice/agent.go
```diff
@@ -192,10 +192,14 @@ func (a *Agent) sendSTUN(msg *stun.Message, local, remote Candidate) {
 }
 
 func (a *Agent) updateConnectionState(newState ConnectionState) {
-	a.connectionState = newState
```
Sean-Der: Can we remove this conditional now? https://github.com/pions/webrtc/blob/master/rtcpeerconnection.go#L786

I added that so the agent wouldn't spam the public API. But since the agent is doing that check now, we should be able to kill in `RTCPeerConnection`

thanks!

PR #174, opened 2018-09-24
SCTP: Notifier & Initialization

backkem:
- SCTP can now initiate a connection.
- SCTP can now notify the upper layers of its connection state.

Relates to #159


Sean-Der:
@backkem This LGTM! Thanks for chipping away at this :)

 I would maybe also get a review from @kc5nra I think he had some ideas around this, he might not have time so if you don't hear from him in 24 feel free to merge!

PR #177, opened 2018-09-25
DataChannels: OpenChannel & OnOpen

backkem:
- OpenChannel messages are now sent after SCTP is Established.
- The OnOpen handler tells the application when a channel has opened.

Relates to #159

### DIFF
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -809,10 +809,15 @@ func (pc *RTCPeerConnection) dataChannelEventHandler(e network.DataChannelEvent)
 	switch event := e.(type) {
 	case *network.DataChannelCreated:
 		id := event.StreamIdentifier()
-		newDataChannel := &RTCDataChannel{ID: &id, Label: event.Label, rtcPeerConnection: pc}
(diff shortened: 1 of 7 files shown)


rtcpeerconnection.go
```diff
@@ -829,6 +834,20 @@ func (pc *RTCPeerConnection) dataChannelEventHandler(e network.DataChannelEvent)
 			fmt.Printf("No datachannel found for streamIdentifier %d \n", e.StreamIdentifier())
 
 		}
+	case *network.DataChannelOpen:
+		for _, dc := range pc.dataChannels {
+			dc.Lock()
+			err := dc.sendOpenChannelMessage()
+			if err != nil {
+				fmt.Println("failed to send openchannel", err)
```
Sean-Der: should we `continue` here if we can't open?

rtcpeerconnection.go
```diff
@@ -829,6 +834,20 @@ func (pc *RTCPeerConnection) dataChannelEventHandler(e network.DataChannelEvent)
...
+			}
+			dc.ReadyState = RTCDataChannelStateOpen
+			dc.Unlock()
+
+			if dc.OnOpen != nil {
```
Sean-Der: I would guard this with a Lock (Or RLock) the race detector might throw warnings otherwise.

PR #188, opened 2018-09-29
Remove IceControlled Attr from ICE keepalive

### LINKED ISSUE #187: ICE Agent improperly includes IceControlled/IceControlling values
keepaliveCandidate always assumes we are controlled, update to be a conditional.

Sean-Der:
This value is incorrect if agent is controlling, instead of fixing
remove instead. We don't need this value in the keepalive.

Resolves #187

### DIFF
--- a/pkg/ice/agent.go
+++ b/pkg/ice/agent.go
@@ -172,8 +172,6 @@ func (a *Agent) pingCandidate(local, remote Candidate) {
 func (a *Agent) keepaliveCandidate(local, remote Candidate) {
 	msg, err := stun.Build(stun.ClassIndication, stun.MethodBinding, stun.GenerateTransactionId(),
 		&stun.Username{Username: a.remoteUfrag + ":" + a.LocalUfrag},
-		&stun.IceControlled{TieBreaker: a.tieBreaker},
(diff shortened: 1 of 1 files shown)


pkg/ice/agent.go
```diff
@@ -172,7 +172,6 @@ func (a *Agent) pingCandidate(local, remote Candidate) {
 func (a *Agent) keepaliveCandidate(local, remote Candidate) {
 	msg, err := stun.Build(stun.ClassIndication, stun.MethodBinding, stun.GenerateTransactionId(),
 		&stun.Username{Username: a.remoteUfrag + ":" + a.LocalUfrag},
-		&stun.IceControlled{TieBreaker: a.tieBreaker},
 		&stun.Priority{Priority: uint32(local.GetBase().Priority(HostCandidatePreference, 1))},
```
backkem: Is the priority attribute needed or does Janus only require the authorization? Otherwise we may as well clean it up now.

Sean-Der: Priority wasn't needed either, update PR! Thanks for catching that.

Sean-Der: I double checked, and every other value was asserted by libnice though

PR #213, opened 2018-11-01
Allow Raw RTP input to RTCTrack

### LINKED ISSUE #212: RTCTrack should accept unmodified *rtp.Packet input
In some cases (e.g. with an existing RTP stream source), it is undesirable to force the input through a full cycle of packet -> sample -> packet. Instead, the RTCTrack implementation should accept "raw" *rtp.Packet input which is sent to networkManager directly.

mjmac:
Resolves #212

Support forwarding packets from existing RTP streams. Adds
the following new methods to RTCPeerConnection:
  * NewRawRTPTrack() -- expects a non-zero SSRC and accepts
    a stream of *rtp.Packet from an existing RTP stream
  * NewRTCSampleTrack() -- generates an SSRC and accepts
    a stream of media.RTCSample

Deprecates NewRTCTrack() in favor of NewRTCSampleTrack().

Exposes a new RawRTP channel on *RTCTrack instances which
accepts *rtp.Packet.

### DIFF
--- a/media.go
+++ b/media.go
@@ -20,4 +20,5 @@ type RTCTrack struct {
 	Codec       *RTCRtpCodec
 	Packets     <-chan *rtp.Packet
 	Samples     chan<- media.RTCSample
+	RawRTP      chan<- *rtp.Packet
 }
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -945,8 +945,7 @@ func (pc *RTCPeerConnection) addDataMediaSection(d *sdp.SessionDescription, midV
 	d.WithMedia(media)
 }
 
-// NewRTCTrack is used to create a new RTCTrack
-func (pc *RTCPeerConnection) NewRTCTrack(payloadType uint8, id, label string) (*RTCTrack, error) {
+func (pc *RTCPeerConnection) newRTCTrack(payloadType uint8, ssrc uint32, id, label string) (*RTCTrack, error) {
 	codec, err := pc.mediaEngine.getCodec(payloadType)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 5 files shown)


media.go
```diff
@@ -20,4 +20,5 @@ type RTCTrack struct {
 	Codec       *RTCRtpCodec
 	Packets     <-chan *rtp.Packet
 	Samples     chan<- media.RTCSample
+	RawRTP      chan<- *rtp.Packet
```
Sean-Der: Is it possible to have a `RTCSampleTrack` and `RTCPacketTrack` ?  Each can embed `RTCTrack` that way we get a little more compile time safety

rtcpeerconnection.go
```diff
@@ -956,31 +955,46 @@ func (pc *RTCPeerConnection) NewRTCTrack(payloadType uint8, id, label string) (*
...
+				}
+			}
+		}()
+		close(rawPackets)
+	} else {
+		// If SSRC is not 0, then we are working with an established RTP stream
+		// and need to accept raw RTP packets for forwarding.
+		go func() {
+			for {
+				p := <-rawPackets
```
Sean-Der: Can/should we copy the packets here?

It may confuse users if they try to send the same packet twice, it will become corrupted (because later we encrypt in place)

rtcpeerconnection.go
```diff
@@ -990,11 +1004,31 @@ func (pc *RTCPeerConnection) NewRTCTrack(payloadType uint8, id, label string) (*
...
+	return pc.newRTCTrack(payloadType, ssrc, id, label)
+}
+
+// NewRTCSampleTrack initializes a new *RTCTrack configured to accept media.RTCSample
+func (pc *RTCPeerConnection) NewRTCSampleTrack(payloadType uint8, id, label string) (*RTCTrack, error) {
+	return pc.newRTCTrack(payloadType, 0, id, label)
+}
+
+// NewRTCTrack is used to create a new RTCTrack -- NB: Deprecated, use NewRTCSampleTrack() instead
+func (pc *RTCPeerConnection) NewRTCTrack(payloadType uint8, id, label string) (*RTCTrack, error) {
```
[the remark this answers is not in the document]
mjmac: Done

media.go
mjmac: I played around with this. In my opinion it adds a lot of complexity to surrounding code without any corresponding benefit. We would need to change `RTCTrack` to an interface in order to be able to pass the two different track types to the same receivers, none of which actually care about where the input is coming from. I understand the concern about potential user confusion, but I think in practice the way this will be used is to get a new track initialized for either Sample or RTP input and then start feeding it very near by to the initialization code. Not sure if you saw it, but the new initializers actually close the inverse input channel, making it impossible to accidentally send the "wrong" type of input. It's the sort of error that would be discovered quickly in development rather than cropping up in weird bugs later.

rtcpeerconnection.go
mjmac: Yeah, this bit me. On the one hand, it would certainly be more convenient for users to not have to deal with it. On the other hand, if we put the code for copying here, it will always impose the copying overhead even when it's not necessary. One way to address this might be to create an SFU implementation which handles this efficiently (c.f. [http/server.go](https://golang.org/src/net/http/server.go#L800)) and explicitly.

mjmac:
I can update the SFU example. I'm not sure about removing SampleBuilder though. It's working code -- right? I'd be inclined to leave it unless it turns into a maintenance drag.

mjmac:
I implemented the SFU stuff in a separate PR (which builds on this one, sorry if it's confusing). I'd like to keep them isolated for easier review/feedback cycles. Hoping that we can get this one merged soon so that I don't have to maintain the stack.

media.go
Sean-Der: That sounds perfect to me! My only goal was to not have silent errors, so I am super happy with this

rtcpeerconnection.go
Sean-Der: Mind adding a comment to the godoc about needing to copy? Beyond that everything LGTM!

PR #219, opened 2018-11-07
Safer Event Callbacks

### LINKED ISSUE #218: Setting callbacks is potentially error prone
The API defines a number of callbacks for handling asynchronous events, e.g. RTCPeerConnection.OnDataChannel, RTCDataChannel.OnMessage, etc. These are intended to mirror the JS spec as closely as possible in order to enable tutorial portability and reduce cognitive burden on developers working in multiple environments.

As currently implemented however, the API requires users to first know about and then remember to handle locking for the callback assignments. The examples show this, but it's not immediately obvious as to why it's necessary. A proposed modification to the API would retain the JS spec names but make them public setter functions instead of struct fields. The actual struct fields for callbacks would become private to the implementation.

Another idea would be to supply the callbacks in the configuration struct, but this is different from the JS spec.

Examples:

```go
peerConnection := webrtc.New(...)

peerConnection.OnDataChannel(func(dc *webrtc.RTCDataChannel) {
    dc.OnOpen(func() {
        fmt.Printf("Opened DataChannel: %s-%d", dc.Label, dc.ID)
    })
})
```

The setter implementations would then take care of locking the RTCPeerConnection and RTCDataChannel instances, like so:

```go
func (pc *RTCPeerConnection) OnDataChannel(f func(*RTCDataChannel)) {
    pc.Lock()
    defer pc.Unlock()
    pc.onDataChannel = f
}

### PR DESCRIPTION
**Pull request by mjmac**
Resolves #218

Change Event Callback APIs to setter functions which take care of
locking so that users don't need to know about or remember
to do this.

### DIFF
--- a/rtcdatachannel.go
+++ b/rtcdatachannel.go
@@ -89,22 +89,65 @@ type RTCDataChannel struct {
 	// OnError             func()
 	// OnClose             func()
 
-	// Onmessage designates an event handler which is invoked on a message
-	// arrival over the sctp transport from a remote peer.
-	//
-	// Deprecated: use OnMessage instead.
-	Onmessage func(datachannel.Payload)
-
-	// OnMessage designates an event handler which is invoked on a message
-	// arrival over the sctp transport from a remote peer.
-	OnMessage func(datachannel.Payload)
-
-	// OnOpen designates an event handler which is invoked when
-	// the underlying data transport has been established (or re-established).
-	OnOpen func()
+	onMessageHandler func(datachannel.Payload)
+	onOpenHandler    func()
 
 	// Deprecated: Will be removed when networkManager is deprecated.
 	rtcPeerConnection *RTCPeerConnection
+}
+
+// OnOpen sets an event handler which is invoked when
+// the underlying data transport has been established (or re-established).
+func (d *RTCDataChannel) OnOpen(f func()) {
+	d.Lock()
+	defer d.Unlock()
+	d.onOpenHandler = f
+}
+
+func (d *RTCDataChannel) onOpen() (done chan struct{}) {
+	d.RLock()
+	hdlr := d.onOpenHandler
+	d.RUnlock()
+
+	done = make(chan struct{})
+	if hdlr == nil {
+		close(done)
+		return
+	}
+
+	go func() {
+		hdlr()
+		close(done)
+	}()
+
+	return
+}
+
+// OnMessage sets an event handler which is invoked on a message
+// arrival over the sctp transport from a remote peer.
+func (d *RTCDataChannel) OnMessage(f func(p datachannel.Payload)) {
+	d.Lock()
+	defer d.Unlock()
+	d.onMessageHandler = f
+}
+
+func (d *RTCDataChannel) onMessage(p datachannel.Payload) {
+	d.RLock()
+	hdlr := d.onMessageHandler
+	d.RUnlock()
+
+	if hdlr == nil || p == nil {
+		return
+	}
+	hdlr(p)
+}
+
+// Onmessage sets an event handler which is invoked on a message
+// arrival over the sctp transport from a remote peer.
+//
+// Deprecated: use OnMessage instead.
+func (d *RTCDataChannel) Onmessage(f func(p datachannel.Payload)) {
+	d.OnMessage(f)
 }
 
 // func (d *RTCDataChannel) generateID() error {
@@ -140,13 +183,4 @@ func (d *RTCDataChannel) Send(p datachannel.Payload) error {
 		return &rtcerr.UnknownError{Err: err}
 	}
 	return nil
-}
-
-func (d *RTCDataChannel) doOnOpen() {
-	d.RLock()
-	onOpen := d.OnOpen
-	d.RUnlock()
-	if onOpen != nil {
-		onOpen()
-	}
 }
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -100,20 +100,12 @@ type RTCPeerConnection struct {
 	// OnIceCandidateError        func() // FIXME NOT-USED
 	// OnSignalingStateChange     func() // FIXME NOT-USED
 
-	// OnIceConnectionStateChange designates an event handler which is called
-	// when an ice connection state is changed.
-	OnICEConnectionStateChange func(ice.ConnectionState)
-
 	// OnIceGatheringStateChange  func() // FIXME NOT-USED
 	// OnConnectionStateChange    func() // FIXME NOT-USED
 
-	// OnTrack designates an event handler which is called when remote track
-	// arrives from a remote peer.
-	OnTrack func(*RTCTrack)
-
-	// OnDataChannel designates an event handler which is invoked when a data
-	// channel message arrives from a remote peer.
-	OnDataChannel func(*RTCDataChannel)
+	onICEConnectionStateChangeHandler func(ice.ConnectionState)
+	onTrackHandler                    func(*RTCTrack)
+	onDataChannelHandler              func(*RTCDataChannel)
 
 	// Deprecated: Internal mechanism which will be removed.
 	networkManager *network.Manager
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

File: rtcdatachannel.go [hunk 1]
```diff
@@ -89,24 +89,61 @@ type RTCDataChannel struct {
...
+
+func (d *RTCDataChannel) onOpen() {
+	d.RLock()
+	defer d.RUnlock()
+	if d.onOpenFn == nil {
+		return
+	}
+
+	go func() {
+		d.onOpenFn()
```
backkem: I think this is a mini-race since this is running in a separate routine the defer statement on the original routine may have already released the lock and someone else could have called ``OnOpen`` in the mean time. I see two options:
1. Store ``d.onOpenFn`` in a variable and pass it to the new routine. This leaves a tiny change of calling an outdated ``onOpenFn``. Seems rather acceptable thought.
2. Only acquire the lock in the new routine. This however means we'll always spawn a routine, even when the callback func is nil.

rtcdatachannel.go
```diff
@@ -89,24 +89,61 @@ type RTCDataChannel struct {
...
+
+func (d *RTCDataChannel) onMessage(p datachannel.Payload) {
+	d.RLock()
+	defer d.RUnlock()
+	if d.onMessageFn == nil || p == nil {
+		return
+	}
+
+	go func(p datachannel.Payload) {
+		d.onMessageFn(p)
```
backkem: See mini-race remark.

rtcpeerconnection.go
```diff
@@ -233,6 +225,67 @@ func (pc *RTCPeerConnection) initConfiguration(configuration RTCConfiguration) e
...
+}
+
+func (pc *RTCPeerConnection) onDataChannel(dc *RTCDataChannel) {
+	pc.RLock()
+	defer pc.RUnlock()
+	if pc.onDataChannelFn == nil || dc == nil {
+		return
+	}
+
+	// Run this synchronously to allow setup done in onDataChannelFn()
```
backkem: In the past doing synchronous callbacks has caused problems by causing a deadlock when calling methods on the peer connection in the callback. E.g.: the peer connection lock is already locked by onDataChannel, any method called in the callback func that also tries to acquire the lock can cause a deadlock.
Maybe it is sufficient to run both the ``onDataChannelFn`` and ``onOpen`` after each other but in a new routine? This way they are called in the right order but still asynchronous. Alternatively, I think we need to at least release the lock when calling the callback funcs.

rtcpeerconnection.go
```diff
@@ -233,6 +225,67 @@ func (pc *RTCPeerConnection) initConfiguration(configuration RTCConfiguration) e
...
+
+func (pc *RTCPeerConnection) onTrack(t *RTCTrack) {
+	pc.RLock()
+	defer pc.RUnlock()
+	if pc.onTrackFn == nil || t == nil {
+		return
+	}
+
+	go func(t *RTCTrack) {
+		pc.onTrackFn(t)
```
backkem: See mini-race remark.

rtcpeerconnection.go
```diff
@@ -233,6 +225,67 @@ func (pc *RTCPeerConnection) initConfiguration(configuration RTCConfiguration) e
...
+
+func (pc *RTCPeerConnection) onICEConnectionStateChange(cs ice.ConnectionState) {
+	pc.RLock()
+	defer pc.RUnlock()
+	if pc.onICEConnectionStateChangeFn == nil {
+		return
+	}
+
+	go func(cs ice.ConnectionState) {
+		pc.onICEConnectionStateChangeFn(cs)
```
backkem: See mini-race remark.

rtcdatachannel.go
mjmac: Ah, yes. You're right, it is a race, and I like your first proposal to fix it.

rtcpeerconnection.go
mjmac: Yeah, doing it this way didn't give me a great feeling... The problem with doing it async though is that there's a race between kicking off the supplied onDataChannelFn and receiving the next DataChannelEvent in dataChannelEventHandler() -- if that event is a message, and the OnMessage handler hasn't been set yet, the message is missed. I'll think about it some more, because this does feel awfully kludgy.

mjmac: I'm entirely willing to believe that I'm thinking about this the wrong way... But I'm having a hard time figuring out how to achieve my goal of ensuring that data channel operations are serialized, at least for setup.

The code I'm trying to support looks like this:
```go
	pc.OnDataChannel(func(dc *webrtc.RTCDataChannel) {
		logs.Infof("New data channel %s-%d", dc.Label, dc.ID)
		dc.OnOpen(func() {
			logs.Infof("Opened data channel %s-%d", dc.Label, dc.ID)
		})

		dc.OnMessage(func(payload datachannel.Payload) {
			logs.Debugf("Got datachannel message: %#v", payload)
			switch p := payload.(type) {
			case *datachannel.PayloadBinary:
				var msg ChannelMessage
				if err := json.Unmarshal(p.Data, &msg); err != nil {
					logs.Errorf("Failed to unmarshal %s into ChannelMessage: %s", p.Data, err)
				}
				prx.handleChannelMessage(client, pc, dc, &msg)
			case *datachannel.PayloadString:
				logs.Noticef("Got string datachannel message: %s", p.Data)
			default:
				logs.Errorf("Unhandled datachannel message: %s-%s", dc.Label, p.PayloadType())
			}
		})
	})
```

So the issue is that without the enforced serialization it's possible (likely, in my testing) that a data channel message comes in before OnMessage is set.

Another kludge that occurs to me is to do something like this:
```go
func (pc *RTCPeerConnection) onDataChannel(dc *RTCDataChannel) chan struct{} {
	pc.RLock()
	defer pc.RUnlock()
	done := make(chan struct{})
	if pc.onDataChannelHandler == nil || dc == nil {
                close(done)
		return done
	}

	// Run this synchronously to allow setup done in onDataChannelFn()
	// to complete before datachannel event handlers might be called.
	go func(f func(*RTCDataChannel), dc *RTCDataChannel) {
		f(dc)
		close(done)
		dc.onOpen()
	}(pc.onDataChannelHandler, dc)

	return done
}
```

And then callers which care about synchronization (only dataChannelEventHandler() for now) can wait for the channel to close before proceeding.

```go
func (pc *RTCPeerConnection) dataChannelEventHandler(e network.DataChannelEvent) {
	switch event := e.(type) {
	case *network.DataChannelCreated:
		id := event.StreamIdentifier()
		newDataChannel := &RTCDataChannel{ID: &id, Label: event.Label, rtcPeerConnection: pc, ReadyState: RTCDataChannelStateOpen}
		pc.Lock()
		pc.dataChannels[e.StreamIdentifier()] = newDataChannel
		pc.Unlock()

		<-pc.onDataChannel(newDataChannel)
...
}
```

mjmac: I'm gonna run with this idea for now... You guys let me know if it's hideous or if better alternatives present themselves to you.

mjmac: After some discussion in the slack channel, I decided to update all of the on* wrappers to return a done channel which callers can check if they care about waiting. This is internal-only, not exposed via public API. It just gives us more flexibility for handling things behind the scenes.

backkem:
Feedback from Slack: The callback mechanism should preserve the order of the messages. Let's take this into account in this PR.

backkem:
Dang, Now I feel like we're just over-engineering this. If we go back to basics for a moment, can we come up with a good reason why this doesn't suffice:
```go
d.RLock()
hdlr := d.onEvent
d.RUnlock()

if hdlr != nil {
    hdlr()
}
```
The good:
1. It's simple.
1. Things are always ordered.

Trade-off:
1. There is a tiny change hdlr() would be outdated. This seems acceptable and this also exists in the current solution.
1. This is blocking but the implementer can add buffering/fan-out/... if they really want.
1. If we ever need buffering in the lib it seems easier to add this between some of the networking layers rather than in these callbacks.
1. A small caveat is that you can't do something blocking in the callback that relies on whatever triggers the callback. E.g. trying to block inside a OnMessage callback until another message arrives.

@Sean-Der @mjmac What do you guys think?

mjmac:
@backkem The problem with making it blocking is that it can introduce deadlocks. The TestRTCDataChannel_EventHandlers test, for example, will deadlock. I could of course write to the out channel in a goroutine in my test handler, but I'm concerned that making this handler synchronous when all of the others are async may cause other problems. Maybe not, I'll defer to you guys as you've been working on this codebase for a lot longer than I have.

Likewise, if I simply run the handler in a new goroutine, e.g.
```go
func (d *RTCDataChannel) onMessage(p datachannel.Payload) {
	d.RLock()
	hdlr := d.onMessageHandler
	d.RUnlock()

	if hdlr == nil || p == nil {
		return
	}
	go func() {
		hdlr(p)
	}()
}
```

Then the TestRTCDataChannel_MessagesAreOrdered test fails miserably with everything out of order. Maybe there's a simpler way to make this work -- I'm totally willing to believe that I over engineered this. ;)

mjmac:
I redid the PR with the simpler approach by making onMessage blocking. I think my concern is probably unfounded, because it would be pretty weird for users to be doing stuff with channels in that handler.

backkem:
``TestRTCDataChannel_EventHandlers`` does seem a bit weird. It could easily be amended by closing the chan instead of sending a bool. Note that you're easily able to create a deadlock in this case because you can manually call the event handler. In a normal case only the library can call the callback, not the user. It would only be possible to create a deadlock when some public API needs the same thread that is calling the callback. I'm not 100% sure this won't happen. However, it feels advisable to try the simple approach first in this case. Otherwise I think we're sneaking a lot of complexity in that may hurt us down the line. Don't hesitate to push back if you think I'm missing something. Do you see any reason that any of the callbacks have to be async?

Lastly, I think we still need to copy the func to a variable while holding the lock and calling the function via that variable after releasing the lock. Otherwise we're accessing d.onEvent without holding the lock which is racy/undefined behavior.

Sean-Der:
Hey @mjmac thanks for all the work on this :) would you mind creating a new tag, just want to make sure we don't bust anything/have an old version people can pin against?

Since we are breaking the API can you also delete https://github.com/pions/webrtc/blob/master/rtcdatachannel.go#L92-L96 and https://github.com/pions/webrtc/blob/master/media.go#L8-L11 just rip the whole band-aid off :) 

@maxmcd and @calebdoxsey just a heads up about the API change! If there is anything we can do to make this less painful, I am all ears

PR #221, opened 2018-11-07
Simple Trickle-ICE support

### LINKED ISSUE #220: Simple Trickle-ICE support
Not entirely sure how much more needs to happen in order to fully support this. I've been maintaining a simple solution to add remote candidates as they trickle in; figured I'd share it as a PR to see what the delta is from my hack to a full solution.

mjmac:
Resolves #220

Adds remote candidates to agent asynchronously. Changes remote
candidates slice to a map keyed by candidate string in order to
avoid duplicate entries.

### DIFF
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -480,9 +480,15 @@ func (pc *RTCPeerConnection) RemoteDescription() *RTCSessionDescription {
 	return pc.CurrentRemoteDescription
 }
 
-// func (pc *RTCPeerConnection) addIceCandidate() {
-// 	panic("not implemented yet") // FIXME NOT-IMPLEMENTED nolint
-// }
+// AddIceCandidate accepts an ICE candidate string and adds it
+// to the existing set of candidates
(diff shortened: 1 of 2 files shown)


rtcpeerconnection.go
```diff
@@ -480,9 +480,15 @@ func (pc *RTCPeerConnection) RemoteDescription() *RTCSessionDescription {
 	return pc.CurrentRemoteDescription
 }
 
-// func (pc *RTCPeerConnection) addIceCandidate() {
-// 	panic("not implemented yet") // FIXME NOT-IMPLEMENTED nolint
-// }
+// AddIceCandidate accepts an ICE candidate string and adds it
+// to the existing set of candidates
+func (pc *RTCPeerConnection) AddIceCandidate(s string) {
```
Sean-Der: Mind updating the signature to `func (pc *RTCPeerConnection) AddIceCandidate(s string) error`

Then instead of printf return an error

mjmac: Ah, yeah. That makes more sense, done.

PR #223, opened 2018-11-09
[WIP] RTCPeerConnection Logging

### LINKED ISSUE #222: Add logging to RTCPeerConnection
RTCPeerConnection should have a logger to debug SDP/Packets and to log other messages at varying verbose levels. RTCPeerConnection should accept an interface so that a user can customize a logger and pass the logger to RTCPeerConnection through a setter. This logging implementation can begin with a simple default logger that will replace all the fmt.print and fmt.printf messages that are currently in rtcpeerconnection.go.

aguilEA:
Resolves #222

### DIFF
--- a/pkg/logging/logging.go
+++ b/pkg/logging/logging.go
@@ -1,0 +1,91 @@
+package logging
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+type logLevel int
+
+const(
+	Error = iota + 1
+	Warn
+	Info
+	Debug
+	Trace
+)
+
+// RTCPeerLogger is the default logging struct for rtcpeerconnection.
+type RTCPeerLogger struct{
+	info, debug, warning,
+	trace, error  *log.Logger
+}
+
+// webrtcLogger is the interface that must be implemented if a custom
+// logger is used.
+type webrtcLogger interface{
+	Info(msg string)
+	Trace(msg string)
+	Debug(msg string)
+	Warning(msg string)
+	Error(msg string)
+}
+
+func (pclog *RTCPeerLogger) Debug(debug string){
+	pclog.debug.Println(debug)
+}
+
+func (pclog *RTCPeerLogger) Error(error string){
+	pclog.error.Println(error)
+}
+
+func (pclog *RTCPeerLogger) Info(info string){
+	pclog.info.Println(info)
+}
+
+func (pclog *RTCPeerLogger) Warning(warning string){
+	pclog.warning.Println(warning)
+}
+
+func (pclog *RTCPeerLogger) Trace(trace string){
+	pclog.trace.Println(trace)
+}
+
+func (pclog *RTCPeerLogger) SetLogLevel(l logLevel) {
+
+	debugOut := ioutil.Discard
+	infoOut := ioutil.Discard
+	traceOut := ioutil.Discard
+	warningOut := ioutil.Discard
+	errorOut := ioutil.Discard
+
+	switch l {
+	case Error:
+		errorOut = os.Stdout
+	case Warn:
+		errorOut = os.Stdout
+		warningOut = os.Stdout
+	case Info:
+		errorOut = os.Stdout
+		warningOut = os.Stdout
+		infoOut = os.Stdout
+	case Debug:
+		errorOut = os.Stdout
+		warningOut = os.Stdout
+		infoOut = os.Stdout
+		debugOut = os.Stdout
+	case Trace:
+		errorOut = os.Stdout
+		warningOut = os.Stdout
+		infoOut = os.Stdout
+		debugOut = os.Stdout
+		traceOut = os.Stdout
+	}
+
+	pclog.debug   = log.New(debugOut, "DEBUG: ", log.Lshortfile)
+	pclog.info    = log.New(infoOut,  "INFO: ", log.Lshortfile)
+	pclog.trace   = log.New(traceOut, "TRACE: ", log.Lshortfile)
+	pclog.warning = log.New(warningOut,"WARNING: ", log.Lshortfile)
+	pclog.error   = log.New(errorOut, "ERROR: ", log.Lshortfile)
+}
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -6,6 +6,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"fmt"
+	"github.com/pions/webrtc/pkg/logging"
 	"net"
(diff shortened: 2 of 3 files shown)


rtcpeerconnection.go
```diff
@@ -27,6 +29,22 @@ import (
 // comparisons when no value was defined.
 const Unknown = iota
 
+// RTCPeerLogger is the default logging struct for rtcpeerconnection.
```
Sean-Der: I would move `RTCPeerLogger` + `webrtcLogger` to a dedicated package

This way other packages can import/use the logger (like `github.com/pions/internal/sctp`) right now we can't pass our logger instance down to other packages.

rtcpeerconnection.go
```diff
@@ -27,6 +29,22 @@ import (
...
+type RTCPeerLogger struct{
+	INFO, DEBUG, WARNING,
```
Sean-Der: I would make these private, otherwise there might be confusion and people could call them directly.

rtcpeerconnection.go
```diff
@@ -145,8 +166,11 @@ func New(configuration RTCConfiguration) (*RTCPeerConnection, error) {
 		mediaEngine:        DefaultMediaEngine,
 		sctpTransport:      newRTCSctpTransport(),
 		dataChannels:       make(map[uint16]*RTCDataChannel),
+		pclog: &RTCPeerLogger{},
 	}
 
+	pc.pclog.setupLogger(1)
```
Sean-Der: This is a FANTASTIC idea, nice work!

This way users of the library can switch logging levels at anytime on the fly (and not even supply their own logger) I would change the name though so people know they can call it at anytime though `setLogLevel` maybe?

rtcpeerconnection.go
```diff
@@ -762,6 +786,46 @@ func (pc *RTCPeerConnection) Close() error {
...
+
+func (pclog *RTCPeerLogger) Warning(warning string){
+	pclog.WARNING.Println(warning)
+}
+
+func (pclog *RTCPeerLogger) Trace(trace string){
+	pclog.TRACE.Println(trace)
+}
+
+func (pclog *RTCPeerLogger) setupLogger(logLevel int) {
```
Sean-Der: Can you make `logLevel` an enum?

Users might be confused by just passing an integer. Would be nice if they could pass `Trace` `Debug` `Info` or `Warning`

rtcpeerconnection.go
aguilEA: Placed them in their own package pkg/logging/. The interface and struct names seem slightly unintuitive. Let me know if you have a better naming convention.

rtcpeerconnection.go
aguilEA: Okay so I changed the function name to setLogLevel and it accepts a logging level enum.  I am increasing the number of loggers that can be used according to the logging level, but if you think that the logging level should just be the type of logger you can use I can change that.

mjmac:
@aguilEA: If you rebase your changes on master and then force-push, it will make it easier to review this patch. As it is now, your patch changes are mixed in with all of the updates that have happened since you first submitted this.

pkg/logging/logging.go
```diff
@@ -0,0 +1,91 @@
...
+
+// RTCPeerLogger is the default logging struct for rtcpeerconnection.
+type RTCPeerLogger struct{
+	info, debug, warning,
+	trace, error  *log.Logger
+}
+
+// webrtcLogger is the interface that must be implemented if a custom
+// logger is used.
+type webrtcLogger interface{
```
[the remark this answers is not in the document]
aguilEA: I was setting it a loging interface in order to have method in RTCPeerConnection which accepts the interface, and the user can pass a struct that allows them to do logging however they want. Ultimately a user would call a function pc.setLogger(myCustomLogger).

pkg/logging/logging.go
```diff
@@ -0,0 +1,91 @@
...
+	Info(msg string)
+	Trace(msg string)
+	Debug(msg string)
+	Warning(msg string)
+	Error(msg string)
+}
+
+func (pclog *RTCPeerLogger) Debug(debug string){
+	pclog.debug.Println(debug)
```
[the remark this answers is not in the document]
aguilEA: Yeah I read "Discard is an io.Writer on which all Write calls succeed without doing anything." and thought that it was not affecting performace. However if this is not the case then yeah we don't want to impact performance. This is my first time messing with the go logger, but if we can set a level in the logger and use that to determine what to print then that sounds good.

pkg/logging/logging.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (pclog *RTCPeerLogger) Warning(warning string){
+	pclog.warning.Println(warning)
+}
+
+func (pclog *RTCPeerLogger) Trace(trace string){
+	pclog.trace.Println(trace)
+}
+
+func (pclog *RTCPeerLogger) SetLogLevel(l logLevel) {
```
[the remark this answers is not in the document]
aguilEA: In the case that we do use levels for loging are you suggesting that we create logers like this?


`
Tracelog := logging.NewLeveledLogger().
      WithLogLevel(logging.LogLevelTrace).
      WithDebugLogger(log.New(os.Stderr, "trace ", log.Lmicroseconds|log.Llongfile|log.LUTC))

Debuglog := logging.NewLeveledLogger().
      WithLogLevel(logging.LogLevelDebug).
      WithDebugLogger(log.New(os.Stderr, "debug ", log.Lmicroseconds|log.Llongfile|log.LUTC))
`

Sean-Der:
@aguilEA  are you still interested in working on this? No rush (would much rather have outside contributors) but would love to see this implemented!

I think this features lends to some bikeshedding, I think we could figure out a minimal version and really should shoot to get it written/merged in under 3 days :) I hate to see peoples work ignored

aguilEA:
Hey yeah I've also been busy with work to have some time to work on it.
However I could work on it this coming week.
I'm not sure if @mjmac is already working on this on his own.
We could work on this together or he can finish it if he already began.

Sean-Der:
This landed with https://github.com/pions/webrtc/commit/b98909a75b254ac58a6be834bde1a5a68cbfe739

This is a great start, but would love to continue to see improvements to the logging if you are interested @aguilEA 

We should right a lint that doesn't allow `fmt.Println` or `fmt.Printf` and covert them all to the logging stuff (and other things like that)

aguilEA:
That is great! Yeah I'll help with what I can. I'll familiarize myself
with @mjmac's implementation and look into lint.

PR #228, opened 2018-11-16
Signaling state

mjmac:
Still a WIP, submitting for early feedback. This should provide a better foundation for implementing renegotiation support.

### DIFF
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -98,11 +98,11 @@ type RTCPeerConnection struct {
 	// OnNegotiationNeeded        func() // FIXME NOT-USED
 	// OnIceCandidate             func() // FIXME NOT-USED
 	// OnIceCandidateError        func() // FIXME NOT-USED
-	// OnSignalingStateChange     func() // FIXME NOT-USED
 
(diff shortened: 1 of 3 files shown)


backkem:
The approach looks good to me. We should finish #219 first so we can use the final callback approach here as well.

rtcpeerconnection.go
```diff
@@ -484,18 +517,148 @@ func (pc *RTCPeerConnection) CreateAnswer(options *RTCAnswerOptions) (RTCSession
...
+	cur := pc.SignalingState
+	setLocal := rtcStateChangeOpSetLocal
+	setRemote := rtcStateChangeOpSetRemote
+	newSdpDoesNotMatchOffer := &rtcerr.InvalidModificationError{Err: errors.New("New sdp does not match previous offer")}
+	newSdpDoesNotMatchAnswer := &rtcerr.InvalidModificationError{Err: errors.New("New sdp does not match previous answer")}
+
+	var nextState RTCSignalingState
+	var err error
+	switch op {
+	case setLocal:
```
Sean-Der: Oh hell yes, thanks for adding all this error handling around signaling!

We have gotten so many issues around this, nice to fix them up.

PR #233, opened 2018-11-19
Use the latest releases

cristaloleg:
Should we also run on `1.10.x`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: go
 
 go:
-  - "1.9.x"
-  - "1.11.x"
+  - "1.x" # use the latest Go release
 
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


Sean-Der:
Hey @cristaloleg thanks for the PR!

I am tempted to ONLY run on 1.11.x, I think multiple runs is breaking coveralls (coverage is really double that)

Do you know what best practices/other projects are doing? I don't want to slow down testing too much (but at the same time not cripple testing)

cristaloleg:
Well, some projects want to have support for old Go versions(sadly not everyone does a migration to the new version), but having previous version supported is treated as a good behaviour.

I doubt that Pion uses new APIs from 1.11 that aren't in 1.10, so having last version looks safe for me.

Sean-Der:
Fixed with 55f1d7d78ae5680c209b021b9d398634f13e1738 thanks for the improvement @cristaloleg

Thanks for joining Slack, and looking forward to future contributions :)

PR #236, opened 2018-11-19
Add Leveled/Scoped logging facility

mjmac:
Not sure about this yet, going to live with it for a bit locally to see how it feels. Updated the Ice agent with debug logging as an example.

### DIFF
--- a/pkg/logging/scoped.go
+++ b/pkg/logging/scoped.go
@@ -1,0 +1,105 @@
+package logging
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+)
+
+type loggerRegistry struct {
+	sync.RWMutex
+	scopeLoggers map[string]*LeveledLogger
+	scopeLevels  map[string]LogLevel
+}
+
+var (
+	registry = &loggerRegistry{
+		scopeLoggers: make(map[string]*LeveledLogger),
+		scopeLevels:  make(map[string]LogLevel),
+	}
+)
+
+// SetLogLevelForScope sets the logging level for the given
+// scope, or all scopes if "all" is provided. If a logger
+// for the scope does not yet exist, the desired logging
+// level is recorded and applied when the scoped logger
+// is created.
+func SetLogLevelForScope(scope string, level LogLevel) {
+	registry.Lock()
+	defer registry.Unlock()
+
+	scope = strings.ToLower(scope)
+	registry.scopeLevels[scope] = level
+
+	if scope == "all" {
+		for _, logger := range registry.scopeLoggers {
+			logger.SetLevel(level)
+		}
+		return
+	}
+
+	if logger, found := registry.scopeLoggers[scope]; found {
+		logger.SetLevel(level)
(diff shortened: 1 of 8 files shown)


Sean-Der:
@mjmac are you still interested in working on this? No rush (would much rather have outside contributors) but would love to see this implemented!

mjmac:
Hi!

Yes, sorry... I've gotten pulled off onto other things for the moment. I should have some downtime this coming week to clean this up and rework it. I think we can combine the approaches @aguilEA and I took to make something useful pretty quickly. I still think that we don't need a full-featured user-facing logging solution, because user-focused logging should ideally happen in the callbacks. But I think we can start with something which is better than nothing and iterate from there.

Sean-Der:
@mjmac no worries at all! We all have full time jobs, and with the holidays spare time is pretty hard to come by :)

Totally agree though, I would love to merge an MVP just so we don't have `fmt.Println` and `fmt.Printf` sprinkled through the library. Users get concerned with the tracing messages we have, and would make development so much easier as well.

no rush

mjmac:
OK, feeling pretty good about this now. Trying to strike a balance between configurability and convention. The most important thing to me is that we should be able to sprinkle low-level logging statements throughout the code without worrying about performance impact when we're not actively debugging.

pkg/logging/scoped.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+		}
+		if level > LogLevelDisabled {
+			registry.loggers[scope].SetLevel(level)
+		}
+	}
+	return registry.loggers[scope]
+}
+
+func init() {
```
Sean-Der: @mjmac @backkem 

What do you think of using the settings engine to init this instead? https://github.com/pions/webrtc/blob/master/settingengine.go

I see this as the our way of setting `non-standard WebRTC things`, also makes it easier to demo in our `examples/` I would have no problem parsing args/env variables in the examples themselves though

wdouglass: one pro-environment-variable argument is this: Gstreamer (and glib) do it, so people used to working with multimedia packages from these libraries will expect environment variables (or at least it's a sideways move)

As long as it's well documented...

mjmac: I don't love it, to be honest. It doesn't feel very ergonomic from a developer perspective. Environment variables have a long history of use for this sort of thing on unix-like platforms. It would be really annoying to me to have to fiddle with a configuration file every time I wanted to tweak logging while I was hacking on something.

That having been said, I'm open to the idea of a compromise that enables both use cases. I don't mind making the internals a bit more complicated in order to make life easier for users and developers.

How would the settings engine work for this? If you guys would propose examples of the way you would like to be able to configure things I can figure out how to implement it.

mjmac:
OK, I can merge it as-is and we can iterate some more as we find out how we want to use it.

I wonder if we should just put this in its own separate package, though. There's nothing webrtc-specific about it. e.g. pions/logging?

PR #243, opened 2018-11-28
Linter Updates

maxhawkins:
* Make commit message linter less annoying
* Add git hooks to auto-run lint scripts before commit and push

### DIFF
--- a/.github/hooks/pre-commit.sh
+++ b/.github/hooks/pre-commit.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+# Redirect output to stderr.
+exec 1>&2
+
+.github/lint-disallowed-functions-in-library.sh
--- a/.github/hooks/pre-push.sh
+++ b/.github/hooks/pre-push.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+.github/assert-contributors.sh
+
+exit 0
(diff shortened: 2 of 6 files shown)


Sean-Der:
This is really great, thanks @maxhawkins IMO it is really important that we encourage new contributors. That is really the 'spirit' of open source, nothing worse then building an ivory tower.

Just so we are all in agreement, what do you think of dropping the `one line change`? I think the commit message may frustrate people, but in the long run make them better contributors! I agree that creating issues is spammy, I think we should still push for it when it makes sense though (but totally subjective)

I will also make it so when opening a new PR/Issue we have a template (and we can list helpful stuff) I see that on other projects so shouldn't be too hard!

.github/hooks/pre-commit.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+        against=HEAD
```
Sean-Der: What is the purpose of setting `against` ?

.github/hooks/pre-push.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -e
+
+remote="$1"
```
Sean-Der: Do we need the `remote` + `url`?

PR #246, opened 2018-11-29
Add Go modules

### LINKED ISSUE #244: Dependency management
Since the project is becoming bigger and has multiple repositories it may be a good time to start thinking about dependency management. It seems that go modules is getting decent support from the go team and community. How do we feel about adopting this strategy?

backkem:
It seems you do lose a tiny bit of freedom since you now have to run go install from the project root (or it's children).

TODO:
- [ ] Update the CI to use the modules.

Resolves #244


Sean-Der:
Will it break anything if we remove the dependencies that are brought in from the examples? It would be nice if we could drop these.
```
	github.com/davecgh/go-spew v1.1.1 // indirect
	github.com/google/go-cmp v0.2.0 // indirect
	github.com/gorilla/websocket v1.4.0 // indirect
	github.com/notedit/janus-go v0.0.0-20180821162543-a152adf0cb7b
```

Also I don't know how much value we are really getting from `github.com/pkg/errors`

I will also create a `v1.0.0` tag for `github.com/pions/pkg`

I need to read up, but do you know if `gomod` bring in the newest dot release? So we can do non-API changes, and only bump `pkg` or `dtls` when really needed.

backkem:
I changed the `janus-gatway` example into it's own module. This avoids the dependencies in the root module. Note that  `janus-gatway` depends on `pions/pkg` because we use ice in the public API. It would be nice to avoid that.

I agree we should minimize the test dependencies. They only provide us with minimal sugar.

You can bump the version of a dependency by manually changing the `go.mod` file or using [go get](https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies).

PR #250, opened 2018-11-30
Handle compound SRTCP messages

### LINKED ISSUE #249: Compound SRTCP messages aren't parsed properly
Chrome sends RR messages with SLI messages packed into it's encrypted payload -- this is legal per the SRTCP standard; lets not throw out the extra messages.

wdouglass:
Make sure any extra packets packed into an
encrypted payload are properly handled
Resolves #249

### DIFF
--- a/internal/network/port-receive.go
+++ b/internal/network/port-receive.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"io"
 	"net"
 
 	"github.com/pions/webrtc/internal/dtls"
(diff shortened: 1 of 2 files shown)


internal/network/port_test.go
```diff
@@ -0,0 +1,115 @@
...
+	0x85, 0xcd, 0x0, 0x2,
+	// sender=0x902f9e2e
+	0x90, 0x2f, 0x9e, 0x2e,
+	// media=0x902f9e2e
+	0x90, 0x2f, 0x9e, 0x2e,
+}
+
+func TestHandleRtcp(t *testing.T) {
+	pktChannel := make(chan rtcp.Packet, 6)
+	gbtStub := func(uint32) *TransportPair {
```
maxhawkins: I didn't know what this variable referred to without looking it up above. Maybe `getBufStub` or something would be clearer?

internal/network/port-receive.go
```diff
@@ -19,6 +20,62 @@ type incomingPacket struct {
...
 }
 
+func handleRTCP(getBufferTransports func(uint32) *TransportPair, buffer []byte) {
+	//decrypted packets can also be compound packets, so we have to nest our reader loop here.
+	compoundPacket := rtcp.NewReader(bytes.NewReader(buffer))
+	for {
+		header, rawrtcp, err := compoundPacket.ReadPacket()
+
+		if err != nil {
+			if err == io.EOF {
```
maxhawkins: Putting this on the previous line would mean less nesting (easier to read)

wdouglass: I get that, but I'd like to leave it, so that in the not-eof case, the error will get printed and we'll return. If i compose the two conditionals, then i'd need a second branch anyway.

internal/network/port_test.go
wdouglass: Changed. Thanks for the review!

PR #258, opened 2018-12-04
Support Extra RTCP messages

wdouglass:
This also avoids a type switch in srtp.go (which may be a bit faster)
I've also added some String() members to aid in debugging

Sorry for the big commit, i can break it up if you guys want that.

### DIFF
--- a/pkg/rtcp/packet.go
+++ b/pkg/rtcp/packet.go
@@ -35,9 +35,9 @@ func Unmarshal(rawPacket []byte) (Packet, Header, error) {
 
 	case TypeTransportSpecificFeedback:
 		switch h.Count {
-		case tlnFMT:
+		case FormatTLN:
 			p = new(TransportLayerNack)
-		case rrrFMT:
+		case FormatRRR:
 			p = new(RapidResynchronizationRequest)
 		default:
 			p = new(RawPacket)
@@ -45,9 +45,9 @@ func Unmarshal(rawPacket []byte) (Packet, Header, error) {
 
 	case TypePayloadSpecificFeedback:
 		switch h.Count {
-		case pliFMT:
+		case FormatPLI:
 			p = new(PictureLossIndication)
-		case sliFMT:
+		case FormatSLI:
 			p = new(SliceLossIndication)
 		default:
 			p = new(RawPacket)
--- a/pkg/rtcp/packet.go
+++ b/pkg/rtcp/packet.go
@@ -3,6 +3,8 @@ package rtcp
 // Packet represents an RTCP packet, a protocol used for out-of-band statistics and control information for an RTP session
 type Packet interface {
 	Header() Header
+	// DestinationSSRC returns an array of SSRC values that this packet refers to.
+	DestinationSSRC() []uint32
 
 	Marshal() ([]byte, error)
 	Unmarshal(rawPacket []byte) error
@@ -32,10 +34,24 @@ func Unmarshal(rawPacket []byte) (Packet, Header, error) {
 		p = new(Goodbye)
 
 	case TypeTransportSpecificFeedback:
-		p = new(RapidResynchronizationRequest)
+		switch h.Count {
+		case tlnFMT:
+			p = new(TransportLayerNack)
+		case rrrFMT:
+			p = new(RapidResynchronizationRequest)
+		default:
+			p = new(RawPacket)
+		}
 
 	case TypePayloadSpecificFeedback:
-		p = new(PictureLossIndication)
+		switch h.Count {
+		case pliFMT:
+			p = new(PictureLossIndication)
+		case sliFMT:
+			p = new(SliceLossIndication)
+		default:
+			p = new(RawPacket)
+		}
 
 	default:
 		p = new(RawPacket)
--- a/pkg/rtcp/packet.go
+++ b/pkg/rtcp/packet.go
@@ -2,6 +2,8 @@ package rtcp
 
 // Packet represents an RTCP packet, a protocol used for out-of-band statistics and control information for an RTP session
 type Packet interface {
+	Header() Header
+
 	Marshal() ([]byte, error)
 	Unmarshal(rawPacket []byte) error
 }
--- a/pkg/rtcp/packet.go
+++ b/pkg/rtcp/packet.go
@@ -36,10 +36,9 @@ func Unmarshal(rawPacket []byte) (Packet, Header, error) {
 		p = new(PictureLossIndication)
 
 	default:
-		return nil, h, errWrongType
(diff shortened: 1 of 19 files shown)


wdouglass:
@maxhawkins i accidentally dragged a couple of your commits from rtcp-api along during my rebase, but i think it makes sense to leave them here. What do you think?

pkg/rtcp/receiver_report.go
```diff
@@ -67,7 +72,10 @@ func (r ReceiverReport) Marshal() ([]byte, error) {
 		return nil, errTooManyReports
 	}
 
+	rawPacket = append(rawPacket, r.ProfileExtensions...)
```
maxhawkins: I can't find the right section in the spec... do these extensions contribute toward the Length in the packet header? If so you'll have to update `len()`

pkg/rtcp/receiver_report_test.go
```diff
@@ -43,6 +43,7 @@ func TestReceiverReportUnmarshal(t *testing.T) {
 					LastSenderReport:   0x9f36432,
 					Delay:              150137,
 				}},
+				ProfileExtensions: []uint8{},
```
maxhawkins: Why `uint8` instead of `byte`?

pkg/rtcp/slice_loss_indication.go
```diff
@@ -0,0 +1,115 @@
...
+
+// The SliceLossIndication packet informs the encoder about the loss of a picture slice
+type SliceLossIndication struct {
+	// SSRC of sender
+	SenderSSRC uint32
+
+	// SSRC of the media source
+	MediaSSRC uint32
+
+	SLI []struct {
```
maxhawkins: It might be helpful to make this an exported struct type instead of an anonymous one.

It's difficult to, for instance, maintain a list of SLIs if there's no type name.

pkg/rtcp/slice_loss_indication.go
```diff
@@ -0,0 +1,115 @@
...
+func (p SliceLossIndication) Marshal() ([]byte, error) {
+
+	if len(p.SLI)+sliLength > math.MaxUint8 {
+		return nil, errTooManyReports
+	}
+
+	rawPacket := make([]byte, sliOffset+(len(p.SLI)*4))
+	binary.BigEndian.PutUint32(rawPacket, p.SenderSSRC)
+	binary.BigEndian.PutUint32(rawPacket[4:], p.MediaSSRC)
+	for i := 0; i < len(p.SLI); i++ {
```
maxhawkins: Just as fast, more idiomatic to use `for i, s := range p.SLI` here

pkg/rtcp/slice_loss_indication.go
```diff
@@ -0,0 +1,115 @@
...
+		sli := ((uint32(p.SLI[i].First) & 0x1FFF) << 19) |
+			((uint32(p.SLI[i].Number) & 0x1FFF) << 6) |
+			(uint32(p.SLI[i].Picture) & 0x3F)
+		binary.BigEndian.PutUint32(rawPacket[sliOffset+(4*i):], sli)
+	}
+	h := Header{
+		Count:  tlnFMT,
+		Type:   TypeTransportSpecificFeedback,
+		Length: uint16(tlnLength + len(p.SLI)),
```
maxhawkins: Does this include the header length? Why not reuse the result from `Header()`?

pkg/rtcp/slice_loss_indication.go
```diff
@@ -0,0 +1,115 @@
+package rtcp
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+)
+
+// The SliceLossIndication packet informs the encoder about the loss of a picture slice
+type SliceLossIndication struct {
```
maxhawkins: No tests?

pkg/rtcp/packet.go
```diff
@@ -3,6 +3,8 @@ package rtcp
 // Packet represents an RTCP packet, a protocol used for out-of-band statistics and control information for an RTP session
 type Packet interface {
 	Header() Header
+	// DestinationSSRC returns an array of SSRC values that this packet refers to.
+	DestinationSSRC() []uint32
```
maxhawkins: Great idea, this is super useful.

pkg/rtcp/transport_layer_nack.go
```diff
@@ -0,0 +1,114 @@
...
+
+// The TransportLayerNack packet informs the encoder about the loss of a transport packet
+type TransportLayerNack struct {
+	// SSRC of sender
+	SenderSSRC uint32
+
+	// SSRC of the media source
+	MediaSSRC uint32
+
+	Nacks []struct {
```
maxhawkins: Same as above: it might make sense to export this struct as a type instead of using an anonymous struct.

pkg/rtcp/transport_layer_nack.go
```diff
@@ -0,0 +1,114 @@
+package rtcp
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+)
+
+// The TransportLayerNack packet informs the encoder about the loss of a transport packet
+type TransportLayerNack struct {
```
maxhawkins: Can you link to the spec or copy some of the diagrams here? It would be convenient to have while reading.

pkg/rtcp/transport_layer_nack.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// SSRC of the media source
+	MediaSSRC uint32
+
+	Nacks []struct {
+		// ID of lost packets
+		PacketID uint16
+
+		// Bitmask of following lost packets
+		BLP uint16
```
maxhawkins: For someone who hasn't read the spec before this name might be hard to understand. What if it was called `LostPackets` or something?

You could even make a type around the bitmap int so the end user doesn't have to do the bit-shifting themselves.

```go
  ...

  LostPackets PacketBitmap
}

type PacketBitmap uint16

func (p PacketBitmap) Lost(i int) bool {
 ...
}
```

pkg/rtcp/receiver_report.go
wdouglass: You're right, i'll dig through the spec and figure out how to do that right (and what to do about padding, etc...)

pkg/rtcp/receiver_report_test.go
wdouglass: old C habit, i'll change it to byte

pkg/rtcp/transport_layer_nack.go
wdouglass: Love this idea, i'll implement it.

Issue #254, opened 2018-12-03
Data Race in SCTP


maxhawkins:
Found this data race in sctp while running tests on Travis.

```
==================
WARNING: DATA RACE
Read at 0x00c0002f7810 by goroutine 110:
  github.com/pions/webrtc/internal/sctp.(*Association).handleChunk()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/sctp/association.go:520 +0x1579
  github.com/pions/webrtc/internal/sctp.(*Association).HandleInbound()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/sctp/association.go:128 +0x1df
  github.com/pions/webrtc/internal/network.(*port).handleSCTP()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/port-receive.go:134 +0x138
  github.com/pions/webrtc/internal/network.(*port).handleDTLS()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/port-receive.go:147 +0xb8b
  github.com/pions/webrtc/internal/network.(*port).networkLoop()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/port-receive.go:216 +0x592
Previous write at 0x00c0002f7810 by goroutine 29:
  github.com/pions/webrtc/internal/sctp.(*Association).setState()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/sctp/association.go:284 +0xa6
  github.com/pions/webrtc/internal/sctp.(*Association).Connect()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/sctp/association.go:298 +0xe6
  github.com/pions/webrtc/internal/network.(*Manager).handleDTLSState()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/manager.go:125 +0xa2
  github.com/pions/webrtc/internal/network.(*Manager).handleDTLSState-fm()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/manager.go:76 +0x4a
Goroutine 110 (running) created at:
  github.com/pions/webrtc/internal/network.newPort()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/port.go:38 +0x232
  github.com/pions/webrtc/internal/network.NewManager()
      /home/travis/gopath/src/github.com/pions/webrtc/internal/network/manager.go:85 +0x5a9
  github.com/pions/webrtc.New()
[...]

PR #267, opened 2018-12-07
SCTP: Transport refactor

### LINKED ISSUE #256: examples/data-channels fires OnOpen twice for the same DataChannel
Here are my logs. This happens over STUN (may be caused by packet loss/retransmissions)

```
New DataChannel foo 842352001600
Data channel 'foo'-'842352001600' open. Random messages will now be sent to any connected DataChannels every 5 seconds
Data channel 'foo'-'842352001600' open. Random messages will now be sent to any connected DataChannels every 5 seconds
Sending vdsLcvepxSEmLFD
Sending XSJpdHGhsEEcjot
```

### LINKED ISSUE #266: SCTP: Transport refactor
Refactor SCTP as discussed on the [wiki](https://github.com/pions/webrtc/wiki/TransportDesign).

backkem:
This PR refactors SCTP.

TODO:
- [x] Analyse thread safety.
- [x] Cleanup & fix tests.
- [x] Test #256. Update: This fixes #256.

I'm leaving shutdown out of scope for now since we didn't have it before either.

Resolves #266, #254


internal/datachannel/datachannel.go
```diff
@@ -0,0 +1,226 @@
...
+	// SCTP does not support the sending of empty user messages.  Therefore,
+	// if an empty message has to be sent, the appropriate PPID (WebRTC
+	// String Empty or WebRTC Binary Empty) is used and the SCTP user
+	// message of one zero byte is sent.  When receiving an SCTP user
+	// message with one of these PPIDs, the receiver MUST ignore the SCTP
+	// user message and process it as an empty message.
+	var ppi sctp.PayloadProtocolIdentifier
+	switch {
+	case !isString && len(p) > 0:
+		ppi = sctp.PayloadTypeWebRTCBinary
```
Sean-Der: The SCTP package shouldn't know about WebRTC IMO. This isn't new to this PR so no reason to block.

But in the future if we make SCTP available to the world this would feel a little bit like a broken Window IMO.

internal/network/manager.go
```diff
@@ -158,12 +140,14 @@ func (m *Manager) Start(isOffer bool,
 	// Check the fingerprint if a certificate was exchanged
 	cert := dtlsConn.RemoteCertificate()
 	if cert != nil {
-		hashAlgo, err := dtls.HashAlgorithmString(fingerprintHash)
+		var hashAlgo dtls.HashAlgorithm
+		hashAlgo, err = dtls.HashAlgorithmString(fingerprintHash)
```
Sean-Der: (major style nit) I assume the linter complained about shadowing?

I stopped to parse this a few times, assuming there was something special going on here. Might be worth putting a lambda or something, just so it doesn't stand out.

rtcpeerconnection_test.go
```diff
@@ -426,49 +422,50 @@ func TestRTCPeerConnection_EventHandlers(t *testing.T) {
...
-
-	// NB: If RTCPeerConnection.dataChannelEventHandler() does not correctly wait for
-	// OnDataChannel() to complete, this will hang until timeout because the handlers aren't set
-	// before the events are processed.
-	assert.EqualValues(t,
-		[]bool{true, true, true},
-		[]bool{<-onDataChannelCalled, <-onOpenCalled, <-onMessageCalled},
-	)
-}
+// TODO: rewrite
```
Sean-Der: I would delete + create issue. We won't come back to this otherwise. This also might make a good noob issue!

internal/datachannel/datachannel.go
backkem: I disagree because:
1. It would mean we lose the String() method.
2. The PPIs are [IANA registered](https://www.iana.org/assignments/sctp-parameters/sctp-parameters.xhtml#sctp-parameters-25) so it seems fine to have them in the SCTP package.

Issue #110, opened 2018-09-09
race conditions


porjo:
I built [my app](https://github.com/porjo/babelcast) with `-race` and found a couple of races inside `webrtc` itself:

```
==================
WARNING: DATA RACE
Write at 0x00c4201bd330 by goroutine 28:
  github.com/pions/webrtc/pkg/ice.(*Agent).updateConnectionState()
      /home/ian/go/src/github.com/pions/webrtc/pkg/ice/agent.go:161 +0x45
  github.com/pions/webrtc/pkg/ice.(*Agent).setValidPair()
      /home/ian/go/src/github.com/pions/webrtc/pkg/ice/agent.go:174 +0x10d
  github.com/pions/webrtc/pkg/ice.(*Agent).handleInboundControlled()
      /home/ian/go/src/github.com/pions/webrtc/pkg/ice/agent.go:305 +0xaa
  github.com/pions/webrtc/pkg/ice.(*Agent).HandleInbound()
      /home/ian/go/src/github.com/pions/webrtc/pkg/ice/agent.go:355 +0x321
  github.com/pions/webrtc/internal/network.(*port).networkLoop()
      /home/ian/go/src/github.com/pions/webrtc/internal/network/port-receive.go:156 +0x546

Previous write at 0x00c4201bd330 by goroutine 38:
  github.com/pions/webrtc/pkg/ice.(*Agent).updateConnectionState()
      /home/ian/go/src/github.com/pions/webrtc/pkg/ice/agent.go:161 +0x45
  github.com/pions/webrtc/pkg/ice.(*Agent).agentTaskLoop()
      /home/ian/go/src/github.com/pions/webrtc/pkg/ice/agent.go:186 +0x74

Goroutine 28 (running) created at:
  github.com/pions/webrtc/internal/network.newPort()
      /home/ian/go/src/github.com/pions/webrtc/internal/network/port.go:38 +0x1e3
  github.com/pions/webrtc/internal/network.NewManager()
      /home/ian/go/src/github.com/pions/webrtc/internal/network/manager.go:68 +0x5b8
  github.com/pions/webrtc.New()
      /home/ian/go/src/github.com/pions/webrtc/rtcpeerconnection.go:171 +0x331
  main.NewPC()
      /home/ian/go/src/github.com/porjo/babelcast/webrtc.go:42 +0x296
  main.(*Conn).setupSession()
      /home/ian/go/src/github.com/porjo/babelcast/conn.go:84 +0x132
  main.wsHandler()
[...]

PR #274, opened 2018-12-09
Fix races in SCTP/ICE

backkem:
Rebase of @mjmac's #260 since it was reverted.

Found a couple of minor races -- just use mutexes to ensure
synchronized access.

Relates to #110


Sean-Der:
Hey @backkem 

Have you tried the examples with this? We hit some deadlocks before, but I didn't investigate where so I just did a speculative revert.

Issue #272, opened 2018-12-09
RTP: Transport refactor


backkem:
Refactor RTP as discussed on the [wiki](https://github.com/pions/webrtc/wiki/TransportDesign).
Current design can be found under [rtp-refactor](https://github.com/pions/webrtc/wiki/TransportDesign#rtp-refactor).

TODO:
* [x] [Optimizations](https://github.com/pions/webrtc/wiki/TransportDesign#optimizations)
* [ ] Separate [raw RTP](https://github.com/pions/webrtc/wiki/TransportDesign#raw-rtp).
* [ ] Move `internal/srtp` -> `github.com/pions/srtp`
* [ ] Move `pkg/rtp` -> `github.com/pions/rtp`
* [ ] Move `pkg/rtcp` -> `github.com/pions/rtcp`

Sean-Der:
I will read the RFC, this sounds perfect as long as the mux/demux doesn't end up in `pions/dtls`!

This demuxing is very WebRTC centric, so might even be worth calling it `webrtc.Mux` or something along those lines since SRTP technically has nothing to do with DTLS (it just gets the encryption keys from it). But I fully support the design of linking transports.
```
                    +----------------+
                    |  127 < B < 192 -+--> forward to RTP
                    |                             |
packet -->  |     19 < B < 64  -+--> forward to DTLS
                    |                             |
                    |                B < 2   -+--> forward to STUN
                    +----------------+
```

Sean-Der:
@backkem 

I started working on this, here are my thoughts and the steps I am going to take. Wanted to make sure I was on the right track.

----

[ ] Make the SRTP generic API generic. Just pass `[]byte` actually useful for public consumption
[ ] Move `internal/srtp` -> `github.com/pions/srtp`
[ ] Move `internal/sdp` -> `github.com/pions/sdp`
[ ] Move `pkg/rtp` -> `github.com/pions/rtp`
[ ] Move `pkg/rtcp` -> `github.com/pions/rtcp`
[ ] Create `internal/transport` and put SRTP wrapper in there (maybe not the right place)
[ ] Wire that all up!
-----

I don't think we should put any transport related abstractions around RTP/SRTP in the public API, we should keep that inside `webrtc`. The transport API makes sense for `SCTP` and `ICE` because those are real connections (reads/writes can happen internally etc..) but SRTP is just when people want to encrypt/decrypt blobs.

backkem:
[...]

PR #278, opened 2018-12-10
Refactor DTLS/SRTP muxing

backkem:
Relates to #272


internal/mux/endpoint.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (e *Endpoint) close() {
+	close(e.doneCh)
+}
+
+// Read reads a packet of len(p) bytes from the underlying conn
+// that are matched by the associated MuxFunc
+func (e *Endpoint) Read(p []byte) (int, error) {
+	select {
+	case e.readCh <- p:
```
Sean-Der: I wish channels could have a different datatype for in/out (so we don't need two unique channels)

This might be a nit, but do you think we could choose a better name. When I first saw this pattern I assumed `readCh` and `sizeCh` were tied to each other (like `readCh` returns data and `sizeCh` returns how much) since we are going to use this a lot might be worth making it less confusing.

backkem: Any suggestions? I'm not sure how to convey the entire idea in the channel names.

PR #282, opened 2018-12-11
Allow local ephemeral port range to be limited

wdouglass:
This allows primitive (non-stateful) firewalls to have less wide-open configuration, while still maintaining some chance of getting UDP packets out...

### DIFF
--- a/internal/network/manager.go
+++ b/internal/network/manager.go
@@ -66,19 +66,9 @@ func (m *Manager) AddTransportPair(ssrc uint32, Rtp chan<- *rtp.Packet, Rtcp cha
 }
 
 // NewManager creates a new network.Manager
-func NewManager(urls []*ice.URL, btg BufferTransportGenerator, ntf ICENotifier) *Manager {
-	iceAgent := ice.NewAgent(urls, ntf)
-
-	return &Manager{
-		IceAgent:                 iceAgent,
-		bufferTransportPairs:     make(map[uint32]*TransportPair),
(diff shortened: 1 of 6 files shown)


backkem:
Some points:
1. I'd prefer if the port range was passed via the ice.Agent constructor, avoiding potential race issues.
2. We need to make it more obvious that:
    - This diverts from plain WebRTC and should only be used is special cases.
    - The implementation is currently limited to host candidates and doesn't impact reflective candidates.

wdouglass:
> Some points:
> 
> 1. I'd prefer if the port range was passed via the ice.Agent constructor, avoiding potential race issues.

I kept it seperate because it *is* a special case, and by default i didn't want to have dummy arguments to the constructor. Should it just be a second constructor?

> 2. We need to make it more obvious that:
>    
>    * This diverts from plain WebRTC and should only be used is special cases.
>    * The implementation is currently limited to host candidates and doesn't impact reflective candidates.

Agreed, i'll update the comments to make this more clear.

backkem:
Do you plan on going with the setting engine approach? If so, I'd also suggest to de-duplicate the constructors. The network package will get removed at some point, just pass everything as an argument. For for ICE agent: It seems smart to bundle all arguments as a ice.Config struct:
```Go
type Config struct {
	Urls []*URL
	Notifier func(ConnectionState)
	PortMin uint16
	PortMax uint16
} 
```
That way it can grow more easily in the future and the port range is optional.

Issue #276, opened 2018-12-10
Add E2E tests that exercise `Close()`


Sean-Der:
We should assert that calling `Close()` works.

* It should be called in all the normal places users would call it.
    - Inside `OnICEConnectionStateChange`
    - At anytime outside a pion-WebRTC callback
    - At random times during signaling (before/between `SetRemoteDescription` and `CreateAnswer`)

thetooth:
I have managed to update to current head and get stable results, but the missing close operation was causing some issues. I tracked down what was going on during tear down that I believe should be done as part of the close operation.

- Right now, as it's not fully implemented attempting to close the PC causes a lockup, it should block for those who need that, but should not block any other peer connections.
- It should dereference all On* callbacks, I have found in my own code it's pretty easy to reuse the same endpoint for the same returning user, this means dying or dead peer connections will still be calling these endpoints with conflicting state.
- Clean up datachannels, this might already be done as I've had no issues to date, but again having OnMessage triggered for a non-existent thing can lead to fun bugs.

backkem:
The E2E test have meanwhile been added for the basic flow of closing the peer connection after use. The test assures that all goroutines are ended. It should therefore not be possible for callbacks to fire afterwards.

Do we want to keep this open for the other tests (different timings) as mentioned in the original issue? Maybe that's better tracked in on of the API concurrency issues?

Sean-Der:
Hey @thetooth I am closing this for now. Me and @backkem added lots of tests around close, so this shouldn't regress again in the future.

I always love getting feedback about how we can be doing stuff better, so if you are still seeing issues/have anything we can do better please let me now!

PR #287, opened 2018-12-12
[WIP] Add PeerConnection Close test.

backkem:
The PeerConnection now closes without complaining. However, a bunch of the network layers leak routines.

Relates to #276


Sean-Der:
Nice! Thank you so much for jumping on this so quick :)

PR #300, opened 2018-12-20
Re-enable SRTP Auth tag verification

### LINKED ISSUE #270: Re-enable SRTP Auth tag verification
This is failing for me with gstreamer-receive disabling for now

Sean-Der:
Auth tag verification before would hash including the old auth tag, now
we remove the old tag before generating the hash

Resolves #270

### DIFF
--- a/internal/srtp/context.go
+++ b/internal/srtp/context.go
@@ -1,6 +1,7 @@
 package srtp
 
 import (
+	"bytes"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/hmac"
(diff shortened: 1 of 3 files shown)


Sean-Der:
@backkem I am able to use this `gstreamer-receive` and Janus now, should be good since SRTP is symmetric encryption!

backkem:
The change looks fine but it seems some test are failing.

Sean-Der:
@backkem passes now!

I needed to update raw+payload fields (didn't do that before because didn't care about auth value)

Issue #277, opened 2018-12-10
API: Expose networking


backkem:
For more demanding use-cases it seems interesting to have direct access to the networking layer. One would still use the WebRTC API to setup a connection since you don't want to duplicate the entire setup procedure. However, afterwards you would have to option to detach the networking, E.g.:
```Go
func (d *RTCDataChannel) Detach() (*datachannel.DataChannel, error)
```
When this method is called, the `RTCDataChannel.readLoop` doesn't get started and it's up to the caller to handle reading from the `datachannel.DataChannel` properly. It throws an error if the readLoop is already started.

Usage examples:
```Go
d, err := peerConnection.CreateDataChannel("data", nil)
util.Check(err)
conn, err := d.Detach()
util.Check(err)
// Manually read/write to conn
```
```Go
peerConnection.OnDataChannel(func(d *webrtc.RTCDataChannel) {
	conn, err := d.Detach()
	util.Check(err)
	// Manually read/write to conn
}
```

TODO:
- [x] Make package datachannel public. TBD: Where?
- [ ] Add the Detach method as described above.
- [ ] Add a Detach example.

PR #304, opened 2018-12-21
Detach: Expose data channel networking

backkem:
This is the initial experiment towards providing direct access to the data channel as described in #277. Once detached the `io.ReadWriter` API feels very familiar. In addition, this seems really powerful for more networking oriented applications. E.g.: a lot of silly glue logic could be removed from `pions/dcnet`.

If you want to build this branch right now, you have to manually move the `sctp` and `datachannel` package under the pions root.

TODO:
- [ ] Finalize API: There is currently no good place to call the detach method. The prerequisites are that the `datachannel.DataChannel` object exists but the `RTCDataChannel.readLoop` is not yet started. Some potential solution:
    - `OnOpen` seems like a good candidate but it is asynchronous ATM.
    - Signal early that you want to detach somehow (settings manager?) and have a method to get the `datachannel.DataChannel` object once it exists.
    - Add an OnDetach callback. However, this would also suffer from #305. Preferably we won't have to solve this twice.
- [x] Correctly move package `sctp` out of tree.
- [x] Correctly move package `datachannel` out of tree.
- [ ] Finalize `data-channels-detach` example.
- [ ] Create `data-channels-detach-create` example.

### DIFF
--- a/rtcdatachannel.go
+++ b/rtcdatachannel.go
@@ -181,7 +181,12 @@ func (d *RTCDataChannel) handleOpen(dc *datachannel.DataChannel) {
 	// Ensure on
 	d.onOpen()
 
-	go d.readLoop()
+	d.Lock()
+	defer d.Unlock()
+
+	if !defaultSettingEngine.Detach.DataChannels {
+		go d.readLoop()
+	}
 }
 
 func (d *RTCDataChannel) readLoop() {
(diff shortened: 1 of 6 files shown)


rtcdatachannel.go
```diff
@@ -224,3 +229,26 @@ func (d *RTCDataChannel) Send(payload sugar.Payload) error {
...
+func (d *RTCDataChannel) Detach() (*datachannel.DataChannel, error) {
+	d.Lock()
+	defer d.Unlock()
+
+	if !defaultSettingEngine.Detach.DataChannels {
+		return nil, errors.New("enable detaching by calling webrtc.DetachDataChannels()")
+	}
+
+	if d.dataChannel == nil {
+		return nil, errors.New("ratachannel not opened yet, try calling Detach from OnOpen")
```
Sean-Der: `ratachannel` -> `datachannel`

PR #308, opened 2018-12-23
Fix crash in RTP, and add test coverage

Sean-Der:


Sean-Der:
@backkem When reading the RFC I also realized I made a mistake when I wrote this originally.

https://tools.ietf.org/html/rfc5285#section-1
```
16-bit length field specifying the length of the
header extension in 32-bit words.
```

PR #318, opened 2018-12-27
Fix for a hang when searching for ICE candidate pairs

wdouglass:


wdouglass:
The only thing that happens when an error gets returned is a write call returns 0 length, and a write gets missed. in the case that a candidate appears late, the connection should recover. @bakkem and I discussed this a bit this morning. I'm going to go ahead and merge. Thanks for all the help!

Issue #324, opened 2018-12-28
ICE Connection timeouts are constant, and very long


wdouglass:
We should add settingEngine hooks to modify them

Sean-Der:
Landed! Thank you @wdouglass

PR #325, opened 2018-12-28
Add settingEngine hooks to modify ICE timeouts

wdouglass:
Relates to #324

### DIFF
--- a/pkg/ice/agent.go
+++ b/pkg/ice/agent.go
@@ -51,7 +51,11 @@ type Agent struct {
 	portmax uint16
 
 	//How long should a pair stay quiet before we declare it dead?
+	//0 means never timeout
 	connectionTimeout time.Duration
+
+	//How often should we send keepalive packets?
+	//0 means never
 	keepaliveInterval time.Duration
 
 	localUfrag      string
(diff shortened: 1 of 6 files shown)


pkg/ice/agent.go
```diff
@@ -126,6 +137,19 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 		portmax:     config.PortMax,
 	}
 
+	// connectionTimeout used to declare a connection dead
+	if config.ConnectionTimeout == 0 {
```
Sean-Der: I can't think of a case, but could 0 ever be an actual value someone wants? Maybe disabling timeout all together? (Seems unlikely though)

rtcpeerconnection.go
```diff
@@ -158,7 +158,9 @@ func New(configuration RTCConfiguration) (*RTCPeerConnection, error) {
 		}
 	}
 
-	pc.networkManager, err = network.NewManager(urls, pc.generateChannel, pc.iceStateChange, defaultSettingEngine.EphemeralUDP.PortMin, defaultSettingEngine.EphemeralUDP.PortMax)
+	pc.networkManager, err = network.NewManager(urls, pc.generateChannel, pc.iceStateChange,
```
Sean-Der: Can we just pass the `ice.AgentConfig` directly here? Having lots of parameters is hard to read (and easy to get them out of order by accident)

pkg/ice/agent.go
wdouglass: That's a good idea, i'll look into that

rtcpeerconnection.go
wdouglass: It's public, I don't see why not

pkg/ice/agent.go
wdouglass: Does it make more sense to create a new null.Duration object, or to just use a pointer and create a copy when the user calls SetConnectionTimeout?

The third option, which i don't like but am including here for completeness, is to use null.Int32 and cast back and forth to time.Duration (it's a typedef underneath it all anyway, but I think this is probably less clear...)

wdouglass: I've gone with the second option to
a) keep the AgentConfig struct a bit smaller
b) avoid creating a new null.Duration type

Changing over to use pkg/null is easy if you'd like; this just fits my instinct better.

PR #327, opened 2018-12-28
 ORTC: Add RTCQuicTransport

### LINKED ISSUE #275: ORTC: Add RTCQuicTransport
One we finish the `RTCIceTransport` in #273 it should be fairly straightforward to add experimental support for [webrtc-quic](https://w3c.github.io/webrtc-quic/). I think this would be a worthwhile experiment. We can use [quic-go](https://github.com/lucas-clemente/quic-go) for the networking itself. It uses the similar interfaces as our own networking code. TBD if we keep it in a separate branch.

backkem:
Resolves #275

### DIFF
--- a/examples/ortc-quic/main.go
+++ b/examples/ortc-quic/main.go
@@ -1,0 +1,140 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"time"
+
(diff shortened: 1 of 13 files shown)


backkem:
@Sean-Der I suggest we use the normal review process for merging into the ORTC branch.

PR #328, opened 2018-12-29
Move SRTP to Transport API

Sean-Der:
Move SRTP to transport based API


backkem:
Possible [optimizations ](https://github.com/pions/webrtc/wiki/TransportDesign#optimizations)

internal/network/manager.go
```diff
@@ -145,27 +102,89 @@ func (m *Manager) startICE(isOffer bool, remoteUfrag, remotePwd string) error {
...
+}
+
+func (m *Manager) startSRTP(isOffer bool) error {
 	keyingMaterial, err := m.dtlsConn.ExportKeyingMaterial([]byte("EXTRACTOR-dtls_srtp"), nil, (srtpMasterKeyLen*2)+(srtpMasterKeySaltLen*2))
 	if err != nil {
 		return err
 	}
 
-	return m.CreateContextSRTP(keyingMaterial, isOffer)
+	offset := 0
```
backkem: Maybe we can consume the keying material inside the srtp package, or is it to WebRTC specific?

internal/network/manager.go
```diff
@@ -247,68 +266,75 @@ func (m *Manager) Close() error {
...
-	}
-
 	raw, err := packet.Marshal()
 	if err != nil {
 		fmt.Printf("SendRTP failed to marshal packet: %s \n", err.Error())
+		return
 	}
 
-	_, err = m.iceConn.Write(raw)
+	writeStream, err := m.srtpSession.OpenWriteStream()
```
backkem: This feels weird. I would explicitly open a stream for each RTCTrack so you can directly write to it instead of implicitly opening it when writing to the network manager.

internal/network/manager.go
Sean-Der: Yea! I I don't see a reason to not move it inside.

I think for our new `*Session` stuff we can make it less flexible, we are just writing abstractions to make our lives easier.

internal/network/manager.go
Sean-Der: Yep agree 100%!

Every `RTCTrack` can call `OpenWriteStream` and keep their stream private. They will all just share one (even if they think they each have their own)

Sean-Der:
@backkem All the tests pass and send/recv RTP+RTCP works, mind taking a look over and suggesting improvements?

I will push hopefully tomorrow and get the code coverage back up, I have to rewrite one test and hopefully that gets up back in a better spot!

backkem:
@Sean-Der Here's what I came up with for Raw RTC: [Raw RTC](https://github.com/pions/webrtc/wiki/TransportDesign#raw-rtp).
I also think we should probably take a look at the [RTC Track API](https://github.com/pions/webrtc/wiki/TransportDesign#rtctrack-api). Maybe we should keep it for another PR thought.

Issue #331, opened 2018-12-31
ICE connection state changes to "Disconnected" 30 seconds after it connected.


enobufs:
I am using data channel only, successfully saw OnOpen callback from RTCDataChannel, but in about 30 seconds, I see ICE Connection State changes from "Connected" to "Disconnected" for a small amount of time (~ a few msec) then back to "Connected" state.

```
[I]00:51:01.946889 ICE Connection State has changed: Checking
[I]00:51:01.947526 Signaling State has changed: stable
[I]00:51:03.954456 ICE Connection State has changed: Connected
[I]00:51:04.962246 New DataChannel data 824637514476
[I]00:51:04.962311 Data channel 'data'-'824637514476'
[I]00:51:04.962328 Data channel (name=data) ready!
[I]00:51:35.950675 ICE Connection State has changed: Disconnected
[I]00:51:35.950753 Peer disconnected
[I]00:51:35.953040 ICE Connection State has changed: Connected
```

This does not occur every time.  I see this about once in 5 times of connections.  When this happens, the value of `time.Since(a.selectedPair.remote.LastReceived())` in [this line](https://github.com/pions/webrtc/blob/v1.2.0/pkg/ice/agent.go#L357) keeps growing until it reaches `connectionTimeout` which is set to 30 seconds.

Also, I noticed that the value of `time.Since(a.selectedPair.remote.LastSent())` in [this line](https://github.com/pions/webrtc/blob/v1.2.0/pkg/ice/agent.go#L374) is rediculously large. Consequently, `keepaliveCandidate()` is called every time checkKeepalive() is called (every 2 seconds in my observation). 

```
validateSelectedPair: sinceLastRcvd=1.997226772s
checkKeepalive: sinceLastSent=2562047h47m16.854775807s
validateSelectedPair: sinceLastRcvd=3.997218202s
checkKeepalive: sinceLastSent=2562047h47m16.854775807s
validateSelectedPair: sinceLastRcvd=5.99869347s
checkKeepalive: sinceLastSent=2562047h47m16.854775807s
validateSelectedPair: sinceLastRcvd=7.99959734s
checkKeepalive: sinceLastSent=2562047h47m16.854775807s
validateSelectedPair: sinceLastRcvd=9.999625507s
checkKeepalive: sinceLastSent=2562047h47m16.854775807s
```

I am using pions/webrtc v1.2.0, running on mac.

enobufs:
[...]

PR #332, opened 2018-12-31
Fix keepalive transmission interval mentioned in issue #331

enobufs:
The `lastSent` value on a remote component of selected-pair never gets updated, which was the reason why LastSent() returned a large duration, causing the keepalive packet sent every 2 seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Oleg Kovalov](https://github.com/cristaloleg) *Use wildcards instead of hardcoding travis-ci config*
 * [Woodrow Douglass](https://github.com/wdouglass) *RTCP, RTP improvements, G.722 support, Bugfixes*
 * [Tobias Fridén](https://github.com/tobiasfriden) *SRTP authentication verification*
+* [Yutaka Takeda](https://github.com/enobufs) *Fix ICE connection timeout*
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/pkg/ice/agent.go
+++ b/pkg/ice/agent.go
@@ -418,7 +418,7 @@ func (a *Agent) checkKeepalive() {
 		return
 	}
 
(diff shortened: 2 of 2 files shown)


Sean-Der:
Nice! Thank you for fixing this @enobufs 

Would you mind putting `Resolves #331` on a newline (with a space between it and the subject) that should make `travis-ci` pass.

Thank you

enobufs:
Resolves #331

Sean-Der:
Sorry bad instructions on my part @enobufs

Can you update your git commit message, a good one would be like.

You can update the message by doing `git commit --amend` and then do a `git push -f`

```
Fix keepalive by checking proper field

Resolves  #331
```

You can then run [this](https://github.com/pions/webrtc/blob/master/.github/lint-commit-message.sh) script it is used to lint commit messages. You can see the rules on line 12-16

backkem:
I'm not convinced yet that this is correct. I think we update LastSent on the local candidate:
https://github.com/pions/webrtc/blob/b98909a75b254ac58a6be834bde1a5a68cbfe739/pkg/ice/candidate.go#L140
From here:
https://github.com/pions/webrtc/blob/b98909a75b254ac58a6be834bde1a5a68cbfe739/pkg/ice/candidatepair.go#L43
And here:
https://github.com/pions/webrtc/blob/b98909a75b254ac58a6be834bde1a5a68cbfe739/pkg/ice/candidatepair.go#L21
And lastReceived on the remote:
https://github.com/pions/webrtc/blob/b98909a75b254ac58a6be834bde1a5a68cbfe739/pkg/ice/agent.go#L617

enobufs:
I should have read "Contributing" page first. Let me fix the last error.

enobufs:
Thanks @backkem.  Happy new year to you all!

Sean-Der:
Nice work @enobufs thanks for contributing :)

You should have a merge button now that all tests are passed, thanks again for using pion-WebRTC! We also have a Slack (instructions in the README) and always happy to help debug 1:1

enobufs:
@Sean-Der I see `Only those with write access to this repository can merge pull requests.`
I am already on Slack by name of "@Yutaka". :)

PR #340, opened 2019-01-07
SRTP optimizations and API stability (so we can export)

Sean-Der:
Relates to #272


internal/srtp/session_srtcp.go
```diff
@@ -0,0 +1,137 @@
...
+
+	encrypted, err := s.localContext.EncryptRTCP(nil, buf, nil)
+	if err != nil {
+		return 0, err
+	}
+	return s.session.nextConn.Write(encrypted)
+}
+
+func (s *SessionSRTCP) decrypt(buf []byte) error {
+	decrypted, err := s.remoteContext.DecryptRTCP(nil, buf, nil)
```
Sean-Der: RTCP packets usually contain multiple SSRces (so we can't decrypt in place I don't think)

rtcpeerconnection.go
```diff
@@ -928,18 +927,15 @@ func (pc *RTCPeerConnection) acceptSRTP() {
 		// RTP
 		go func() {
 			for {
-				rtpPacket := &rtp.Packet{}
-				rtpLen, err = r.Read(rtpBuf)
+				rtpBuf := make([]byte, receiveMTU)
```
Sean-Der: We have to alloc for now because we dupe packets for our user (in existing API)

When we do API breakage we can make strict rules about who owns what data wise

rtcpeerconnection.go
```diff
@@ -1539,8 +1523,8 @@ func (pc *RTCPeerConnection) newRTCTrack(payloadType uint8, ssrc uint32, id, lab
 	}
 
 	trackInput := make(chan media.RTCSample, 15) // Is the buffering needed?
-	rawPackets := make(chan *rtp.Packet)
-	rtcpPackets := make(chan rtcp.Packet)
+	rawPackets := make(chan *rtp.Packet, 15)     // Is the buffering needed?
```
Sean-Der: I am just making this consistent with the others ones.

I noticed on the SFU example a poorer experience, this needs to be visited fully.

PR #342, opened 2019-01-07
ORTC

backkem:
This PR adds the ORTC constructors and the Quic transport. It backs WebRTC by ORTC and phases out most of the network manager except for SRTP.


rtcdtlstransport.go
```diff
@@ -1,13 +1,166 @@
...
+// NewRTCDtlsTransport creates a new RTCDtlsTransport.
+// This constructor is part of the ORTC API. It is not
+// meant to be used together with the basic WebRTC API.
+func NewRTCDtlsTransport(transport *RTCIceTransport, certificates []RTCCertificate) (*RTCDtlsTransport, error) {
+	t := &RTCDtlsTransport{iceTransport: transport}
+
+	if len(certificates) > 0 {
+		now := time.Now()
+		for _, x509Cert := range certificates {
+			if !x509Cert.Expires().IsZero() && now.After(x509Cert.Expires()) {
```
Sean-Der: It worries me doing security stuff like this, I don't feel confident we catch all the edge cases. So if we start catching some, I think people will assume we caught everything.

Before we did nothing, so easy enough to say `please check things yourself` but we can talk about this further outside the PR!

rtcquictransport_test.go
```diff
@@ -0,0 +1,222 @@
+package webrtc
+
+// TODO: figure out how to avoid leaking
```
Sean-Der: Can we re-enable this but just disable the leak testing (and ticket it) these E2E tests are just super valuable.

backkem: @Sean-Der I tried that but the leaked routine also breaks any test that checks for leaks that runs after this one.

PR #345, opened 2019-01-08
Isolate some global variables, add a settingengine test

wdouglass:
Hopefully we can keep all of our globals in a central place, and refer to them by internal reference. This will increase the library's flexibility.


rtcdatachannel.go
```diff
@@ -93,13 +93,23 @@ type RTCDataChannel struct {
 
 	sctpTransport *RTCSctpTransport
 	dataChannel   *datachannel.DataChannel
+
+	// A reference to the associated setting engine used by this datachannel
+	settingEngine *settingEngine
```
backkem: Nit: prefer linking to API directly.

Issue #347, opened 2019-01-08
Update go.mod and go.sum for new QUIC libs and ICE changes


Sean-Der:
Currently hitting these errors

```
8c859026d445:sfu seaduboi$ go run main.go 
# github.com/pions/webrtc
../../rtcicecandidate.go:26:34: undefined: sdp.ICECandidate
../../rtcicecandidate.go:47:34: undefined: sdp.ICECandidate
../../rtcicecandidate.go:48:9: undefined: sdp.ICECandidate
../../rtcpeerconnection.go:742:8: a.IsICECandidate undefined (type sdp.Attribute has no field or method IsICECandidate)
../../rtcpeerconnection.go:743:27: a.ToICECandidate undefined (type sdp.Attribute has no field or method ToICECandidate)
../../rtcpeerconnection.go:997:32: attribute.ToICECandidate undefined (type sdp.Attribute has no field or method ToICECandidate)
../../rtcpeerconnection.go:1464:79: undefined: sdp.ICECandidateAttribute
../../rtcpeerconnection.go:1466:8: media.WithICECandidate undefined (type *sdp.MediaDescription has no field or method WithICECandidate)
../../rtcpeerconnection.go:1468:8: media.WithICECandidate undefined (type *sdp.MediaDescription has no field or method WithICECandidate)
../../rtcpeerconnection.go:1499:79: undefined: sdp.ICECandidateAttribute
../../rtcpeerconnection.go:1499:79: too many errors

```

backkem:
We should probably force enable go mix in CI.

Sean-Der:
@backkem fixed this with https://github.com/pions/webrtc/commit/4c45b5863c289baaf844ea05f632f31e7a3b43ae

PR #348, opened 2019-01-08
Modules: Update dependencies

backkem:
`go-quic` is a rather large dependency apparently.

Relates to #347


Sean-Der:
I don't think it is an issue right now, but would be cool if we could make things 'pluggable' if in the future people may only want SCTP+DTLS or QUIC and ICE.

Not a burning task would be cool, and would bring down binary size probably!

PR #351, opened 2019-01-10
[WIP] Add NTP time types

maxhawkins:
_This is a breaking API change. It's a work in progress, but I want to get it on the radar for 2.0. Let me know what you think._

Using a custom type rather than raw integers  makes it easier
to parse RTP and RTCP timestamps and convert them into Go types.
This will be useful for writing jitter buffers, track synchronization,
and statistics tabulation code.

Easy conversion from the Go time package also allows us to use
a time.Duration in the RTCSample struct. This makes it more difficult
to produce invalid packets with short timestamps while using
the SampleTrack interface.

### DIFF
--- a/pkg/rtp/packet.go
+++ b/pkg/rtp/packet.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 
+	"github.com/pions/webrtc/pkg/ntp"
 	"github.com/pkg/errors"
 )
 
@@ -18,7 +19,7 @@ type Packet struct {
 	PayloadOffset    int
 	PayloadType      uint8
 	SequenceNumber   uint16
-	Timestamp        uint32
+	Timestamp        ntp.Time32
 	SSRC             uint32
 	CSRC             []uint32
 	ExtensionProfile uint16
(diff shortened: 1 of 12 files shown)


pkg/rtp/packet.go
```diff
@@ -169,7 +170,7 @@ func (p *Packet) Marshal() ([]byte, error) {
 	rawPacket[1] |= p.PayloadType
 
 	binary.BigEndian.PutUint16(rawPacket[seqNumOffset:], p.SequenceNumber)
-	binary.BigEndian.PutUint32(rawPacket[timestampOffset:], p.Timestamp)
+	binary.BigEndian.PutUint32(rawPacket[timestampOffset:], uint32(p.Timestamp))
```
wdouglass: This conversion (and the marshall version, above) may not be correct by virtue of the fact that the RTP timestamp may not be in the same units for every packet type. for instance, H264 requires a timestamp clocked at 90 Khz, while many audio payload types (at least μlaw and G.722) have a clock rate of 8 Khz. maybe you can add the clock rate as a property of the payloader, and pass the payloader to the marshal and unmarshal members? At any rate, this needs to be more complicated then it is.

wdouglass: another idea is to have time.Duration<->uint32 conversion functions themselves as members of the payloader interface.

maxhawkins:
It seems I misread the RTP spec! Thanks for correcting me @wdouglass.

I'm going to back out these changes and try again after re-reading RFC 3550.

Sean-Der:
@maxhawkins even with small technical issues, I think this is awesome (and the way we need to go)

Thanks for putting in the hard work and approaching this from another angle, I think this will make things a lot better. @kc5nra is also a great resource here, he still works full time in this area so has racked up more time then me even :)

maxhawkins:
After giving it some thought, I don't think this approach is a good one. I might add some helper functions for parsing rtp time later, but for now I'm just going to close this PR.

Issue #335, opened 2019-01-03
ICE: Support connecting to peer reflexive candidates


backkem:
Support connecting to peer reflexive candidates.

cc @enobufs

backkem:
I suggest we first verify we can connect to peer reflexive candidates correctly. Allow peer reflexive candidates to be added via `PeerConnection.AddIceCandidate()` and enable the ICE agent to connect to them. This may already work but it isn't handled correctly since `ice.CandidateTypePeerReflexive` is not fully implemented:
https://github.com/pions/webrtc/blob/361cef680f416a35570259222922e1b1336ebd65/pkg/ice/candidatetype.go#L10

Next, we can start generating our own peer reflexive candidates. For this case I think you are right. The peer reflexive candidates will only be generated once the ICE checks have started so they can only be passed via the `OnICECandidate()` callback. Internally the the logic for detecting peer reflexive candidates should be added to package `pkg/ice`.

enobufs:
Attempted to connect from pion to Chrome, using pion's AddIceCandidate() for new candidates arrived after setRemoteDescription() was called, but ICE wouldn't succeed. (I confirmed, non-trickle mode in [my code](https://github.com/enobufs/webrtc-audio-streaming/tree/feature/pion) works).  I will look into this further.

enobufs:
Ok, it was fixed by #350.  Confirmed trickle works.  Moving on to prflx issue...

enobufs:
I noticed in the following configuration, pion ignores all the STUN-binding requests from NATed node. Consequently, all connectivity checks fail.
```
Chrome Browser ----> [NAT] ---((  Internet ))----> Pion
```

[This code](https://github.com/pions/webrtc/blob/20619e871258b242c19a8ed90dfbca6696492269/pkg/ice/agent.go#L650-L652) needs to be fixed.

Connectivity checks from a NAT'ed node will have translated source tranport address by NAT. Therefore, the receiving end won't be able to find the address in the remote candidate list (expected). It should respond to the request.  In the case for the receiving peer not being able to find the transport address in the current remote candidate list, it should be treated as a `peer-reflexive` candidate as described in -RFC 8445 - sec 7.3.1.3](https://tools.ietf.org/html/rfc8445#section-7.3.1.3)

enobufs:
[...]

PR #354, opened 2019-01-12
[WIP] Detect then add prflx candidate.

enobufs:
This PR is made for early review on the solution for #335 I am working on.

What I did so far were:
* Detect prflx candidate when a STUN binding request is received from unknown source.
* Add the candidate to the remote candidate list, which would cause connectivity checks against the added candidate next time.

> I left comments with `TODO` for which I was not 100% sure, and need a careful review. 

With this change, I was able to connect Chrome browser behind a NAT and pion with no NAT as described in #335.

TODO: write tests


backkem:
Looking good so far!

pkg/ice/agent.go
```diff
@@ -646,22 +647,60 @@ func (a *Agent) handleInboundControlling(m *stun.Message, localCandidate, remote
...
 func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr) {
+	iceLog.Tracef("inbound STUN from %s to %s", remote.String(), local.String())
 	remoteCandidate := a.findRemoteCandidate(local.NetworkType, remote)
 	if remoteCandidate == nil {
-		iceLog.Debugf("Could not find remote candidate for %s ", remote)
-		return
-	}
-
-	remoteCandidate.seen(false)
+		//iceLog.Debugf("Could not find remote candidate for %s ", remote)
```
backkem: Small suggestion: Maybe move this entire section to a separate method so you can return errors to make the code flatter:
```go
if err != nil {
  return err 
}
doSomething()
```
You can still log the error in `handleInbound`.

enobufs: Good idea. Will do.

enobufs:
oops..  I totally forgot about that. :p  Do I need to amend all commits I made?

backkem:
> Do I need to amend all commits I made?
I think the lint only checks the last one but we do like to keep the commit history nice and tidy. Maybe you can squash a couple of them so you have to edit less.

enobufs:
Fixed the commit messages. (`git rebase -i` is awesome)
One test (quic/transport related) is failing. I think it is an isolated issue...

enobufs:
Thanks @Sean-Der .

backkem:
We fixed the Quic issue in master. Do you mind re-basing your changes?

enobufs:
No.  Please go ahead, and thank you.

pkg/ice/agent.go
```diff
@@ -644,11 +645,55 @@ func (a *Agent) handleInboundControlling(m *stun.Message, localCandidate, remote
...
+		0,  // unknown at this moment. TODO: need a review
+	)
+
+	if err != nil {
+		return errors.Wrapf(err, "failed to create peer-reflexive candidate: %v", remote)
+	}
+
+	// Add pflxCandidate to the remote candidate list
+	// TODO: review if this wouldn't cause any race
+	networkType := pflxCandidate.NetworkType
```
Sean-Der: Are you able to call `AddRemoteCandidate` directly here? It handles all the locking/modifying for you

```
return a.AddRemoteCandidate(pflixCandidate)
```

enobufs: Just confirmed, the thread has already in agent.run() routine.  Calling another agent.run() would be a dead lock.  (I wonder if there's away to allow "recursivce mutex locking".)

Sean-Der: Oops sorry my mistake!

I would make a private version `addRemoteCandidate` that assumes you have the lock so it could look like this.

```
// AddRemoteCandidate adds a new remote candidate
func (a *Agent) AddRemoteCandidate(c *Candidate) error {
  return a.run(func (agent *Agent) {
    a.addRemoteCandidate(c)
  })
}

// addRemoteCandidate assumes you are holding the lock (must be execute using a.run)
func (a *Agent) addRemoteCandidate(c *Candidate) error {
        networkType := c.NetworkType
        set := agent.remoteCandidates[networkType]

        for _, candidate := range set {
            if candidate.Equal(c) {
                return
            }
        }

        set = append(set, c)
        agent.remoteCandidates[networkType] = set
}
```

enobufs: Iterating over remoteCandidates has already been done earlier by findRemoteCandidate(). So, it would redundant. Having said that, I agree using the new `addRemoteCandidate` might be idiomatically better.  Let me try.

enobufs:
opps, wrong agent used. fixing it.

Sean-Der:
Merged, nice work @enobufs :) Thanks for the hard work, it is so great to have stuff like this.

Also! You have access to the pions/webrtc repo, so you can push your branches directly here (if you want) you just can't push to master. That can only be done when someone else approves the PR.

enobufs:
@Sean-Der Thanks for allowing me to contribute to your great work.  Ok, I will create a branch directly in pions/webrtc next time - that makes my work easier. I actually tried it but I could not push my branch tho... I will try again.

PR #357, opened 2019-01-13
Quic fix

backkem:
Fix broken dependency that caused go get to fail finding qtls.

This is a bit random but fixed as follows:
- Take the [revision ](https://github.com/lucas-clemente/quic-go/blob/13067877038ad52efdd473e66fd52d81b9ed6708/vendor/vendor.json#L50) from go-quic's vendor.json.
- This leads to [this commit](https://github.com/marten-seemann/qtls/commit/26b223ad36d4436ed3eeb843041b66ac21dcee34). However, it turns out `go get` doesn't find commits in [some cases](https://github.com/golang/go/issues/27191).
- I knew they are following the [qtls](https://github.com/marten-seemann/qtls/tree/qtls) branch.
- I tried the latest commit in that branch but that doesn't work.
- I looked up the same commit on this branch (looks re-based), and found [this commit](https://github.com/marten-seemann/qtls/commit/7a090b30c3abf4651a84ee4fd7aa0774ec777054) which does works.
- With `go` command you can apply this as follows:
  - `go get -u github.com/marten-seemann/qtls@7a090b30c3abf4651a84ee4fd7aa0774ec777054`
  - `go mod tidy`


Sean-Der:
I think this will fail in CI @backkem :( I had the same exact on my SRTP branch

* Our linters will check the vendored code (and fail) I think we just need to set an explicit path maybe? 
* building the `examples` out of tree will fail as well (I think we just need to `go get` in each dir)

PR #370, opened 2019-01-21
Agent handles any inbound package (as opposed to STUN only)

hugoArregui:
This way, we update lastReceived properly, not just for STUN messages.

We are using lastReceived to check for a connection timeout (https://github.com/pions/webrtc/blob/master/pkg/ice/agent.go#L442) but the problem was we were not updating lastReceived for non STUN messages

### DIFF
--- a/pkg/ice/agent.go
+++ b/pkg/ice/agent.go
@@ -684,7 +684,7 @@ func (a *Agent) handleNewPeerReflexiveCandidate(local *Candidate, remote net.Add
 	return nil
 }
 
-// handleInbound processes traffic from a remote candidate
+// handleInbound processes STUN traffic from a remote candidate
 func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr) {
 	iceLog.Tracef("inbound STUN from %s to %s", remote.String(), local.String())
 	remoteCandidate := a.findRemoteCandidate(local.NetworkType, remote)
@@ -708,6 +708,14 @@ func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr
 		a.handleInboundControlling(m, local, remoteCandidate)
 	} else {
 		a.handleInboundControlled(m, local, remoteCandidate)
+	}
+}
+
+// noSTUNSeen processes non STUN traffic from a remote candidate
+func (a *Agent) noSTUNSeen(local *Candidate, remote net.Addr) {
+	remoteCandidate := a.findRemoteCandidate(local.NetworkType, remote)
+	if remoteCandidate != nil {
+		remoteCandidate.seen(false)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/ice/candidate.go
```diff
@@ -104,20 +102,12 @@ func (c *Candidate) recvLoop() {
...
-				agent.handleInbound(m, c, srcAddr)
-			})
-			if err != nil {
-				iceLog.Warnf("Failed to handle message: %v", err)
-			}
-
-			continue
+		m := buffer[:n]
+		err = c.agent.run(func(agent *Agent) {
+			agent.handleInbound(m, c, srcAddr)
```
backkem: There is a known problem here: `run` is executed asynchronously. This means the buffer will be used concurrently in the `recvLoop` and the code executed by `run`. This can cause corruption of the buffer. This change makes this problem worse since it now occurs with every package, not just the STUN packages.
We intend to solve the original problem by making a copy of the buffer before passing it to the `run` method. However, this solution doesn't seem appropriate for this case since we would have to make a copy for all incoming data and the copy would mostly not be used. I suggest to split this up again: have a separate call to `run` for user data that finds the right remote candidate and marks it as seen.

pkg/ice/agent.go
```diff
@@ -711,6 +711,15 @@ func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr
 	}
 }
 
+// NoSTUNSeen processes non STUN traffic from a remote candidate
+func (a *Agent) NoSTUNSeen(local *Candidate, remote net.Addr) {
```
backkem: I would keep this private:
```suggestion
func (a *Agent) noSTUNSeen(local *Candidate, remote net.Addr) {
```

pkg/ice/agent.go
```diff
@@ -711,6 +711,15 @@ func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr
...
+	iceLog.Tracef("inbound (not STUN) from %s to %s", remote.String(), local.String())
```
backkem: This will probably get overly rather crazy when a lot of data is sent.
```suggestion
```

pkg/ice/agent.go
```diff
@@ -711,6 +711,15 @@ func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr
 	}
 }
 
+// NoSTUNSeen processes non STUN traffic from a remote candidate
```
backkem:
```suggestion
// noSTUNSeen processes non STUN traffic from a remote candidate
```

pkg/ice/candidate.go
```diff
@@ -118,6 +118,13 @@ func (c *Candidate) recvLoop() {
 			}
 
 			continue
+		} else {
+			err := c.agent.run(func(agent *Agent) {
+				agent.NoSTUNSeen(c, srcAddr)
```
backkem:
```suggestion
				agent.noSTUNSeen(c, srcAddr)
```

hugoArregui:
@backkem done! I now realize there is a way to apply your suggestion directly in GH :-D

PR #372, opened 2019-01-23
Return error if AddIceCandidate is called but no remote description is set

hugoArregui:
Return proper error and prevent nil pointer dereference

### DIFF
--- a/rtcpeerconnection.go
+++ b/rtcpeerconnection.go
@@ -1063,6 +1063,10 @@ func (pc *RTCPeerConnection) RemoteDescription() *RTCSessionDescription {
 // to the existing set of candidates
 func (pc *RTCPeerConnection) AddIceCandidate(s string) error {
 	// TODO: AddIceCandidate should take RTCIceCandidateInit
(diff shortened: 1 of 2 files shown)


rtcpeerconnection.go
```diff
@@ -1063,6 +1063,10 @@ func (pc *RTCPeerConnection) RemoteDescription() *RTCSessionDescription {
 // to the existing set of candidates
 func (pc *RTCPeerConnection) AddIceCandidate(s string) error {
 	// TODO: AddIceCandidate should take RTCIceCandidateInit
+	if pc.CurrentRemoteDescription != nil {
```
mjmac: I don't think this test is correct... It's checking that pc.CurrentRemoteDescription is not nil and returning an error if it is not nil, which is the opposite of the intended effect if I understand correctly. I am getting errors from addition of trickle-ICE candidates with the code as it is currently.

I think the correct test here should be:

`if pc.RemoteDescription() == nil {`

This correctly evaluates whether or not pc.PendingRemoteDescription or pc.CurrentRemoteDescription is nil and returns an error if so.

PR #385, opened 2019-02-01
sfu example using 2 files

jinleileiking:
this example use sdp.txt and sdp1.txt to let the sfu program run. 
Very simple example.


Sean-Der:
I think this is a great feature @jinleileiking 

What do you think of making this work with every example? We can make it so every example can accept an SDP or `file://` maybe? So if the input was `file://sdp.txt` it would read that instead

Sean-Der:
Hey @jinleileiking talking to some other devs we should spin up a simple HTTP server on `:0` that has a form people can submit the SDPs on. We can leave the SDP pasting, but every example will just start with a message `You can submit your SDP on 127.0.0.1:$RANDOM_PORT` and we can also include instructions on how to use curl for people that are using servers etc...

* The issue with stdin is that we always are running up against length restrictions, so even with compression we will see random errors
* The issue with files on disk is that the examples all will sorta work differently, and we will create garbage files everywhere

I am happy to explain, add any guidance!

@backkem does that sound like a ok idea?

Sean-Der:
Hey @jinleileiking 

I am going to close this for now, but if you are still interested in getting involved I would love to work together!

Issue #180, opened 2018-09-26
Graceful shutdown


backkem:
We need to investigate graceful shutdown of a peer connection. E.g.: Close all tracks and data channels when closing the peer connection. In addition, we can provide API for:
- Closing/un-registering a track.
- Closing a data channel.

Re-negotiations, E.g.: adding/removing tracks or data channels at runtime, seem less crucial and can be tracked in a future issue.

backkem:
Closing since this is mostly covered by #389 and older work on the data channel side. Any last remains will be picked up during the redesign of the Track API (#405).

PR #388, opened 2019-02-02
RTCRtpReceiver.Stop() better teardown

rob-deutsch:
Closes the channels it creates and the underlying connections.

This is a contribution towards #180.

It would be appreciated it someone with more experience/knowledge than me could review.

Thanks!

### DIFF
--- a/rtcrtpreceiver.go
+++ b/rtcrtpreceiver.go
@@ -49,6 +49,7 @@ func (r *RTCRtpReceiver) Receive(parameters RTCRtpReceiveParameters) chan bool {
 		}()
 
 		srtpSession, err := r.transport.getSRTPSession()
+		defer srtpSession.Close()
 		if err != nil {
 			pcLog.Warnf("Failed to open SRTPSession, RTCTrack done for: %v %d \n", err, parameters.encodings.SSRC)
 			return
@@ -83,6 +84,7 @@ func (r *RTCRtpReceiver) Receive(parameters RTCRtpReceiveParameters) chan bool {
 			select {
 			case r.rtpOut <- &rtpPacket:
 			default:
+				break
 			}
 		}
 	}()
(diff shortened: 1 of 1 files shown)


Sean-Der:
Hey @rob-deutsch thanks for the PR!

The downside is that this makes it possible for users to create a panic when using the public API. I eventually want to add `Read` and `Write` to the RTCTrack (then it can handle these errors gracefully) we should probably add those methods and then make these members private before closing them (@backkem agree/disagree?)

I would love to see https://github.com/pions/webrtc/blob/master/rtcrtpsender.go#L31 torn down properly (and add some tests around that) would be happy to merge that right away!

rtcrtpreceiver.go
```diff
@@ -117,6 +120,7 @@ func (r *RTCRtpReceiver) Receive(parameters RTCRtpReceiveParameters) chan bool {
 			select {
 			case r.rtcpOut <- rtcpPacket:
 			default:
+				break
```
Sean-Der: I don't believe we need/want this `break`

This exists so when giving packets to the user we discard them if they aren't listening/running behind. We should probably make this a switch, but currently it is lossy so we can always be live.

rtcrtpreceiver.go
```diff
@@ -49,6 +49,7 @@ func (r *RTCRtpReceiver) Receive(parameters RTCRtpReceiveParameters) chan bool {
 		}()
 
 		srtpSession, err := r.transport.getSRTPSession()
+		defer srtpSession.Close()
```
Sean-Der: nice!

Mind moving this after the `err` handling, we don't need to close if this fails.

rtcrtpreceiver.go
```diff
@@ -127,4 +131,6 @@ func (r *RTCRtpReceiver) Receive(parameters RTCRtpReceiveParameters) chan bool {
 // Stop irreversibly stops the RTCRtpReceiver
 func (r *RTCRtpReceiver) Stop() {
```
Sean-Der: Can you add a `mu Sync.mutex` and `closed bool` member to `RTCRtpReceiver`

Then only allow people to call close once, and if it is already closed just early return. Calling `close()` twice on a channel causes a panic

Sean-Der:
Closed in favor of #389

Thanks for all the work on this! I added you as a co-author on that commit, I used some of your work on that one.

thanks

PR #391, opened 2019-02-05
Fix Codacy warnings

Sean-Der:
Run standardjs across all js files, fix all other issues by hand


Sean-Der:
@kc5nra can you approve my CR

PR #396, opened 2019-02-10
Fixed minor pipeline/element memory leak in examples/util

### LINKED ISSUE #394: Pipeline reference leak in examples/util/gstreamer-src
https://github.com/pions/webrtc/blob/95fcd7b3a03677c7ccd850c8750990ae07d2afd0/examples/util/gstreamer-src/gst.c#L76-L81

According to [this thread](http://gstreamer-devel.966125.n4.nabble.com/delete-a-pipeline-the-right-way-td4684015.html); "`gst_bin_get_by_name()` increases the refcount to your appsink. After you set your appsink callbacks/signals you should also `gst_object_unref()` this object again."

Should `gst_object_unref()` be called?

rob-deutsch:
When an element is selected by `gst_bin_get_by_name` we need to make sure that we also `gst_object_unref` it.

Fixes #394

### DIFF
--- a/examples/util/gstreamer-sink/gst.c
+++ b/examples/util/gstreamer-sink/gst.c
@@ -57,5 +57,6 @@ void gstreamer_receive_push_buffer(GstElement *pipeline, void *buffer, int len) 
     gpointer p = g_memdup(buffer, len);
     GstBuffer *buffer = gst_buffer_new_wrapped(p, len);
     gst_app_src_push_buffer(GST_APP_SRC(src), buffer);
+    gst_object_unref(src);
   }
 }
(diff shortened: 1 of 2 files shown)


examples/util/gstreamer-sink/gst.c
```diff
@@ -58,4 +58,5 @@ void gstreamer_receive_push_buffer(GstElement *pipeline, void *buffer, int len)
     GstBuffer *buffer = gst_buffer_new_wrapped(p, len);
     gst_app_src_push_buffer(GST_APP_SRC(src), buffer);
   }
+  gst_object_unref(src);
```
Sean-Der: Can you move this inside the `src != null` check, thanks!

PR #402, opened 2019-02-13
Set pions/dtls SRTPProtectionProfiles

Sean-Der:
SRTP Profile used to be set implicitly, but pions/dtls was updated to
require setting it explicitly. For now we only support one profile so
we don't need to query it later.


Sean-Der:
@backkem I won't be landing this for a day or so also, don't wanna merge pions/dtls until I am positive it fixed the reporters issue!

I will tag `github.com/pions/dtls v1.2.0` when I push this to master also

PR #412, opened 2019-02-16
Gstreamer-receive reserves main thread for Glib loop

rob-deutsch:
The Glib loop is run in the main thread (in the OS point of view)

This is a simpler solution than [Russ Cox's solution](https://groups.google.com/forum/#!msg/golang-nuts/IiWZ2hUuLDA/SNKYYZBelsYJ) because of the simpler nature of gstreamer-receive.

### DIFF
--- a/examples/gstreamer-receive/main.go
+++ b/examples/gstreamer-receive/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


Sean-Der:
nice work, thanks again for digging into this. I think you just need to shorten the commit subject and it should be all green!

PR #414, opened 2019-02-17
Enable certificate validation when server

### LINKED ISSUE #413: Enable peer certificate validation when running as DTLS Server
Currently this isn't a fatal error if the remote doesn't offer a certificate

Sean-Der:
We currently only check the peer certificate when running as a client.
pions/dtls added support for ClientAuth, so start using this feature and
checking the remote certificate.

Resolves #413


Sean-Der:
@backkem @kixelated all the tests pass, just having issues uploading the coverage results.

rtcdtlstransport.go
```diff
@@ -181,7 +181,12 @@ func (t *RTCDtlsTransport) Start(remoteParameters RTCDtlsParameters) error {
 	// TODO: handle multiple certs
 	cert := t.certificates[0]
 
-	dtlsCofig := &dtls.Config{Certificate: cert.x509Cert, PrivateKey: cert.privateKey, SRTPProtectionProfiles: []dtls.SRTPProtectionProfile{dtls.SRTP_AES128_CM_HMAC_SHA1_80}}
+	dtlsCofig := &dtls.Config{
+		Certificate:            cert.x509Cert,
+		PrivateKey:             cert.privateKey,
+		SRTPProtectionProfiles: []dtls.SRTPProtectionProfile{dtls.SRTP_AES128_CM_HMAC_SHA1_80},
+		ClientAuth:             dtls.RequireAnyClientCert,
```
kixelated: Does this always require client certs? Is that supposed to be a requirement?

Sean-Der: Yep it is a requirement!

The peers should always validate the certificate using the fingerprint in the SDP

PR #425, opened 2019-02-20
Examples: Make examples/util internal

### LINKED ISSUE #424: Make examples/util internal
Avoid exposing examples/util. We've seen this leaking into other projects which is definitely not intended.

backkem:
Resolves #424


maxhawkins:
+1 on moving to internal.

Do we need lazy.Check at all? Go programmers are used to a lot of `if err != nil` noise. Would make it easier for people to copy-paste.

backkem:
Yea, I guess you're right.

PR #429, opened 2019-02-21
API: Review DataChannel

### LINKED ISSUE #427: Review DataChannel API
Since we're breaking the API it's a good time to review the data channel API. Known limitations of the current API:
- You have to import `pkg/datachannel`.
- The binary/string type switch isn't handy.
- The payload size is hardcoded.

Some alternatives are already explored in #364 & #365.

backkem:
This seems to make sense. I didn't address the hardcoded payload size issue. This is inherit to the WebRTC API and can be circumvented by detaching the underlying pions/datachannel.

Resolves #427


backkem:
I guess this does mean you'll lose message order between string & binary messages. Not sure how important that is.

maxhawkins:
> I guess this does mean you'll lose message order between string & binary messages. Not sure how important that is.

I could see that being a big downside depending on the application.

What about this?

```go
dc.OnMessage(func(data []byte, binary bool) {
  if binary {
    fmt.Printf("%x", data)
  } else {
    fmt.Printf("%s", string(data))
  }
})
```
A `[]byte` is kind of like a string right?


Alternatively we could draw inspiration from [gorilla/websocket](https://godoc.org/github.com/gorilla/websocket). They use a reader rather than a callback and an integer message type:

```go
for {
    messageType, data, err := conn.ReadMessage()
    if err != nil {
        panic(err)
    }
    if messageType == websocket.BinaryMessage {
        fmt.Printf("%x", data)
    } else {
        fmt.Printf("%s", string(data))
    }
}
```

Sean-Der:
I think we should keep the callbacks, just so it looks like WebRTC as much as possible.

Another thing you could do @backkem is have the output be more like https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/onmessage and be a `MessageEvent` (or something like that) with `[]byte` and `type`

I also worry that people will copy examples and only define one CB (and then miss the other)

maxhawkins:
One problem with a callback in Go vs Javascript is that Go is multithreaded and Javascript is not.

The callback style is more comfortable in Javascript because you know one callback invocation is going to complete before the next one fires. You don't get this kind of guarantee in Go.

It's hard to reason about (and not intuitive with a callback api) what will happen if a message arrives before the previous one has finished processing. Especially if the user expects an ordered datachannel this could lead to unexpected bugs.

Not sure if that outweighs the confusion people might face while porting javascript code into Go.

backkem:
We already have the io.ReadWriter style API I'd you detach the datachannel. I'll try one of your suggestions next.

backkem:
I went for the struct approach, let me know what you think.

PR #433, opened 2019-02-21
Remove global API for settings engine

backkem:
Limit global API to the basics needed to construct a PeerConnection.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,9 +1,5 @@
 package webrtc
 
-import (
-	"time"
-)
-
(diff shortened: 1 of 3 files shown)


api.go
```diff
@@ -53,26 +49,6 @@ func WithSettingEngine(s SettingEngine) func(a *API) {
 // in the future.
 var defaultAPI = NewAPI()
```
maxhawkins: This can go away right?

backkem: Yea, next PR.

Issue #434, opened 2019-02-21
No globals


backkem:
This ticket tracks avoiding globals in the library.

**defaultAPI**
We currently use a global variable `defaultAPI` in `api.go` to enable a global API. This global API allows for simple examples covering the basic use-cases. This reduces the learning curve of the library.
In these basic cases we mostly do:
```go
webrtc.RegisterDefaultCodecs()
pc, err := webrtc.NewPeerConnection(config)
```
I suggest to change the global `NewPeerConnection` so it:
- Constructs it's own API object.
- Always registers the default codecs, supporting most browsers.

This way we can avoid the `defaultAPI` global variable and still have a nice API for simple use-cases. Users with more specific use-cases can construct their own API object and configure it to their needs.
Example:
```go
// Create a MediaEngine and set your codecs
m := webrtc.MediaEngine{}
m.RegisterCodec(webrtc.NewRTPOpusCodec(webrtc.DefaultPayloadTypeOpus, 48000, 2))

// Create an API object with the engine
api := webrtc.NewAPI(webrtc.WithMediaEngine(m))

// Create the PeerConnection
peerConnection, err := api.NewPeerConnection(config)

// Or use ORTC
gatherer, err := api.NewICEGatherer(iceOptions)
```

**Logging**
See #361

TODO:
- [x] Avoid `DefaultAPI`
- [x] Avoid global logging
- [ ] Avoid MuxFuncs as global variables
- [ ] Enable `gochecknoinits` & `gochecknoglobals` linters

maxhawkins:
This makes sense to me. Global state is bad news.

I'd like to propose making the settings stuff more explicit by passing it in as a parameter. That way it will be more discoverable via godoc for people just discovering the library. It also has the benefit that if we rename or remove a parameter users will be notified via their type checker.

Here's an API that would eliminate the need for `SettingsEngine` and `MediaEngine` by providing an explicit `Settings` object.

`webrtc.Settings` would contain all the non-RFC options that can be passed to a peer connection.

```go
pc, err := webrtc.NewPeerConnection(
    webrtc.Configuration{
        IceServers: iceServers,
        RTCPMuxPolicy: webrtc.RTCPMuxPolicyNegotiate
    },
    &webrtc.Options{
        Codecs: webrtc.DefaultCodecs, // omitting this defaults to webrtc.DefaultCodecs
        UDPPortMin: 8080,
        UDPPortMax: 0, // omitting or setting to zero means there is no max
[...]

PR #436, opened 2019-02-21
Avoid defaultAPI

backkem:
Relates to #434

I'll leave this open for feedback a little while longer.


Sean-Der:
Nice, really love this. 

I think we do need to start taking documentation more seriously. Especially with the non-standard stuff, maybe it would be helpful to have just a page for `where we diverge`

PR #437, opened 2019-02-22
Move to new Track API

### LINKED ISSUE #405: Pushing to RTCTrack can cause panics/drops
The current interface uses the channels `RTCTrack.Samples chan<- media.RTCSample` and `RTCTrack.RawRTP chan<- *rtp.Packet` to send data over the connection.

If the connection is closed, these channels are closed. Sending to a closed channel will cause the code to panic. I've also seen a lot of packet drops because the buffered channel (size 15) is not large enough to handle surges. 

I would love to change the API such that you call `WriteSample` or `WritePacket` instead of interfacing with channels. I also don't think this should be part of the `RTPTrack`, but instead the `RTCRtpSender` returned by `AddTrack`.

This should also be a blocking call which would remove a goroutine and some context switching. It would go a long way for my project which creates thousands of peer connections.

Sean-Der:
The Track API has been rewritten to remove Channels from the public API. This was done because we ran into the following issues.

* Use tight loops to read/write from Channels to exchange RTP/RTCP packets is more expensive then I realized
* We were unable to return errors when reading/write RTP or RTCP
* We were unable to close channels (or risk causing a panic if the user writes to them)

We also gained some unforeseen benefits moving to the new API.

* A Track can be added to multiple `PeerConnections` now. The most common use-case so far has been building SFUs, and this feature will reduce complexity for everyone. This also more closely follows the browser implementation of WebRTC.

# Changes

### Before
```
peerConnection.OnTrack(func(track *webrtc.Track) {
})
```

### After
```
peerConnection.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
})
```

### Before
```
peerConnection.OnTrack(func(track *webrtc.Track) {
    <-track.RTCPPackets
})
```

### After
```
peerConnection.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
    pkt, header, err := receiver.ReadRTCP(make([]byte, 1400))
})
````

### Before
```
peerConnection.OnTrack(func(track *webrtc.Track) {
    <-track.Packets
})
```

### After
```
peerConnection.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
    pkt, err := track.ReadRTP(make([]byte, 1400))
})
````

### Before
```
peerConnection.OnTrack(func(track *webrtc.Track) {
    vp8Track.RawRTP <- &rtp.Packet{}
})
```

### After
```
peerConnection.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
                       if i, err := track.WriteRTP(&rtp.Packet{})
})
````

## Raw and Sample tracks now share a constructor
The only behavior difference is that the user must supply a SSRC for Sample tracks. A random uint32 is all that should be needed in most cases.

### Before
```
peerConnection.NewRawRTPTrack(outboundPayloadType, outboundSSRC, "video", "pion")
pcOffer.NewSampleTrack(DefaultPayloadTypeVP8, "video", "pion")
```

### After
```
peerConnection.NewTrack(outboundPayloadType, outboundSSRC, "video", "pion")
```

### DIFF
--- a/datachannel_test.go
+++ b/datachannel_test.go
@@ -156,8 +156,10 @@ func TestDataChannel_MessagesAreOrdered(t *testing.T) {
 	out := make(chan int)
 	inner := func(msg DataChannelMessage) {
 		// randomly sleep
-		// NB: The big.Int/crypto.Rand is overkill but makes the linter happy
+		// math/rand a weak RNG, but this does not need to be secure. Ignore with #nosec
+		/* #nosec */
 		randInt, err := rand.Int(rand.Reader, big.NewInt(int64(max)))
+		/* #nosec */
 		if err != nil {
 			t.Fatalf("Failed to get random sleep duration: %s", err)
 		}
--- a/examples/gstreamer-send/main.go
+++ b/examples/gstreamer-send/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"math/rand"
 
 	"github.com/pions/webrtc"
 
@@ -39,7 +40,7 @@ func main() {
 	})
 
 	// Create a audio track
-	opusTrack, err := peerConnection.NewSampleTrack(webrtc.DefaultPayloadTypeOpus, "audio", "pion1")
+	opusTrack, err := peerConnection.NewTrack(webrtc.DefaultPayloadTypeOpus, rand.Uint32(), "audio", "pion1")
 	if err != nil {
 		panic(err)
 	}
@@ -49,7 +50,7 @@ func main() {
 	}
 
 	// Create a video track
-	vp8Track, err := peerConnection.NewSampleTrack(webrtc.DefaultPayloadTypeVP8, "video", "pion2")
+	vp8Track, err := peerConnection.NewTrack(webrtc.DefaultPayloadTypeVP8, rand.Uint32(), "video", "pion2")
 	if err != nil {
 		panic(err)
 	}
@@ -84,8 +85,8 @@ func main() {
 	fmt.Println(signal.Encode(answer))
 
 	// Start pushing buffers on these tracks
-	gst.CreatePipeline(webrtc.Opus, opusTrack.Samples, *audioSrc).Start()
-	gst.CreatePipeline(webrtc.VP8, vp8Track.Samples, *videoSrc).Start()
+	gst.CreatePipeline(webrtc.Opus, opusTrack, *audioSrc).Start()
+	gst.CreatePipeline(webrtc.VP8, vp8Track, *videoSrc).Start()
 
 	// Block forever
 	select {}
--- a/lossy_stream.go
+++ b/lossy_stream.go
@@ -1,0 +1,93 @@
+package webrtc
+
+import (
+	"fmt"
+	"io"
+	"sync"
+)
+
+// lossyReader wraps an io.Reader and discards data if it isn't read in time
+// Allowing us to only deliver the newest data to the caller
+type lossyReadCloser struct {
+	nextReader io.ReadCloser
+	mu         sync.RWMutex
+
+	incomingBuf chan []byte
+	amountRead  chan int
+
+	readError  error
+	hasErrored chan interface{}
+
+	closed chan interface{}
+}
+
+func newLossyReadCloser(nextReader io.ReadCloser) *lossyReadCloser {
+	l := &lossyReadCloser{
+		nextReader: nextReader,
+
+		closed: make(chan interface{}),
+
+		incomingBuf: make(chan []byte),
+		hasErrored:  make(chan interface{}),
+		amountRead:  make(chan int),
+	}
+
+	go func() {
+		readBuf := make([]byte, receiveMTU)
+		for {
+			i, err := nextReader.Read(readBuf)
+			if err != nil {
+				l.mu.Lock()
+				l.readError = err
+				l.mu.Unlock()
+
+				close(l.hasErrored)
+				break
+			}
+
+			select {
+			case in := <-l.incomingBuf:
+				copy(in, readBuf[:i])
+				l.amountRead <- i
+			default: // Discard if we have no inbound read
+			}
+		}
+	}()
+
+	return l
+}
+
+func (l *lossyReadCloser) Read(b []byte) (n int, err error) {
+	select {
+	case <-l.closed:
+		return 0, fmt.Errorf("lossyReadCloser is closed")
+	case <-l.hasErrored:
+		l.mu.RLock()
+		defer l.mu.RUnlock()
+		return 0, l.readError
+
+	case l.incomingBuf <- b:
+	}
+
+	select {
+	case <-l.closed:
+		return 0, fmt.Errorf("lossyReadCloser is closed")
+	case <-l.hasErrored:
+		l.mu.RLock()
+		defer l.mu.RUnlock()
+		return 0, l.readError
+
+	case i := <-l.amountRead:
+		return i, nil
+	}
+}
+
+func (l *lossyReadCloser) Close() error {
+	select {
+	case <-l.closed:
+		return fmt.Errorf("lossyReader is already closed")
+	default:
+	}
+	close(l.closed)
+	return l.nextReader.Close()
+}
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/pions/rtcp"
-	"github.com/pions/rtp"
 	"github.com/pions/sdp/v2"
 	"github.com/pions/webrtc/pkg/ice"
 	"github.com/pions/webrtc/pkg/logging"
@@ -103,7 +102,7 @@ type PeerConnection struct {
 
 	onSignalingStateChangeHandler     func(SignalingState)
 	onICEConnectionStateChangeHandler func(ICEConnectionState)
-	onTrackHandler                    func(*Track)
+	onTrackHandler                    func(*Track, *RTPReceiver)
 	onDataChannelHandler              func(*DataChannel)
 
 	iceGatherer   *ICEGatherer
@@ -279,13 +278,13 @@ func (pc *PeerConnection) OnDataChannel(f func(*DataChannel)) {
 
 // OnTrack sets an event handler which is called when remote track
 // arrives from a remote peer.
-func (pc *PeerConnection) OnTrack(f func(*Track)) {
+func (pc *PeerConnection) OnTrack(f func(*Track, *RTPReceiver)) {
 	pc.mu.Lock()
 	defer pc.mu.Unlock()
 	pc.onTrackHandler = f
 }
 
-func (pc *PeerConnection) onTrack(t *Track) (done chan struct{}) {
+func (pc *PeerConnection) onTrack(t *Track, r *RTPReceiver) (done chan struct{}) {
 	pc.mu.RLock()
 	hdlr := pc.onTrackHandler
 	pc.mu.RUnlock()
@@ -298,7 +297,7 @@ func (pc *PeerConnection) onTrack(t *Track) (done chan struct{}) {
 	}
 
 	go func() {
-		hdlr(t)
+		hdlr(t, r)
 		close(done)
 	}()
 
@@ -856,18 +855,21 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 			return
 		}
 
-		if pc.onTrackHandler != nil {
-			pc.openSRTP()
-		} else {
-			pcLog.Warnf("OnTrack unset, unable to handle incoming media streams")
-		}
+		pc.openSRTP()
 
 		for _, tranceiver := range pc.rtpTransceivers {
 			if tranceiver.Sender != nil {
-				tranceiver.Sender.Send(RTPSendParameters{
+				err = tranceiver.Sender.Send(RTPSendParameters{
 					encodings: RTPEncodingParameters{
-						RTPCodingParameters{SSRC: tranceiver.Sender.Track.SSRC, PayloadType: tranceiver.Sender.Track.PayloadType},
+						RTPCodingParameters{
+							SSRC:        tranceiver.Sender.track.SSRC(),
+							PayloadType: tranceiver.Sender.track.PayloadType(),
+						},
 					}})
+
+				if err != nil {
+					pcLog.Warnf("Failed to start Sender: %s", err)
+				}
 			}
 		}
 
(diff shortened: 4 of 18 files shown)

### CODE REVIEW

File: track.go [hunk 1]
```diff
@@ -1,76 +1,173 @@
...
+	t.mu.RLock()
+	if t.receiver != nil {
+		t.mu.RUnlock()
+		return 0, fmt.Errorf("this is a remote track and must not be written to")
+	}
+	senders := t.senders
+	t.mu.RUnlock()
+
+	for _, s := range senders {
+		if _, err := s.sendRTP(append([]byte{}, b...)); err != nil {
```
Sean-Der: TODO -- Should we always copy when write is called?

track.go
```diff
@@ -1,76 +1,173 @@
...
+
+// WriteSample packetizes and writes to the track
+func (t *Track) WriteSample(s media.Sample) error {
+	packets := t.packetizer.Packetize(s.Data, s.Samples)
+	for _, p := range packets {
+		buf, err := p.Marshal()
+		if err != nil {
+			return err
+		}
+		if _, err := t.Write(append([]byte{}, buf...)); err != nil {
```
Sean-Der: TODO -- Should we always copy when write is called?

.golangci.yml
```diff
@@ -7,6 +7,7 @@ linters-settings:
 linters:
   enable-all: true
   disable:
+    - dupl
```
Sean-Der: I will fix this before I land. The lossy reading in Sender/Receiver is duplicated

track.go
```diff
@@ -1,76 +1,173 @@
...
 	"github.com/pions/rtp"
 	"github.com/pions/webrtc/pkg/media"
-	"github.com/pkg/errors"
 )
 
-// Track represents a track that is communicated
+const rtpOutboundMTU = 1400
+
+// Track represents a single media track
 type Track struct {
```
wdouglass: If all of the members are unexposed, would it make sense for this to be `track` and make `Track` an interface? I'm not sure why, but that *feels* like the right thing to do here...

Sean-Der: Yea I don't have an issue with it!

Is the goal to make code easier for the reader? I don't think we will have an internal behavior differences. 

In the future we could also add different Track behaviors (people could pass their own also) which might be interesting!

wdouglass: Yeah, i think the smaller we keep the public interface, the easier it is to read. Also, the "different track behaviors" idea is interesting (but i'm not sure how it would work, or why)

Sean-Der: You could then maybe create a `Track` constructor that automatically pulled from a webcam. So if someone called `camera.NewTrack()` it would return something that users don't have to do the read/write on at all. But would implement our interface.

Just spitballing, no idea if that would all work out (or even make sense)

rtpreceiver.go
```diff
@@ -14,149 +13,190 @@ type RTPReceiver struct {
...
+	select {
+	case i := <-r.rtcpReadLen:
+		return i, nil
+	case <-r.chansDone:
+		return 0, fmt.Errorf("RTPReceiver is stopped")
+	}
+}
+
+// ReadRTCP is a convenience method that wraps Read and unmarshals for you
+func (r *RTPReceiver) ReadRTCP(b []byte) (rtcp.Packet, rtcp.Header, error) {
```
maxhawkins: `rtcp.Packet` is an interface that includes `Header()` so we don't need to split out like this anymore. You can just return the packet.

kixelated: YAY

track.go
Sean-Der: I had a pretty hard time making this happen, there are some things I don't want to export that would need to be to make this happen `determinePayloadType` and setting the values of members? Mind if we table this for now.

If you feel strongly I can try some other stuff with this!

peerconnection.go
```diff
@@ -949,18 +968,18 @@ func (pc *PeerConnection) openSRTP() {
...
+			receiver.Track().mu.Lock()
+			receiver.Track().kind = codec.Type
+			receiver.Track().codec = codec
+			receiver.Track().mu.Unlock()
 
-			pc.onTrack(receiver.Track)
+			if pc.onTrackHandler != nil {
+				pc.onTrack(receiver.Track(), receiver)
+			} else {
+				pcLog.Warnf("OnTrack unset, unable to handle incoming media streams")
```
mjmac: Would it make more sense to check this right away at the top of the function and return an error if it's not set? Seems like it might reduce user confusion if they forget to set an OnTrack handler.

rtpsender.go
```diff
@@ -1,154 +1,143 @@
...
-		packets := packetizer.Packetize(in.Data, in.Samples)
-		for _, p := range packets {
-			r.sendRTP(p)
+// Stop irreversibly stops the RTPSender
+func (r *RTPSender) Stop() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.stopCalled {
+		return fmt.Errorf("Stop has already been called")
```
mjmac: If Stop() was already called is it really an error to call it again? Could just exit early. Likewise for Close()...

track.go
mjmac: It's least-surprising. I'd vote for keeping it this way for now and exploring performance optimizations later.

kixelated: `io.Writer: Write must not modify the slice data, even temporarily`. Same is true when you write to a `net.Conn` and other stuff like that. Not always true when dealing with pions which should be fixed.

It's better this way because it's on the writer to perform the copy when it knows it will modify the data. Otherwise you have cases like there where I have to assume the underlying implementation will modify my data.

track.go
```diff
@@ -1,76 +1,186 @@
...
+	return nil
+}
+
+// WriteRTP writes RTP packets to the track
+func (t *Track) WriteRTP(p *rtp.Packet) error {
+	buf, err := p.Marshal()
+	if err != nil {
+		return err
+	}
+	if _, err := t.Write(append([]byte{}, buf...)); err != nil {
```
kixelated: Same comment for this entire file. `Write` should never modify the given buffer.

In fact, this append is completely unnecessary and a good example of why this convention exists.

track.go
Sean-Der: @kixelated this is currently the behavior, so I don't think anything to do here.

It would be nice if Go had move semantics, or some official way to pass ownership of a buffer.

examples/gstreamer-send/main.go
```diff
@@ -38,8 +40,13 @@ func main() {
 		fmt.Printf("Connection State has changed %s \n", connectionState.String())
 	})
 
+	randSSRC := make([]byte, 4)
+	if _, err = rand.Read(randSSRC); err != nil {
```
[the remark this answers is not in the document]
Sean-Der: Fixed with https://github.com/pions/webrtc/pull/437/commits/0bdcaa9d71d70d32b224a1c60d4d33c61541b4f7

lossy_stream.go
```diff
@@ -0,0 +1,93 @@
...
+				l.amountRead <- i
+			default: // Discard if we have no inbound read
+			}
+		}
+	}()
+
+	return l
+}
+
+func (l *lossyReadCloser) read(b []byte) (n int, err error) {
```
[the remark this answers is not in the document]
Sean-Der: `lossyReadCloser` isn't public so just hiding everything. Will make uppercase

rtpreceiver.go
```diff
@@ -5,185 +5,141 @@ import (
...
+	select {
+	case <-r.closed:
+		return 0, fmt.Errorf("RTPSender has been stopped")
+	case <-r.received:
+		return r.rtcpReadStream.read(b)
+	}
+}
+
+// ReadRTCP is a convenience method that wraps Read and unmarshals for you
+func (r *RTPReceiver) ReadRTCP(b []byte) (rtcp.Packet, error) {
```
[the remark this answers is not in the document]
Sean-Der: nice, yea when we have that can drop all of these cases

rtpreceiver.go
```diff
@@ -5,185 +5,141 @@ import (
...
+	return nil
+}
+
+// Read reads incoming RTCP for this RTPReceiver
+func (r *RTPReceiver) Read(b []byte) (n int, err error) {
+	select {
+	case <-r.closed:
+		return 0, fmt.Errorf("RTPSender has been stopped")
+	case <-r.received:
+		return r.rtcpReadStream.read(b)
```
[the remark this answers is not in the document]
Sean-Der: What else should it be reading? 

This  follows the ORTC/WebRTC model, so we will never have `RTP` available via the `RTPSender` or `RTPReceiver` not something we can change.

track.go
Sean-Der: Fixed, SRTP actually allocates for us I forgot https://github.com/pions/webrtc/pull/437/commits/e112cd42c71a9afe8fdc34ec3fc83b4e292270e4

rtpreceiver.go
```diff
@@ -5,185 +5,141 @@ import (
...
-	<-r.rtcpOutDone
-	<-r.rtpOutDone
-
-	r.closed = true
+	close(r.closed)
 	return nil
 }
+
+// readRTP should only be called by a track, this only exists so we can keep state in one place
+func (r *RTPReceiver) readRTP(b []byte) (n int, err error) {
```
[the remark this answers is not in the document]
Sean-Der: I think `rtpengine` will be pluggable on startup. Using the `api` stuff.

Sorry I don't see that happening :( it just breaks how other implementations behave, the fanout exists in every one but pions/webrtc right now :(

rtpsender.go
Sean-Der: Agree! Fixed with https://github.com/pions/webrtc/pull/437/commits/a40fc1f385f17a35285edf720bfa33c9ac9fe243

peerconnection.go
```diff
@@ -856,18 +855,16 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
 		for _, tranceiver := range pc.rtpTransceivers {
 			if tranceiver.Sender != nil {
-				tranceiver.Sender.Send(RTPSendParameters{
+				if err = tranceiver.Sender.Send(RTPSendParameters{
 					encodings: RTPEncodingParameters{
-						RTPCodingParameters{SSRC: tranceiver.Sender.Track.SSRC, PayloadType: tranceiver.Sender.Track.PayloadType},
-					}})
+						RTPCodingParameters{SSRC: tranceiver.Sender.track.SSRC(), PayloadType: tranceiver.Sender.track.PayloadType()},
+					}}); err != nil {
+					pcLog.Warnf("Failed to start Sender: %s", err)
```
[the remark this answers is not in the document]
Sean-Der: we can't return an error unfortunately, this is in a detached thread :(

We do need to fix that though, I would like to make it so no errors happen in a background thread and we just leave the user hanging

peerconnection.go
```diff
@@ -856,18 +855,16 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 			return
 		}
 
-		if pc.onTrackHandler != nil {
-			pc.openSRTP()
-		} else {
-			pcLog.Warnf("OnTrack unset, unable to handle incoming media streams")
-		}
+		pc.openSRTP()
```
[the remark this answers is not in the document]
Sean-Der: `openSRTP` doesn't return any errors, everything is done in detached threads.

I would love to make a larger story out of `How do we handle background errors` though!

rtpsender.go
```diff
@@ -1,154 +1,144 @@
...
 }
 
 // Send Attempts to set the parameters controlling the sending of media.
-func (r *RTPSender) Send(parameters RTPSendParameters) {
-	if r.Track.isRawRTP {
-		go r.handleRawRTP(r.Track.rawInput)
-	} else {
-		go r.handleSampleRTP(r.Track.sampleInput)
+func (r *RTPSender) Send(parameters RTPSendParameters) error {
+	r.mu.Lock()
```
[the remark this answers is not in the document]
Sean-Der: This one is a little harder to fix, I am trying to make it so one person can call `Send` at a time also. 

I only close the channel if no errors have happened. I don't think this is worth fixing yet, but if you think so I can look at it again!

peerconnection.go
```diff
@@ -856,18 +855,16 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
 
 		for _, tranceiver := range pc.rtpTransceivers {
 			if tranceiver.Sender != nil {
-				tranceiver.Sender.Send(RTPSendParameters{
+				if err = tranceiver.Sender.Send(RTPSendParameters{
 					encodings: RTPEncodingParameters{
-						RTPCodingParameters{SSRC: tranceiver.Sender.Track.SSRC, PayloadType: tranceiver.Sender.Track.PayloadType},
-					}})
+						RTPCodingParameters{SSRC: tranceiver.Sender.track.SSRC(), PayloadType: tranceiver.Sender.track.PayloadType()},
```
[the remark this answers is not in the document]
Sean-Der: Fixed with a40fc1f385f17a35285edf720bfa33c9ac9fe243

peerconnection.go
```diff
@@ -931,15 +928,37 @@ func (pc *PeerConnection) openSRTP() {
 
 	for i := range incomingSSRCes {
 		go func(ssrc uint32, codecType RTPCodecType) {
-			receiver := pc.api.NewRTPReceiver(codecType, pc.dtlsTransport)
-			<-receiver.Receive(RTPReceiveParameters{
+			receiver, err := pc.api.NewRTPReceiver(codecType, pc.dtlsTransport)
+			if err != nil {
+				pcLog.Warnf("Could not create RTPReceiver %s", err)
```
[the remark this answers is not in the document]
Sean-Der: Agree, we will work on it. I don't have any plans yet, but we can make it our next big thing post `v2.0.0`

rtpsender.go
```diff
@@ -1,154 +1,143 @@
...
+	case r.stopCalled:
+		return 0, fmt.Errorf("RTPReceiver is stopped")
+	default:
+		return r.rtcpReadStream.read(b)
 	}
+}
 
-	readStream, err := srtcpSession.OpenReadStream(r.Track.SSRC)
+// ReadRTCP is a convenience method that wraps Read and unmarshals for you
+func (r *RTPSender) ReadRTCP(b []byte) (rtcp.Packet, rtcp.Header, error) {
```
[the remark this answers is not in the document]
Sean-Der: Fixed with 2d3dec5

lossy_stream.go
```diff
@@ -0,0 +1,93 @@
...
+
+				close(l.hasErrored)
+				break
+			}
+
+			select {
+			case in := <-l.incomingBuf:
+				copy(in, readBuf[:i])
+				l.amountRead <- i
+			default: // Discard if we have no inbound read
```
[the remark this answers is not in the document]
Sean-Der: Mind if we create an issue and do this in another PR, I don't think we can solve this short term without discussing a bit more (and it isn't a regression)

I think this ties into designing a good API + `rtpengine` not a hard problem, but I don't see us solving it in the next 48 hours

rtpreceiver.go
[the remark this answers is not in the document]
kixelated: Yeah, you can have a `sync.Pool` of buffers, reusing them instead of allocating.

lossy_stream.go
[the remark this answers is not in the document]
kixelated: I mean, if you're comfortable enough with the state of the code to merge, then sure go for it. I don't think there's a rush but it's your repo.

datachannel_test.go
```diff
@@ -157,7 +157,9 @@ func TestDataChannel_MessagesAreOrdered(t *testing.T) {
 	inner := func(msg DataChannelMessage) {
 		// randomly sleep
 		// NB: The big.Int/crypto.Rand is overkill but makes the linter happy
```
wdouglass: i would change this to read `The big.Int/crypto.Rand is overkill, so disable that rule with #nosec` or something like that

(i had to look up what #nosec means)

lossy_stream.go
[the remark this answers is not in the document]
wdouglass: I think i agree with Sean here, since it isn't a regression (it's new functionality), something that works (but may be sub-optimal) can always be improved upon later

lossy_stream.go
```diff
@@ -0,0 +1,93 @@
...
+		l.mu.RLock()
+		defer l.mu.RUnlock()
+		return 0, l.readError
+
+	case i := <-l.amountRead:
+		return i, nil
+	}
+}
+
+func (l *lossyReadCloser) close() error {
```
wdouglass: same as @kixelated said above, make this `Close` so it's an `io.ReadCloser`

peerconnection_media_test.go
```diff
@@ -158,38 +168,41 @@ This test adds an input track and asserts
...
 		t.Fatal(err)
 	}
 
-	opusTrack, err := pcOffer.NewSampleTrack(DefaultPayloadTypeOpus, "audio", "pion1")
+	opusTrack, err := pcOffer.NewTrack(DefaultPayloadTypeOpus, rand.Uint32(), "audio", "pion1")
 	if err != nil {
 		t.Fatal(err)
 	}
-	vp8Track, err := pcOffer.NewSampleTrack(DefaultPayloadTypeVP8, "video", "pion2")
+	vp8Track, err := pcOffer.NewTrack(DefaultPayloadTypeVP8, rand.Uint32(), "video", "pion2")
```
wdouglass: `rand.Uint32` is fine for allocating SSRCs, but maybe we should have a call somewhere that allocates and keeps track, to avoid SSRC collisions. It's really a non-problem, but i'm just thinking out loud here

rtptranceiver.go
```diff
@@ -17,7 +17,7 @@ type RTPTransceiver struct {
 }
 
 func (t *RTPTransceiver) setSendingTrack(track *Track) error {
-	t.Sender.Track = track
+	t.Sender.track = track
```
wdouglass: should this be a locked operation?

PR #441, opened 2019-02-24
Allow multiple Tracks with the same ID

### LINKED ISSUE #440: Allow duplicated Track IDs
Currently we don't allow two Tracks with the same ID. This restriction means we can't mux any media. 

If we remove this resriction two tracks have the same ID then are delivered to the client muxed as one MediaStream.

Sean-Der:
A PeerConnection is now able to mux multiple tracks. by removing this
resriction two tracks with the same ID are delivered to the client
muxed as one MediaStream.

Resolves #440

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1135,14 +1135,6 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 	if pc.isClosed {
 		return nil, &rtcerr.InvalidStateError{Err: ErrConnectionClosed}
 	}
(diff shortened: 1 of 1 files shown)


Sean-Der:
Also I think [this](https://github.com/pions/webrtc/blob/master/peerconnection.go#L1136) check is wrong?

It is trying to add a sender if the RTPTransceiver has none, but I think it should also be comparing lablels and IDs? I don't fully understand the Transceiver right now, but right now it is just combining if we have two codecs that are the same.

PR #445, opened 2019-02-25
Feat/opuswriter

Antonito:
- Implemented a common Writer interface
- Made IFWriter compliant
- Implemented a OGG Opus writer (generated files run fine in browsers and `ffmpeg`)
- Updated the save-file example (using goroutines in case of multiple streams)
- ~Added a WAV writer (I'm worried about the "gopkg.in/hraban/opus.v2" dependency used to extract PCM from Opus packets, is it ok ?)~ -> **Removed it**


Sean-Der:
Fantastic work!

OpusWriter + IVFWriter I would love to accept. I am hesitant to accept the WAV one because of the dependency sorry :( 

I will do a full review right after work!

Sean-Der:
This is amazing, it works perfectly. Really nice work @Antonito :) 
------

* Can you add yourself to the README
* Can you update `janus-gateway/streaming` to use your new stuff. Don't worry about testing! I will confirm for you.
* What do you think of having `Media.Writer` having a `Open` (maybe takes a *File)
   - I can take care of the IVFWriter, but we need to write some tests :( it should be quick though!

examples/save-to-disk/main.go
```diff
@@ -56,23 +72,22 @@ func main() {
...
-			for {
-				packet, err := track.ReadRTP()
-				if err != nil {
-					panic(err)
-				}
-
-				err = i.AddPacket(packet)
+			go saveToDisk(i, track)
+		} else if codec.Name == webrtc.VP8 {
+			if codec.Name == webrtc.VP8 {
```
Sean-Der: Duplicated `webrtc.VP8` check

Antonito:
I'm going to update the `janus-gateway/streaming` right now.
I can add an `Open` method to `Media.Writer`, but I thought it was convenient to have it done automatically during the `New` call. What do you think ?

examples/save-to-disk/main.go
```diff
@@ -56,23 +72,22 @@ func main() {
...
+				fmt.Println("Got VP8 track, saving to disk as output.ivf")
+				i, err := ivfwriter.New("output.ivf")
 				if err != nil {
 					panic(err)
 				}
+				go saveToDisk(i, track)
```
Sean-Der: `OnTrack` is a fresh goroutine, so this can block! (don't need the `go`)

examples/save-to-disk/main.go
Antonito: Woops, merge error, nice spot!

examples/save-to-disk/main.go
```diff
@@ -56,23 +72,22 @@ func main() {
...
 			}
-
-			for {
-				packet, err := track.ReadRTP()
-				if err != nil {
-					panic(err)
-				}
-
-				err = i.AddPacket(packet)
+			go saveToDisk(i, track)
```
Sean-Der: Drop the `go`

pkg/media/opuswriter/opus-writer.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// Comment Header
+	oggCommentHeader := make([]byte, 21)
+	copy(oggCommentHeader[0:], []byte("OpusTags"))          // Magic Signature 'OpusTags'
+	binary.LittleEndian.PutUint32(oggCommentHeader[8:], 5)  // Vendor Length
+	copy(oggCommentHeader[12:], []byte("pions"))            // Vendor name 'pions'
+	binary.LittleEndian.PutUint32(oggCommentHeader[17:], 0) // User Comment List Length
+
+	// RFC specifies that the page where the CommentHeader completes should have a granule position of 0
```
Sean-Der: Is it easy to link to the RFC here? Just nice for the reader, if it is a PDF or anything don't worry about it!

pkg/media/opuswriter/opus-writer.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Should be equivalent to sampleRate * duration
+	if i.previousTimestamp != 0 {
+		increment := packet.Timestamp - i.previousTimestamp
+		i.previousGranulePosition += uint64(increment)
+	}
+	i.previousTimestamp = packet.Timestamp
+
+	data := i.createPage(payload, 0, i.previousGranulePosition)
+	if _, err := i.fd.Write(data); err != nil {
```
Sean-Der: You can just return `err` here (no `err != nil` check needed)

pkg/media/opuswriter/opus-writer.go
Antonito: I used an online RFC, I can add the link to it :)

Sean-Der:
@Antonito Passing the filename works for me also!

My only goal is to make this code testable, if we do `File` we could mock it out and then assert things, writing to `/tmp` could always fail in weird ways so would be nice to avoid.

examples/save-to-disk/main.go
```diff
@@ -56,23 +72,22 @@ func main() {
 			}
 		}()
 
-		if track.Codec().Name == webrtc.VP8 {
-			fmt.Println("Got VP8 track, saving to disk as output.ivf")
-			i, err := ivfwriter.New("output.ivf")
+		codec := track.Codec()
+		if codec.Name == webrtc.Opus {
+			fmt.Println("Got Opus track, saving to disk as output.opus")
+			i, err := opuswriter.New("output.opus", codec.ClockRate, codec.Channels)
```
Sean-Der: Mind changing the filename to `output.ogg` (unless that is wrong) just so the user understands.

Antonito: The RFC specifies that the recommanded extension is `.opus`
https://tools.ietf.org/html/rfc7845#section-9
But if you think `.ogg` will be clearer for people, that's fine by me!

Sean-Der: Nope I am wrong here! thanks for the pushback

backkem:
@Sean-Der Just a thought: Does it make sense to move this to pions/rtpengine?

examples/janus-gateway/streaming/main.go
```diff
@@ -52,29 +89,7 @@ func main() {
...
-				panic(err)
-			}
-
-			err = i.AddPacket(packet)
-			if err != nil {
-				panic(err)
-			}
-		}
-	})
+	peerConnection.OnTrack(handleTrack)
```
Sean-Der: Do you mind unfolding `handleTrack` it just makes it easier for newer programmers to read (I think)

examples/save-to-disk/main.go
```diff
@@ -49,32 +86,13 @@ func main() {
...
-					panic(err)
-				}
-
-				err = i.AddPacket(packet)
-				if err != nil {
-					panic(err)
-				}
-			}
-		}
+		handleTrack(track)
```
Sean-Der: I would unfold this a bit, only use a function if it can save some duplication (the indirection makes things a little harder for new devs to read)

pkg/media/ivf-writer_test.go
```diff
@@ -0,0 +1,151 @@
...
+func TestIVFWriter_FileCreation(t *testing.T) {
+	assert := assert.New(t)
+
+	// Create a IVFWriter with an empty filename
+	writer, err := NewIVFWriter("")
+	assert.NotNil(err, "NewIVFWriter shouldn't create file with an empty name")
+	assert.Nil(writer, "NewIVFWriter should return a nil value for the writer, in case of error")
+
+	// Creates a valid file, and tries to open it twice
+	writer, err = NewIVFWriter("/tmp/pions_testCreateIVFWriter1.ivf")
```
Antonito: Is there any better way to test that wouldn't imply the creation of a file ? (via mocking the `open()` method, as you suggested ?)

pkg/media/ivf-writer_test.go
```diff
@@ -0,0 +1,151 @@
...
+					PayloadType:      defaultPayloadTypeOpus,
+					SequenceNumber:   27023,
+					Timestamp:        3653407706,
+					SSRC:             476325762,
+					CSRC:             []uint32{},
+				},
+				Payload: rawPkt[20:],
+				Raw:     rawPkt,
+			},
+			err:      nil, // TODO: Update expected value
```
Antonito: Currently, when unmarshaling a packet, `pions/rtp` doesn't check the `PayloadType`, so it doesn't fail (it should IMHO)

pkg/media/ivf-writer_test.go
```diff
@@ -0,0 +1,151 @@
...
+	assert.Nil(err, "IVFWriter should be created")
+	assert.NotNil(writer, "Writer shouldn't be nil")
+	addPacketTestCase[3].writer = writer
+
+	for _, t := range addPacketTestCase {
+		res := t.writer.AddPacket(t.packet)
+		assert.Equal(res, t.err, t.message)
+	}
+
+	for _, t := range addPacketTestCase {
```
Antonito: Maybe I should do that in a single loop ?

pkg/media/opus-writer_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/pions/rtp"
+	"github.com/stretchr/testify/assert"
+)
+
+// PayloadTypes for the default codecs,
+// taken from pions/webrtc/mediaengine.go, to avoid cyclic dependency
+const (
+	defaultPayloadTypeOpus = 111
+	defaultPayloadTypeVP8  = 96
+)
```
Antonito: I'd love to have a way of not hardcoding those values, and use the ones in `mediaengine.go` instead.

Sean-Der: I would remove the payload restriction, so those numbers are just chosen by the Chromium team. Opus and VP8 are [dynamic](https://en.wikipedia.org/wiki/RTP_audio_video_profile) so nothing you can assert here :(

pkg/media/ivf-writer_test.go
Sean-Der: IMO this is good! it is testing code, so don't worry about making it terse as long as it is manageable :)

Sean-Der:
Hey @Antonito 

I moved this PR to a branch directly on the `pions/webrtc` repo. You have access so you can push/pull from this branch also :) The PR is #452 and the branch is `opuswriter`

This just makes it easier for me to run tests and review things, and we can work together!

PR #446, opened 2019-02-26
Expose Encodings variable for ORTC

kixelated:
It's not possible to use a RTPReceiver without it.


kixelated:
I've been trying to figure out if the SSRC is actually required and I'm not sure. There doesn't seem to be a convenient way to get this information from the API like some of the other information required for signaling. Do you even need to signal the track SSRC?

```
c.receiver.Receive(webrtc.RTPReceiveParameters{
  Encodings: webrtc.RTPDecodingParameters{
    webrtc.RTPCodingParameters{SSRC: ssrc},
  },
})
```

Issue #449, opened 2019-02-27
CreateAnswer sdp doesn't explicity reject m= offers


maxhawkins:
According to the jsep spec [5.3.1](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-25#section-5.3.1):

```
   For each offered m= section, if any of the following conditions are
   true, the corresponding m= section in the answer MUST be marked as
   rejected by setting the port in the m= line to zero, as indicated in
   [RFC3264], Section 6, and further processing for this m= section can
   be skipped:

   o  The associated RtpTransceiver has been stopped.

   o  None of the offered media formats are supported and, if
      applicable, allowed by codec preferences.

   o  The bundle policy is "max-bundle", and this is not the first m=
      section or in the same bundle group as the first m= section.

   o  The bundle policy is "balanced", and this is not the first m=
      section for this media type or in the same bundle group as the
      first m= section for this media type.

   o  This m= section is in a bundle group, and the group's offerer
      tagged m= section is being rejected due to one of the above
      reasons.  This requires all m= sections in the bundle group to be
      rejected, as specified in
      [I-D.ietf-mmusic-sdp-bundle-negotiation], Section 7.3.3.
```

We don't implement this behavior.

We do handle the second case (no media formats match) but don't send an explicit rejection. We just drop the m= section from the answer. My client (webrtc on ios) is rejecting the answer for that reason.

PR #450, opened 2019-02-27
Explicitly reject unsupported m= lines

maxhawkins:
CreateAnswer currently omits any m= lines from the offer that contain
unsupported codecs. The jsep spec says that unsupported lines in the
offer need to be explicitly rejected by setting the port number to 0.

Related to #449


maxhawkins:
This is really hard to write tests for! Do any of you have ideas about how to refactor PeerConnection to be more testable?

PR #452, opened 2019-02-27
Add OpusWriter and testing for IVFWriter

Sean-Der:


Sean-Der:
Hey @Antonito 

Things look really good, I have some nits though sorry :( if you are burned out I can take it the rest of the way though (and you will keep the authorship credit) this turned into quite the big task. When these get done it is an instant approval from me!

* We have to keep the `ivfwriter` at `github.com/pions/webrtc/pkg/media/ivfwriter` since users depend on it, we don't want to break the public API for them. I would do the old layout you had before, that worked really well!

* We can't create any files during the tests, we don't want to liter the users file system (and you just don't know what a CI system looks like) I would put a `io.Writer` inside the structs (or something similar) or just something that run to disk and a `[]byte{}` if needed. It is also possible that multiple test suites are running on the same machine at the same time.

* You can remove checking for PayloadType, this number doesn't actually mean anything for Opus/VP8! If the parse fails we can just return an error you can see some of them [here](https://en.wikipedia.org/wiki/RTP_audio_video_profile)

* Use underscores when naming files. I will add a lint for this! This isn't documented anywhere, sorry about that :( 

* When you are done mind squashing these into a single commit? They just make it easy to revert in case we have an issue.

Sorry for having so many issues, but this is the biggest improvement I have ever seen from a first time contributor :) you bit a huge chunk off!

Sean-Der:
@Antonito that all looks perfect to me :D fantastic work!

If you `git fetch && git rebase origin/master` and squash everything into a single commit it is a merge from me :)

Antonito:
@Sean-Der Thank you! It's all done, should be OK after the CI ends :D

pkg/media/opuswriter/opuswriter.go
```diff
@@ -0,0 +1,228 @@
...
+
+// NewWith initialize a new OGG Opus writer with an io.Writer output
+func NewWith(out io.Writer, sampleRate uint32, channelCount uint16) (*OpusWriter, error) {
+	if out == nil {
+		return nil, fmt.Errorf("file not opened")
+	}
+
+	r := rand.NewSource(time.Now().UnixNano())
+	r1 := rand.New(r)
+	serial := uint32(math.Ceil(r1.Float64() * math.Pow(2, 32)))
```
kixelated: `serial := rand.Uint32()`

pkg/media/opuswriter/opuswriter.go
```diff
@@ -0,0 +1,228 @@
...
+*/
+
+func (i *OpusWriter) writeHeaders() error {
+	// ID Header
+	oggIDHeader := make([]byte, 19)
+
+	copy(oggIDHeader[0:], []byte("OpusHead"))                     // Magic Signature 'OpusHead'
+	oggIDHeader[8] = 1                                            // Version
+	oggIDHeader[9] = uint8(i.channelCount)                        // Channel count
+	binary.LittleEndian.PutUint16(oggIDHeader[10:], 0)            // pre-skip, don't need to skip any value
```
kixelated: This is optional because memory defaults to 0, but I like how it's being explicit (even at a tiiiiinny performance hit).

pkg/media/opuswriter/opuswriter.go
```diff
@@ -0,0 +1,228 @@
...
+	data = i.createPage(oggCommentHeader, 0, 0)
+	if _, err := i.stream.Write(data); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CRC32
+func (i *OpusWriter) initChecksum() {
```
kixelated: https://golang.org/pkg/hash/crc32

This looks like `crc32.MakeTable(0x04c11db7)` although the implementation is slightly different but probably the same value.

pkg/media/opuswriter/opuswriter.go
```diff
@@ -0,0 +1,228 @@
...
+				r = (r << 1) ^ 0x04c11db7
+			} else {
+				r <<= 1
+			}
+		}
+		i.checksumTable[ndx] = r & 0xffffffff
+	}
+}
+
+func (i *OpusWriter) getChecksum(payload []uint8) uint32 {
```
kixelated: Same as above, use the `crc32` package.

PR #454, opened 2019-02-27
Docs: Build Janus example on OSX

backkem:
Documented how to build the janus-gateway example on OSX.


backkem:
Thanks for the contribution @feixiao!

PR #455, opened 2019-02-27
Eliminate API object, pass settings explicitly

maxhawkins:
By passing non-standard settings to constructors explicitly:

* options will become easier to discover in godoc
* you won't have to lug around an API object anymore, especially in the default case
```go
// BEFORE:
api := webrtc.NewAPI()
api.SetEphemeralUDPPortRange(2000,3000)
api.NewPeerConnection(opts)

// AFTER:
webrtc.NewPeerConnection(opts, &PeerConnectionOptions{
  ICEAgentOptions: ICEAgentOptions{
    PortMin: 2000,
    PortMax: 3000,
  },
})

// Or in the default case:
webrtc.NewPeerConnection(opts, nil)
```
* option structs will be more focused and easier to understand (`DataChannelOptions` and `ICEAgentOptions` rather than everything combined in an unweildy `SettingEngine`)


Sean-Der:
Two nice things about the `api` object that I personally like.

- All non-standard behavior is in one place. So if users wan't something that doesn't work with WebRTC by default they just have one bucket they need to look at
- We can have an option work across multiple objects/things. If we do options per constructor you will duplicate some fields (and then what happens if the user doesn't set both properly)

maxhawkins:
For the average (non-ORTC) user, all the non-standard behavior is still all in one place: `PeerConnectionOptions`

```go
webrtc.NewPeerConnection(params, &webrtc.PeerConnectionOptions{
  DataChannelOptions: DataChannelOptions{
    Detach: true,
  },
  ICEAgentOptions: ICEAgentOptions{
    PortMin: 1000,
    PortMax: 2000,
    ConnectionTimeout: 1 * time.Minute,
    KeepaliveInterval: 1 * time.Minute,
  },
  Codecs: webrtc.DefaultCodecs,
})
```

It's now closer to the site of use instead of being tied to an API object instantiated earlier in the program.

As for the duplication issue, think of it this way: Keeping track of a config object is the same LOC-wise as keeping track of an api object.

```go
api := webrtc.NewAPI()
api.NewPeerConnection(params)
// vs
opts := webrtc.PeerConnectionOptions{}
webrtc.NewPeerConnection(params, opts)
```

but in the default case (which most people will be using) it's much simpler and you don't have to keep track of any variables:

```go
api.NewPeerConnection(params, nil)
```

mjmac:
> ```go
> api.NewPeerConnection(params, nil)
> ```
I hate to bikeshed on this, but I really dislike this pattern.

Personally, I see nil parameters in API methods as an anti-pattern. They're easy to screw up (oops, I didn't meant set that parameter to nil!) and go against the principle of self-documenting code.

If we really wanted to avoid using the API object for configuration (which I don't feel strongly about one way or the other), then I'd suggest instead to add a PionsOptions member to the Configuration struct. This way all of the configuration is in one place, and the non-standard options are clearly distinguishable.

maxhawkins:
I agree. With a couple days distance I like the nil default parameter a lot less.

I'll update the PR shortly.

PR #464, opened 2019-03-01
Add support for batched RTP writes

kixelated:
Depends on SRTP supporting the same batched interface.

The only API change is that `WriteRTP` can now take multiple packets at
once instead of a single one. This lets the API consumer decide what
level of batching to perform.

`WriteSample` will automatically batch packets into a single send.

Under the hood, we use the `sendmmsg` syscall if supported.
Other platforms will default to multiple `sendmsg` calls and won't see a
performance difference. This is faster because it avoids extra
transitions between kernel and user-space which has a significant cost
especially after spectre patches.

For my write-heavy application, this improved performance by 1.4x.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,4 +13,5 @@ require (
 	github.com/pions/transport v0.2.0
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.3.0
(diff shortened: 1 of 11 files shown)


kixelated:
Depends on https://github.com/pions/srtp/pull/13

kixelated:
Turns out it's a performance decrease.

PR #465, opened 2019-03-02
Update LocalInterfaces() to comply with spec

trivigy:


internal/ice/util.go
```diff
@@ -12,28 +12,39 @@ func localInterfaces() (ips []net.IP) {
 	}
 
 	for _, iface := range ifaces {
-		if iface.Flags&net.FlagUp == 0 {
```
Sean-Der: The old way is easier to parse IMO. By putting them in the same conditional I took a second parsing them trying to figure out if they were related.

I would keep as two if branches or `if/elsif`

internal/ice/util.go
```diff
@@ -12,28 +12,39 @@ func localInterfaces() (ips []net.IP) {
...
+
 			}
+
 			if ip == nil || ip.IsLoopback() {
 				continue
 			}
-			ips = append(ips, ip)
+
+			// The conditions of invalidation written below are defined in
+			// https://tools.ietf.org/html/rfc8445#section-5.1.1.1
```
Sean-Der: can you keep the append at the end, and do early returns like the preceding code just to stay consistent

internal/ice/util.go
trivigy: Changing it. I was just copy pasting from my version of the library.

internal/ice/util.go
```diff
@@ -18,21 +18,36 @@ func localInterfaces() (ips []net.IP) {
...
+
 			}
+
 			if ip == nil || ip.IsLoopback() {
 				continue
 			}
+
+			// The conditions of invalidation written below are defined in
+			// https://tools.ietf.org/html/rfc8445#section-5.1.1.1
+			if _ip := ip.To4(); _ip != nil {
```
Sean-Der: I would do this, so we can keep this function in the same style. I don't have an opinion either way, but just so we don't introduce bugs.

Also can you squash the three commits, just so if anything happens it is easier to debug in logical chunks. Thank you!

```
if ipv4 := ip.To4(); ipv4 == nil {
  if !isSupportedIPv6(ip) {
￼      continue
￼  }
}

   ips = append(ips, _ip)
```

PR #467, opened 2019-03-03
Issue 460

### LINKED ISSUE #460: Slow connection times
We are using pion/webrtc to set up communications between mobile apps and a backend server.

We are seeing connection times in the range of 5-6 seconds from the time the offer/answer exchange happens to the time the server receives a datachannel opened by the client.

Do you have suggestions for how we can get that to 1 second (or less, ideally).

trivigy:
This is a temporary solution. We should not have this lying around normally. Once v2.0 is out and fixes the underlying issue with specific ipv6 issues for the server reflexive candidates, we should remove the support for this.

Resolves #460

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -26,7 +26,7 @@ type ICEGatherer struct {
 // This constructor is part of the ORTC API. It is not
 // meant to be used together with the basic WebRTC API.
 func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
-	validatedServers := []*ice.URL{}
+	var validatedServers []*ice.URL
 	if len(opts.ICEServers) > 0 {
 		for _, server := range opts.ICEServers {
 			url, err := server.validate()
@@ -62,6 +62,15 @@ func (g *ICEGatherer) Gather() error {
 		PortMax:           g.api.settingEngine.ephemeralUDP.PortMax,
 		ConnectionTimeout: g.api.settingEngine.timeout.ICEConnection,
 		KeepaliveInterval: g.api.settingEngine.timeout.ICEKeepalive,
+	}
+
+	requestedNetworkTypes := g.api.settingEngine.candidates.ICENetworkTypes
+	if len(requestedNetworkTypes) == 0 {
+		requestedNetworkTypes = supportedNetworkTypes
+	}
+
+	for _, typ := range requestedNetworkTypes {
+		config.NetworkTypes = append(config.NetworkTypes, ice.NetworkType(typ))
 	}
 
 	agent, err := ice.NewAgent(config)
--- a/internal/ice/util.go
+++ b/internal/ice/util.go
@@ -5,10 +5,21 @@ import (
 	"sync/atomic"
 )
 
-func localInterfaces() (ips []net.IP) {
+func localInterfaces(networkTypes []NetworkType) (ips []net.IP) {
 	ifaces, err := net.Interfaces()
 	if err != nil {
 		return ips
+	}
+
+	var IPv4Requested, IPv6Requested bool
+	for _, typ := range networkTypes {
+		if typ.IsIPv4() {
+			IPv4Requested = true
+		}
+
+		if typ.IsIPv6() {
+			IPv6Requested = true
+		}
 	}
 
 	for _, iface := range ifaces {
@@ -33,17 +44,18 @@ func localInterfaces() (ips []net.IP) {
 				ip = addr.IP
 
 			}
-
 			if ip == nil || ip.IsLoopback() {
 				continue
 			}
 
-			// The conditions of invalidation written below are defined in
-			// https://tools.ietf.org/html/rfc8445#section-5.1.1.1
 			if ipv4 := ip.To4(); ipv4 == nil {
-				if !isSupportedIPv6(ip) {
+				if !IPv6Requested {
+					continue
+				} else if !isSupportedIPv6(ip) {
 					continue
 				}
+			} else if !IPv4Requested {
+				continue
 			}
 
 			ips = append(ips, ip)
@@ -62,6 +74,8 @@ func (a *atomicError) Load() error {
 	return err.error
 }
 
+// The conditions of invalidation written below are defined in
+// https://tools.ietf.org/html/rfc8445#section-5.1.1.1
 func isSupportedIPv6(ip net.IP) bool {
 	if len(ip) != net.IPv6len ||
 		!isZeros(ip[0:12]) || // !(IPv4-compatible IPv6)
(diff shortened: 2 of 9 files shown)


configuration.go
```diff
@@ -12,6 +12,10 @@ type Configuration struct {
 	// ICE, such as STUN and TURN servers.
 	ICEServers []ICEServer
 
+	// IgnoreIPv6 is an optional configuration for disabling support for ipv6
+	// server reflexive candidates. Default is false.
+	IgnoreIPv6 bool
```
maxhawkins: I think we're putting non-standard options in SettingEngine instead of directly in these structs. Maybe move this there?

trivigy: Can you show me? A link.

maxhawkins: https://github.com/pions/webrtc/blob/master/settingengine.go#L12

trivigy: Thank you. I will try moving it there but it might be challenging since the only way for a user to pass in configurations is via `type Configuration struct{}`. Unless there is another way I am not thinking about.

trivigy: Okay so I just looked at it an unfortunately there is not way to avoid not putting this at least in `type Configuration struct{}`. I am going to add a deprecation comment to discourage people from using it.

maxhawkins: ICEGatherer is constructed by API and API has a SettingEngine right? A little convoluted but you can get the settings there.

Let's see what @Sean-Der thinks. He has stronger opinions about keeping `struct Configuration` standard than I do.

trivigy: Yeah. He actually was the one that said for now to just add the opt-out config. I'll wait for him to review.

Personally I am okay with completely removing ipv6 support right now but I know others use this library in other ways so I am taking a more "backward compatible kinda solution".

trivigy: One other thing I should probably do is make that variable local. This way it will not show up in the documentations at all and can be used only by people reading through the source code.

I'll just go ahead and do it.

Sean-Der:
Reading this it looks like the only real solution is to implement trickle ICE, because we could still have an issue if we had busted `ipv4` :( it doesn't look like we are that far away either.

We could do these short term, I don't think they are adding tech debt though (and could still provide value after trickle is done) @backkem @maxhawkins @trivigy 

* Add a configurable timeout to srflx candidates and put it [here](https://github.com/pions/webrtc/blob/master/settingengine.go#L23)

* Add an option to select `ICECandidateTypes` and users can pass an array of `UDP4` `UDP6` `TCP4` or `TCP6` I would prefer that to a boolean (would give more people flexibility)

trivigy:
@Sean-Der @maxhawkins Is there an example that shows how to pass `SettingEngine` from the user side? All I know is via setting the `Configuration`.

Sean-Der:
Hey @trivigy 

We use the `MediaEngine` in `examples/save-to-disk` and this patch shows using the `SettingEngine` this is where we are putting all our non-standard behavior so far.

```
diff --git a/examples/save-to-disk/main.go b/examples/save-to-disk/main.go
index 24fac9e..56d1e46 100644
--- a/examples/save-to-disk/main.go
+++ b/examples/save-to-disk/main.go
@@ -41,8 +41,11 @@ func main() {
        m.RegisterCodec(webrtc.NewRTPOpusCodec(webrtc.DefaultPayloadTypeOpus, 48000))
        m.RegisterCodec(webrtc.NewRTPVP8Codec(webrtc.DefaultPayloadTypeVP8, 90000))

+       s := webrtc.SettingEngine{}
+       s.SetEphemeralUDPPortRange(1000, 1500)
+
        // Create the API object with the MediaEngine
-       api := webrtc.NewAPI(webrtc.WithMediaEngine(m))
+       api := webrtc.NewAPI(webrtc.WithMediaEngine(m), webrtc.WithSettingEngine(s))

        // Everything below is the pion-WebRTC API! Thanks for using it ❤.
```

trivigy:
@Sean-Der correct me if I am wrong. The `API` is basically a webrtc builder class meant for optimization? So I put the non-standard stuff into the `SettingEngine{}` and `API`, from there initialize those, and they have a `NewPeerConnection` func to act as a PeerConnection producer with the customizations? If I got this correct then it's a fairly easy change.

Sean-Der:
yep that is exactly right @trivigy!

It is just a container for any non-standard behavior we want to add.

icegatherer.go
```diff
@@ -26,7 +26,7 @@ type ICEGatherer struct {
 // This constructor is part of the ORTC API. It is not
 // meant to be used together with the basic WebRTC API.
 func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
-	validatedServers := []*ice.URL{}
+	var validatedServers []*ice.URL
```
trivigy: This is just an autolinter change.

internal/ice/util.go
```diff
@@ -38,8 +38,31 @@ func localInterfaces() (ips []net.IP) {
...
+				}
+
+				if typ.IsIPv6() {
+					IPv6Supported = true
+				}
+			}
+
 			// The conditions of invalidation written below are defined in
 			// https://tools.ietf.org/html/rfc8445#section-5.1.1.1
+			if !IPv4Supported && !IPv6Supported {
```
trivigy: I worked out the logic below with a binary conditional table. This still maintains the logic you wanted with the last line being `ips = append(ips, ip)` but all the other statements work through eliminating the ips that do not fit the supported passed list.

networktype.go
```diff
@@ -0,0 +1,40 @@
+package webrtc
```
trivigy: I created this entire file, which is a mini version of the one found inside of `ice` package because the `ice` package is internal and now that the config can be passed via `SettingEngine`, it had to be externalized. Rather then moving stuff around I just created a copy of that enum.

peerconnection.go
```diff
@@ -112,7 +112,9 @@ type PeerConnection struct {
 func NewPeerConnection(configuration Configuration) (*PeerConnection, error) {
 	m := MediaEngine{}
 	m.RegisterDefaultCodecs()
-	api := NewAPI(WithMediaEngine(m))
+	s := SettingEngine{}
```
trivigy: Had to add this due to the default supported networks that need to be added.

peerconnection.go
```diff
@@ -468,7 +471,6 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
 func (pc *PeerConnection) createICEGatherer() (*ICEGatherer, error) {
 	g, err := pc.api.NewICEGatherer(ICEGatherOptions{
 		ICEServers: pc.configuration.ICEServers,
-		// TODO: GatherPolicy
```
trivigy: Do we need to keep this? I am pretty sure I removed this by mistake.

PR #477, opened 2019-03-04
Check that RTPTransceiver has .Sender on PeerConnection.AddTrack

### LINKED ISSUE #476: PeerConnection.AddTrack: skip transceivers with nil .Sender?
Looks like PeerConnection's transceivers may have nil `.Sender` - for remote streams/tracks (right?)

https://github.com/pions/webrtc/blob/15d0ae0808446a7c1a083ba43b74fc49b13f1a6f/peerconnection.go#L1138-L1148

I'm getting a nil-ptr panic when trying to send a stream, using smth like this (not sure if it's Ok to copy streams this way, but still):

```go
peerConnection.OnTrack(func(track *webrtc.Track, receiver *webrtc.RTPReceiver) {
  // interested only in video streams:
  if track.Kind() != webrtc.RTPCodecTypeVideo {
    return
  }

  // PictureLossIndication sending on Ticker
  go func() { ... }()

  videoTrack, err := peerConnection.NewTrack(track.PayloadType(), rand.Uint32(), "video", "transformedvideo") // panic: runtime error: invalid memory address or nil pointer dereference
  if err != nil {
    panic(err)
  }
  _, err = peerConnection.AddTrack(videoTrack)
  if err != nil {
    panic(err)
  }
  go io.Copy(videoTrack, track)
})
```

Would it be fine to skip `.Stream == nil` transceivers?

mxmCherry:
Resolves #476

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1138,8 +1138,8 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 	var transceiver *RTPTransceiver
 	for _, t := range pc.rtpTransceivers {
 		if !t.stopped &&
-			t.Sender != nil && // TODO: check that the sender has never sent
-			t.Sender.track == nil &&
+			t.Sender != nil &&
+			!t.Sender.hasSent() &&
 			t.Receiver.Track() != nil &&
 			t.Receiver.Track().Kind() == track.Kind() {
 			transceiver = t
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1138,7 +1138,7 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 	var transceiver *RTPTransceiver
 	for _, t := range pc.rtpTransceivers {
 		if !t.stopped &&
-			// t.Sender == nil && // TODO: check that the sender has never sent
+			t.Sender != nil && // TODO: check that the sender has never sent
 			t.Sender.track == nil &&
 			t.Receiver.Track() != nil &&
 			t.Receiver.Track().Kind() == track.Kind() {
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Luke Curley](https://github.com/kixelated)
 * [Antoine Baché](https://github.com/Antonito) - *OGG Opus export*
 * [frank](https://github.com/feixiao) - *Building examples on OSX*
+* [mxmCherry](https://github.com/mxmCherry)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/rtpsender.go
+++ b/rtpsender.go
@@ -56,10 +56,9 @@ func (r *RTPSender) Transport() *DTLSTransport {
 func (r *RTPSender) Send(parameters RTPSendParameters) error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	select {
-	case <-r.sendCalled:
+
+	if r.hasSent() {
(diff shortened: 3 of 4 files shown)


peerconnection.go
```diff
@@ -1138,7 +1138,7 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 	var transceiver *RTPTransceiver
 	for _, t := range pc.rtpTransceivers {
 		if !t.stopped &&
-			// t.Sender == nil && // TODO: check that the sender has never sent
+			t.Sender != nil && // TODO: check that the sender has never sent
```
mxmCherry: Regarding this TODO - I think, a helper can be added and used here:

```go
func (r *RTPSender) sent() bool {
  r.mu.Lock()
  defer r.mu.Unlock()

  select {
    case <-r.sendCalled:
      return true
    default:
       return false
  }
}
```

Looks like `sendCalled` chan is made in constructor and closed on `.Send` call:
https://github.com/pions/webrtc/blob/f46ebf7effae3f373aa8fe8e0aa7c279572a49ad/rtpsender.go#L55-L81

Sean-Der:
Fantastic, thank you so much for the PR @mxmCherry !

Would you mind fixing the TODO? I thinking adding the helper `hasSent` and I think that can lower duplication also, other people do the same thing in that file iirc

peerconnection.go
```diff
@@ -1138,8 +1138,8 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 	var transceiver *RTPTransceiver
 	for _, t := range pc.rtpTransceivers {
 		if !t.stopped &&
-			// t.Sender == nil && // TODO: check that the sender has never sent
-			t.Sender.track == nil &&
```
mxmCherry: Initially, I thought that `RTPSender.track` cannot be nil, as it's validated in `PeerConnection.NewRTPSender`:

https://github.com/pions/webrtc/blob/f46ebf7effae3f373aa8fe8e0aa7c279572a49ad/rtpsender.go#L24-L27

But then I have double-checked the code and looks like, theoretically, track CAN be nil-ed:

https://github.com/pions/webrtc/blob/f46ebf7effae3f373aa8fe8e0aa7c279572a49ad/rtptransceiver.go#L19-L20

This method is called only once and only if we've found a matching transceiver in this loop (just few lines down the code):

https://github.com/pions/webrtc/blob/f46ebf7effae3f373aa8fe8e0aa7c279572a49ad/peerconnection.go#L1149-L1153

Would be correct to assume, that ever adding a nil track does not make any sense? Probably, it's even dangerous? Even more - there's a (commented out for now) FIXME for `PeerConnection.RemoveTrack`, so I assume `AddTrack` should never accept a nil track.

So if that's correct ^^ I'd like to add a nil-check for `track` argument for this method.

mxmCherry: Or, probably, would be better to add this `track != nil` check right in `RTPTransceiver.setSendingTrack`

Sean-Der: Yea this makes sense to me! We should never have a `nil` Track.

`setSendingTrack` also will break things. If we change the track, it currently doesn't change all the state required.

I am going to work on adding/removing tracks after signaling is complete but currently I think things will break :/

mxmCherry: So not sure if `setSendingTrack` will change much (or will be replaced with something other), but since I'm on it right now - will add a nil check there, why not.

Sean-Der:
Nice work @mxmCherry 

You have been invited to the `pions` organization. Once you accept you will have a `rebase and merge` button on this page.

If you are interested you should also join our Slack (mentioned in the README) I would love your continued involvement in the project. We are making lots of changes, the more people involved the better :)

Thanks again.

PR #479, opened 2019-03-05
Add JavaScript/WASM bindings

albrow:
This is a very WIP PR to add WASM support. The goal is to have an isomorphic API that will work in the browser or in standalone executables. This will be implemented by using build tags and the `syscall/js` package.

### DIFF
--- a/datachannel_js.go
+++ b/datachannel_js.go
@@ -1,0 +1,256 @@
+// +build js
+
+package webrtc
+
+import (
+	"syscall/js"
+)
+
+const dataChannelBufferSize = 16384 // Lowest common denominator among browsers
+
+// DataChannel represents a WebRTC DataChannel
+// The DataChannel interface represents a network channel
+// which can be used for bidirectional peer-to-peer transfers of arbitrary data
+type DataChannel struct {
+	// Pointer to the underlying JavaScript RTCPeerConnection object.
+	underlying js.Value
+
+	// Keep track of handlers/callbacks so we can call Release as required by the
+	// syscall/js API. Initially nil.
+	onOpenHandler    *js.Func
+	onCloseHandler   *js.Func
+	onMessageHandler *js.Func
+}
+
+// OnOpen sets an event handler which is invoked when
+// the underlying data transport has been established (or re-established).
+func (d *DataChannel) OnOpen(f func()) {
+	if d.onOpenHandler != nil {
+		oldHandler := d.onOpenHandler
+		defer oldHandler.Release()
+	}
+	onOpenHandler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+		go f()
+		return js.Undefined()
+	})
+	d.onOpenHandler = &onOpenHandler
+	d.underlying.Set("onopen", onOpenHandler)
+}
+
+// OnClose sets an event handler which is invoked when
+// the underlying data transport has been closed.
+func (d *DataChannel) OnClose(f func()) {
+	if d.onCloseHandler != nil {
+		oldHandler := d.onCloseHandler
+		defer oldHandler.Release()
+	}
+	onCloseHandler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+		go f()
+		return js.Undefined()
+	})
+	d.onCloseHandler = &onCloseHandler
+	d.underlying.Set("onclose", onCloseHandler)
+}
+
+// OnMessage sets an event handler which is invoked on a binary message arrival
+// from a remote peer. Note that browsers may place limitations on message size.
+func (d *DataChannel) OnMessage(f func(msg DataChannelMessage)) {
+	if d.onMessageHandler != nil {
+		oldHandler := d.onMessageHandler
+		defer oldHandler.Release()
+	}
+	onMessageHandler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+		msg := valueToDataChannelMessage(args[0].Get("data"))
+		go f(msg)
+		return js.Undefined()
+	})
+	d.onMessageHandler = &onMessageHandler
+	d.underlying.Set("onmessage", onMessageHandler)
+}
+
+// Send sends the binary message to the DataChannel peer
+func (d *DataChannel) Send(data []byte) (err error) {
+	defer func() {
+		if e := recover(); e != nil {
+			err = recoveryToError(e)
+		}
+	}()
+	array := js.TypedArrayOf([]uint8(data))
+	defer array.Release()
+	d.underlying.Call("send", array.Value)
+	return nil
+}
+
+// SendText sends the text message to the DataChannel peer
+func (d *DataChannel) SendText(s string) (err error) {
+	defer func() {
+		if e := recover(); e != nil {
+			err = recoveryToError(e)
+		}
+	}()
+	d.underlying.Call("send", s)
+	return nil
+}
+
+// Close Closes the DataChannel. It may be called regardless of whether
+// the DataChannel object was created by this peer or the remote peer.
+func (d *DataChannel) Close() (err error) {
+	defer func() {
+		if e := recover(); e != nil {
+			err = recoveryToError(e)
+		}
+	}()
+
+	d.underlying.Call("close")
+
+	// Release any handlers as required by the syscall/js API.
+	if d.onOpenHandler != nil {
+		d.onOpenHandler.Release()
+	}
+	if d.onCloseHandler != nil {
+		d.onCloseHandler.Release()
+	}
+	if d.onMessageHandler != nil {
+		d.onMessageHandler.Release()
+	}
+
+	return nil
+}
+
+// Label represents a label that can be used to distinguish this
+// DataChannel object from other DataChannel objects. Scripts are
+// allowed to create multiple DataChannel objects with the same label.
+func (d *DataChannel) Label() string {
+	return d.underlying.Get("label").String()
+}
+
+// Ordered represents if the DataChannel is ordered, and false if
+// out-of-order delivery is allowed.
+func (d *DataChannel) Ordered() bool {
+	return d.underlying.Get("ordered").Bool()
+}
+
+// MaxPacketLifeTime represents the length of the time window (msec) during
+// which transmissions and retransmissions may occur in unreliable mode.
+func (d *DataChannel) MaxPacketLifeTime() *uint16 {
+	return valueToUint16Pointer(d.underlying.Get("maxPacketLifeTime"))
+}
+
+// MaxRetransmits represents the maximum number of retransmissions that are
+// attempted in unreliable mode.
+func (d *DataChannel) MaxRetransmits() *uint16 {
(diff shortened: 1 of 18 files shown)


backkem:
I actually did come up with another advantage of using a common interface: It makes sure that both implementations share the same public API. I still believe we shouldn't expose such an interface to the public but it may be worth investigation options to make sure the public API is the same for both the native and JS/WASM implementations.

albrow:
@backkem I was thinking the same thing. It's a good idea to make an unexported interface to enforce that the JavaScript/WASM bindings follow the same API.

albrow:
There's still a lot of work to do, but I've reached the point where there is a runnable example in __examples/data-channels-wasm__. The example simply ports the JavaScript code from __examples/data-channels__ to Go. There are potentially more interesting examples we could build down the road, e.g., having the same __main.go__ file run in both the browser and the server.

backkem:
The example is really cool already!

Things I noticed:
- It would indeed be amazing to be able to build the exact same example for both targets. That definitely seems feasible.
- We provide a web server to host the examples locally as documented [here](https://github.com/pions/webrtc/tree/master/examples#install). We should restructure the example so it can be served like this as well.
- I would like to avoid duplicating the accompanying go code example. The same example is already available. Maybe we can find a good way to merge the example with the corresponding original example. (There may be a relation to the previous point. E.g.: make the examples server detect a `wasmfiddle` folder to serve wasm examples.)

Let me know if you would like any help with that.

albrow:
Cool I'm happy to get the WASM example more in line with the other examples. Will reach out to you on Slack for some of the details.

I'd like to focus on the minimum set of changes to get this PR ready to merge.

1. Currently, the PR supports DataChannels only. That means methods like `OnTrack` and `NewTrack` are not implemented. I would like to keep this PR small and address these features in a future PR. I'm also not sure if this is something I can implement myself since we only need DataChannels for the project I'm working on (but I would be happy to help point someone in the right direction).
2. There are not any tests for the WASM bindings (unless you count the example). Testing them will require some significant changes and new testing infrastructure which is not in place. I have some ideas for how to do this and it *is* something I want to help build, but I would prefer to do it in a separate PR in order to make changes easier to review.
3. As we discussed on Slack, it is not possible to implement certain properties of `DataChannel` and `PeerConnection` in the WASM bindings (e.g., `PeerConnection.CurrentLocalDescription` and `PeerConnection.PendingLocalDescription`). So unlike the Go version, these have been converted to methods in the WASM bindings. I think we are aligned on a plan to update the Go version to use methods too. I chose not to do that in this PR in order to make it easier to review. Additionally, my preference would be for someone else who is more familiar with the internals of the Go implementation to convert the properties to methods.
4. I would also like to delay handling Certificates and Credentials as they are a lot trickier to implement. Cryptography in JavaScript land is completely different from how it works in Go. It's not as simple as converting basic data structures.
5. I added two methods to `PeerConnection` in the WASM bindings that don't exist in the Go implementation: `OnICECandidate` and `OnNegotiationNeeded`. Both of which are part of the WebRTC spec and I would like to add these to the Go implementation too at some point.

With that said, here's what I have left to do:

- [ ] Update WASM example to be more in line with existing examples.
- [ ] Add better error handling in some places, especially recovering from exceptions in JavaScript land.

@backkem how does all this sound?

backkem:
I agree with everything you said. I think we're fine with landing this in increments.

datachannel_js.go
```diff
@@ -0,0 +1,185 @@
...
+func (d *DataChannel) OnMessage(f func(msg DataChannelMessage)) {
+	if d.onMessageHandler != nil {
+		oldHandler := d.onMessageHandler
+		defer oldHandler.Release()
+	}
+	onMessageHandler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+		event := args[0]
+		rawData := event.Get("data").String()
+		msg := DataChannelMessage{
+			IsString: false,
```
backkem: Can we figure out if we're getting a string or binary data? This information is normally included. I think this depends on the datatype of event.data in the JS world.

datachannel_js.go
```diff
@@ -0,0 +1,185 @@
...
+		go f(msg)
+		return js.Undefined()
+	})
+	d.onMessageHandler = &onMessageHandler
+	d.underlying.Set("onmessage", onMessageHandler)
+}
+
+// Send sends the binary message to the DataChannel peer
+func (d *DataChannel) Send(data []byte) error {
+	return d.SendText(string(data))
```
backkem: This seems wrong. There is actually a difference in sending a string vs a binary format in the data channel spec. I think this depends on the datatype passed to the method in the JS world. It would be nice if we could support that here as well.

albrow: Hmm.. I've been using the MDN docs as a reference but they are light on the details of [what happens under the hood when you send different data types](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/send). Is there another reference you have been using? How is the Go implementation able to distinguish between the different types of data?

albrow: Actually never mind. I don't necessarily need to understand everything that's going on under the hood, just how to convert between the JavaScript types and Go representation of those types. This is sort of a huge can of worms and I'm not going to be able to support everything. I'm going to aim to support `USVString` <-> `string` and `ArrayBuffer` <-> `[]byte`, then return a reasonable error if we see a type we didn't expect.

backkem: Ok, that sounds great. We can extend when we run into other types.

albrow: Take a look at the latest commit. I think we are handling this more correctly now.

albrow:
@backkem I'm not sure how to address this issue in Travis:

```
syscall/js
go build syscall/js: build constraints exclude all Go files in /home/travis/.gimme/versions/go1.12.linux.amd64/src/syscall/js
```

Any ideas?

backkem:
@albrow I figured out the CI problem. At some point the CI (and go modules as well) seems to run `go get ./...` to download the dependencies. Because the `examples/data-channels-wasm/browser` package doesn't have the build flag it also tries to download and build `syscall/js` as well. The fix is pretty simple: ensure that all the uses of `syscall/js` are behind the `js` build flag. In this case that means adding the build flag at the top of  `examples/data-channels-wasm/browser/main.go`

backkem:
FYI, The CI pointed out the following after fixing the dependency issue:
- Please add yourself to the contributor section in the root readme.md.
- We don't allow the use of 'panic'. Please change that to just returning an error or something similar.

peerconnection_js.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+	return js.ValueOf(maps)
+}
+
+func iceServerToValue(server ICEServer) js.Value {
+	return js.ValueOf(map[string]interface{}{
+		"urls":           stringsToValue(server.URLs), // required
+		"username":       stringToValueOrUndefined(server.Username),
+		"credential":     interfaceToValueOrUndefined(server.Credential),
+		"credentialType": stringToValueOrUndefined(server.CredentialType.String()),
```
backkem: Apparently Firefox doesn't allow passing `unknown` to `credentialType`:
```
panic: JavaScript error: 'credentialType' member of RTCIceServer 'unknown' is not a valid value for enumeration RTCIceCredentialType. wasm_exec.js:85:11
wasm_exec.js:85:11
goroutine 1 [running]: wasm_exec.js:85:11
syscall/js.Value.New(0x7ff800030000001c, 0xc06be48, 0x1, 0x1, 0x0) wasm_exec.js:85:11
	syscall/js/js.go:330 +0x9 wasm_exec.js:85:11
github.com/pions/webrtc.NewPeerConnection(0xc02a4c0, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...) wasm_exec.js:85:11
	github.com/pions/webrtc/peerconnection_js.go:50 +0xf wasm_exec.js:85:11
main.main() wasm_exec.js:85:11
	github.com/pions/webrtc/examples/data-channels-wasm/browser/main.go:20 +0x8
```
Maybe there is a difference between passing undefined and not creating the attribute?

backkem: I think we can ticket this one out if you prefer to land first.

albrow: No I know how to address this. It'll be a quick fix.

albrow: @backkem I think this is fixed now. Mind trying again?

You were on the right track and all I had to do was convert `"unknown"` to `js.Undefined`.

backkem: Cool, I can confirm this works 👍

datachannel_js.go
```diff
@@ -0,0 +1,256 @@
...
+		return DataChannelMessage{
+			IsString: true,
+			Data:     []byte(val.String()),
+		}
+	}
+
+	// For other types, we need to first determine val.constructor.name.
+	constructor := val.Get("constructor")
+	if constructor == js.Null() || constructor == js.Undefined() {
+		panic(JavaScriptTypeConverstionError{
```
albrow: @backkem I'm not sure how to avoid panicking here. For context, this is called during the `OnMessage` handler function. The API doesn't give us any way to return an error. Any ideas?

peerconnection_js.go
```diff
@@ -0,0 +1,499 @@
...
+	onICECandidateHandler            *js.Func
+	onNegotiationNeededHandler       *js.Func
+}
+
+// NewPeerConnection creates a peerconnection with the default
+// codecs. See API.NewRTCPeerConnection for details.
+func NewPeerConnection(configuration Configuration) (pc *PeerConnection, err error) {
+	configMap := configurationToValue(configuration)
+
+	// TOOD(albrow): Recover from possible panic
```
fabioberger:
```suggestion
	// TODO(albrow): Recover from possible panic
```

datachannel_js.go
backkem: Maybe we just fall back to a default type instead of throwing an error?

albrow: Hmm.. the problem here arises when we see a JavaScript type for `Message.Data` that we don't know how to convert to a Go type (typically either `[]byte` or `string`). It's not really possible to have a default type in this case.

We could do something like return an empty message, but IMO it would be better to fail fast than cause potentially confusing issues down the line.

backkem: I wonder if we need to create this many options. AFAIK the DataChannel (protocol level) spec only allows for 2 types of data: string and binary. In combination with the [binaryType](https://w3c.github.io/webrtc-pc/#dom-datachannel-binarytype) attribute of the WebRTC API the binary data may be interpreted as a "blob" or "arraybuffer". So maybe we don't have to account for a large amount of data types. If we can just cover string and whatever the binary-"blob" datatype is we should be good for now. We can look into the "arraybuffer" when implementing the "binaryType" attribute.

albrow: Okay check the latest commit and let me know what you think. We should be covering every case in the spec.

backkem: That conversion step seems a bit involved but I don't if/how to do it differently so we can optimize later. LGTM.

PR #482, opened 2019-03-06
Use args instead of stdin in save-to-disk example

alucic:
When running `save-to-disk` example, there are two sets of instructions, one for Windows and the other for Linux/macOS. This PR is trying to bridge the platform gap by using arguments instead of `stdin` for getting Browser's Session Description.

* pass in browser session as an argument instead of stdin
* update readme file with additional information

```diff
- echo BROWSER_SDP | save-to-disk
+ save-to-disk BROWSER_SDP
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Antoine Baché](https://github.com/Antonito) - *OGG Opus export*
 * [frank](https://github.com/feixiao) - *Building examples on OSX*
 * [mxmCherry](https://github.com/mxmCherry)
+* [Ante Lucic](https://github.com/alucic)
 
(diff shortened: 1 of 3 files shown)


Sean-Der:
@backkem 

You good with this change as well? I don't want to rush into anything (I personally think this is an improvement and makes it more cross platform friendly)

----

@alucic I would move your stdin parsing into `examples/internal/signal` maybe. Also I would prefer if we could move every example (happy to help with that if you don't have time!) 

I would really like them to all be consistent so the are pleasant to use.

backkem:
Yea, I'm fine with this.

Sean-Der:
Hey @alucic 

I am just clearing our backlog so closing this for now, if you get some time I would love to land this still! Feel free to reach out here or on Slack, and always happy to help land this.

thanks

Issue #471, opened 2019-03-03
Hardcoded header in IVFWriter


kacperjurak:
Hi,

Why header in IVFWriter is hardcoded? When i try to save a video from using examples, then the video is set to 640x480 with 30fps. Number of frames is set to 900, so the video is always 30 seconds long, even i record a shorter one. Recorded video is accelerated when open in desktop player.

Thanks

```
func (i *IVFWriter) writeHeader() error {
	header := make([]byte, 32)
	copy(header[0:], []byte("DKIF"))                // DKIF
	binary.LittleEndian.PutUint16(header[4:], 0)    // Version
	binary.LittleEndian.PutUint16(header[6:], 32)   // Header Size
	copy(header[8:], []byte("VP80"))                // FOURCC
	binary.LittleEndian.PutUint16(header[12:], 640) // Version
	binary.LittleEndian.PutUint16(header[14:], 480) // Header Size
	binary.LittleEndian.PutUint32(header[16:], 30)  // Framerate numerator
	binary.LittleEndian.PutUint32(header[20:], 1)   // Framerate Denominator
	binary.LittleEndian.PutUint32(header[24:], 900) // Frame count
	binary.LittleEndian.PutUint32(header[28:], 0)   // Unused

	_, err := i.stream.Write(header)
	return err
}
```

Sean-Der:
Hey @kacperjurak 

The `IVFWriter` atm is really just a quick hack so we can write to disk, I would love a PR to improve it some more! Maybe when you call `Close()` it will finalize the header values.

Right now it is hard to fill in a lot of detail since we are dealing with live video, and things like resolution are impossible to fill in. The resolution can change at anytime during the stream.

Sean-Der:
@kacperjurak 

master has been updated to count the number of frames written, does this fix the playback speed for you? Also note that you need to refresh/close the tab for the file to finish writing (otherwise we write a temporary header)

Please re-open if you are still have issues! thanks for using pion :)

cc: @Antonito who did all the work here!

thanks!

PR #484, opened 2019-03-06
Hardcoded header in IVFWriter

Antonito:
Relates to #471

The framecount of the IVF header is now updated when we call the `Close` method.
I'm considering passing the `width` and `height` fields as parameters to `New()` method, and keeping a constan 30 framerate. Any thoughts on this?

### DIFF
--- a/examples/save-to-disk/main.go
+++ b/examples/save-to-disk/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/pions/rtcp"
@@ -21,12 +22,11 @@ func saveToDisk(i media.Writer, track *webrtc.Track) {
 	}()
 
 	for {
-		packet, err := track.ReadRTP()
+		rtpPacket, err := track.ReadRTP()
 		if err != nil {
 			panic(err)
 		}
-
-		if err := i.AddPacket(packet); err != nil {
+		if err := i.AddPacket(rtpPacket); err != nil {
 			panic(err)
 		}
 	}
@@ -61,6 +61,15 @@ func main() {
 		panic(err)
 	}
 
+	opusFile, err := opuswriter.New("output.opus", 48000, 2)
+	if err != nil {
+		panic(err)
+	}
+	ivfFile, err := ivfwriter.New("output.ivf")
+	if err != nil {
+		panic(err)
+	}
+
 	// Set a handler for when a new remote track starts, this handler saves buffers to disk as
 	// an ivf file, since we could have multiple video tracks we provide a counter.
 	// In your application this is where you would handle/process video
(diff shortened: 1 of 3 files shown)


Sean-Der:
Nice work @Antonito 

I think this makes the example more complicated that it needs to be (IMO) feel free to push back.

Can you try watch the `webrtc.ICEConnectionState` instead? I think the code might read cleaner if when we get `failed` we `.Close()` both files. https://github.com/pions/webrtc/blob/master/examples/save-to-disk/main.go#L100

We can then tell users 'refresh the demo page and when the client times out it will exit' also the code will be easier to follow (less LOC and no channels etc..)

Sean-Der:
Also the `width` and `height` I don't think will be possible to pass in :( 

WebRTC allows clients to change resolution on the fly (to adjust with available bandwidth) so we don't have anything we can do.

examples/save-to-disk/main.go
```diff
@@ -64,6 +64,7 @@ func main() {
 	// Set a handler for when a new remote track starts, this handler saves buffers to disk as
 	// an ivf file, since we could have multiple video tracks we provide a counter.
 	// In your application this is where you would handle/process video
+	mediaRecorder := []media.Writer{nil, nil}
```
Sean-Der: Can you put a `sync.Mutex` around this. Just need to lock since ICE+OnTrack fire in threads

examples/save-to-disk/main.go
```diff
@@ -98,6 +101,21 @@ func main() {
 	// This will notify you when the peer has connected/disconnected
 	peerConnection.OnICEConnectionStateChange(func(connectionState webrtc.ICEConnectionState) {
 		fmt.Printf("Connection State has changed %s \n", connectionState.String())
+
+		if connectionState == webrtc.ICEConnectionStateConnected {
+			fmt.Println("Ctrl+C the client to stop the demo")
```
Sean-Der: This means `.Close()` will never get called right? If you want to support `Ctrl-C` that would be really cool also?

I would put a note in the `README` the created `ivf and opus files will exist, but headers will only have temporary values and may not play properly` if they don't end the example properly.

Sean-Der:
Looks much better! This is great, just a few nits/questions but we could merge now if you don't want to support `Ctrl-C` !

examples/save-to-disk/main.go
Antonito: Oh this needs to be more explicit then. By client, I meant the JS fiddle thing for example. If you "ctrl+c" it (or close the tab), the Go program will detect the connection is closed, will update the headers, close the files and terminate.

I thought about supporting ctrl+c in the Go example, but also wanted to keep the example simple. I can add it if you think it's relevant!

PR #488, opened 2019-03-07
Fix deadlock in DTLS shutdown

Sean-Der:
DTLS shutdown deadlocks if Close is called before startup completes,
because the DTLS connection hasn't finished yet we don't have handles
to close anything.

This updates DTLS to follow how SCTP is shutdown, by shutting down the
nextConn (ICE in this case) we can shutdown the subsystem. By closing
ICE first, DTLS (and then SCTP) close properly no matter what state
they are in.

### DIFF
--- a/internal/mux/mux.go
+++ b/internal/mux/mux.go
@@ -1,10 +1,13 @@
 package mux
 
 import (
-	"fmt"
 	"net"
 	"sync"
+
+	"github.com/pions/webrtc/pkg/logging"
 )
+
+var muxLog = logging.NewScopedLogger("mux")
 
 // Mux allows multiplexing
 type Mux struct {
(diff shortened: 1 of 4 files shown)


Sean-Der:
I should have a fix soon, but wanted to make sure I had reproduce first

peerconnection_media_test.go
```diff
@@ -289,10 +289,8 @@ func TestPeerConnection_Media_Sender_Transports_OnSelectedCandidatePairChange(t
 
 	pcAnswer.OnICEConnectionStateChange(func(iceState ICEConnectionState) {
 		if iceState == ICEConnectionStateConnected {
-			go func() {
-				time.Sleep(3 * time.Second) // TODO PeerConnection.Close() doesn't block for all subsystems
-				close(iceComplete)
-			}()
+			time.Sleep(3 * time.Second) // TODO PeerConnection.Close() doesn't block for all subsystems
```
Sean-Der: This is fired in a `goroutine` already, I don't want users to see this and assume they need the same.

internal/mux/mux.go
```diff
@@ -100,7 +103,7 @@ func (m *Mux) dispatch(buf []byte) {
 	m.lock.Unlock()
 
 	if endpoint == nil {
-		fmt.Printf("Warning: mux: no endpoint for packet starting with %d\n", buf[0])
+		muxLog.Warnf("Warning: mux: no endpoint for packet starting with %d\n", buf[0])
```
Sean-Der: Just moving everything off of `fmt.`

PR #492, opened 2019-03-07
Examples: serve WASM examples locally

### LINKED ISSUE #491: WASM: Serve WASM examples locally
Adapt the server under `webrtc/examples` to serve WASM examples.

backkem:
The examples server now detects 'demo.wasm' files in
the jsfiddles. Examples layout updated to match Pion style.

**Example**
If there is a `demo.wasm` file in the examples folder:
image

The example overview will give you the option to run it:
image

Result (the gzipped SDP indicates we're using WASM):
image

Resolves #491

cc @albrow


backkem:
I didn't write documentation for this yet. It seems best to adjust the documentation when actually adding the WASM examples.

examples/wasm_exec.js
```diff
@@ -0,0 +1,535 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
Sean-Der: Can you put this in a `examples/vendor` folder maybe? Just so people know it isn't our code (and maybe include the SHA1 you got it from)

Issue #509, opened 2019-03-10
Race in  PeerConnection.newRTCTranceiver()


enobufs:
It appears that there's a race in PeerConnection.newRTCTranceiver().
Used master at rev 3426d
```
=== RUN   TestPeerConnection_Media_Sample
==================
WARNING: DATA RACE
Write at 0x00c0004d2248 by goroutine 67:
  github.com/pions/webrtc.(*PeerConnection).newRTPTransceiver()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection.go:1496 +0x20c
  github.com/pions/webrtc.(*PeerConnection).openSRTP.func1()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection.go:943 +0x2b3

Previous read at 0x00c0004d2248 by goroutine 85:
  github.com/pions/webrtc.(*PeerConnection).SetRemoteDescription.func2()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection.go:853 +0x346

Goroutine 67 (running) created at:
  github.com/pions/webrtc.(*PeerConnection).openSRTP()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection.go:928 +0x579
  github.com/pions/webrtc.(*PeerConnection).SetRemoteDescription.func2()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection.go:851 +0x32e

Goroutine 85 (running) created at:
  github.com/pions/webrtc.(*PeerConnection).SetRemoteDescription()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection.go:815 +0x1358
  github.com/pions/webrtc.signalPair()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection_test.go:45 +0xf4
  github.com/pions/webrtc.TestPeerConnection_Media_Sample()
      /Users/ytakeda/Projects/pions/webrtc/peerconnection_media_test.go:128 +0x63e
  testing.tRunner()
      /Users/ytakeda/go/current/src/testing/testing.go:827 +0x162
==================
```

```go
func (pc *PeerConnection) newRTPTransceiver(
	receiver *RTPReceiver,
	sender *RTPSender,
	direction RTPTransceiverDirection,
) *RTPTransceiver {

	t := &RTPTransceiver{
		Receiver:  receiver,
		Sender:    sender,
		Direction: direction,
	}
	pc.mu.Lock()
	defer pc.mu.Unlock()
[...]

PR #510, opened 2019-03-10
[WIP] Fix race in DataChannel.open()

### LINKED ISSUE #505: Race in DataChannel.open()
In my app, I am seeing a race condition at the following line:
https://github.com/pions/webrtc/blob/90b85e8c7a5bf92405d2a07fde0069bcfd3ab526/datachannel.go#L189
```go
	dc, err := datachannel.Dial(d.sctpTransport.association, *d.ID, cfg)
	if err != nil {
		d.mu.RUnlock()
		return err
	}

	d.ReadyState = DataChannelStateOpen // <------------ race
	d.mu.RUnlock()

	d.handleOpen(dc)
	return nil
}
```
I believe, we need to use write lock instead of read lock here.

enobufs:
Now it uses write lock in this PR.
No regression in unit tests, race detection nor golangci-lint in my local run.

I noticed the was a race (reported as #509), not caused by this modification.

Resolves #505

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -138,11 +138,11 @@ func (api *API) newDataChannel(params *DataChannelParameters) (*DataChannel, err
 
 // open opens the datachannel over the sctp transport
 func (d *DataChannel) open(sctpTransport *SCTPTransport) error {
-	d.mu.RLock()
+	d.mu.Lock()
 	d.sctpTransport = sctpTransport
 
 	if err := d.ensureSCTP(); err != nil {
-		d.mu.RUnlock()
+		d.mu.Unlock()
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


datachannel.go
```diff
@@ -182,12 +182,12 @@ func (d *DataChannel) open(sctpTransport *SCTPTransport) error {
 
 	dc, err := datachannel.Dial(d.sctpTransport.association, *d.ID, cfg)
 	if err != nil {
-		d.mu.RUnlock()
+		d.mu.Unlock()
 		return err
 	}
 
 	d.ReadyState = DataChannelStateOpen
```
Sean-Der: Can you make this so we just take the writelock for setting `d.ReadyState` I just don't want to be wasteful having a write lock for the whole function. If this is a bad idea please push back!

```
d.mu.RUnlock()
d.mu.Lock()
d.ReadyState = DataChannelStateOpen
d.mu.Unlock()
```

enobufs: I think, it wouldn't give us much because:
1. This method would virtually 100% succeed in production use (bug-free).
1. Sticking with only one kind of lock is relatively less error-prone.
1. DataChannel.open() won't be called that often anyway (I wouldn't worry)

Sean-Der: A agree with all that, I was overthinking it. More complicated code for no real gain :)

PR #514, opened 2019-03-12
Return io.EOF when reading/writing media

### LINKED ISSUE #511: RTPReceiver.Read() EOF behaviour
Perhaps I'm using the wrong pattern here, but I have a goroutine reading packets out of a `Track` and pumping them elsewhere.

At some point I will want to close the `PeerConnection`, and I think that the code below could signal this in two different ways to my goroutine:

1. The function will return `fmt.Errorf("RTPSender has been stopped")`
2. The function will return `err = EOF` from `r.rtcpReadStream.Read()`

https://github.com/pions/webrtc/blob/3426d00e2cda2eba9dfff3cb71be6ffda68d813c/rtpreceiver.go#L101-L108

Is my evaluation correct? Can this be done a different way such that I can more gracefully handle a `PeerConnection` being torn down?

Sean-Der:
Using this enum allows people to more gracefully handle shutdown errors

Resolves #511

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -8,6 +8,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"fmt"
+	"io"
 	"net"
 	"strconv"
 	"strings"
@@ -1330,7 +1331,10 @@ func (pc *PeerConnection) WriteRTCP(pkt rtcp.Packet) error {
 	if _, err := writeStream.Write(raw); err != nil {
 		if err == ice.ErrNoCandidatePairs {
 			return nil
+		} else if err == ice.ErrClosed {
+			return io.ErrClosedPipe
 		}
+
 		return fmt.Errorf("WriteRTCP failed to write: %v", err)
 	}
 	return nil
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -5,6 +5,7 @@ package webrtc
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"math/rand"
 	"sync"
 	"testing"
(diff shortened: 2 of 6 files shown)


Sean-Der:
@rob-deutsch @backkem 

Following the conversation I *think* this is the right thing to do? You can detect graceful shutdown now that you have an enum

Sean-Der:
Nice thank you @backkem :)

@rob-deutsch would you mind testing this locally and seeing if that works for you behavior wise? I will also add some tests afterwards to assert it behaves as expected!

rob-deutsch:
Hey, sorry I haven't had a chance to engage in this conversation.

I'm not 100% convinced this is the best way forward yet. I'm not saying that it isn't. Just that I'm having some broader thoughts. I'll post in #511 shortly.

kixelated:
I'm about to merge `read-buffer`, which conflicts but also makes this change.

rob-deutsch:
Thanks for putting together this PR!

I've posted a design question in #511 which might impact this PR. If this path forward is chosen, though, then should `RTPReceiver.readRTP()` and `RTPSender.sendRTP()` get the same treatment?

rob-deutsch:
Thanks for putting together this PR!

I've posted a design question in #511 which might impact this PR. If this path forward is chosen, though, then should RTPSender.sendRTP() should return `io.ErrClosedPip()`?

This is getting complicated. Maybe RTPReceiver and RTPSender should act more like `io.PipeReader` and `io.PipeWriter` than just `io.Reader` and `io.Writer`?

I'm sorry. I don't mean to make this so difficult. I think that, ultimately, this PR is strictly better than not having this PR, but I'm not yet convinced that it's the 'right' way to go.

What I plan to do now is just go back to "first principles" thinking about how `Track` _should_ behave (or, at least, how I'd suggest that `Track` should behave) and post my thoughts in #511.

Sean-Der:
Hey @backkem @maxhawkins @rob-deutsch (and mind adding anyone else that cares about this area)

This adds all the error handling/logic around media and PeerConnection shutdown. Here is what we are going to do.

* Reading from a closed PeerConnection (RTP or RTCP) returns `io.EOF`
* Writing to a closed track returns `io.ErrClosedPipe`
   - Writing a track that hasn't opened yet (or is in the process of connecting) does NOT return an error. IMO this is consistent with browser behavior, we don't want to expose the ICE state at all via Media APIs

PR #522, opened 2019-03-17
Implement RTPDump Reader/Writer

maxhawkins:
#### Description

RTPDump is a widely-implemented file format for saving RTP packet dumps
without the overhead of UDP and IP headers found in pcap dumps. libWebRTC,
Wireshark, and RTPTools all have an implementation.

For more information see https://www.cs.columbia.edu/irt/software/rtptools

This commit enables users to save RTP media streams to a file and play them
back later.

#### Reference issue
Relates to #522

### DIFF
--- a/pkg/media/rtpdump/reader.go
+++ b/pkg/media/rtpdump/reader.go
@@ -1,0 +1,103 @@
+package rtpdump
+
+import (
+	"bufio"
+	"io"
+	"regexp"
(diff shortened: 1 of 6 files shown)


maxhawkins:
I want to make one change before merging this. When writing some other code I found that Packet.Offset is easier to use if it's a time.Duration rather than an integer.

Issue #526, opened 2019-03-19
Change PeerConnection public properties to methods 


enobufs:
## Summary
PeerConnection has public properties. We should change those to methods so that we could better control thread-safety for those parameters.

## Motivation
These properties are currently not protected by mutex, but we may like to make API completely thread-safe in the future. (See #525)
There's a PR for changing DataChannel properties to methods already (#524), and in fact, some of those had race condition and those had to be a method (#505). For consistency, we should do the same for PeerConnection.

## Describe alternatives you've considered
Any other options not doing this now could bite us back in the future.

## Additional context
* Naming convention: https://golang.org/doc/effective_go.html#Getters
* @albrow is working on JS API (WASM) now. Good have this done asap.

PR #527, opened 2019-03-19
Change PeerConnection public props to methods

enobufs:
The change is based on the PR #524 since both of these are about turning public properties to methods. Please review and merge the PR 524 first.

Relates to #526
Blocked by PR #524


peerconnection_go_test.go
```diff
@@ -357,3 +357,25 @@ func TestPeerConnection_ShutdownNoDTLS(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestPeerConnection_PeropertyGetters(t *testing.T) {
```
albrow: Small typo here: `Peroperty` -> `Property`.

PR #536, opened 2019-03-20
Add feedback support to SDP generation.

kixelated:
These are optional RTCP packets that will be sent if enabled in the
offer/answer. They currently all default to disabled because
pions/webrtc has no built-in support to handle these packets. We can
change the default values in the future but until then, they use too
much bandwidth.

The API may be controversial because I'm extending the WebRTC spec. I
don't think that there's any issue providing a superset of the API. We
already do it today by adding methods to RTPSender, and it makes the
interface a lot nicer.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -197,10 +197,11 @@ func NewRTPCodec(
 
 // RTPCodecCapability provides information about codec capabilities.
 type RTPCodecCapability struct {
-	MimeType    string
-	ClockRate   uint32
-	Channels    uint16
-	SDPFmtpLine string
+	MimeType     string
+	ClockRate    uint32
+	Channels     uint16
+	SDPFmtpLine  string
+	RTCPFeedback []RTCPFeedback
 }
 
 // RTPHeaderExtensionCapability is used to define a RFC5285 RTP header extension supported by the codec.
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1373,6 +1373,10 @@ func (pc *PeerConnection) addRTPMediaSection(d *sdp.SessionDescription, codecTyp
 
 	for _, codec := range pc.api.mediaEngine.getCodecsByKind(codecType) {
 		media.WithCodec(codec.PayloadType, codec.Name, codec.ClockRate, codec.Channels, codec.SDPFmtpLine)
+
+		for _, feedback := range codec.RTPCodecCapability.RTCPFeedback {
+			media.WithValueAttribute("rtcp-fb", fmt.Sprintf("%d %s %s", codec.PayloadType, feedback.Type, feedback.Parameter))
+		}
 	}
 
 	weSend := false
--- a/rtcpfeedback.go
+++ b/rtcpfeedback.go
@@ -1,0 +1,14 @@
+package webrtc
+
+// RTCPFeedback signals the connection to use additional RTCP packet types.
+// https://draft.ortc.org/#dom-rtcrtcpfeedback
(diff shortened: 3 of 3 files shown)


backkem:
I think we should follow the ORTC API to implement this:
- [RTCRtcpFeedback ](https://draft.ortc.org/#dom-rtcrtcpfeedback)
- [RTCRtpCodecCapability](https://draft.ortc.org/#dom-rtcrtpcodeccapability)
- [RTCRtpCodecParameters](https://draft.ortc.org/#dom-rtcrtpcodecparameters)

Work on this was started in [issue-272](https://github.com/pions/webrtc/compare/issue-272). Sadly this branch got outdated and needs resurrection. It tries to follow [webrtcHacks/adapter](https://github.com/webrtcHacks/adapter) for layering WebRTC over ORTC.

There still may be need for extending the API but the core should follow these specs IMO.

kixelated:
@backkem 

You could implement this yourself with the ORTC API, by you know, literally parsing/generating the SDP yourself. But the point of the WebRTC API is to perform the SDP crap for you. If it's not possible to add a single field to the SDP, then it is not a very useful API.

I do have a branch that switches my entire project from the WebRTC API to the ORTC API. I perform all of the SDP manipulation myself. This is basically a fork of `peerconnection.go` and costs me the ability to contribute or merge upstream patches because I needed to add a single line to the SDP. 

I would love an adapter layer that converts ORTC params <=> SDP. However, this doesn't address the issue of the WebRTC API not being extensible and forcing people to use the ORTC API.

trivigy:
@backkem @kixelated  I have a local sdp version that does all of the `rtcp-fb` attribute parsing and stuff completely implemented already. I just cannot seem to finish up my own work to get that stuff upstream into the `pions/sdp`. I am happy to share my version of the library if someone wants to spend the time making sure that everything works backward compatible.

backkem:
I do agree that you should be able to influence this behavior, also when using the WebRTC API. Extrapolating from the existing conventions it seems this should be doable as follows:
- Extend the `RTPCodecCapability` with [RTCRtcpFeedback](https://draft.ortc.org/#dom-rtcrtpcodeccapability-rtcpfeedback) as detailed in the ORTC API.
- The `RTPCodecCapability` is part of a [RTPCodec](https://github.com/pions/webrtc/blob/7cd3d00a38de36b47ea013b160b70732eacc345a/mediaengine.go#L167).
- You can construct your own `RTPCodec` and add it to the media engine using the [RegisterCodec](https://github.com/pions/webrtc/blob/7cd3d00a38de36b47ea013b160b70732eacc345a/mediaengine.go#L26) method.

kixelated:
@backkem I'll give that a shot. I do like the abstraction that you're signalling support via the track/codec, rather than directly controlling the SDP offer/answer.

That being said, this overlap between WebRTC and ORTC is confusing and I really think the APIs need to be split into separate packages at some point.

kixelated:
Heh, I forgot to say that I'm trying to make a better interface. What about these as helpers?

```
var RTCPFeedbackNACK = RTCPFeedback{Type: "nack"}
var RTCPFeedbackPLI = RTCPFeedback{Type: "nack", Parameter: "pli"}
var RTCPFeedbackSLI = RTCPFeedback{Type: "nack", Parameter: "sli"}
var RTCPFeedbackREMB = RTCPFeedback{Type: "goog-remb"}
var RTCPFeedbackTransportCC = RTCPFeedback{Type: "transport-cc"}
```

PR #537, opened 2019-03-21
Improve sfu-ws broadcast datachannel example

imalic3:
First of all, thanks for the best lib for WebRTC in Go.

I'm doing with sfu-ws example and trying to improve the example with datachannel broadcasting.


imalic3:
Thank you @backkem, I will edit the changed.

PR #539, opened 2019-03-21
Fix error reporting

zigazeljko:
Wrong error variable was used in data-channels-create example.


Sean-Der:
Hey @zigazeljko 
Nice work, thanks for catching this :)

Would you mind updating the README to include yourself in the contributors? After that the build should be green, and we can merge it.

thanks

Sean-Der:
Fantastic this landed with ee5aa24977f88fc652007f257e4d0b4ef7d57d51

thank you so much for the contribution @zigazeljko

PR #542, opened 2019-03-22
Improve build tags for Wasm

### LINKED ISSUE #515: Add any missing build tags to include or exclude files in WASM environments
I've been adding the appropriate build tags in an ad hoc fashion. Really we should go through every file and make sure that the files that get included with ` GOOS=js GOARCH=wasm` are only those which are actually supported.

albrow:
#### Description

This includes a few small and closely related changes:

1. All occurrences of the build tag `+build js` have been changed to the
   more precise `+build js,wasm`. This will exclude the files from being
   included by third-party compilers like GopherJS, with which they are
   incompatible.
2. Some files which are incompatible with JavaScript/Wasm now have the
   correct build tag (`+build -js`) so they will be excluded from Wasm
   builds.
3. Some configuration options which are incompatible with
   JavaScript/Wasm (or at least the current bindings) will now no longer
   appear in Wasm builds. This meant creating new files with new struct
   definitions and the appropriate build tags.

#### Reference issue
Fixes #515

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,3 +1,5 @@
+// +build !js
+
 package webrtc
 
 // API bundles the global funcions of the WebRTC and ORTC API.
--- a/certificate.go
+++ b/certificate.go
@@ -1,3 +1,5 @@
+// +build !js
+
 package webrtc
 
 import (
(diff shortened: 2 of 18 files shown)


albrow:
@Sean-Der I'm not sure if you have already seen it, but after @backkem's suggestion, I created https://github.com/pions/webrtc/issues/519 to track the remaining divergences.

PR #543, opened 2019-03-22
Don't notify users via Track of ICE errors

### LINKED ISSUE #523: Handle disconnected peers with track sharing
### Your environment.
  - Version: ab81640

### What did you do?
Using the basic SFU example with some custom signalling we connect a source(pions) and several consumers(browsers), and let them run. At least 1 in every 10 connections will transition to disconnected after approx 30 seconds which is fine but as soon as one of the peer connections using the sources track terminates all of the remaining peers stop receiving data. The Write method on the local track returns "No Valid Candidate Pairs Available" immediately and all consumers must manually reconnect after the source peer connection has been recreated from scratch (consumers enter an eternal connected state and cannot be closed, source also can't be cleaned up).

### What did you expect?
One transitioning ICE state should not down the whole SFU. Not sure of the best way to fix this, but it would be very useful if we had a "don't give up" parameter for this use case or provide an iterator for the senders that we can extract the offending peer connection from.

Something like
```
for _, s := range localTrack.Senders() {
	if err := s.SendRTP(rtpBuf[:i]); err != nil {
		// User defined
	}
}
or
localTrack.ForEachSender(func(s *webrtc.RTPSender) bool {
	if err := s.SendRTP(rtpBuf[:i]); err != nil {
		return true / false // Where true means break
	}

	return false
})
```

Sean-Der:
Writing to a Track shouldn't return errors for an individual
RTPSender. This filters ErrNoCandidatePairs from being returned
and instead just returns nil

Resolves #523

----

@thetooth mind checking out this PR and seeing if it works for you? thanks!

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -27,9 +27,10 @@ import (
 type DTLSTransport struct {
 	lock sync.RWMutex
 
-	iceTransport     *ICETransport
-	certificates     []Certificate
-	remoteParameters DTLSParameters
+	iceTransport      *ICETransport
+	certificates      []Certificate
+	remoteParameters  DTLSParameters
+	remoteCertificate []byte
 	// State     DTLSTransportState
 
 	// OnStateChange func()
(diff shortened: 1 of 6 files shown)


Sean-Der:
Adding @backkem @rob-deutsch 

You both have context in this area, I believe this is the right change describe [here](https://github.com/pions/webrtc/issues/523#issuecomment-474677501)

thetooth:
Just confirmed working well. Additional errors after this change are "failed to verify auth tag" and "Failed to retransmit init (nRtos=5): EOF", however these actually stop once the dead peers get collected.

Sean-Der:
fantastic, thanks @thetooth! I am going to write a test and get this merged.

We didn't have leveled logging available yet (so SCTP had to printf everything) but in the next ~week those should go away!

Sean-Der:
@backkem @rob-deutsch Mind giving this a quick review/sign-off? thanks!

Sean-Der:
Added more people who work with media APIs

PR #546, opened 2019-03-22
Add ICE candidate event handlers

albrow:
#### Description

Add OnICECandidate and OnICEGatheringStateChange methods to PeerConnection. The main goal of this change is to improve API compatibility with the JavaScript/Wasm bindings. It does not actually add trickle ICE support or change the ICE candidate gathering process, which is still synchronous in the Go implementation. Rather, it fires the appropriate events similar to they way they would be fired in a true trickle ICE process.


albrow:
This PR doesn't add all methods. Before we merge this PR, I plan to make another pass to see if there are any other easy changes to make the Go and JavaScript/Wasm APIs more similar.

albrow:
There are still potentially more event handlers or methods to add in the future, but I've done all I want to do for this PR. The Go implementation and the JavaScript/Wasm bindings now have matching function signatures for `OnICECandidate` and `OnICEGatheringStateChange`.

peerconnection.go
```diff
@@ -238,6 +238,61 @@ func (pc *PeerConnection) OnDataChannel(f func(*DataChannel)) {
...
+		pc.onICECandidateHandler(nil)
+	}
+
+	pc.iceGatheringState = ICEGatheringStateComplete
+
+	// Also trigger the onICEGatheringStateChangeHandler
+	if pc.onICEGatheringStateChangeHandler != nil {
+		// Note: Gathering is already done at this point, but some clients might
+		// still expect the state change handler to be triggered.
+		pc.onICEGatheringStateChangeHandler()
```
Sean-Der: I think we need to call this in a goroutine (copying how `onTrack` does it) otherwise someone via the API could do something blocking (and then `pc.mu.Lock()` could be held for a long time)

peerconnection.go
```diff
@@ -238,6 +238,61 @@ func (pc *PeerConnection) OnDataChannel(f func(*DataChannel)) {
...
+
+// signalICECandidateGatheringComplete should be called after ICE candidate
+// gathering is complete. It triggers the appropriate event handlers in order to
+// emulate a trickle ICE process.
+func (pc *PeerConnection) signalICECandidateGatheringComplete() error {
+	pc.mu.Lock()
+	defer pc.mu.Unlock()
+
+	// Call onICECandidateHandler for all candidates.
+	if pc.onICECandidateHandler != nil {
```
Sean-Der: Same as below! Just make sure not to hold any locks when hitting callbacks for public API

albrow:
@Sean-Der I applied changes based on your feedback. Ready for another look.

albrow:
> Do you feel good about it testing wise? It might be nice to assert we get candidates or we got a nil, also another useful test is the blocking in the callbacks doesn't break anything.

Testing could definitely be improved, but I think it's good enough for now. The updated `signalPair` function uses the `OnICECandidate` event handler, and `signalPair` is used in a lot of places. `OnICEGatheringStateChange` doesn't have any test coverage but it's also really simple.

albrow:
And yes, I'll squash before merging.

PR #547, opened 2019-03-23
Rewrite examples to avoid onnegotiationneeded

### LINKED ISSUE #545: libwebrtc has broken onnegotiationneeded
### Your environment.
  - Version: *N/A*
  - Browser: *Chromium 73*
  - Other Information - *N/A*

### What did you do?
Use JSfiddle examples

### What did you expect?
Them to work

### What happened?
Chrome throws an exception, libwebrtc broken onnegotiationneeded. A upstream thread was started [here](https://groups.google.com/forum/#!topic/discuss-webrtc/3-TmyjQ2SeE)

```
OperationError: Failed to execute 'setLocalDescription' on 'RTCPeerConnection': Failed to set local offer sdp: The order of m-lines in subsequent offer doesn't match order from previous offer/answer.
```

Sean-Der:
onnegotiationneeded is broken in libwebrtc M73, so
just rewrite all examples to only set the offer once

Resolves #545


Sean-Der:
Threw on a bunch of people who have edited these files it said, just need a quick sign-off to fix an annoying issue!

thanks

examples/gstreamer-receive/jsfiddle/demo.js
```diff
@@ -12,8 +12,10 @@ var log = msg => {
 }
 
 navigator.mediaDevices.getUserMedia({ video: true, audio: true })
-  .then(stream => pc.addStream(document.getElementById('video1').srcObject = stream))
-  .catch(log)
+  .then(stream => {
+    pc.addStream(document.getElementById('video1').srcObject = stream)
+    pc.createOffer().then(d => pc.setLocalDescription(d)).catch(log)
```
svranesevic: Add something along the lines of:
```js
document.getElementById('localSessionDescription').value = btoa(JSON.stringify(d))
```
That way SD will be displayed right away in the browser

Sean-Der: Hey @svranesevic thanks for the review :)

The SD isn't complete yet, we are still waiting for ICE to finish. If we populate here we could generate an offer with no candidates (and Pion would have nothing to connect too). It might coincidentally work if things were fast enough or something blocked JS elsewhere though!

When `onicecandidate` returns NULL it means we have all our candidates [MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onicecandidate#Value)

Sean-Der: Hey @svranesevic if everything looks good would you be up for signing-off on this?

I would love to get this merged so other people don't run into the same issues. Our CI just doesn't let me merge without one review.

thanks!

svranesevic: Hey @Sean-Der, sure!

PR #554, opened 2019-03-27
Let sfu-ws example uses latest webrtc codes And introduce prommods

jinleileiking:
And uses ice standalone package.


jinleileiking:
@backkem @Sean-Der  Could you honour give me a review?

backkem:
@jinleileiking Can you explain your reasoning for adding promod? I think it makes the example less to the point. These aren't meant to be fully fledged apps, just examples of using the library.

PR #562, opened 2019-03-28
Custom Loggers

hugoArregui:
LeveledLogger is an interface now


backkem:
Awesome, that looks even cleaner 👍

PR #565, opened 2019-03-29
WASM Detach

backkem:
Support for Detach when targeting JS/WASM. The goal is to enable the same idiomatic API in both Native and WASM. This is a POC of the shim to turn the JS OnMessage Callback into the ReadWriteCloser interface used by the DataChannel.Detach. I tried to avoid leaking JS code into pions/datachannel by introducing an interface. Therefore this depends on pions/datachannel#11.

#### TODO
- [x] The WASM example deadlocks. Have to look into it further.
- [x] In order to use the same code in Native and WASM we would have to find a common solution for the way Detach is enabled in the SettingEngine in Native ATM.
- [ ] Performance improvements. Maybe you can port your POC over @albrow?

### DIFF
--- a/datachannel_js.go
+++ b/datachannel_js.go
@@ -5,6 +5,8 @@ package webrtc
 import (
 	"fmt"
 	"syscall/js"
+
+	"github.com/pions/datachannel"
 )
 
 const dataChannelBufferSize = 16384 // Lowest common denominator among browsers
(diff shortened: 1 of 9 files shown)


backkem:
Fixed some small bugs in the way `Blob` messages are handled. This fixes the deadlock issue.

datachannel_js.go
```diff
@@ -60,8 +66,15 @@ func (d *DataChannel) OnMessage(f func(msg DataChannelMessage)) {
...
 	onMessageHandler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
-		msg := valueToDataChannelMessage(args[0].Get("data"))
-		go f(msg)
+		// TODO: Ensure message order?
+		data := args[0].Get("data")
+		go func() {
+			// valueToDataChannelMessage may block when handling 'Blob' data
+			// so we need to call it from a new routine. See:
+			// https://godoc.org/syscall/js#FuncOf
+			msg := valueToDataChannelMessage(data)
```
albrow: 👍 Nice catch!

datachannel_js.go
```diff
@@ -243,6 +273,8 @@ func valueToDataChannelMessage(val js.Value) DataChannelMessage {
 			return js.Undefined()
 		}))
 
+		reader.Call("readAsArrayBuffer", val)
```
albrow: I'm guessing the purpose of this call is to flush the buffer? Maybe this is worth a small comment becuase it's not immediately obvious.

backkem: I think this was just missing from the conversion code. It's mentioned in the MDN docs linked a couple lines above:
https://developer.mozilla.org/en-US/docs/Web/API/Blob#Example_for_extracting_data_from_a_Blob

Without this line `val` is never passed to the reader. Maybe we never tested the Blob type in the end.

PR #579, opened 2019-04-03
Migrate SDP generation to Unified Plan

Sean-Der:
This commit does cause some breaking changes. This API change means we
can no longer support an arbitrary number of receivers. For every track
you want to receive you MUST call PeerConnection.AddTransceiver

We do now support sending an multiple audio/video feeds. You can see
this behavior via gstreamer-receive and gstreamer-send currently.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -7,6 +7,8 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
+	mathRand "math/rand"
+
 	"fmt"
 	"io"
 	"net"
@@ -135,7 +137,7 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
 	pc.iceTransport = iceTransport
 
 	// Create the DTLS transport
-	dtlsTransport, err := pc.createDTLSTransport()
+	dtlsTransport, err := pc.api.NewDTLSTransport(pc.iceTransport, pc.configuration.Certificates)
 	if err != nil {
 		return nil, err
 	}
@@ -431,10 +433,6 @@ func (pc *PeerConnection) GetConfiguration() Configuration {
 	return pc.configuration
 }
 
-// ------------------------------------------------------------------------
-// --- FIXME - BELOW CODE NEEDS REVIEW/CLEANUP
-// ------------------------------------------------------------------------
-
 // CreateOffer starts the PeerConnection and generates the localDescription
 func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription, error) {
 	useIdentity := pc.idpLoginURL != nil
@@ -461,16 +459,18 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
 	}
 
 	bundleValue := "BUNDLE"
-
-	if pc.addRTPMediaSection(d, RTPCodecTypeAudio, "audio", iceParams, RTPTransceiverDirectionSendrecv, candidates, sdp.ConnectionRoleActpass) {
-		bundleValue += " audio"
-	}
-	if pc.addRTPMediaSection(d, RTPCodecTypeVideo, "video", iceParams, RTPTransceiverDirectionSendrecv, candidates, sdp.ConnectionRoleActpass) {
-		bundleValue += " video"
+	bundleCount := 0
+	appendBundle := func() {
+		bundleValue += " " + strconv.Itoa(bundleCount)
+		bundleCount++
 	}
 
-	pc.addDataMediaSection(d, "data", iceParams, candidates, sdp.ConnectionRoleActpass)
-	d = d.WithValueAttribute(sdp.AttrKeyGroup, bundleValue+" data")
+	for _, t := range pc.GetTransceivers() {
+		pc.addTransceiverSDP(d, t, bundleCount, iceParams, candidates, sdp.ConnectionRoleActpass)
+		appendBundle()
+	}
+	pc.addDataMediaSection(d, bundleCount, iceParams, candidates, sdp.ConnectionRoleActive)
+	appendBundle()
 
 	for _, m := range d.MediaDescriptions {
 		m.WithPropertyAttribute("setup:actpass")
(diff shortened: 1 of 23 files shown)


Sean-Der:
I added everyone who works with media APIs. Don't feel obligated to do a full review, but want to make sure I am not doing anything stupid. 

Also I assume we will get some people who hit issues because of this commit, so just to make sure we are all on the same page for people who need help in Slack :)

mjmac:
A point of clarification... This is not intended to be backwards-compatible with plan-b, right? i.e. When this change lands, everything will need to be updated to use unified plan? I think that's probably fine (because 2.0.0), but it's going to be a big breaking change and worth pointing out clearly.

peerconnection.go
```diff
@@ -957,64 +968,72 @@ func (pc *PeerConnection) openSRTP() {
...
+					pc.onTrack(receiver.Track(), receiver)
+				} else {
+					pc.log.Warnf("OnTrack unset, unable to handle incoming media streams")
+				}
+			}(ssrc, t.Receiver)
+			break
+		}
 	}
+
+	// for i := range incomingSSRCes {
```
maxhawkins: Left this comment in

maxhawkins:
Will this still support clients running plan-b? Many of my application's clients are running older versions of WebRTC.

peerconnection.go
```diff
@@ -567,43 +562,63 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
...
-				peerDirection = RTPTransceiverDirectionSendonly
-			case strings.HasPrefix(*a.String(), "recvonly"):
-				peerDirection = RTPTransceiverDirectionRecvonly
+			case t.kind != kind:
+				continue
+			case direction == RTPTransceiverDirectionSendrecv && t.Direction != RTPTransceiverDirectionSendrecv:
+				continue
+			case direction != RTPTransceiverDirectionSendrecv && direction == t.Direction:
+				continue
+			case direction == RTPTransceiverDirectionInactive:
```
mjmac: My application code is send-only (just calls AddTrack()), so I had to add the following statement in order to get this to work with chrome as offerer (direction.recvOnly):
```go
case direction == RTPTransceiverDirectionRecvonly && t.Direction == RTPTransceiverDirectionSendonly:
        fallthrough
```

mjmac:
> Will this still support clients running plan-b? Many of my application's clients are running older versions of WebRTC.

FWIW, in my quick testing with an existing plan-b offerer, it failed. I didn't look into why, just started working on migrating my clients to unified. There may be a way to make our side compatible with both, but not sure how much extra work that is.

mjmac:
This shows the offer/answer SDP generation as it currently stands (all plan b):
[offer-plan-b-answer-plan-b.txt](https://github.com/pions/webrtc/files/3040114/offer-plan-b-answer-plan-b.txt)

This shows the offer/answer SDP generation with the PR code (offer is plan b, pions generates unified):
[offer-plan-b-answer-unified.txt](https://github.com/pions/webrtc/files/3040111/offer-plan-b-answer-unified.txt)

Chromium rejects this answer with the following error:
```(peer_connection.cc:2492): Failed to set remote answer sdp: The order of m-lines in answer doesn't match order in offer. Rejecting answer.```

Note that if I change my client to unified, it works (with the patch I included in my inline comment).

peerconnection.go
Sean-Der: Fixed, thanks max!

Sean-Der:
Oh yea @mjmac multi-track is a hard `this will break` during the migration from Chromium.

* For 'Pion Sending' I think it would be easy the transform the SDP though (and would keep the guts of Pion unified only)

* For 'Pion receiving' I don't have a good answer :/ we could support exposing arbitrary SRTP streams to the user (and throw an error message out maybe?) or maybe it is worth exploring ORTC for this use case?

peerconnection.go
Sean-Der: I am feeling dense, but which case here is causing a `continue` when remote is `recvonly` and we have a local `sendonly`? Reading I think it should fail all cases (and should return a local `Transceiver`)

I will add a test though, and reproduce locally?

mjmac: Sorry! I should have pasted in patch form; easier to see:

```diff
--- a/vendor/github.com/pions/webrtc/peerconnection.go
+++ b/vendor/github.com/pions/webrtc/peerconnection.go
@@ -582,6 +582,8 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
                                continue
                        case direction != RTPTransceiverDirectionSendrecv && direction == t.Direction:
                                continue
+                       case direction == RTPTransceiverDirectionRecvonly && t.Direction == RTPTransceiverDirectionSendonly:
+                               fallthrough
                        case direction == RTPTransceiverDirectionInactive:
                                continue
                        }```

Hmm. Did you update your PR? The last case (inactive) didn't have a continue when I looked at this earlier. I thought the logic was that if the offer's direction was inactive, it would satisfy that media entry with a localTransceiver. 

In any case, I'm probably not understanding the overall intent here. My patch was attempt to allow a localTransceiver with direction set to sendOnly to satisfy a remote media entry set to recvOnly. This worked in my testing.

Sean-Der: That was my mistake! I messed up the `inactive` branch earlier.

That switch is to skip a local Transceiver if it can't satisfy what the remote needs (so only sendrecv can work for sendrecv, sendonly and recvonly can satisify each other and inactive can't satisfy anything)

mjmac:
@Sean-Der: please take a look at the attached patch:

[0002-WIP-support-plan-b-peer-offer.txt](https://github.com/pions/webrtc/files/3040600/0002-WIP-support-plan-b-peer-offer.txt)

It works in my quick local testing. If you think that this is a worthwhile way to proceed, then maybe the right way to do this is for you to go ahead and get this PR merged and then I can make my own PR for plan-b compatibility.

Sean-Der:
@mjmac I am 100% behind that patch

It is minimal, so I would be fine keeping it for a long time. Also not a place we work in often, so I don't see it conflicting ever.

Sean-Der:
@mjmac You ok with moving forward without Plan-B?

My #1 concern is impacting people that just drop on Pion for first time (or just users that don't know WebRTC that well) by not having Unified Plan atm we are behind the curve :(

I say we add the force Plan-B/Unified, that is probably the best path forward. Also once this lands I can start exporting all our examples that use 3rd party libs.

Issue #361, opened 2019-01-16
Logging cleanup


mjmac:
Creating an umbrella issue for log messages which are deemed by consensus to be at the wrong log level (e.g. at Debug when Trace would be more reasonable). Refer to this issue when creating PRs for logging adjustments.

gissleh:
Does this issue also include removing std outputs like this one from `gatherCandidatesLocal()` in `agent.go`? These are very unhelpful when using stdin and stdout for IPC.
 
```
could not listen udp fe80::2c36:8bff:fea6:ab50
```

mjmac:
Hi @gissleh!

Sure, I think so. We've been gradually replacing all of those with proper logging statements but I'm not surprised that a few got missed.

backkem:
Other points that need investigating:
- Logging is now global. This may lead to issues when using multiple instances of the library.
- Logging is now limited to package webrtc. We should roll this out to the other ones.

backkem:
After #535 lands we need to move all variants of `fmt.Println` over to `pions/logging`.

PR #585, opened 2019-04-05
Add lint to disallow fmt.* functions

Sean-Der:
Relates to #361 

----

When this lands I am going to do one for every other library used by pion/webrtc!

### DIFF
--- a/.github/lint-disallowed-functions-in-library.sh
+++ b/.github/lint-disallowed-functions-in-library.sh
@@ -4,7 +4,7 @@ set -e
 # Disallow usages of functions that cause the program to exit in the library code
 SCRIPT_PATH=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 EXCLUDE_DIRECTORIES="--exclude-dir=examples --exclude-dir=.git --exclude-dir=.github "
-DISALLOWED_FUNCTIONS=('os.Exit(' 'panic(' 'Fatal(' 'Fatalf(' 'Fatalln(')
+DISALLOWED_FUNCTIONS=('os.Exit(' 'panic(' 'Fatal(' 'Fatalf(' 'Fatalln(' 'fmt.Println(' 'fmt.Printf(' 'log.Print(' 'log.Println(' 'log.Printf(')
 
 
 for disallowedFunction in "${DISALLOWED_FUNCTIONS[@]}"
--- a/quictransport.go
+++ b/quictransport.go
@@ -8,12 +8,12 @@ import (
 	"crypto/rand"
 	"crypto/x509"
 	"errors"
-	"fmt"
 	"strings"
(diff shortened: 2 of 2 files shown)


certificate.go
```diff
@@ -93,24 +93,23 @@ func (c Certificate) Expires() time.Time {
...
 	fingerprintAlgorithms := []dtls.HashAlgorithm{dtls.HashAlgorithmSHA256}
 	res := make([]DTLSFingerprint, len(fingerprintAlgorithms))
 
 	i := 0
 	for _, algo := range fingerprintAlgorithms {
 		value, err := dtls.Fingerprint(c.x509Cert, algo)
 		if err != nil {
-			fmt.Printf("Failed to create fingerprint: %v\n", err)
-			continue
+			return nil, fmt.Errorf("failed to create fingerprint: %v", err)
```
hugoArregui: Are you sure you want to do this change here? This seems to change the behaviour. If the change is correct maybe clarify it on the commit and/or PR for traceability?

maxhawkins: Doesn't this mean the function will fail if any of the fingerprints fail rather than choosing the last working fingerprint like before?

examples/ortc-quic/main.go
```diff
@@ -73,7 +73,10 @@ func main() {
 		panic(err)
 	}
 
-	quicParams := qt.GetLocalParameters()
+	quicParams, err := qt.GetLocalParameters()
+	if err != nil {
+		panic(err)
```
maxhawkins: I thought panic wasn't allowed

peerconnection.go
```diff
@@ -1475,11 +1479,16 @@ func (pc *PeerConnection) iceStateChange(newState ICEConnectionState) {
 	pc.onICEConnectionStateChange(newState)
 }
 
-func (pc *PeerConnection) addFingerprint(d *sdp.SessionDescription) {
+func (pc *PeerConnection) addFingerprint(d *sdp.SessionDescription) error {
 	// TODO: Handle multiple certificates
-	for _, fingerprint := range pc.configuration.Certificates[0].GetFingerprints() {
+	fingerprints, err := pc.configuration.Certificates[0].GetFingerprints()
```
maxhawkins: What if `len(certificates) == 0`?

examples/ortc-quic/main.go
Sean-Der: It is currently whitelisted in examples

certificate.go
Sean-Der: @maxhawkins 

I would rather fail fast on this one (I just can't think of a case where something like this would fail and I wouldn't want to know)

Also I have never seen this error message printed before (and haven't gotten user feedback around it) but I do think I would rather things fail then move past it.

Happy to split into its own commit though, if you both are good with this change.

PR #589, opened 2019-04-05
Disable quic by default; enable with build tag

mjmac:
Taking a more extreme approach than #588... This patch
disables quic by default. The advantage here would be
that webrtc builds with older versions of go than
quic-go supports (currently 1.12+).


albrow:
@mjmac as an alternative I think you can use `go1.12` instead of creating a custom build tag. See https://golang.org/pkg/go/build/

PR #590, opened 2019-04-05
Support plan-b (fallback or explicit)

mjmac:
By default, we should be using unified-plan as the
SDP format of choice. This patch adds a PeerConnection
configuration option to allow the user to specify that
they want plan-b only, unified-plan only, or a special
compatibility mode where a plan-b answer will be generated
IFF a plan-b offer is received (this is the default).


peerconnection.go
```diff
@@ -1502,7 +1564,12 @@ func (pc *PeerConnection) addFingerprint(d *sdp.SessionDescription) error {
 	return nil
 }
 
-func (pc *PeerConnection) addTransceiverSDP(d *sdp.SessionDescription, t *RTPTransceiver, midValue string, iceParams ICEParameters, candidates []ICECandidate, dtlsRole sdp.ConnectionRole) {
+func (pc *PeerConnection) addTransceiverSDP(d *sdp.SessionDescription, midValue string, iceParams ICEParameters, candidates []ICECandidate, dtlsRole sdp.ConnectionRole, transceivers ...*RTPTransceiver) {
+	if len(transceivers) < 1 {
+		return
```
Sean-Der: Should this be an error? Do you ever expect to add a media section without a `Transceiver` (adding an error might help us catch weird stuff in the future)

peerconnection.go
```diff
@@ -1002,51 +1153,25 @@ func (pc *PeerConnection) openSRTP() {
...
-				} else {
-					pc.log.Warnf("OnTrack unset, unable to handle incoming media streams")
-				}
-			}(ssrc, t.Receiver)
+			go startReceiver(ssrc, t.Receiver)
 			break
 		}
 	}
+
+	if remoteIsPlanB {
```
Sean-Der: Feel free to push back on this (might be a nit)

Do you think it is worth creating the `incomingSSRCes` at the very beginning of `SetRemoteDescription` ? It would be nice to convert some of those warnings into errors we can return back to the user. Then just pass the slice into `openSRTP`

IMO this LGTM merge wise! I would rather unblock you, and we can move forward. I should have time to test more tomorrow and since this isn't the default path I am not worried about users hitting weird cases at all!

maxhawkins:
This fixes my problem @mjmac!

Issue #610, opened 2019-04-10
Implement PeerConnection GetStats API


maxhawkins:
## Summary

Add support for the PeerConnection GetStats method, conforming to the [dom-rtcstats](https://www.w3.org/TR/webrtc/#dom-rtcstats) spec.

## Motivation

Exposing this API makes it easier to debug connection problems by exposing statistics on lower-level components of the API. Using this standard format will allow our implementation to be interoperable with stats analysis tooling created for the browser's getStats method.

## Mandatory Stats

An implementation MUST support generating statistics of the following types when the corresponding objects exist on a PeerConnection, with the attributes that are listed when they are valid for that object:

- [ ] `RTCRTPStreamStats`
- [ ] `RTCReceivedRTPStreamStats`
- [ ] `RTCInboundRTPStreamStats`
- [ ] `RTCRemoteInboundRTPStreamStats`
- [ ] `RTCSentRTPStreamStats`
- [ ] `RTCOutboundRTPStreamStats`
- [ ] `RTCRemoteOutboundRTPStreamStats`
- [x] `RTCPeerConnectionStats`
- [x] `RTCDataChannelStats`
- [ ] `RTCMediaStreamStats`
- [ ] `RTCMediaStreamTrackStats`
- [ ] `RTCMediaHandlerStats`
- [ ] `RTCAudioHandlerStats`
- [ ] `RTCVideoHandlerStats`
- [ ] `RTCVideoSenderStats`
- [ ] `RTCVideoReceiverStats`
- [x] `RTCCodecStats`
- [ ] `RTCTransportStats`
- [x] `RTCIceCandidatePairStats`
- [x] `RTCIceCandidateStats`
- [x] `RTCCertificateStats`

## Missing attributes on implemented stats

The fields are there, but they are not actually properly loaded

### RTCICECandidatePairStats

- [ ] PacketsSent
- [ ] PacketsReceived
- [ ] BytesSent
- [ ] BytesReceived
- [ ] LastPacketSentTimestamp
- [ ] LastPacketReceivedTimestamp
- [ ] FirstRequestTimestamp 
- [ ] LastRequestTimestamp 
- [ ] LastResponseTimestamp 
- [ ] TotalRoundTripTime
- [ ] CurrentRoundTripTime 
- [ ] AvailableOutgoingBitrate 
- [ ] AvailableIncomingBitrate 
- [ ] CircuitBreakerTriggerCount
- [ ] RequestsReceived
- [ ] RequestsSent
[...]

PR #611, opened 2019-04-10
WIP: Add Stats structs

maxhawkins:
Add structs from [rtc-domstats](https://www.w3.org/TR/webrtc/#dom-rtcstats) spec so we can begin implementing the GetStats API.

Related to #610

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -6,19 +6,7 @@ import (
 
 // A Stats object contains a set of statistics copies out of a monitored component
 // of the WebRTC stack at a specific time.
-type Stats interface {
-	// Timestamp returns the timestamp associated with this object.
-	//
-	// For statistics that came from a remote source (e.g., from received RTCP packets),
-	// timestamp represents the time at which the information arrived at the local endpoint.
-	Timestamp() StatsTimestamp
-	// Type returns the object's StatsType
-	Type() StatsType
-	// ID returns a unique id that is associated with the component inspected to produce
-	// this Stats object. Two Stats objects will have the same ID if they were produced
-	// by inspecting the same underlying object.
-	ID() string
-}
+type Stats interface{}
 
 // StatsType indicates the type of the object that a Stats object represents.
 type StatsType string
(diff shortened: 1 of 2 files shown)


maxhawkins:
@Sean-Der Do you think it would be useful or confusing to merge these struct definitions?

I don't have time to implement the stats just yet, but this will make it easier for someone to do it in the future. Alternately I could hang onto this PR until I have time to add a stat.

Sean-Der:
@maxhawkins 

IMO I would merge them! someone might get interested and finish the work.

Would you mind adding some dumb tests that just assert things exist? just so coverage doesn't drop

maxhawkins:
Will do

maxhawkins:
PTAL

stats.go
```diff
@@ -0,0 +1,1306 @@
...
+func (s CodecStats) ID() string {
+	return s.StatsBase.ID
+}
+
+// ReceivedRTPStreamStats contains stats measured at the receiving end of an RTP stream,
+// known either because they're measured locally or transmitted via an RTCP Receiver
+// Report (RR) or Extended Report (XR).
+type ReceivedRTPStreamStats struct {
+	// PacketsReceived is the total number of RTP packets received for this SSRC.
+	PacketsReceived uint32 `json:"packetsReceived"`
```
Sean-Der: NIT -- What do you think of putting a newline after every member? Visually I break chunks apart by newline, so easier to consume when reading

stats.go
```diff
@@ -0,0 +1,1306 @@
...
+	millis := float64(s)
+	nanos := int64(millis * float64(time.Millisecond))
+
+	return time.Unix(0, nanos).UTC()
+}
+
+// StatsReport collects Stats objects indexed by their ID.
+type StatsReport map[string]Stats
+
+// StatsBase contains all the fields Stats objects share
```
Sean-Der: Does this need to be public?

stats.go
```diff
@@ -0,0 +1,1306 @@
+package webrtc
+
+import (
+	"time"
+)
+
+// A Stats object contains a set of statistics copies out of a monitored component
+// of the WebRTC stack at a specific time.
+type Stats interface {
```
Sean-Der: Does this need to be public?

Sean-Der: Not a big deal, but we are putting some implementation details in the users face by having these public.

I don't feel strongly, but would be nice if `godoc` doesn't show anything we did to make our code shorter

maxhawkins: I do want to keep the type around. Maybe i'll get rid of the methods:

```go
type Stats interface{}
```

I'm not sure Type() and Timestamp() are so useful anyway. Everyone will end up casting to the base type and they can find those values there.

The initial idea for this had to do with keeping the results of the not-yet-implemented `GetStats` method typesafe. It returns a map with polymorphic values `map[string]Stats`.

stats.go
maxhawkins: On second thought, let's just delete it.

If I get rid of the `Stats` methods like I mention below all of these fields can just be duplicated in the base stat objects.

stats.go
maxhawkins: will do

jinleileiking:
Is this done？exciting

PR #624, opened 2019-04-17
Bump pion/ice

Sean-Der:
Fix candidates being able to create prflx candidates
without a full request/response cycle


Sean-Der:
@maxhawkins would you mind testing this out with your libwebrtc implementation? Everything is tested/works on pion/ice itself, but ICE changes always make me nervous.

I am going to confirm that Firefox, Chrome and Janus still work.

maxhawkins:
Yep, I'll give it a try this afternoon.

maxhawkins:
Working for me

PR #625, opened 2019-04-17
Fix CreateAnswer

ChristopherFry:
Allow recvonly direction with sendrecv offer

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -621,6 +621,8 @@ func (pc *PeerConnection) addAnswerMediaTransceivers(d *sdp.SessionDescription) 
 			switch {
 			case t.kind != kind:
 				continue
+			case direction == RTPTransceiverDirectionSendrecv && t.Direction == RTPTransceiverDirectionSendonly:
+				continue
 			case direction != RTPTransceiverDirectionSendrecv && direction == t.Direction:
 				continue
 			case direction == RTPTransceiverDirectionInactive:
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -621,8 +621,6 @@ func (pc *PeerConnection) addAnswerMediaTransceivers(d *sdp.SessionDescription) 
 			switch {
 			case t.kind != kind:
 				continue
-			case direction == RTPTransceiverDirectionSendrecv && t.Direction != RTPTransceiverDirectionSendrecv:
-				continue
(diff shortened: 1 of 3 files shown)


peerconnection.go
```diff
@@ -621,8 +621,6 @@ func (pc *PeerConnection) addAnswerMediaTransceivers(d *sdp.SessionDescription)
 			switch {
 			case t.kind != kind:
 				continue
-			case direction == RTPTransceiverDirectionSendrecv && t.Direction != RTPTransceiverDirectionSendrecv:
```
Sean-Der: Will this currently allow a  `RTPTransceiverDirectionSend` to satisfy a `RTPTransceiverDirectionSendrecv` ?

ChristopherFry: In the future yes which I believe is alright, but currently no since `AddTransceiver` only supports `sendrecv` and 'recvonly` (https://github.com/pion/webrtc/blob/master/peerconnection.go#L1414).

Sean-Der: Will this break the other side? I need to confirm how Chromium acts.

If the other side is offering to send us video (via a SendRecv) candidate and we return with a (SendOnly) will it throw an error because we aren't able to satisfy? Or will it it gracefully just not send.

Also if we make this change we should probably prioritize giving them a `SendRecv`, `Recv` and then if we can't do anything a `Send` candidate. 

Sorry don't want to throw more work on you, our SDP parsing has to do a lot more with UnifiedPlan + multi-track send+recv so sometimes I burn a few cycles on it :)

ChristopherFry: That's a good question. For the moment, I can update this pull request so only `sendrecv` and `recvonly` is accepted here. Once we confirm `sendonly` works, we can create a new pull request that adds this functionality.

ChristopherFry: Updated the pull request to only accept `sendrecv` and `recvonly` here. Let me know if this looks better.

Sean-Der:
Hey @ChristopherFry 

Thank you so much for working on this, I merged this (with my changes as well in https://github.com/pion/webrtc/pull/630)

I hope w/e you are building is going well with Pion. If you need anything feel free to file here, and also we are available via [Slack](https://pion.ly/slack) thanks for opening the PR, and look forward to future PRs :)

PR #646, opened 2019-04-27
Organize ice related files into internal/ice package

### LINKED ISSUE #645: Organize ice related files into an ice package
## Summary

Currently the webrtc package is very large and hard to understand/develop. Splitting it up into smaller packages would make development easier. As a first step I suggest moving ice related code into internal/ice package, and using go aliases for providing public access for them for the users of the webrtc lib.

masterada:
#### Description

Moved ice related files to internal/ice package.

#### Reference issue
Fixes #645


gen/genaliasdocs.go
```diff
@@ -0,0 +1,152 @@
+package main
```
Sean-Der: Can we put this in `internal/gen` ?

ice.go
```diff
@@ -0,0 +1,274 @@
...
+import (
+	"github.com/pion/sdp/v2"
+	"github.com/pion/webrtc/v2/internal/ice"
+)
+
+//go:generate go run gen/genaliasdocs.go -- $GOFILE
+
+const (
+
+	// ICETransportPolicyRelay is an alias for ice.ICETransportPolicyRelay
```
Sean-Der: If the thing we are aliasing has a doc, can we use that instead?

If I am reading the godoc these comments aren't helpful currently.

masterada: Yea, that's the plan, i just didn't get there yet (TODO: "finish the script for generating docs for the aliases").

On a related note: how should we handle wasm binding aliases?
1. create ice.go for the common aliases, ice_go.go for go only aliases and ice_js.go for wasm aliases
  - if we choose this, where should we put aliases like ICEServer, that are present in both but defined separately for go/js (also, they might have different docs, so in this case which one should we use?)
2. create ice_go.go and ice_js.go only for go/wasm alises

Should I also create a tests for comparing js and wasm aliases?
- for types, if the type is present in both go and wasm, check if it has the same exported methods with the same signature
- for constants, if the constant is present in both go and wasm, check if they have the same value
- for functions, if the function is present in both go and wasm, check if the functions have the same signature

masterada:
@Sean-Der it's mostly in the form i think would be the best (maybe apart from some more tweaks to the docs generation), please take a look.

Sean-Der:
@masterada 

This is looking great to me!

* Should/Can we run something in CI to make sure the auto-generated files aren't edited by hand?
* Can you open a separate PR for all the small bugs you have fixed? If something happens we need to be able to revert/bisect those still.

Sean-Der:
Hey @masterada 

I still would love to land this :) sorry there is so much friction. I just need to make sure everything falls into the right place :)

The only thing I would like to see in this PR is if it would be possible to bring down the amount of duplication. We have `ice_go` and `ice_js` that mostly duplicate each other (with things like `	ICETransportPolicyRelay = ice.TransportPolicyRelay` in their twice)

Beyond that it LGTM! I think this would be a very welcome change, @backkem agree/disagree anything I am missing?

masterada:
Yea, sorry, it was a busy week. Will probably have time for this this weeken or on Monday. One question though: if the method is declared both in wasm and go code, but have different comments, what should we do? Display the go and show a warning maybe?

Sean-Der:
No worries at all @masterada busy on my end also, sorry I don't respond on PRs quicker :(

That sounds like a good plan to me (re: comment) my only goal is to cut down on the duplication. I was working with the code and confused when grepping/jumping to enums that were the same.

masterada:
And it's finally done. If anyone checks this, I recommend checking the two commits separatly:.
- file moving: https://github.com/pion/webrtc/pull/646/commits/a634801ff9e5d8d67cd21f7d239b7ff820ecb6f0
- dropping of ICE prefixs: https://github.com/pion/webrtc/pull/646/commits/90d1b7708937143181f468e601e1f45bebd65fc7

Sean-Der:
Nice this looks really good to me @masterada!

I will review fully tomorrow night, but I am pretty sure this is a merge from me :) just lots of LoC I want to make sure to read everything

Sean-Der:
Started merging this and ran out of time :( this should be done tomorrow @masterada 

pion/turn was actively burning people so needed to get that done. sorry about that!

Sean-Der:
Also! Can/should we follow this up and do the same for `dtls` `sctp` and `rtp` ?

masterada:
> 
> 
> Started merging this and ran out of time :( this should be done tomorrow @masterada
> 
> pion/turn was actively burning people so needed to get that done. sorry about that!

don't worry, i will do it :)

> 
> 
> Also! Can/should we follow this up and do the same for `dtls` `sctp` and `rtp` ?

yea, sure, let's merge it first then i will do the rest (i think those will be much smaller tasks)

Sean-Der:
Fantastic work, thanks for doing this! Sorry this went slower then it should have (with getting merged/approved) I will try and make this better :(

Sean-Der:
An unfortunate side effect of this is that the godoc isn't as useful now

On the main page we just have aliases (and no listing of functions). We should maybe also try aliases, and see if those render better.

Sean-Der:
@masterada actually the issue is that things are in `internal` 

What do you think of moving things to `pkg` that way users can still get access to the documentation via godoc

Sean-Der:
@masterada should we add a check to CI to ensure that people don't modify these generated files manually?

masterada:
> What do you think of moving things to pkg that way users can still get access to the documentation via godoc

@Sean-Der Sure, we can put it in pkg, only put it in internal because "why not" :) 

> should we add a check to CI to ensure that people don't modify these generated files manually?

Actually currently it only generates the comments. We could generate the aliases too, although the multiple build envs complicate things a bit (go/wasm). But, yea, sure we can add a check to travis, it's relatively simple.

> Also! Can/should we follow this up and do the same for dtls sctp and rtp ?

 I slept on it. I think we should wait at least a couple weeks to see if any other issues arise with this solution (like the godocs issue), and how ppl like it in general.

PR #653, opened 2019-05-05
[WIP] Fix datachannel not returning EOF

### LINKED ISSUE #652: EOF returns earlier than the last message on receipt of Reset
### Your environment.
  - Version: sctp e0c13 (webrtc@v2.0.10, sctp@v1.6.1)
  - Browser: n/a
 
### What did you do?
libp2p test (SubTestPingPong) is failing.
The test uses ioutil.ReadAll() in datachannel's detached mode which reads data until EOF is received. 
See slack thread for more details: https://gophers.slack.com/archives/CAK2124AG/p1556658527035200

### What did you expect?
When the sender side's SCTP stream is closed, Reset chunk will be received with indicating the last TSN in the chunk. Receiver side should forward all the data received that are equal to or smaller than the reset's TSN. 

### What happened?
stream.ReadSCTP() returns EOF before the last message (which has TSN <= the last TSN reset chunk indicates) becomes available to read.

enobufs:
This updates sctp and datachannel that fix EOF related issues.

> sctp and datachannel needs to be merged and tagged before we merge this. 

Fixes #652


datachannel_go_test.go
```diff
@@ -345,3 +352,218 @@ func TestDataChannelBufferedAmount(t *testing.T) {
 		assert.True(t, nCbs > 0, "callback should be made at least once")
 	})
 }
+
+func TestEOF(t *testing.T) {
+	log := logging.NewDefaultLoggerFactory().NewLogger("test")
+	label := "test-channel"
+	testData := []byte("this is some test data")
+
+	checkError := func(t *testing.T, err error) {
```
albrow: Are we trying to eventually remove the `stretchr/testify/assert` dependency? Because if not, you can use `assert.NoError(t, err)` instead of writing your own `checkError` function. It does the same thing.

datachannel_go_test.go
```diff
@@ -345,3 +352,218 @@ func TestDataChannelBufferedAmount(t *testing.T) {
...
+			} else {
+				log.Debug("Received pong GOOD!")
+			}
+		}()
+
+		wg.Wait()
+	})
+
+	t.Run("No detach", func(t *testing.T) {
+		// streams := 1
```
albrow: We should remove some of these outdated comments. `// streams := 1` and the comments about detached mode.

datachannel_go_test.go
enobufs: `assert.NoError(t, err)` is good.  will switch to that.

datachannel_go_test.go
enobufs: True. will remove it.

backkem:
As mentioned in pion/datachannel#18 this now fails TestEOF/Detach. What are your thoughts @enobufs?

datachannel_go_test.go
enobufs: Done

PR #659, opened 2019-05-06
Remove direct usage of unexported fields in ICETransport

masterada:
There are some direct usages of unexported fields of ICETransport
from non ice-related methods. This would be problematic when ice
once ice related code is moved to a separate packet. Added proxy
methods to ICETransport to avoid this.

#### Reference issue
Rel #646

### DIFF
--- a/icetransport.go
+++ b/icetransport.go
@@ -25,6 +25,8 @@ type ICETransport struct {
 	onConnectionStateChangeHdlr       func(ICETransportState)
 	onSelectedCandidatePairChangeHdlr func(*ICECandidatePair)
 
+	state ICETransportState
+
 	gatherer *ICEGatherer
 	conn     *ice.Conn
 	mux      *mux.Mux
@@ -62,6 +64,7 @@ func (api *API) NewICETransport(gatherer *ICEGatherer) *ICETransport {
 		gatherer: gatherer,
 		api:      api,
 		log:      api.settingEngine.LoggerFactory.NewLogger("ortc"),
+		state:    ICETransportStateNew,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Sean-Der:
Hey @masterada 

Want to tag+bump `pion/ice` and rebase off of that? This has that annoying PreICE fail :/

masterada:
you got to be kidding me :D

icetransport.go
```diff
@@ -142,17 +150,24 @@ func (t *ICETransport) Start(gatherer *ICEGatherer, params ICEParameters, role *
...
 	if t.mux != nil {
-		return t.mux.Close()
+		err = t.mux.Close()
 	} else if t.gatherer != nil {
-		return t.gatherer.Close()
+		err = t.gatherer.Close()
 	}
 
-	return nil
+	if t.state != ICETransportStateClosed {
```
Sean-Der: I would just return the message via `fmt.Errorf` 

This seems like it would be easy to miss as a message, and this isn't something that should ever happen in normal usage.

masterada: played around with it a bit, and found some problems with this solution:
1. the agent's state update callback, which updates the transport state is asynchron, and uses the same lock as the transport close, meaning when i call gatherer.close, which calls agent.close, which should change the transport state to closed, can't happen as long as the transport lock is hold in the close
2. i tried releasing the lock in ICETransport.Close, and waiting for the agent to fire the statechange event with closed value, but it seems like the agent does not fire that event (it seems like the agent never fires the OnConnectionStateChange with closed value, but im not 100% sure of that yet)

So for now i just set the state to closed when the ICETransport is closed. I will debug the ice package code to see why the agent doesn't fire the closed event, but i think as long as the gatherer.Close works properly, this solution should be ok. 

 peerconnection_close_test runs, (which also means agent is properly shut down, no leaking gorutines)

Sean-Der: @masterada correct me if I am wrong, but this feels a bit tech debty to me. I think this would possibly be the right solution (and I can help!)

* Move `ICETransportState` to a atomic. Then we don't have to worry about mutex juggling
* Fix `pion/ice` to properly emit state

masterada: atomic - good idea
fix pion/ice - already created an issue for that and assigned myself, although im not sure when i will have time for it, not sure how hard it will be to write tests for it

there is still a question: do we need to make sure that once Close() call returns, State() will return closed (because if we wait for the agent's event to set the closed state for the transport, this will not be the case)

PR #662, opened 2019-05-07
Add stargazer

### LINKED ISSUE #654: Add stargazer :)
https://starchart.cc/pion/webrtc

Check this :)

jinleileiking:
Resolves #654

#### Reference issue
Fixes #...

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
 </p>
 <br>
 
+
 See [DESIGN.md](DESIGN.md) for an overview of features and future goals.
 
 ### Breaking Changes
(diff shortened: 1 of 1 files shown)


backkem:
Just my 2 cents:
- When I'm looking for a library, I'm usually looking for info on how to get started ASAP. This pushes this info down.
- Featuring the stargazer so prominently feels overly vain to me and may give the wrong idea. As fun/important as popularity is, I think we should aim to inspire quality over quantity at this stage.

Maybe these concerns would be less pronounced if the graph was featured less prominently.

Sean-Der:
I agree with @backkem 

I am all for adding Stargarzers + issues opened/closed but it should be under a dedicated `Project Status` probably in the `Community Section`

This data might be important for someone when evaluating the project, but that other info is way more important.

thanks

jinleileiking:
I will move to community section

jinleileiking:
@Sean-Der Please check again

Sean-Der:
Pushed with 6a0b7020b1724dcb302ddfadab0c80fabc144c97 thanks @jinleileiking :)

PR #673, opened 2019-05-13
Immediately fire onOpenHandler if the DataChannel is already open

albrow:
#### Description

This can help fix some race conditions where the data channel is opened before `onOpenHandler` is set up.


albrow:
Just fixing the commit message. Hang on..

PR #674, opened 2019-05-16
Fix DC open handler been called twice

hugoArregui:
I'm not 100% sure we need the onceMutex, I believe so just because we could have a race condition between the once invokation and a new session (in which the once is instantiated again). Furthermore we cannot use the global mutex since we need to unlock it during the openHandler execution

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -44,6 +44,8 @@ type DataChannel struct {
 	// OnError             func()
 
 	onMessageHandler    func(DataChannelMessage)
+	onceMutex           sync.Mutex
+	openHandlerOnce     sync.Once
 	onOpenHandler       func()
 	onCloseHandler      func()
 	onBufferedAmountLow func()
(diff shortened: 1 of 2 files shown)


datachannel.go
```diff
@@ -44,6 +44,8 @@ type DataChannel struct {
 	// OnError             func()
 
 	onMessageHandler    func(DataChannelMessage)
+	onceMutex           sync.Mutex
```
albrow: You don't need this mutex. `sync.Once` is already goroutine safe.

datachannel_test.go
```diff
@@ -56,6 +56,41 @@ func closeReliabilityParamTest(t *testing.T, pc1, pc2 *PeerConnection, done chan
...
+			})
+		})
+
+		_, err = offerPC.CreateDataChannel(expectedLabel, nil)
+		assert.NoError(t, err)
+
+		assert.NoError(t, signalPair(offerPC, answerPC))
+
+		// wait for open handlers to be executed
+		time.Sleep(1 * time.Second)
```
albrow: Tests which rely sleeping for a certain amount of time like this can be finicky. For example, on slower machines, it might take longer than 1 second for the open handlers to be executed. Is there anyway we could avoid this?

datachannel.go
hugoArregui: yes, but check the PR description, is not about the execution but the fact that it can be replaced

datachannel_test.go
hugoArregui: yes, I don't like it either, but couldn't find a way to avoid it, I'm open to suggestions

Sean-Der: This is a tough one, I have no idea how to properly test. You are almost trying to assert `This thing shouldn't happen, but I don't know when it will happen`

Oh, what do you think of making a `maybeFireOnOpenHandler` (or a better name) and call it 3 times (asserting that your callback only gets fired once) 

Someone can always circumvent it and call the handler manually, but maybe just adding a note is good enough.

PR #678, opened 2019-05-19
Update framerate using frame count and time duration

raphaelpereira:
Instead of keeping fixed 30fps, get current timestamp
on frame count zero and update framerate using full
video duration upon closing file


raphaelpereira:
I had to do this to be able to get a consistent video from a remote WebRTC browser camera. Otherwise the video track file would be out of sync with audio track file. I know this is not optimal as the ideal would be to get this information (together with width and height) from WebRTC track, which I couldn't find in the code

PR #685, opened 2019-05-22
Fix DC open

hugoArregui:
it's expected for a open Datachannel to have an associated datachannel.Datachannel. Handling everything under the same lock guarantee that

I'm currently random getting this error (which this PR fixes), when trying to detach a channel in the open handler `datachannel not opened yet, try calling Detach from OnOpen`


Sean-Der:
This passed in travis-ci :( not sure what broke with updating here

PR #689, opened 2019-05-25
Default ICECredentialType to password

Sean-Der:
https://www.w3.org/TR/webrtc/#dom-rtciceserver has the following default
```
credentialType = "password";
```

Update our enum to match that, currently we require users to specify
password


Sean-Der:
@enobufs needed for TURN

PR #691, opened 2019-05-26
Add ice trickle support

trivigy:
resolves pion/ice#51

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -43,7 +43,6 @@ type PeerConnection struct {
 	currentRemoteDescription *SessionDescription
 	pendingRemoteDescription *SessionDescription
 	signalingState           SignalingState
-	iceGatheringState        ICEGatheringState
 	iceConnectionState       ICEConnectionState
 	connectionState          PeerConnectionState
 
@@ -69,8 +68,6 @@ type PeerConnection struct {
 	onICEConnectionStateChangeHandler func(ICEConnectionState)
 	onTrackHandler                    func(*Track, *RTPReceiver)
 	onDataChannelHandler              func(*DataChannel)
-	onICECandidateHandler             func(*ICECandidate)
-	onICEGatheringStateChangeHandler  func()
 
 	iceGatherer   *ICEGatherer
 	iceTransport  *ICETransport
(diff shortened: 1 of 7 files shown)


Sean-Der:
@trivigy Ok ready for a first pass!

It toggles between `trickle`/`non-trickle` if you have the callback set. When we do `/v3` eventually we can break that behavior (and only do trickle)

I moved most of the behavior into `ICEGatherer` and had things follow ORTC stuff.

peerconnection.go
```diff
@@ -641,6 +589,12 @@ func (pc *PeerConnection) addAnswerMediaTransceivers(d *sdp.SessionDescription)
 		return nil, err
 	}
 
+	if !pc.iceGatherer.agentIsTrickle {
+		if err = pc.iceGatherer.Gather(); err != nil {
```
trivigy: Are you should it should start gathering here? It used to start gathering during initialization before.

peerconnection.go
```diff
@@ -882,23 +834,19 @@ func (pc *PeerConnection) SetLocalDescription(desc SessionDescription) error {
...
-	// is complete. In reality it completed a while ago, but triggering these
-	// events helps maintain API compatibility with the JavaScript/Wasm bindings.
-	if err := pc.signalICECandidateGatheringComplete(); err != nil {
-		return err
+	if !pc.iceGatherer.agentIsTrickle {
+		return nil
 	}
-
-	return nil
+	return pc.iceGatherer.Gather()
```
trivigy: Again same deal. This is how it should be but this is a change from original setup. Will this make a performance impact for people?

peerconnection.go
Sean-Der: We can't do it during init because people need time to set the `OnICECandidate` callback.

Yea changing what function blocks isn't great, but I don't think users will notice in the normal case.

peerconnection.go
Sean-Der: This will only fire for users that are doing `Trickle` and will not block. So shouldn't have an impact!

peerconnection.go
trivigy: ahhhh... makes total sense. In my local version I really struggled with that.

Sean-Der:
> One other thing if you don't mind addressing. There is inconsistency regarding what `SessionDescription`. Previously session description was extracted from `CreateOffer()` where it was returned as a struct instance. Now with trickle changes we need to extract it from `LocalDescription()` where it is returned as a pointer. This makes things inconsistent.

we need a way to express `null`, we can update this to return an error (but having no currentLocalDescription I don't believe is an error)

PR #693, opened 2019-05-26
Prepare for internal/ice package

masterada:
Removed SDP dependency from ice files. Exported string to ice enum
functions. Exported ICEServer.Validate.

#### Reference issue
Rel #645

### DIFF
--- a/ice_go.go
+++ b/ice_go.go
@@ -2,6 +2,8 @@
 
 package webrtc
 
+import "github.com/pion/sdp/v2"
+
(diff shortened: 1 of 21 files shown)


ernado:
BTW should we eventually move the code that parses ICE candidates to pion/ice package?

IMO it's ice package responsibility :thinking:

masterada:
@ernado if you mean sdp.ICECandidate  <-> webrtc.ICECandidate <-> ice.Candidate conversions, that's a good question, not sure where these would be best at.

PR #700, opened 2019-06-03
Expose new ICE options

hugoArregui:
Expose new ICE timeouts in the setting engine

Comments: 

- I moved the ICEGathererOptions to icegatherer.go because that seems more standard for us (the config and the proper struct in the same file, but I can roll it back)
- Should I add setters for the timeouts?


icegatherer.go
```diff
@@ -11,6 +11,17 @@ import (
 	"github.com/pion/logging"
 )
 
+// ICEGatherOptions provides options relating to the gathering of ICE candidates.
+type ICEGatherOptions struct {
```
Sean-Der: This comes from the RFC. If possible I would avoid diverging!

https://draft.ortc.org/#dom-rtcicegatheroptions

hugoArregui: sure, do you want me to add an extra config object instead?

Sean-Der: Can you pass the options to `NewGatherer` [here](https://github.com/pion/webrtc/blob/master/internal/ice/gatherer.go#L41)

Sean-Der: It is ugly having some many arguments to that function, but I think it is important we don't diverge from the standard (I think)

I would love to figure out how we can push these things into the WebRTC standard.

hugoArregui: cool!, I changed it, however regarding the bigger picture I think there is an argument to be made regarding how adding additional options is not "diverging" from the standard. Specially in this case, because this new options (as the standard specifies) are up to the implementation to decide.

PR #708, opened 2019-06-07
Fix race issue caused by multiple gathering

### LINKED ISSUE #707: Gather() causing a race condition
Too many consecutive calls to `iceGatherer.Gather()` are causing a race condition because `pion/ice` operates asynchronously.

https://github.com/pion/webrtc/blob/master/peerconnection.go#L413
https://github.com/pion/webrtc/blob/master/peerconnection.go#L592
https://github.com/pion/webrtc/blob/master/peerconnection.go#L840
https://github.com/pion/webrtc/blob/master/peerconnection.go#L864

trivigy:
resolves #707

### DIFF
--- a/internal/ice/gatherer.go
+++ b/internal/ice/gatherer.go
@@ -3,7 +3,6 @@
 package ice
 
 import (
-	"errors"
 	"sync"
 	"time"
 
@@ -54,6 +53,7 @@ func NewGatherer(
 	prflxAcceptanceMinWait,
 	relayAcceptanceMinWait *time.Duration,
 	loggerFactory logging.LoggerFactory,
+	agentIsTrickle bool,
 	networkTypes []NetworkType,
 	opts GatherOptions,
 ) (*Gatherer, error) {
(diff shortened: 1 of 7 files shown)


internal/ice/gatherer.go
```diff
@@ -68,7 +68,7 @@ func NewGatherer(
 		}
 	}
 
-	candidateTypes := []ice.CandidateType{}
+	candidateTypes := make([]ice.CandidateType, 0)
```
trivigy: This change has to do with Jetbrain's Goland go linting complaining that this is better.

peerconnection.go
```diff
@@ -429,8 +428,8 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
 	}
 
 	if pc.configuration.SDPSemantics == SDPSemanticsPlanB {
-		video := []*RTPTransceiver{}
-		audio := []*RTPTransceiver{}
+		video := make([]*RTPTransceiver, 0)
```
trivigy: Same issue as above regarding jetbrain goland linting.

peerconnection.go
```diff
@@ -475,14 +474,14 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
 		m.WithPropertyAttribute("setup:actpass")
 	}
 
-	sdp, err := d.Marshal()
```
trivigy: `sdp` conflicted with the pkg `sdp`

PR #715, opened 2019-06-15
Fix panic in openSRTP

### LINKED ISSUE #563: nil pointer in openSRTP
### Your environment.
  - Version: 8ef5b7034ff4664530938fa48873d34d3f9672fc

### What did you do?
Created a PeerConnection between two Pion instances and sent data using `track.WriteSample`.

### What happened?

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x42ed3dd]

goroutine 32 [running]:
github.com/pions/webrtc.(*PeerConnection).openSRTP.func1(0xc000132480, 0xc0000004d2, 0x2)
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.1-0.20190328075210-8ef5b7034ff4/peerconnection.go:991 +0x2bd
created by github.com/pions/webrtc.(*PeerConnection).openSRTP
        /Users/max/go/pkg/mod/github.com/pions/webrtc@v1.2.1-0.20190328075210-8ef5b7034ff4/peerconnection.go:962 +0x3fe
```

This is the line
```
			sdpCodec, err := pc.currentLocalDescription.parsed.GetCodecForPayloadType(receiver.Track().PayloadType())
			if err != nil {
				pc.log.Warnf("no codec could be found in RemoteDescription for payloadType %d", receiver.Track().PayloadType())
				return
			}
```

mchlrhw:
#### Description

The fact that SetLocalDescription isn't called is erroneous, but should not result in a panic. It may be that we shouldn't even be processing incoming tracks until the local description has been set.

**Edit**: The approach taken to fix the panic is to just log and bail as is done in the rest of the function, rather than moving around the code to handle incoming tracks. I would appreciate suggestions for alternatives.

#### Reference issue
Fixes #563

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1102,6 +1102,11 @@ func (pc *PeerConnection) openSRTP() {
 		pc.mu.RLock()
 		defer pc.mu.RUnlock()
 
+		if pc.currentLocalDescription == nil {
+			pc.log.Warnf("SetLocalDescription not called, unable to handle incoming media streams")
+			return
+		}
+
 		sdpCodec, err := pc.currentLocalDescription.parsed.GetCodecForPayloadType(receiver.Track().PayloadType())
 		if err != nil {
 			pc.log.Warnf("no codec could be found in RemoteDescription for payloadType %d", receiver.Track().PayloadType())
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [xsbchen](https://github.com/xsbchen)
 * [Alex Harford](https://github.com/alexjh)
 * [Aleksandr Razumov](https://github.com/ernado)
+* [mchlrhw](https://github.com/mchlrhw)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/peerconnection_setlocaldescription_test.go
+++ b/peerconnection_setlocaldescription_test.go
@@ -1,0 +1,127 @@
+// +build !js
+
+package webrtc
+
+import (
+	"bufio"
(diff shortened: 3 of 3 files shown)


mchlrhw:
Unfortunately the test for this is fail fast, pass slow. Anyone got any ideas how we can rewrite to make it fast in both directions? Or is this just something to chalk up to "asnyc testing is _hard_"? :smile:

Sean-Der:
Yea that is a tough one, I don't think we can add that much time to the testing.

I think we can set a custom logger and when we get that message it is a pass! We just need to assert your log message is seen, I think that is 100% good for now. Just so we can make sure this doesn't regress in the future

go.sum
```diff
@@ -43,6 +43,7 @@ github.com/pion/sdp/v2 v2.1.1 h1:i3fAyjiLuQseYNo0BtCOPfzp91Ppb7vasRGmUUTog28=
 github.com/pion/sdp/v2 v2.1.1/go.mod h1:idSlWxhfWQDtTy9J05cgxpHBu/POwXN2VDRGYxT/EjU=
```
Sean-Der: Mind removing `go.sum` from PR, thanks!

peerconnection.go
```diff
@@ -1099,6 +1099,11 @@ func (pc *PeerConnection) openSRTP() {
 			return
 		}
 
+		if pc.currentLocalDescription == nil {
```
Sean-Der: Should this be after we take the lock?

mchlrhw: I wasn't sure abiut that, I thought it could probably be either. My reasoning behind putting it there is that we can do the check as soon as we return from the previous function call, so no need to wait for the lock.

Issue #619, opened 2019-04-12
Add observability metrics


maxhawkins:
## Summary

Add hooks in the PeerConnection that allow metrics tools like Prometheus or statsd to get information about the inner-workings of Pion.

Some stats that should be implemented:
* Number of open PeerConnections
* Number of failed PeerConnections
* Ports allocated
* ICE connection state count
* DTLS/ICE error count
* RTP statistics (jitter, loss, packets sent, etc)
* (others ?) comment below.

Export to [expvar](https://golang.org/pkg/expvar/) by default, but support common metrics packages like Prometheus.

## Motivation

Exposing metrics can help find bugs and configuration issues in production, optimize complicated Pion installations, and catch issues early.

Some metrics (like those related to low-level DTLS metrics) are difficult to reach from the application layer because they sit at the far end of the stack. Exposing metrics would give users better observability into the low level parts of the library.

## Describe alternatives you've considered

Perhaps this can be moved further up the stack and help us avoid straying from the RTCWEB spec. Could the same functionality be built with the GetStats API (#610) and Prometheus?

jinleileiking:
Prometheus is the basic

maxhawkins:
Closing in favor of #610

PR #721, opened 2019-06-20
Initial GetStats: get some statistics about ice candidates, ice candidate pairs and data channels

hugoArregui:
Related to #619 

# References

- Spec: https://www.w3.org/TR/webrtc-stats/
- We are generating internal IDS similar as how libwertc does:
    - [DataChannel ID](https://webrtc.googlesource.com/src/+/refs/heads/master/pc/rtc_stats_collector.cc#1228)
    - [PeerConnection ID](https://webrtc.googlesource.com/src/+/refs/heads/master/pc/rtc_stats_collector.cc#1445)

# Related PRs

- ICE: https://github.com/pion/ice/pull/71
- ICE: https://github.com/pion/ice/pull/73
- Datachannel: https://github.com/pion/datachannel/pull/21


maxhawkins:
What are all the identifers we need? Is it more than just track id?

hugoArregui:
yes, there are a lot, every object has an ID

```
id of type DOMString
A unique id that is associated with the object that was inspected to produce this RTCStats object. Two RTCStats objects, extracted from two different RTCStatsReport objects, MUST have the same id if they were produced by inspecting the same underlying object. User agents are free to pick any format for the id as long as it meets the requirements above.
```

So we connection ID, datachannel id, candidate ID, candidate pair ID,transport ID, codec ID, etc, etc

maxhawkins:
I'm trying to think of reasons why we would allow users to make their own IDs rather than generating them internally.

Maybe if you had a distributed system where you wanted IDs to be consistent across service boundaries for easier stats aggregation...

hugoArregui:
Yes, not a lot of reasons really, I think it could be useful to allow the user to set the connection ID and then prefix everything as connection-id+...., to make debugging easier, but that's it.

icecandidate.go
```diff
@@ -8,6 +8,7 @@ import (
 
 // ICECandidate represents a ice candidate
 type ICECandidate struct {
+	ObjectID       string           `json:"objectID"`
```
Sean-Der: Is it possible to not export `ObjectID` (looking at the RFC I don't see this key mentioned)

peerconnection.go
```diff
@@ -391,6 +396,11 @@ func (pc *PeerConnection) GetConfiguration() Configuration {
 	return pc.configuration
 }
 
+// ObjectID returns the connection object id
+func (pc *PeerConnection) ObjectID() string {
```
Sean-Der: The W3C has pushed back pretty hard on having ids assigned to anything https://www.w3.org/2018/04/12-webrtc-minutes#item06

I would avoid if possible, we are going to break WASM/Go interop if we do go down that route.

hugoArregui:
@Sean-Der the ID is mentioned in here: https://www.w3.org/TR/webrtc-stats/ section `4. Basic concepts`:

```
    dictionary RTCStats {
DOMHighResTimeStamp timestamp;
RTCStatsType type;
DOMString id;
    };
```

The ID is only used for stats purposes, we could rename it. The thing is you look at the usage the ID is pretty useful, you can

```
report := conn.GetStats()

// get connection stats
report[conn.ObjectID()]
 
// get DC stats
report[dc.ObjectID()]

// if you subscribe to the Candidate Pair change listener
pairStats := report[pair.ObjectID]

// and then get stats for its candidates
report[pairStats.LocalCandidateID]
report[pairStats.RemoteCandidateID]
```

I think that's how the webrtc-stats spec is meant to be used

@maxhawkins wdyt?

Sean-Der:
@hugoArregui @maxhawkins 

No push back from me if it is following the standard! I think we can make the ID public for the `RTCStats` but I don't think it specifies for the `source objects` themself.

I could be wrong though! Harald is super friendly, I am sure if you emailed the RFC list (or him directly) you would get a response pretty quick.

hugoArregui:
> No push back from me if it is following the standard! I think we can make the ID public for the `RTCStats` but I don't think it specifies for the `source objects` themself.

That's true, I think the spec assumes:

a) every object subject of monitoring has an ID
b) the stats for that object are referenced by the ID

It doesn't say the ID should be publicly available, so we could leave it out, but that seems much less useful.

> I could be wrong though! Harald is super friendly, I am sure if you emailed the RFC list (or him directly) you would get a response pretty quick.

I could ask what's the intended usage or if they have any suggestion for implementers.

Sean-Der:
That all 100% to me!

My only push back is on the ID stuff. If you want maybe we can land that stuff and put the ID changes in a `WIP` PR. 

Maybe your feedback could influence the RFC! Now that would be an amazing accomplishment, I have never done any RFC stuff :)

hugoArregui:
Thread: https://lists.w3.org/Archives/Public/public-webrtc/2019Jun/0054.html

Sean-Der:
Awesome job starting the conversation on the mailing list @hugoArregui 

`StatsID` sounds great to me, and we can always change later.

PR #730, opened 2019-07-05
Fix routine leak in drainSRTP

### LINKED ISSUE #729: Receiving unknown SSRC causes leaked SSRC on shutdown
`drainSRTP` doesn't destroy its streams on shutdown.

Sean-Der:
Store orphaned SRTP/SRTCP streams and discard them
when we close the PeerConnection

Resolves #729

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -86,6 +86,7 @@ func (t *DTLSTransport) ICETransport() *ICETransport {
 	return t.iceTransport
 }
 
+// onStateChange requires the caller holds the lock
 func (t *DTLSTransport) onStateChange(state DTLSTransportState) {
 	t.state = state
 	hdlr := t.onStateChangeHdlr
@@ -224,6 +225,10 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error {
 
 	if err := t.ensureICEConn(); err != nil {
 		return err
+	}
+
+	if t.state != DTLSTransportStateNew {
+		return &rtcerr.InvalidStateError{Err: fmt.Errorf("attempted to start DTLSTransport that is not in new state: %s", t.state)}
 	}
 
 	dtlsEndpoint := t.iceTransport.NewEndpoint(mux.MatchDTLS)
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -31,10 +31,9 @@ type DTLSTransport struct {
 	certificates      []Certificate
 	remoteParameters  DTLSParameters
 	remoteCertificate []byte
-	// State     DTLSTransportState
+	state             DTLSTransportState
 
-	// OnStateChange func()
-	// OnError       func()
+	onStateChangeHdlr func(DTLSTransportState)
 
 	conn *dtls.Conn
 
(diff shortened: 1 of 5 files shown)


Sean-Der:
Not ready for merge yet, need to write a better test

Sean-Der:
@maxhawkins @mchlrhw @mjmac @wdouglass added all the media devs :)

Thanks for the review! Hopefully not controversial, everything I added exists in WebRTC spec.

peerconnection.go
```diff
@@ -1207,6 +1212,10 @@ func (pc *PeerConnection) drainSRTP() {
 				return
 			}
 
+			pc.mu.Lock()
```
maxhawkins: Should orphanSRTP have its own mutex? Not sure what this mutex is shared with and if it would cause contention/deadlocks.

peerconnection.go
```diff
@@ -1207,6 +1212,10 @@ func (pc *PeerConnection) drainSRTP() {
...
+			pc.orphanSRTP = append(pc.orphanSRTP, r)
```
maxhawkins: Why not close these immediately?

For a long-running PC wouldn't this grow without bound?

dtlstransport.go
```diff
@@ -214,21 +240,26 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error {
 		ClientAuth:             dtls.RequireAnyClientCert,
 		LoggerFactory:          t.api.settingEngine.LoggerFactory,
 	}
+
+	t.onStateChange(DTLSTransportStateConnecting)
```
maxhawkins: Is it possible to transition from failed to connecting? I know in a PeerConnection that state transition is forbidden but I'm not sure about ORTC.

dtlstransport.go
```diff
@@ -83,6 +86,29 @@ func (t *DTLSTransport) ICETransport() *ICETransport {
 	return t.iceTransport
 }
 
+func (t *DTLSTransport) onStateChange(state DTLSTransportState) {
+	t.state = state
```
maxhawkins: is this missing a mutex?

dtlstransport.go
```diff
@@ -31,10 +31,9 @@ type DTLSTransport struct {
 	certificates      []Certificate
 	remoteParameters  DTLSParameters
 	remoteCertificate []byte
-	// State     DTLSTransportState
+	state             DTLSTransportState
```
maxhawkins: Gonna be so great to have this in the api. Much faster to know if the connection has closed or not.

dtlstransport.go
Sean-Der: That has to be without a mutex (all the callers are holding one) will add a comment!

dtlstransport.go
Sean-Der: Do you think the right thing to do here is `if state == failed` that bails out?

peerconnection.go
Sean-Der: Oh yes good point, that seems like an excellent for a DoS. 

This just exists for being helpful for debugging but serves no purpose beyond that. Will re-push

dtlstransport.go
Sean-Der: fixed!

peerconnection.go
Sean-Der: Fixed!

peerconnection.go
Sean-Der: Fixed!

dtlstransport.go
maxhawkins: I don't know. Does the spec say? Couldnt find any more information.

Sean-Der: That sounds right to me https://github.com/pion/webrtc/pull/730/files#diff-af8b99e79ad0bc29625704279b2d48e5R230

Will read spec and see if they say anything

Sean-Der: `Start` doesn't exist for WebRTC so nothing there.

I would say this is the right behavior, I bet internally they do have the check.

Sean-Der: Matched ORTC here `If start is called after a previous start call, or if state is closed, throw an InvalidStateError.`

PR #739, opened 2019-07-15
Implement permissive STUN URL parsing

### LINKED ISSUE #735: Make STUN URL parsing more permissive
## Summary

The STUN URL parser in the `ice` package is strict about not allowing URLs in the `stun:` scheme with a query, e.g `stun:global.stun.twilio.com:3478?transport=udp`, and returns `ErrSTUNQuery` if it finds one.

The current implementation of `configuration.go` rejects any URL that errors during parsing:
```go
url, err := ice.ParseURL(rawURL)
if err != nil {
    return nil, err
}
```

Both behaviours are technically correct, but they prevent apps from using ICE configuration generated by a third party without modification (unless the response is completely to spec), e.g. unmarshaling the response to a request to the Twilio Network Traversal API straight into the `Configuration`.

We could solve this by either implementing a more permissive parser that detects queries and strips them before returning the URL, or we could make `getICEServers` more resistant to invalid ICE server config by allowing it to omit invalid URLs, i.e. simply `continue` when we hit an error.

Either solution is more user friendly than a hard fail and could include appropriate WARN-level logging to notify the user that "something is wrong, but we're carrying on with what we've got anyway".

## Motivation

The suggested change would make it easier for users to work with non-compliant ICE server configurations without having to implement any parsing or sanitisation of their own.

The alternative is to leave the code as-is and leave it up to app developers to sanitise URLs before adding the `ICEServer` struct to the `Configuration`.

mchlrhw:
### Description
The parser included in the `ice` package strictly adheres to RFC 7064, however, this can be problematic when attempting to use ICE server URLs that were automatically generated by a third party.

Twilio, for example, has been seen to send the following in its list of ICE servers:

    stun:global.stun.twilio.com:3478?transport=udp

which would require user intervention to sanitise before passing to Pion.
This patch side-steps this aspect of `ice.ParseURL` by pre-stripping any queries that may be present so that we can allow URLs of this form.

#### Reference issue
Fixes #735


configuration_common.go
```diff
@@ -0,0 +1,29 @@
...
+// getICEServers side-steps the strict parsing mode of the ice package
+// (as defined in https://tools.ietf.org/html/rfc7064) by stripping any
+// erroneous queries from "stun(s):" URLs before parsing.
+func (c Configuration) getICEServers() (*[]*ice.URL, error) {
+	var iceServers []*ice.URL
+	for _, server := range c.ICEServers {
+		for _, rawURL := range server.URLs {
+			if strings.HasPrefix(rawURL, "stun") {
+				// strip the query from "stun(s):" if present
+				parts := strings.Split(rawURL, "?")
```
trivigy: Why not move these changes into `ice.ParseURL()`. Seems like the extra parsing should be inside the parsing function.

Sean-Der: So me and @mchlrhw were looking at [this rfc](https://tools.ietf.org/html/rfc7064) and these URLs are malformed. However since so many big TURN/STUN providers use them wanted to get rid of this sharp edge.

Maybe the wrong decision, but I pushed to have it in `pion/webrtc` so that we don't introduce the wrongness(?) into `pion/ice`. My thinking was `pion/ice` will be users with a higher level of technical knowledge.

trivigy: Makes sense. Is is malformed just because of the query param? Like these should only apply to stun servers right? https://tools.ietf.org/html/rfc7065#appendix-A

Sean-Der: Yea! Nice call out, I completely didn't think to check that case.

I have to say I really appreciate the review @trivigy :) it really feels great that we are starting to have a full community around this. Thanks for pushing to keep the quality up, really makes my day to see stuff like this!

Issue #106, opened 2018-09-06
Add lint to TODO/FIXME


Sean-Der:
Require that each comment with a TODO/FIXME has an associated github issue (otherwise we will lose track of them)

trivigy:
@Sean-Der can we lose some of the restrictions here? Personally I use TODO and FIXME a lot during "work in progress" to indicate things that I put in place but still require additional coding. Those are not stories within themselves. I fear is that this kind of strict linting is discouraging me from putting indications of things that don't fully work by requiring me to have to open tens of issues for things people are unlikely to actually be able to help with.

One proposal is maybe adding additional marker like "ISSUE" or maybe doing this sort of linting only on the TODO since that one is a bit more like: "Here is another thing that needs to be completed."

maxhawkins:
We should definitely do this. I just had the idea to add this and found you had already thought of it.

enobufs:
@Sean-Der 
I checked the current state of max data channel message size.
As of today, our implementation allows up to 65536 bytes, or math.MaxUint16
RFC for WebRTC `https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13` states:
```
   The SCTP base protocol specified in [RFC4960] does not support the
   interleaving of user messages.  Therefore sending a large user
   message can monopolize the SCTP association.  To overcome this
   limitation, [I-D.ietf-tsvwg-sctp-ndata] defines an extension to
   support message interleaving, which SHOULD be used.  As long as
   message interleaving is not supported, the sender SHOULD limit the
   maximum message size to 16 KB to avoid monopolization.
``` 
so, 16KB is recommended. Since it is not a `MUST`, I guess we are ok.
Next gen SDP draft specifies an attribute, `a=max-message-size`, but I don't think it is used in WebRTC.

Therefore, you can remove all these params (and the helper method) all together in sctptransport.go:
```
remoteMaxMessageSize
canSendSize
r.maxMessageSize
calcMessageSize()
```

Can you do that in this branch?

PR #745, opened 2019-07-21
Add godox to .travis.yml

Sean-Der:
Eventually this will be added to golangci-lint, but for now
we have to do it manually

Relates to #106

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -295,7 +295,7 @@ func (d *DataChannel) readLoop() {
 			d.readyState = DataChannelStateClosed
 			d.mu.Unlock()
 			if err != io.EOF {
-				// TODO: Throw OnError
+				// pion/webrtc#746
 				d.log.Errorf("Failed to read from data channel %v", err)
 			}
 			d.onClose()
--- a/datachannel.go
+++ b/datachannel.go
@@ -140,7 +140,7 @@ func (d *DataChannel) open(sctpTransport *SCTPTransport) error {
 
 	cfg := &datachannel.Config{
 		ChannelType:          channelType,
-		Priority:             datachannel.ChannelPriorityNormal, // TODO: Wiring
+		Priority:             datachannel.ChannelPriorityNormal,
 		ReliabilityParameter: reliabilityParameteer,
 		Label:                d.label,
 		LoggerFactory:        d.api.settingEngine.LoggerFactory,
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -851,8 +851,7 @@ func (pc *PeerConnection) LocalDescription() *SessionDescription {
 
 // SetRemoteDescription sets the SessionDescription of the remote peer
 func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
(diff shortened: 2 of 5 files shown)


datachannel.go
```diff
@@ -140,7 +140,7 @@ func (d *DataChannel) open(sctpTransport *SCTPTransport) error {
 
 	cfg := &datachannel.Config{
 		ChannelType:          channelType,
-		Priority:             datachannel.ChannelPriorityNormal, // TODO: Wiring
```
Sean-Der: This seems to be respected/fully implemented?

peerconnection.go
```diff
@@ -1567,27 +1573,15 @@ func (pc *PeerConnection) Close() error {
 	pc.signalingState = SignalingStateClosed
 
 	// https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close (step #11)
-	// pc.ICEConnectionState = ICEConnectionStateClosed
-	pc.iceStateChange(ice.ConnectionStateClosed) // FIXME REMOVE
```
Sean-Der: We don't need to set this ourselves `pion/ice` sets this to Closed for us

Sean-Der:
@trivigy @albrow @backkem @enobufs 

I am moving all our `TODO` comments into real bugs. I have noticed that most of our TODOs don't get worked (until someone hits a bug related to them!)

This PR is going to take a long time to finish, but would you mind looking through the commits and signing off on what I have now? I will cherry-pick to master, this is probably going to be a week long process.

rtptransceiver.go
```diff
@@ -8,7 +8,6 @@ import (
 
 // RTPTransceiver represents a combination of an RTPSender and an RTPReceiver that share a common mid.
 type RTPTransceiver struct {
-	Mid       string
```
Sean-Der: See https://github.com/pion/webrtc/pull/745/commits/5492a221456a8bdd4dce014d0512bd0c3bd0de0c for why this was removed

peerconnection.go
Sean-Der: I also added a test that confirms that `Closed` is properly set. Users should see no behavior difference

.travis.yml
```diff
@@ -21,6 +21,7 @@ install:
   - export PATH=$GOROOT/bin:$PATH
   - go version
   - go env
+  - go get -u github.com/matoous/godox
```
Sean-Der: `godox` may be added to `golangci-lint`

If not I might just extend our current `panic` finder to look for `TODO`, `FIXME` and `BUG`

I think this is the right decision. Currently people contribute code, but don't drive the tech debt down. This is also a good way to get things in the bug tracker so others can be aware/help out!

trivigy:
@Sean-Der You think you could separate the MID work from the TODO stuff? Will be easier to reverse later in case of debugging.

Sean-Der:
Done! @trivigy https://github.com/pion/webrtc/pull/745/commits/48963f5458503063a609f27ffb44225bf9cb9f67 put it in its own commit

PR #757, opened 2019-07-26
Bump pion/dtls

Sean-Der:
pion/dtls verifies certificates by default now. Disable verification
since WebRTC uses self-signed

Relates to pion/dtls#81


dtlstransport.go
```diff
@@ -247,6 +247,7 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error {
 		SRTPProtectionProfiles: []dtls.SRTPProtectionProfile{dtls.SRTP_AES128_CM_HMAC_SHA1_80},
 		ClientAuth:             dtls.RequireAnyClientCert,
 		LoggerFactory:          t.api.settingEngine.LoggerFactory,
+		InsecureSkipVerify:     true,
```
daenney: Should this really be set by default?

Sean-Der: @daenney I think so (but please correct me if I am wrong!)

We put the fingerprint of the cert in the SessionDescription, so that verifies that we did a handshake with the Client/Server we expected. 

The user could pass a cert, but there isn't a way to communicate `I am passing a cert that will pass the verify` so I don't think we can ever conditionally set it. 

Also looked at other projects and they don't verify either https://github.com/meetecho/janus-gateway/blob/master/dtls.c#L962

daenney: Mmm, alright then. I'm not familiar enough with WebRTC to be able to say something sensible here. It just struck me as odd b/c in other contexts you'd normally not want to do this.

Sean-Der:
@daenney you are fast :)

thank you so much for the quick reviews!

PR #760, opened 2019-07-29
Implemented AddSender method on PeerConnection

rviscarra:
This allows to have send only tracks

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Slugalisk](https://github.com/slugalisk)
 * [Agugua Kenechukwu](https://github.com/spaceCh1mp)
 * [Ato Araki](https://github.com/atotto)
+* [Rafael Viscarra](https://github.com/rviscarra)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1336,23 +1336,29 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 }
 
 // AddTransceiver Create a new RTCRtpTransceiver and add it to the set of transceivers.
+// Deprecated: Use AddTrack, AddTransceiverFromKind or AddTransceiverFromTrack
 func (pc *PeerConnection) AddTransceiver(trackOrKind RTPCodecType, init ...RtpTransceiverInit) (*RTPTransceiver, error) {
+	return pc.AddTransceiverFromKind(trackOrKind, init...)
+}
+
+// AddTransceiverFromKind Create a new RTCRtpTransceiver(SendRecv or RecvOnly) and add it to the set of transceivers.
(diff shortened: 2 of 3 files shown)


Sean-Der:
Awesome work @rviscarra really excited to see the PR! Getting more people involved the better :)

Unfortunately I don't think we can add a `AddSender` method since that doesn't exist in the WebRTC RFC. But I think we can add this method (just need to change the name). The root issue here is that Go doesn't support dispatching off of types `AddTransceiver` should take a `RTPCodecType` OR `RTCTrack`. 

What would you think of an API that looked like this
```
// Mark as deprecated, calls AddTransceiverFromKind
func (pc *PeerConnection) AddTransceiver(kind RTPCodecType, init ...RtpTransceiverInit) (*RTPTransceiver, error)

// This is our old AddTransceiver, we keep the old one around just so we don't break the API
func (pc *PeerConnection) AddTransceiverFromKind(kind RTPCodecType, init ...RtpTransceiverInit) (*RTPTransceiver, error)

// Your new method 
func (pc *PeerConnection) AddTransceiverFromTrack(track RTCTrack, init ...RtpTransceiverInit) (*RTPTransceiver, error) {
```

rviscarra:
I thought of doing something like that but then I just focused on covering all the possible cases:

- Support SendRecv (in AddTrack and AddTransceiver)
- Support RecvOnly (in AddTransceiver)
- Support SendOnly (in AddSender, I don't like the name but that's the best I could come up with)

I like your API better, I only see one potential issue: don't you think `AddTransceiverFromTrack` should support `SendRecv` and `SendOnly` for consistency? That means there would be three different ways to add a SendRecv transceiver which may be confusing.

Let me know what you think so I can update the PR.

Sean-Der:
@rviscarra agree! If they want to do `SendRecv` we suggest `AddTrack`. But then provide these two methods for `*Only`

We will for sure need to document this, I wish the API was simpler. My big goal here is to resemble the WebRTC API, so feel free to mold/change in the process if you see anything better!

rviscarra:
I decided to go with:

```go
func (pc *PeerConnection) AddTransceiver(kind RTPCodecType, init ...RtpTransceiverInit) {}

func (pc *PeerConnection) AddTransceiverFromKind(kind RTPCodecType, init ...RtpTransceiverInit) {}

func (pc *PeerConnection) AddTransceiverFromTrack(track *Track, init ...RtpTransceiverInit) {}
```
I thought I could be a good idea to put a warning telling the user that they should use AddTrack to create a SendRecv transceiver but I realized it could be a valid use case if they need the transceiver instead of just the sender.

Sean-Der:
Fantastic work, this LGTM @rviscarra 

You just need to move your tests to `peerconnection_media_test.go` these are media tests that we only run in Go (and not WASM via browser)

I am adding you to the Pion organization so when you get the build green you can merge!

Sean-Der:
@rviscarra Totally your call, but I would squash commits into auditable bits! Totally subjective, but always nice when reading history to have one commit per feature/bug

rviscarra:
@Sean-Der Definitely, one more force push can't hurt :smile:.

It took a bit of trial and error to make the CI happy, I have a semi-automated process to test locally now.

PR #762, opened 2019-07-30
Add ToJSON ICECandidate method

hugoArregui:
With this change we can always exchange ICECandidateInit when signaling

### DIFF
--- a/icecandidate.go
+++ b/icecandidate.go
@@ -137,3 +137,26 @@ func (c ICECandidate) String() string {
 	}
 	return ic.String()
 }
+
+// ToJSON returns an ICECandidateInit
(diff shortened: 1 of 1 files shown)


backkem:
Looks good. Mind adding a simple test?

Sean-Der:
@hugoArregui I would export https://github.com/pion/sdp/blob/master/ice.go#L41 from the SDP package and then just use that.

I just tried this out and works great for me! If you don't mind adding one minor test, and exporting from SDP package (maybe rename it `String`) in the process. It is a merge from me!

If you don't have the bandwidth I can grab also!

Sean-Der:
cherry-picked into https://github.com/pion/webrtc/pull/763

PR #763, opened 2019-07-31
Trickle Example

Sean-Der:
This cherry-picks @hugoArregui work and adds an example

I had to update the gathering code to only start when signaling is in a stable enough state. If someone emits candidates before the other side is ready it causes Chrome/Pion to fail

### DIFF
--- a/examples/pion-to-pion-trickle/README.md
+++ b/examples/pion-to-pion-trickle/README.md
@@ -1,9 +1,9 @@
-# pion-to-pion
-pion-to-pion is an example of two pion instances communicating directly!
+# pion-to-pion-trickle
+pion-to-pion-trickle is an example of two pion instances communicating directly!
+This example uses Trickle ICE, this allows communication to begin before gathering
+has completed.
 
-To see an example of `pion-to-pion` that uses Trickle ICE see `pion-to-pion-trickle`.
-This may connect faster (and will eventually become the default API) but requires
-more code.
+See `pion-to-pion` example of a non-Trickle version of this.
 
 The SDP offer and answer are exchanged automatically over HTTP.
 The `answer` side acts like a HTTP server and should therefore be ran first.
@@ -21,10 +21,3 @@ offer
 ```
 
 You should see them connect and start to exchange messages.
-
-## You can use Docker-compose to start this example:
-```sh
-docker-compose up -d
-```
-
-Now, you can see message exchanging, using `docker logs`.
--- a/examples/pion-to-pion-trickle/answer/main.go
+++ b/examples/pion-to-pion-trickle/answer/main.go
@@ -1,0 +1,154 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/pion/webrtc/v2"
+
+	"github.com/pion/webrtc/v2/examples/internal/signal"
+)
+
+func main() {
+	offerAddr := flag.String("offer-address", "localhost:50000", "Address that the Offer HTTP server is hosted on.")
+	answerAddr := flag.String("answer-address", ":60000", "Address that the Answer HTTP server is hosted on.")
+	flag.Parse()
+
+	// Everything below is the Pion WebRTC API! Thanks for using it ❤️.
+
+	// Prepare the configuration
+	config := webrtc.Configuration{
+		ICEServers: []webrtc.ICEServer{
+			{
+				URLs: []string{"stun:stun.l.google.com:19302"},
+			},
+		},
+	}
+
+	// Create a new API with Trickle ICE enabled
+	// This SettingEngine allows non-standard WebRTC behavior
+	s := webrtc.SettingEngine{}
+	s.SetTrickle(true)
+	api := webrtc.NewAPI(webrtc.WithSettingEngine(s))
+
+	// Create a new RTCPeerConnection
+	peerConnection, err := api.NewPeerConnection(config)
+	if err != nil {
+		panic(err)
+	}
+
+	// When an ICE candidate is available send to the other Pion instance
+	// the other Pion instance will add this candidate by calling AddICECandidate
+	peerConnection.OnICECandidate(func(c *webrtc.ICECandidate) {
+		if c == nil {
+			return
+		}
+
+		payload := []byte(c.ToJSON().Candidate)
+		resp, onICECandidateErr := http.Post(fmt.Sprintf("http://%s/candidate", *answerAddr), "application/json; charset=utf-8", bytes.NewReader(payload))
+		if onICECandidateErr != nil {
+			panic(onICECandidateErr)
+		}
+		defer func() {
+			closeErr := resp.Body.Close()
+			if closeErr != nil {
+				panic(closeErr)
+			}
+		}()
+	})
+
+	// A HTTP handler that allows the other Pion instance to send us ICE candidates
+	// This allows us to add ICE candidates faster, we don't have to wait for STUN or TURN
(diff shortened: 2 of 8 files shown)


cohosh:
Hey!

I was pointed to this PR while debugging some problems we recently started having that were caused by commit 7c18bbcceed776b547a44a0b921284db7d4bc69a.

The changes there delay gathering of ICE candidates for the trickle method to when the peer is preparing an SDP **answer**. The reason your trickle example works, is that you're relying on the other peer to send you your candidates directly: https://github.com/pion/webrtc/pull/763/commits/9be2a9fd172f726df145d49132aa5b4e214a69ea#diff-f70c62dbae11cfb1979e93d74673e9c3R73
and not on the gathering of ICE candidates provided in your library. In this example, only one side is actually trickling candidates, the other is adding them manually.

In fact, if you were to modify your example to rely on STUN, the **offer** peer would never connect because `Gather` is never being called on their side of the connection.

This is problematic I think not just for us, but also for others who are trying to use WebRTC with STUN or TURN. We rely on STUN and TURN for the design of our system and on a centralized signaling broker. Without the ability for a client to gather candidates for an offer, we'd have to add an additional round trip which would cause a lot of additional engineering and headache.

I'd like to suggest a fix for this example that we've also implemented in our code to prevent premature sharing of trickled candidates. We wait for the ICE gathering state to change to completed before sending candidates to peers: https://github.com/cohosh/snowflake/blob/pion/client/lib/webrtc.go#L183

Sean-Der:
Hey @cohosh 

So I think the *right* thing to do here is match whatever `libwebrtc`/Chromium does (I am not exactly sure what they/where I might have accidentally diverged)

The nice thing is we have WASM tests! So we can write Go code and test them via libwebrtc, this makes compliance a lot easier :)

----

For Trickle to work each side should set `OnICECandidate` and send the candidates to the remote peer. You can get away having only one side send because WebRTC will create peer reflexive candidates.

If you are waiting for `ICEGathererStateComplete` I don't think trickle is actually providing you any value. When trickle is off we basically don't return until `ICEGathererStateComplete`

----

Hopefully I am understanding the issue! Sorry it has been a little bit since I was in this area, so may be missing a few points. Please fire away and I will make sure to get everything fixed up ASAP.

cohosh:
> So I think the _right_ thing to do here is match whatever `libwebrtc`/Chromium does (I am not exactly sure what they/where I might have accidentally diverged)
> 
> The nice thing is we have WASM tests! So we can write Go code and test them via libwebrtc, this makes compliance a lot easier :)

Agreed :) I'll take a look at the issue you reported above and see if I can figure out why sending candidates was breaking things. Thanks for pointing out these tests, I'll be sure to use them when creating a patch

> 
> For Trickle to work each side should set `OnICECandidate` and send the candidates to the remote peer. You can get away having only one side send because WebRTC will create peer reflexive candidates.
> 
> If you are waiting for `ICEGathererStateComplete` I don't think trickle is actually providing you any value. When trickle is off we basically don't return until `ICEGathererStateComplete`

Yeah, we use it in a weird way here. The reason being that one of the callbacks we rely on has not been implemented in this library for the old non-trickle method. Perhaps the best way for us to move forward is to submit a patch implementing it. You can see relevant comments here: https://trac.torproject.org/projects/tor/ticket/28942#comment:28

Regardless, I think there's a better way to go about solving the issues you were having than the changes that were merged in this PR, due to the breaks outlined above where peers can no longer collect ICE candidates when crafting offers. I'll take a look at the bugs you saw with Chromium and see if I can suggest a way to fix them that also allows you to use the trickle method as intended.

Thanks again for the quick response and your help with this!

cohosh:
TL;DR is that the problem you were seeing with your example before merging ab0c239a9b4094dd9ecf7216b07fa525880b0902 was not a result of bugs in the implementation of SetLocalDescription and SetRemoteDescription, but rather because the implementation of your example allowed the answering peer to add ICE candidates before receiving an initial offer.

---

Okay, after digging into this farther, it seems that the problems you were seeing stem from the pion-to-pion-trickle example itself, which is implemented in a way that is unlikely to happen in practice. The issue you referenced above is caused by the offering peer sending their ice candidates to the answering peer before an initial offer has been provided **and** the fact that the answering peer accepted these candidates before accepting an offer. This (rightfully) throws an error because the answering peer calls `AddICECandidate` before making a call to `SetRemoteDescription` with the initial offer.

The code in `answer` specifically allows this behaviour, which should probably not happen in practice. Here's the RFC for trickle ice: https://tools.ietf.org/html/draft-ietf-mmusic-trickle-ice-02

Note:
```
   Trickled Candidates:  Candidates that a trickle ICE agent is sending
      subsequently to but within the context defined by an offer or an
      answer.  Trickled candidates can be sent in parallel with
      candidate harvesting and connectivity checks

```
the answering candidate has to process an `initial offer` and then call `SetRemoteDescription(offer)` before subsequently receiving trickled candidates and adding them via `AddICECandidate`. More information on initial offers here: https://tools.ietf.org/html/draft-ietf-mmusic-trickle-ice-02#section-5

In what I suspect would be a typical use case of trickle ice, the answering peer would wait for an initial offer before creating a PeerConnection and would record some kind of session id for that offer, making sure that the receipt of trickled ice candidates happens **after** the receipt of the offer and comparing the session id to make sure they are from the same remote peer. This example is constructed to make the assumption that there is one connecting peer, and in fact running the offer program twice causes it to fail.

The root of the problem for this example is therefore in the application, and not in the implementation of the WebRTC API, and so did not warrant the changes above. There are two problems:
1) The answering peer should not add trickled candidates before receiving an offer, and
2) The offering peer should hold off sending trickled candidates until they POST their offer.
I've included a patched version of the pion-to-pion-trickle example that makes use of synchronization primitives to solve the issues you were seeing here: https://github.com/pion/webrtc/pull/816

Going back to your earlier point that we should do whatever the libwebrtc implementation does, you'll note that in the google libwebrtc implementation:
- Candidates are not gathered in SetRemoteDescription: https://webrtc.googlesource.com/src/+/refs/heads/master/pc/peer_connection.cc#2606
- Candidates are gathered in SetLocalDescription regardless of whether it was an offer or answer: https://webrtc.googlesource.com/src/+/refs/heads/master/pc/peer_connection.cc#2225

Your comment above that our method of signaling is not the intended use case for trickle ice is true, I'm going to look into changing back to the old method of gathering candidates. However in the meantime, I think it's important to revert the changes you've made to the core API in this PR for the reasons outlined above. Because of our use case, Snowflake requires a fair amount of stability in the API. Should we expect to see fundamental changes like this being made for the purpose of single use cases (in this case a constructed example)? As far as I can tell, this example uses the webRTC API in a very unusual way that violates the specification of trickle ICE.

Sean-Der:
Hey @cohosh, traveling today! I will merge and give you a full answer as soon as I am in my hotel!

> Because of our use case, Snowflake requires a fair amount of stability in the API. Should we expect to see fundamental changes like this being made for the purpose of single use cases (in this case a constructed example)? As far as I can tell, this example uses the webRTC API in a very unusual way that violates the specification of trickle ICE.

This was a mistake on my part. I didn't intend to break compatibility with WebRTC (I actually thought my commit was fixing it!) I didn't do enough research/write tests and that caused this. In the future I will make sure to be more deliberate! It is never ok for us to break downstream, so I really want to make this always work for you!

I am really grateful Tor is using Pion. My intention is to have zero divergence from `W3C/IETF WebRTC` anytime we do is entirely accidental.

cohosh:
Thank you for all your help and for the feedback on our use of the library. I'm going to take a look at your comments on our pion trac ticket and make some fixes on our end. I think you're right that we shouldn't be using trickle. I'll back to you if we have PRs for anything else we might need there.

BTW, nice GopherCon talk!

PR #764, opened 2019-07-31
Fixes examples/echo on Safari and latest Firefox

fivebats:
#### Summary

- Modifies demo.js to support `addTrack` if available.
- Preloads codecs into mediaEngine taken from the offer SDP.

#### Description

The previous code only worked on browsers such as Chrome that happened
to favor VP8 on dynamic payload type 96. Safari and recent versions of
Firefox do not do this. Answer SDP really needs to follow the
codec/dynamic payload type assignments from the offer. Also the
javascript was modified to use `addTrack` if present, since `addStream` is
not available on Safari.

This code takes advantage of current `peerconnection.CreateAnswer`
behavior by preloading the sender's codecs into the mediaEngine,
knowing that the codecs will be added to the appropriate transceivers
in order. 

### Updates:

- No longer needs changes to sdp module.
- Simplified to take advantage of more existing code.


examples/echo/jsfiddle/demo.js
```diff
@@ -13,7 +13,13 @@ var log = msg => {
 
 navigator.mediaDevices.getUserMedia({ video: true, audio: true })
   .then(stream => {
-    pc.addStream(stream)
+    if (pc.addTrack) {
```
Sean-Der: IMO we can drop `addStream` and just do `addTrack`!

Mind changing it everywhere? Also feel free to use arrow functions (if you want) your PR :)

`stream.getTracks().forEach(track => pc.AddTrack(stream, stream)`

examples/echo/main.go
```diff
@@ -107,3 +128,55 @@ func main() {
 	// Block forever
 	select {}
 }
+
+// setCodecsFromOffer finds all codecs in a session description and adds them to a MediaEngine, using dynamic
```
Sean-Der: What do you think of adding a `populateMediaEngineFromSDP` for lack of a better name?

Otherwise people are going to copy this around a lot, that might be ok though if you think this is an uncommon use case. We can always push it into the library later.

Also we can/should go back and update other examples using this new helper function. Will be a big improvement (I assume `gstreamer-send` will need it!)

examples/echo/jsfiddle/demo.js
fivebats: OK, I will do that. I was not sure if that would break something really old...

examples/echo/main.go
fivebats: Will do. I'm terrible at naming things, so am always quite happy to change names of stuff. I haven't gotten to any other examples yet, but will start looking at them now.

examples/echo/jsfiddle/demo.js
Sean-Der: Yea totally!

I would say don't worry about using modern features in the examples. The only time I really care is when we change the API, refactoring aggressively in the examples or internal API is a fun free for all :)

examples/echo/main.go
Sean-Der: awesome! So I know https://github.com/pion/example-webrtc-applications/blob/master/gstreamer-send/main.go will probably blow up.

Also https://github.com/pion/webrtc/tree/master/examples/sfu-minimal probably

fivebats:
Just a note that the fix I made above (1bbad9c) doesn't work properly with Firefox nightly, but is OK in Safari and Chrome. Investigating...

fivebats:
Missed that you wanted `populateMediaEngineFromSDP` moved to `mediaEngine.PopulateFromSDP`. Will do that and test those other two examples next.

Sean-Der:
Nice! I really like this/think this is the best thing we can do.

Adding @mchlrhw he has opinions/has ran into this as well.

mediaengine.go
```diff
@@ -41,6 +42,47 @@ func (m *MediaEngine) RegisterDefaultCodecs() {
 	m.RegisterCodec(NewRTPVP9Codec(DefaultPayloadTypeVP9, 90000))
 }
 
+// PopulateFromSDP finds all codecs in a session description and adds them to a MediaEngine, using dynamic
+// payload types and parameters from the sdp.
+func (m *MediaEngine) PopulateFromSDP(sd sdp.SessionDescription) error {
```
Sean-Der: Can this take `webrtc.SessionDescription`. I want to avoid having people importing other packages in the happy case and have them run into type issue etc..

examples/echo/main.go
```diff
@@ -107,3 +128,21 @@ func main() {
 	// Block forever
 	select {}
 }
+
+// firstVideoPayloadType finds the dynamic payload type of the first video
+// codec in the session description
+func firstVideoPayloadType(sd sdp.SessionDescription) (uint8, error) {
```
mchlrhw: I like this approach and I think this can stay here for now, but it might be good to move this to `sdp` eventually because it seems like it could be more generally useful.
The RFC says:
> If the <proto> sub-field is "RTP/AVP" or "RTP/SAVP" the <fmt>
      sub-fields contain RTP payload type numbers.  When a list of
      payload type numbers is given, this implies that all of these
      payload formats MAY be used in the session, but the first of these
      formats SHOULD be used as the default format for the session.

so we could code up this logic as `func (s *SessionDescription) GetDefaultPayloadType(mediaType string) (uint8, error) {` in `sdp/util.go` and use it for audio as well. Maybe create a mediaType enum while we're at it :wink:

mediaengine.go
mchlrhw: This looks great :+1:

examples/echo/main.go
Sean-Der: 👍 I would prefer this as well. My goal  is be that users shouldn't import anything but `pion/webrtc` in the usual cases.

It sounds like there are even more issue we gotta deal with. Feel free to land anything as WIP. It can be frustrating to have long running PRs I don't want to burn you out.

Sean-Der: If you can get the build green, and then want to keep iterating on it.

fivebats:
I seem to have botched up the history. Will close this one and create a new one. Sorry!

PR #772, opened 2019-08-06
Fix echo for Safari and Firefox

fivebats:
#### Description

This fixes the echo program so it works properly on Safari
and Firefox, where the preferred offered dynamic media type
is not 96/VP8. It loads MediaEngine with codecs found in the
offer and then uses the payload type of the offer's preferred
video codec in the answer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Agugua Kenechukwu](https://github.com/spaceCh1mp)
 * [Ato Araki](https://github.com/atotto)
 * [Rafael Viscarra](https://github.com/rviscarra)
+* [Mike Coleman](https://github.com/fivebats)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/examples/echo/jsfiddle/demo.js
+++ b/examples/echo/jsfiddle/demo.js
@@ -13,7 +13,7 @@ var log = msg => {
 
 navigator.mediaDevices.getUserMedia({ video: true, audio: true })
   .then(stream => {
-    pc.addStream(stream)
+    stream.getTracks().forEach(track => pc.addTrack(track, stream));
     pc.createOffer().then(d => pc.setLocalDescription(d)).catch(log)
   }).catch(log)
 
--- a/examples/echo/main.go
+++ b/examples/echo/main.go
@@ -7,12 +7,35 @@ import (
 
 	"github.com/pion/rtcp"
 	"github.com/pion/webrtc/v2"
-
 	"github.com/pion/webrtc/v2/examples/internal/signal"
 )
 
 func main() {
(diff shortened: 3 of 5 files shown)


Sean-Der:
This LGTM! Really nice @fivebats 

Would you be up for talking why Track/MediaStream doesn't actually work? Probably will need to be `/v3` breakage, but I would love to see this happen. I have been blissfully unaware of these poor design choices :(

fivebats:
> Would you be up for talking why Track/MediaStream doesn't actually work?

I think I may have to write a document rather than describe it all here. But beware of the fact that I have only begun learning the pion code base and in the past I had little need to use the W3C API, since I was mostly working on the media server end of things.  But I can see how some of the concepts from media serving seem to map to the W3C stuff.

The first disconnect from my understanding of W3C was that pion's webrtc.Track has session-specific fields like payloadType and codec. Seems like those things are negotiated on a per-session basis, and because tracks can be shared among sessions (which could be completely different browsers or devices) they shouldn't contain that stuff. That stuff really belongs in the Sender or Transceiver I think, especially if we add things like transcoding.

I would be interested in seeing how other contributors on pion are thinking about media flow as it relates to this API. I guess that a lot of what I normally do is hidden from the W3C API, left for the browser implementers to do. For example, to get media in W3C you simply add tracks from a MediaStream to a PeerConnection. The media flow is hidden from you. So where should a developer really write media samples with this API? Seems to me that this could happen in the MediaStream, which is responsible for synchronizing media and getting it to all senders properly, using the session-specific settings for each sender. So it feels to me like there is too much in Track, MediaStream should exist, and I would expect to see more code in the transceiver, receiver, and sender if we want to support media server type functionality.

Anyway, those are some initial thoughts. Probably for some later version. I'm sure this comment will seem really naive to future me, after I get more experience with pion and its API ;-)

fivebats:
> However, in the general case I wouldn't expect the media engine to be populated from the offer, but rather from the supported codecs of the answering peer.

That is why what PopulateFromSDP was originally local to the echo program. The echo program is truly codec-agnostic since it reflects the sender's packets back without touching the media.

The way pion currently generates answers is broken IMO since it needs to be aware of the offer's mapping of payload types to codecs. If the offer says VP8 is payload 100, the answer must follow that choice. So yes, in general pion should answer with the subset of codecs it can handle from the offer. But echo can handle them all...

Sean-Der:
@fivebats you should have a `Rebase and merge` button, feel free to do that w/e you want! It should be unblocked since this PR has an approval (you just need one)

thanks!

fivebats:
> you should have a Rebase and merge button, feel free to do that w/e you want! It should be unblocked since this PR has an approval (you just need one)

Nope. Been waiting to see something like that, but I can only close or comment. 

It says "Only those with write access to this repository can merge pull requests."

Sean-Der:
@fivebats I sent you another invite to join the Pion org. When you accept that should work.

Make sure to enable 2FA that is required to Join I believe

PR #780, opened 2019-08-11
Parse DTLS setup in SetRemoteDescription

### LINKED ISSUE #494: Support For Creating an Answer with Passive Negotiation
RFC 5763 - Section 5
```
The endpoint MUST use the setup attribute defined in [RFC4145].
The endpoint that is the offerer MUST use the setup attribute
value of setup:actpass and be prepared to receive a client_hello
before it receives the answer.  The answerer MUST use either a
setup attribute value of setup:active or setup:passive.  Note that
if the answerer uses setup:passive, then the DTLS handshake will
not begin until the answerer is received, which adds additional
latency. setup:active allows the answer and the DTLS handshake to
occur in parallel.  Thus, setup:active is RECOMMENDED.  Whichever
party is active MUST initiate a DTLS handshake by sending a
ClientHello over each flow (host/port quartet).
```

Currently, pions/webrtc assumes that if the remote SDP is an `RTCSdpTypeOffer` that we must reply with `setup:active` and initiate the DLTS handshake. There should be an option to be the passive party in the connection if we're creating an answer.

I could probably make a PR for this if need be.

Sean-Der:
Take into consideration if remote is running DTLS as a
client/server. Before we ignored this value and we could
enter cases where DTLS would never connect.

Resolves #494

### DIFF
--- a/dtlsrole.go
+++ b/dtlsrole.go
@@ -1,5 +1,9 @@
 package webrtc
 
+import (
+	"github.com/pion/sdp/v2"
+)
+
 // DTLSRole indicates the role of the DTLS transport.
 type DTLSRole byte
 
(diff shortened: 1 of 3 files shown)


dtlsrole.go
```diff
@@ -28,3 +32,29 @@ func (r DTLSRole) String() string {
 		return unknownStr
 	}
 }
+
+// Iterate a SessionDescription from a remote to determine if an explicit
+// role can been determined for local connection. The decision is made from the first role we we parse.
+// If no role can be found we return DTLSRoleAuto
+func dtlsRoleFromRemoteSDP(sessionDescription *sdp.SessionDescription) DTLSRole {
```
hugoArregui: I need to read more about this to be able to make a good review, because this sounds good to me:

(maybe we should like to this in the code, is always nice to be able to compare it to the spec)

```
     The endpoint that is the offerer MUST use the setup attribute
      value of setup:actpass and be prepared to receive a client_hello
      before it receives the answer.  
```

Done, it was the previous PR.

```
The answerer MUST use either a
      setup attribute value of setup:active or setup:passive.  Note that
      if the answerer uses setup:passive, then the DTLS handshake will
      not begin until the answerer is received, which adds additional
      latency. setup:active allows the answer and the DTLS handshake to
      occur in parallel.  Thus, setup:active is RECOMMENDED.  Whichever
      party is active MUST initiate a DTLS handshake by sending a
      ClientHello over each flow (host/port quartet).
```

This is this code right here.. but.. how do we factor multiple values for different media descriptions?

Sean-Der: We use the first value we find. 

If media descriptions disagree we can't handle that. Since we only have one DTLSTransport I don't think WebRTC can handle them being different values ever. Should we add an assert to make sure all values are the same?

PR #782, opened 2019-08-12
Modified sfu-minimal to work with heterogeneous clients

fivebats:
#### Description

The original code only worked with Chrome as sender and receiver. This
change permits the sender and receivers to all be different types of
clients, such as Chrome, Safari, and Firefox.

* Changed local track so it always gets VP8 from remote client using
their dynamic payload type instead of assuming 96, enabling interop
with Safari and recent Firefox.

* Changed remote tracks so they also negotiate the correct VP8 payload
  type.

* Added MediaEngine.GetCodecsByName() to assist the above.

* Changed RTPSender so it writes its MediaEngine's VP8 payload type
  into the RTP header, rather than using the sender's VP8 payload type
  when the packets are written to the remote track.

#### Compatibility

This change to RTPSender will affect all code that
currently uses RTPSender.sendRTP(). That method has not written into
the RTP header until now.

### DIFF
--- a/examples/sfu-minimal/README.md
+++ b/examples/sfu-minimal/README.md
@@ -29,4 +29,9 @@ You can change the listening port using `-port 8011`
 
 You can `Join the broadcast` as many times as you want. The `sfu-minimal` Golang application is relaying all traffic, so your browser only has to upload once.
 
+When using `curl` Instead of pasting the SDP into the terminal, you may prefer using a paste-from-clipboard command:
+
+* Linux: `xclip -o | curl http://localhost:8080/sdp -d @-`
+* macOS: `pbpaste | curl http://localhost:8080/sdp -d @-`
+
 Congrats, you have used Pion WebRTC! Now start building something cool
--- a/examples/sfu-minimal/jsfiddle/demo.js
+++ b/examples/sfu-minimal/jsfiddle/demo.js
@@ -21,7 +21,8 @@ window.createSession = isPublisher => {
   if (isPublisher) {
     navigator.mediaDevices.getUserMedia({ video: true, audio: false })
       .then(stream => {
-        pc.addStream(document.getElementById('video1').srcObject = stream)
+        stream.getTracks().forEach(track => pc.addTrack(track, stream));
+        document.getElementById('video1').srcObject = stream;
         pc.createOffer()
           .then(d => pc.setLocalDescription(d))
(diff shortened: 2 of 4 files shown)


Sean-Der:
My vote is to make this explicit @fivebats. I think having the `RTPSender` modify packets is a good temporary fix to the issue!

I doubt this behavior change will effect anyone, but I am worried that people will depend on the behavior. What do you think of people explicitly calling `RewritePayloadType` or something like that?

Sean-Der:
Sorry this took so long to respond! I prototyped a few things locally and played with it.

fivebats:
> What do you think of people explicitly calling `RewritePayloadType` or something like that?

I think you as benevolent dictator of the API should probably decide that. The W3C APIs for `RTCRtpSender` does have a `setParameters` method that takes a dictionary, so that may be a more standards-aligned way of setting up the `RTPSender`.

And this code is temporary, I hope, because Tracks should not have codecs or payload types - they are too high of a layer for that. As I said before I think there needs to be some better codec interface in MediaStream, and that MediaStream, maybe through the transceiver, should set the parameters on the senders.

I also just found a bug in this code: It will not properly choose the payload type (identify the proper Codec) if the media is not using the preferred encoding for its type. This could come up with codecs like H264 where perhaps the answerer had to choose the sender's second choice of profile/level. The current code just assumes the preferred one is chosen. This should be an easy fix.

fivebats:
This has been idle almost a month now, so I am closing it.

Sean-Der:
Shit, sorry @fivebats 

I spent some time with this, and here were my thoughts.

* We should rework all the media APIs. Everything should be driven via `AddTransceiver`. `AddTrack` is deprecated, but you can't really remove things on the web (but we can)

* `AddTransceiver` should always accept a `MediaStreamTrack`. On the `MediaStreamTrack` you set the codec, and that is it. During `SetRemoteDescription` we should validate if the remote supports the codecs we wish to use.

* We should handle/rewrite PayloadTypes completely. We have to copy anyway for encryption, so why not.

----

Sorry I haven't been more hands on. Since this is a bigger more ambiguous task it is hard to get time for it. I really think it is important/I would love your involvement with the project. I just can't be as hands on (but in the same vein if you have anything you want to do you have sign-off quicker) I trust you to make the right decision

thanks

PR #787, opened 2019-08-14
Remove doubling of connection setup in SDP creation

suhasgaddam:
#### Description
I noticed `a:setup=actpass` showing up twice in each media description in an SDP. It is already being added in the `addTransceiverSDP` and `addDataMediaSection` methods.


Sean-Der:
@suhasgaddam great work!

* Can you update your commit message so it is `subject`/`body`

You could do something like
```
Remove doubling of setup attribute in SDP creation

Before a:setup=actpass was showing up twice in each media description. It is already being added in the addTransceiverSDP and addDataMediaSection methods. This commit removes 
the extra call.
```

* Add yourself to the README

* Add your test to `peerconnection_go_test.go` instead. Our media APIs diverge from the browser, so we can't compile them to WASM.

After that it should be good to merge! Really excited to see you involved :)

PR #789, opened 2019-08-15
Fix OpusWriter issues that break VLC

Sean-Der:
* Correctly calculate page CRC
Ogg uses slightly non-standard version of CRC, we are unable to
use the Go's version. You can see the details here[0], in summary
"direct algorithm, initial val and final XOR = 0,
generator polynomial=0x04c11db7"

[0] https://xiph.org/vorbis/doc/framing.html

* Properly set EOS value
Before we created a new page with an EOS. Instead seek backwards
and update the last page with valid Opus to have EOS header and
re-generate the CRC

* Only use 0 timestamp/granule for headers
Audio itself should start from 0


pkg/media/oggwriter/oggwriter.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// New builds a new OGG Opus writer
+func New(fileName string, sampleRate uint32, channelCount uint16) (*OggWriter, error) {
+	f, err := os.Create(fileName)
+	if err != nil {
+		return nil, err
+	}
+	writer, err := NewWith(f, sampleRate, channelCount)
+	if err != nil {
```
hugoArregui: should be close the descriptor f here?

pkg/media/oggwriter/oggwriter.go
```diff
@@ -0,0 +1,248 @@
...
+    'Beginning Of Stream'
+
+   Figure 1: Example Packet Organization for a Logical Ogg Opus Stream
+*/
+
+func (i *OggWriter) writeHeaders() error {
+	// ID Header
+	oggIDHeader := make([]byte, 19)
+
+	copy(oggIDHeader[0:], []byte("OpusHead"))                       // Magic Signature 'OpusHead'
```
hugoArregui: maybe it will be worthy to put all the magic signatures as constants? maybethe compiler already optimizes this as it is

pkg/media/oggwriter/oggwriter.go
```diff
@@ -0,0 +1,248 @@
...
+	for index := range page {
+		checksum = (checksum << 8) ^ i.checksumTable[byte(checksum>>24)^page[index]]
+	}
+	binary.LittleEndian.PutUint32(page[22:], checksum) // Checksum - generating for page data and inserting at 22th position into 32 bits
+
+	return page
+}
+
+// WriteRTP adds a new packet and writes the appropriate headers for it
+func (i *OggWriter) WriteRTP(packet *rtp.Packet) error {
```
hugoArregui: Something I think it may be worth looking at (but I know nothing about this): it seems to me (by just reading the write-to-disk test), this WriteRTP method will be called in a loop and this function calls createPage which will alocate the page. This could be easily avoided by defining a buffer before the loop and using it in every iteration. Maybe is not worthy though, I'm talking by just reading at the code, don't really know much about the use case.

pkg/media/oggwriter/oggwriter.go
Sean-Der: Fixed!

pkg/media/oggwriter/oggwriter.go
Sean-Der: Fixed! I think it just good for readability, nice to have things be explicit

pkg/media/oggwriter/oggwriter.go
Sean-Der: Yea I don't know either. IMO the optimization isn't worth it (yet) just because we haven't measured.

I don't think the `oggwriter` is really going to see much usage besides examples either. More just a fun way for people to see it work

PR #792, opened 2019-08-17
matching the video MTU from chromium

epes:
#### Description
Setting the MTU to be the same as the video MTU from Chromium. This should fix issues where certain networks would fragment RTP packets from Pion but not from Chromium.

Discussion: https://groups.google.com/forum/#!topic/discuss-webrtc/gH5ysR3SoZI
Chromium constant: https://codesearch.chromium.org/chromium/src/third_party/webrtc/media/engine/constants.cc?type=cs&q=1200+file:webrtc&sq=package:chromium&l=16


Sean-Der:
Merged with d05491f52b7ef54708c4ee0b289474ad1ebe57fb, rebased because of merge conflict with other commit today.

thanks again for upstreaming this @epes! Had no idea this was an issue

PR #801, opened 2019-08-23
Add testing for 32-bit OS

### LINKED ISSUE #750: Add CI test for 32-bit OS
This happens after pion/webrtc@2.0.23...
- v2.0.23 - OK
- v2.0.24 - fail (TestDataChannelBufferedAmount/set_before_datachannel_becomes_open fail)
- v2.0.25 - fail (TestDataChannelBufferedAmount/set_before_datachannel_becomes_open fail)

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x804a7cc]

goroutine 48 [running]:
runtime/internal/atomic.Xadd64(0x90526ec, 0x4, 0x0, 0xffffffff, 0x0)
	/root/go/current/src/runtime/internal/atomic/asm_386.s:105 +0xc
github.com/pion/datachannel.(*DataChannel).WriteDataChannel(0x90526c0, 0x903dd90, 0x4, 0x4, 0x8070b00, 0x91204d0, 0x91d60c0, 0xc0000000)
	/root/go_ws/pkg/mod/github.com/pion/datachannel@v1.4.4/datachannel.go:294 +0x7e
github.com/pion/webrtc/v2.(*DataChannel).Send(0x9118360, 0x9130d90, 0x4, 0x4, 0x9120518, 0x9120a88)
	/root/go_ws/pkg/mod/github.com/pion/webrtc/v2@v2.0.24/datachannel.go:320 +0x72
```

I believe, the atomic operation (for datachannel stats we recently added) has some alignment issue. (not 100% sure yet)

This is causing crashes on win32 and android - also with 32bit.

Sean-Der:
We are hitting 32bit specific errors for things like atomics.
This adds another step that runs our test suite via a 32bit Docker
image.

Resolves #750

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,8 @@ cache:
 env:
   - GO111MODULE=on
 
-notifications:
-  email: false
-  slack:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -22,20 +20,20 @@ install:
...
   - yarn install
 
-before_install:
-  - wget https://raw.githubusercontent.com/creationix/nvm/v0.31.0/nvm.sh -O ~/.nvm/nvm.sh
-
-before_script:
-  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.17.1
-
 script:
+  - bash .github/assert-contributors.sh
```
Sean-Der: @hugoArregui moved these to the top so they should fail faster! Hopefully cause less frustration with these :)

.travis.yml
```diff
@@ -6,10 +6,8 @@ cache:
 env:
   - GO111MODULE=on
 
-notifications:
```
Sean-Der: This was a Slack instance that was just for Pion, hasn't been used for almost two years :)

We moved everything to Gophers Slack since it had paid features.

.travis.yml
hugoArregui: hehe, thanks!

PR #802, opened 2019-08-23
Bump wrtc used for WASM testing

Sean-Der:
Upstream has fixed an issue that was blocking one of our tests. Enable
the test and bump the wrtc version.

### DIFF
--- a/datachannel_test.go
+++ b/datachannel_test.go
@@ -223,44 +223,39 @@ func TestDataChannelParameters(t *testing.T) {
 	defer report()
 
 	t.Run("MaxPacketLifeTime exchange", func(t *testing.T) {
-		// Note(albrow): See https://github.com/node-webrtc/node-webrtc/issues/492.
-		// There is a bug in the npm wrtc package which causes this test to fail.
-		// TODO(albrow): Uncomment this once issue is resolved.
-		t.Skip("Skipping because of upstream issue")
+		ordered := true
+		maxPacketLifeTime := uint16(3)
+		options := &DataChannelInit{
+			Ordered:           &ordered,
+			MaxPacketLifeTime: &maxPacketLifeTime,
+		}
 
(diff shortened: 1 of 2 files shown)


datachannel_test.go
```diff
@@ -223,39 +223,34 @@ func TestDataChannelParameters(t *testing.T) {
...
-		// 	// Check if parameters are correctly set
-		// 	assert.True(t, d.Ordered(), "Ordered should be set to true")
-		// 	if assert.NotNil(t, d.MaxPacketLifeTime(), "should not be nil") {
-		// 		assert.Equal(t, maxPacketLifeTime, *d.MaxPacketLifeTime(), "should match")
-		// 	}
-		// 	done <- true
-		// })
-
-		// closeReliabilityParamTest(t, offerPC, answerPC, done)
+		var ordered = true
```
albrow: Maybe this is just a stylistic source but I see no reason not to use the `:=` shorthand here.

```suggestion
		ordered := true
```

peerconnection_media_test.go
```diff
@@ -372,7 +372,6 @@ func TestPeerConnection_Media_Shutdown(t *testing.T) {
 	pcAnswer.OnICEConnectionStateChange(func(iceState ICEConnectionState) {
 		if iceState == ICEConnectionStateConnected {
 			go func() {
-				time.Sleep(3 * time.Second) // TODO PeerConnection.Close() doesn't block for all subsystems
```
Sean-Der: Unrelated, but noticed this when reading code! 

We shouldn't need this anymore. In `pion/transport` we sleep and check, so don't need this

datachannel_test.go
Sean-Der: agree, fixed! The `var` is just extra to distract the reader, and offers no real value :)

PR #805, opened 2019-08-27
Pass all fingerprints from SDP to dtls transport

suhasgaddam:
Before pion was only checking against the first fingerprint found in
the SDP, now all fingerprints from the SDP are used to validate
during the dtls handshake.

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -849,6 +849,30 @@ func (pc *PeerConnection) LocalDescription() *SessionDescription {
 	return pc.currentLocalDescription
 }
 
+func createDTLSFingerprintFrom(fingerprint string) (DTLSFingerprint, error) {
+	dtlsFingerprint := DTLSFingerprint{}
+
+	parts := strings.Split(fingerprint, " ")
+	if len(parts) != 2 {
+		return dtlsFingerprint, fmt.Errorf("invalid fingerprint")
+	}
+
+	dtlsFingerprint.Algorithm = parts[0]
+	dtlsFingerprint.Value = parts[1]
+
+	return dtlsFingerprint, nil
+}
+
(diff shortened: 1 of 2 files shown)


Sean-Der:
@suhasgaddam 

Here is the text from the RFC
```
   The fingerprint attribute may be either a session-level or a media-
   level SDP attribute.  If it is a session-level attribute, it applies
   to all TLS sessions for which no media-level fingerprint attribute is
   defined.
```

So it sounds like we need to ignore the session level attribute, for each of the individual ones. I *think* this is the proper logic then.

* Loop through each media sections. If you find a fingerprint use it, confirm that all the fingerprints match though. Currently we only support one `DTLSTransport`, so that is an important error to communicate to the user.
* If we didn't find anything at the media level, then check the session level. If there isn't anything at the session level return an error.

Sorry I didn't respond sooner!

Sean-Der:
@suhasgaddam hey! no rush, but anything I can do to help here?

Hopefully you are still using Pion for your project. If not I am happy to write/land this change myself, but would much rather see contributors :)

thanks (and no rush!)

suhasgaddam:
@Sean-Der, I am still using pion. The current changeset was a quick and dirty fix for me(and I didn't know what should happen if there are multiple fingerprints). I can still fix this up according to the RFC, as long as no one is in a rush for the fix.

suhasgaddam:
I still need to test this against an actual application. What should happen if N media sections, 1 section does not have a fingerprint?

Sean-Der:
> I still need to test this against an actual application. What should happen if N media sections, 1 section does not have a fingerprint?

It should use the session fingerprint.

If the session doesn't match the other Media fingerprints then we need to error out. In the future we can support multiple fingerprints (multiple DTLS sessions) but for the time being we just support one DTLSTransport.

Sean-Der:
No rush either! I think this is a great fix (and I am sure people are hitting this today) so will be great to get in :)

Sean-Der:
Closing for #1023

PR #812, opened 2019-08-31
Added examples/data-channels-flow-control

### LINKED ISSUE #811: Add a data channel example with `bufferedAmount` control
## Summary
Add a data channel example using `bufferedAmount` and `bufferedAmountLowThreshold`.

## Motivation
There was a multiple of times pion user stumbled upon the flow control issues with data channel.  I think it is nice to have an example that demonstrates how to use `bufferedAmount` and `bufferedAmountLowThreshold`.

Just points to some external sources.

## Additional context
https://godoc.org/github.com/pion/webrtc#DataChannel

enobufs:
Resolves #811

### DIFF
--- a/examples/data-channels-flow-control/README.md
+++ b/examples/data-channels-flow-control/README.md
@@ -1,0 +1,65 @@
+# data-channels-flow-control
+This example demonstrates how to use the following property / methods.
+
+* func (d *DataChannel) BufferedAmount() uint64
+* func (d *DataChannel) SetBufferedAmountLowThreshold(th uint64)
+* func (d *DataChannel) BufferedAmountLowThreshold() uint64
+* func (d *DataChannel) OnBufferedAmountLow(f func())
+
+These methods are equivalent to that of JavaScript WebRTC API.
+See https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel for more details.
+
+## When do we need it?
+Send or SendText methods are called on DataChannel to send data to the connected peer.
+The methods return immediately, but it does not mean the data was actually sent onto
+the wire. Instead, it is queued in a buffer until it actually gets sent out to the wire.
+
+When you have a large amount of data to send, it is an application's responsibility to
+control the buffered amount in order not to indefinitely grow the buffer size to eventually
+exhaust the memory.
+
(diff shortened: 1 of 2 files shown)


enobufs:
@hugoArregui I can try whatever that makes this example easier to understand. Which part did you think it was confusing? What do you mean by "removing endpoint"?

I actually tried to localize signaling exchange in the main() (so that you wouldn't see it in the logic above). Also, avoid having separate implementation for each endpoint by having Endpoint, otherwise, it would have redundant lines in the code to have shorter code. (I am just trying to find the best approach..)

hugoArregui:
> @hugoArregui I can try whatever that makes this example easier to understand. Which part did you think it was confusing? What do you mean by "removing endpoint"?
> 
> I actually tried to localize signaling exchange in the main() (so that you wouldn't see it in the logic above). Also, avoid having separate implementation for each endpoint by having Endpoint, otherwise, it would have redundant lines in the code, more lines in the code. (I am just trying to find the best approach..)

Maybe is not possible to avoid it, for me it was distracting to read about the Endpoint struct, I was trying to understand what and endpoint was. But I'm not saying the code is not clear, I like the implementation a lot, I just think in this particullar case it may be worth to have a objectively worse code but focus on the control flow. 

How much code are you reusing though, since you are not starting the second endpoint it seems to me there is not a lot in common. I can give it a shot myself later tonight

enobufs:
@hugoArregui Let me flatten the code to see if it looks better then...

hugoArregui:
@enobufs perhaps something like this: https://gist.github.com/hugoArregui/fcfba97fb5f4dc06a2497d9370c413c8

Which is objectively worst than yours btw!, I just think is more straight to the point, feel free to disagree and merge as it is thought, this is very subjective.

enobufs:
@hugoArregui I am ok with it. Can you push the code and I can approve it. Thanks!

hugoArregui:
There you go!

enobufs:
Oh that was fast!  It looks much better! Approved back at you! ;)

enobufs:
Oh.. I have to update the README to align with the new code.
An unrelated test is failing also.

PR #814, opened 2019-09-02
Use @daenney improvements for .travis.yml

Sean-Der:
* Split into lint/test stages so we can fail fast
* cache modules
* Only upload coverage if testing passes


Sean-Der:
I am going to come back to this eventually, but I am hoping that travis-ci fixed WASM with 1.13

PR #815, opened 2019-09-03
Negotiated datachannels

AeroNotix:
#### Description

Adds the Negotiated fields to the right places.

I feel like there's more to do with this PR but just looking for review.


AeroNotix:
The behaviour I see from the browser WebRTC implementation is that if a datachannel is created with the options `{negotiated: true, id: 0}` where ID is any value for a stream id. Then there is no associated event triggered for `ondatachannel`.

What is supposed to happen is that when both peers create a datachannel that is negotiated, with the same ID on both sides - _then_ the datachannel is said to be `opened`. The implementation in `pion` seems to be that all datachannels are opened as _non-negotiated_ which means in pion you cannot control whether a datachannel gets opened at all. If one peer decides to open a datachannel then the other peer must accept it and the local peer always calls its `DataChannel.OnOpen` call back.

I can build a small JS example and the same example in `pion` to show the differences but looking for feedback on the above to ensure that I've understood the various RFCs around this.

AeroNotix:
For example, if we read: https://www.w3.org/TR/webrtc/#rtcdatachannel

Then it states:

> [...] The second way is to let the application negotiate the RTCDataChannel. To do this, create an RTCDataChannel object with the negotiated RTCDataChannelInit dictionary member set to true, and signal out-of-band (e.g. via a web server) to the other side that it SHOULD create a corresponding RTCDataChannel with the negotiated RTCDataChannelInit dictionary member set to true and the same id. This will connect the two separately created RTCDataChannel objects. The second way makes it possible to create channels with asymmetric properties and to create channels in a declarative way by specifying matching ids.

I've played around with this in JS just in the browser, seeing how it deals with this and now I'm more confused.

```javascript
      // Create the data channel and establish its event listeners
      sendChannel = localConnection.createDataChannel("sendChannel", {negotiated: true, id: 0});
      sendChannel.onopen = (event) => {
          console.log("local sendChannel on open");
      };
      sendChannel.onclose = (event) => {
          console.log("local sendChannel on close");
      };

      localConnection.ondatachannel = (event) => {
          console.log('local ondatachannel');
      };

      // Create the remote connection and its event listeners
      sendChannel.addEventListener("open", (event) => {
          console.log("addEventListener open");
          console.log(event);
      });

      remoteConnection = new RTCPeerConnection();
      remoteConnection.ondatachannel = (event) => {
          console.log("remote ondatachannel");
      };
```

We get events from the onopen function and the addEventListener but the remoteConnection.ondatachannel call back isn't called. 

What really is the expected behaviour? How do you control which datachannels are opened either side of a WebRTC connection without having to support and open all channels requested by both parties? The documentation for WebRTC around the web in various forms suggests that the `{negotiated: true, id:0}` method is the way to do this but no matter what - the channels are always opened, at least on the local side.

AeroNotix:
from w3c.github.io: 

> Each RTCDataChannel has an associated underlying data transport that is used to transport actual data to the other peer. In the case of SCTP data channels utilizing an RTCSctpTransport (which represents the state of the SCTP association), the underlying data transport is the SCTP stream pair. The transport properties of the underlying data transport, such as in order delivery settings and reliability mode, are configured by the peer as the channel is created. The properties of a channel cannot change after the channel has been created. The actual wire protocol between the peers is specified by the WebRTC DataChannel Protocol specification [RTCWEB-DATA].

An RTCDataChannel can be configured to operate in different reliability modes. A reliable channel ensures that the data is delivered at the other peer through retransmissions. An unreliable channel is configured to either limit the number of retransmissions ( maxRetransmits ) or set a time during which transmissions (including retransmissions) are allowed ( maxPacketLifeTime ). These properties can not be used simultaneously and an attempt to do so will result in an error. Not setting any of these properties results in a reliable channel.

An RTCDataChannel, created with createDataChannel or dispatched via an RTCDataChannelEvent, MUST initially be in the connecting state. Each RTCDataChannel has an associated underlying data transport that is used to transport actual data to the other peer. In the case of SCTP data channels utilizing an RTCSctpTransport (which represents the state of the SCTP association), **the underlying data transport is the SCTP stream pair.** The transport properties of the underlying data transport, such as in order delivery settings and reliability mode, are configured by the peer as the channel is created. The properties of a channel cannot change after the channel has been created. The actual wire protocol between the peers is specified by the WebRTC DataChannel Protocol specification [RTCWEB-DATA].

An RTCDataChannel can be configured to operate in different reliability modes. A reliable channel ensures that the data is delivered at the other peer through retransmissions. An unreliable channel is configured to either limit the number of retransmissions ( maxRetransmits ) or set a time during which transmissions (including retransmissions) are allowed ( maxPacketLifeTime ). These properties can not be used simultaneously and an attempt to do so will result in an error. Not setting any of these properties results in a reliable channel.

An RTCDataChannel, created with createDataChannel or dispatched via an RTCDataChannelEvent, MUST initially be in the connecting state. **When the RTCDataChannel object's underlying data transport is ready, the user agent MUST announce the RTCDataChannel as open.**

Emphasis mine.

It seems browsers also implement this differently as well - owing to the fact that the `onopen` event is fired even if the actual stream is not actually available for us (i.e. the underlying data transport is _NOT READY_)

So question is @Sean-Der what should `pion` do? Realistically if you create a stream that a peer is not expecting, that is, one side creates a stream with `negotiated: true` and yet your `onopen` event fires - you are just casting data into the abyss (this actually happens in the browsers I tested) which eats resources or has a peer perform costly operations.

Sean-Der:
[pcap.zip](https://github.com/pion/webrtc/files/3573177/pcap.zip)

This should be a pretty simple change! It looks like when someone passes `negotiated: true` each side just needs to not fire a `DATACHANNEL_OPEN_REQUEST` in the attached pcap you can see it if you decode the packets as SCTP.

Sean-Der:
In the `pion/datachannel` project don't call this write https://github.com/pion/datachannel/blob/master/datachannel.go#L116

https://gist.github.com/5a166e16b5383daf3788a776d7845d60 is helpful for debugging. I just grabbed a capture via `tcpdump` to confirm the differences.

I think our current implementation of `onopen` is fine. Since the association is ready no data should be lost!

AeroNotix:
@Sean-Der reason I mention about the `onopen` implementation is that I wonder if there's an issue with emitting the `onopen` event when the datachannel may not exist on the peer.

Sean-Der:
Aye good question.

So we can wait/emit until the association is ready. But we can't really know if the other side accepts/acknowledges our DataChannel.

This is probably worth another PR. If you update `pion/datachannel` to take into account the value of negotiated did everything work ok?

Oh! We also have our WASM tests, so that would be a great way to test the compliance of all our DataChannel methods.

Sean-Der:
Hey @AeroNotix 

Anything I can do to help?

AeroNotix:
@Sean-Der not sure - mostly I am wondering if the behaviour of how the `OnOpen` event is triggered for negotiated datachannels even when the remote peer isn't going to create one for it is the right behaviour.

E.g. peer A attempts to connect to peer B. Peer A offers a bunch of datachannels - yet peer B refuses or does not want to handle, or doesn't understand how to handle these datachannels but Peer A's `OnOpen` event still fires. Without some mechanism via signalling or a "control plane" datachannel that is always requested - what exactly is the solution here?

Reason I started looking into negotiated data channel support is because I want a generic application to be able to support specific datachannels for specific streams of data. Having them "negotiated" should mean that either peer opens the channels that they will need for the session, if one side doesn't open that channel then that session doesn't support that stream - and my impression was that the `OnOpen` event should not fire.

Thoughts?

Sean-Der:
Closing in favor of #855 

`pion/datachannel` change has landed, just need to wire up for Go

PR #830, opened 2019-09-15
Add PeerConnectionState

### LINKED ISSUE #614: Shut down PeerConnection when async errors occur
### What did you do?
Made a PeerConnection Chrome <=> Pion. Due to an unrelated bug, [dtlsTransport.Start](https://github.com/pion/webrtc/blob/7f6c01d5bc8dfcb6f0aaa70d12c02d8049a3fb7b/peerconnection.go#L996) returned `UnexpectedMessage`. The ICE connection is still stable, but packets sent from Pion are not delivered to Chrome.

### What did you expect?
When the DTLS connection fails I would expect the ICE connection to go down as well. Otherwise there's no way for Chrome to know that there's a problem. Not sure what the spec says here.

We should fix the TODO at [peerconnection.go:1001](https://github.com/pion/webrtc/blob/7f6c01d5bc8dfcb6f0aaa70d12c02d8049a3fb7b/peerconnection.go#L1001).

(reported by @adwpc)

Sean-Der:
Start properly setting the PeerConnectionState and make
OnConnectionState available

Resolves #614

----

TODO
* [x] Add test if remote gives invalid fingerprint
* [x] Assert that PeerConnectionState goes through checking/connected/closed (Not possible libwebrtc is inconsistent so we can't test this across both)

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -485,7 +485,7 @@ func (pc *PeerConnection) createICEGatherer() (*ICEGatherer, error) {
 
 // Update the PeerConnectionState given the state of relevant transports
 // https://www.w3.org/TR/webrtc/#rtcpeerconnectionstate-enum
-func (pc *PeerConnection) updateConnectionState() {
+func (pc *PeerConnection) updateConnectionState(iceConnectionState ICEConnectionState, dtlsTransportState DTLSTransportState) {
 	pc.mu.Lock()
 	defer pc.mu.Unlock()
 
@@ -496,22 +496,22 @@ func (pc *PeerConnection) updateConnectionState() {
 		connectionState = PeerConnectionStateClosed
 
 	// Any of the RTCIceTransports or RTCDtlsTransports are in a "failed" state.
-	case pc.iceConnectionState == ICEConnectionStateFailed || pc.dtlsTransport.State() == DTLSTransportStateFailed:
+	case iceConnectionState == ICEConnectionStateFailed || dtlsTransportState == DTLSTransportStateFailed:
 		connectionState = PeerConnectionStateFailed
 
 	// Any of the RTCIceTransports or RTCDtlsTransports are in the "disconnected"
 	// state and none of them are in the "failed" or "connecting" or "checking" state.  */
-	case pc.iceConnectionState == ICEConnectionStateDisconnected:
+	case iceConnectionState == ICEConnectionStateDisconnected:
 		connectionState = PeerConnectionStateDisconnected
 
 	// All RTCIceTransports and RTCDtlsTransports are in the "connected", "completed" or "closed"
 	// state and at least one of them is in the "connected" or "completed" state.
-	case pc.iceConnectionState == ICEConnectionStateConnected && pc.dtlsTransport.State() == DTLSTransportStateConnected:
+	case iceConnectionState == ICEConnectionStateConnected && dtlsTransportState == DTLSTransportStateConnected:
 		connectionState = PeerConnectionStateConnected
 
 	//  Any of the RTCIceTransports or RTCDtlsTransports are in the "connecting" or
 	// "checking" state and none of them is in the "failed" state.
-	case pc.iceConnectionState == ICEConnectionStateChecking && pc.dtlsTransport.State() == DTLSTransportStateConnecting:
+	case iceConnectionState == ICEConnectionStateChecking && dtlsTransportState == DTLSTransportStateConnecting:
 		connectionState = PeerConnectionStateConnecting
 	}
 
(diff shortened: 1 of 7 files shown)


peerconnection.go
```diff
@@ -1875,6 +1819,90 @@ func (pc *PeerConnection) GetStats() StatsReport {
 	return statsCollector.Ready()
 }
 
+// Start all transports. PeerConnection now has enough state
+func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, remoteUfrag, remotePwd, fingerprint, fingerprintHash string) {
```
Sean-Der: This was just moved out of `SetRemoteDescription` no actual code changes.

Sean-Der:
Ok this is all done!

Users are now able to be notified about `PeerConnectionState` going to failed by listening to this event. After this lands will cut another PR to update all the examples.

This PR is already bulky enough so don't want to make reviewers go through a bunch of mechanical changes.

peerconnection_test.go
```diff
@@ -386,6 +388,14 @@ func TestPeerConnection_EventHandlers(t *testing.T) {
 			wg.Done()
 		})
 	})
+	pcOffer.OnConnectionStateChange(func(PeerConnectionState) {
```
at-wat:
```suggestion
	pcOffer.OnConnectionStateChange(func(s PeerConnectionState) {
              if s2 := pcOffer.ConnectionState(); s != s2 {
                      t.Errorf("State in callback argument is different from ConnectionState(): arg=%s, state=%s", s.String(), s2.String())
              }
```
results
```
$ go test . -run TestPeerConnection_EventHandlers
--- FAIL: TestPeerConnection_EventHandlers (2.01s)
    peerconnection_test.go:393: State in callback argument is different from ConnectionState(): arg=connected, state=new
    peerconnection_test.go:393: State in callback argument is different from ConnectionState(): arg=connected, state=new
FAIL
FAIL	github.com/pion/webrtc/v2	2.037s
```
`pc.connectionState` is not updated and `pc.onConnectionStateChangeHandler` called duplicately.

peerconnection.go
```diff
@@ -501,9 +483,51 @@ func (pc *PeerConnection) createICEGatherer() (*ICEGatherer, error) {
...
+	case pc.iceConnectionState == ICEConnectionStateConnected && pc.dtlsTransport.State() == DTLSTransportStateConnected:
+		connectionState = PeerConnectionStateConnected
+
+	//  Any of the RTCIceTransports or RTCDtlsTransports are in the "connecting" or
+	// "checking" state and none of them is in the "failed" state.
+	case pc.iceConnectionState == ICEConnectionStateChecking && pc.dtlsTransport.State() == DTLSTransportStateConnecting:
+		connectionState = PeerConnectionStateConnecting
+	}
+
+	if pc.connectionState == connectionState {
```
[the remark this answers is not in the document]
Sean-Der: Oops nice catch, fixed!

peerconnection_test.go
Sean-Der: Great catch! Thanks for the test suggestion, added it

peerconnection.go
```diff
@@ -526,6 +551,7 @@ func (pc *PeerConnection) createICETransport() *ICETransport {
 			return
 		}
 		pc.onICEConnectionStateChange(cs)
+		pc.updateConnectionState()
```
at-wat: This is called from https://github.com/pion/webrtc/blob/81d03573bcaa2b824c3521d73f16c6c132b6b689/icetransport.go#L188-L195
from https://github.com/pion/webrtc/blob/81d03573bcaa2b824c3521d73f16c6c132b6b689/icetransport.go#L84-L91
from `ice.Agent.updateConnectionState`
https://github.com/pion/ice/blob/bd60240fdc34f14ed511109467a2c930c0b6467e/agent.go#L508-L513

It's called in goroutine and can be parallel.
In very rare case, `pc.iceConnectionState` can be updated twice and then `pc.updateConnectionState()` called twice.
In this case, the first state change may be skipped.

Passing `cs` as an argument of `updateConnectionState()` would be more safe?
(I'm not sure it can be happens actually.)

Sean-Der: Done! https://github.com/pion/webrtc/pull/830/commits/6ffbea04a93a9c3231fd068a64511936268f2f82

Sean-Der: I think that makes sense, also just nice to make the function a little simpler.

Sean-Der:
Thank you so much for the reviews @at-wat!

w/e you get a chance mind approving and I will be able to merge :)

Sean-Der:
Fantastic! That is great to hear

Issue #207, opened 2018-10-23
remoteDescription is already defined, SetRemoteDescription can only be called once


ast:
Not sure if this is a bug or a feature but according to [MDN documentation][1] calling setRemoteDescription after a connection is already in place should trigger renegotiation. However this is not the case with pions/webrtc and I get:

    remoteDescription is already defined, SetRemoteDescription can only be called once

[1]: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setRemoteDescription

mxmCherry:
Related to / depends on https://github.com/pions/webrtc/issues/70

(just to keep track)

mxmCherry:
I'm curious what could happen if this check would be removed before (re) negotiation is implemented.

Theoretically, it should work just fine, right? Negotiation handling will be up to consumer, but nothing should really break? Probably, adding some logging would be nice, just to make sure consumers know what they do.

Or this rabbit hole is way too deep (just removing a check wouldn't be enough)?

I'd vote to have a possibility to use this lib for such case right now instead of waiting for proper re-negotiation implementation.

ciceroverneck:
Is this not planned for the next release anymore?

Sean-Der:
Hey I do want to do this @ciceroverneck!

Is this something that is blocking you short term? I am going to be releasing a new version of `pion/dtls` this weekend, and then moving back to `pion/webrtc`.

I don't see this happening in the short term, but would be happy to move faster if it helps.

thanks

guilhermehubner:
@Sean-Der it would be amazing if this fix came with the new version of `pion/webrtc`, it is blocking us to deliver our SFU system. There's a work around, it's not optimised though so not production ready :disappointed:

Sean-Der:
@guilhermehubner that is exciting, glad to hear you are using Pion :)

I started a branch with this work! I am thinking this is ~week of work. When it is ready for alpha testing will update this thread so everyone can play with it!

thanks for the patience everyone. Also if you get a chance join [Slack](https://pion.ly/slack). I work on w/e issue people are most excited for, so giving feedback/chatting is the best way for me to know.

Sean-Der:
I have finished implement Adding tracks! I will be working on removal tomorrow.

Currently I have only tested against Pion. I am sure bugs will show up when testing against FireFox/Chromium. If anyone is feeling brave testing out my branch would be really helpful!

PR #836, opened 2019-09-17
Add PeerConnection renegotiate tests

Sean-Der:
These will pass in WASM, but fail in Go for now. Later we will work on
making these tests pass as we implement re-negotiation

Relates to #207


Sean-Der:
Switching to draft

PR #842, opened 2019-09-24
Remove unnecessary `defer`s

ZizhengTai:
In some simple cases `defer` doesn't simplify the code and is probably not worth it for its negative impact on performance. I only removed `defer`s from functions that obviously can't go wrong between locking and unlocking.

This might or might not be a tiny bit less readable, depending on the reader's personal preference, but should bring actual performance improvement.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -225,8 +225,8 @@ func (pc *PeerConnection) onSignalingStateChange(newState SignalingState) {
 // channel message arrives from a remote peer.
 func (pc *PeerConnection) OnDataChannel(f func(*DataChannel)) {
 	pc.mu.Lock()
-	defer pc.mu.Unlock()
 	pc.onDataChannelHandler = f
+	pc.mu.Unlock()
 }
 
 // OnICECandidate sets an event handler which is invoked when a new ICE
@@ -245,8 +245,8 @@ func (pc *PeerConnection) OnICEGatheringStateChange(f func(ICEGathererState)) {
 // arrives from a remote peer.
 func (pc *PeerConnection) OnTrack(f func(*Track, *RTPReceiver)) {
 	pc.mu.Lock()
-	defer pc.mu.Unlock()
 	pc.onTrackHandler = f
+	pc.mu.Unlock()
 }
 
 func (pc *PeerConnection) onTrack(t *Track, r *RTPReceiver) {
@@ -264,8 +264,8 @@ func (pc *PeerConnection) onTrack(t *Track, r *RTPReceiver) {
 // when an ICE connection state is changed.
 func (pc *PeerConnection) OnICEConnectionStateChange(f func(ICEConnectionState)) {
 	pc.mu.Lock()
-	defer pc.mu.Unlock()
 	pc.onICEConnectionStateChangeHandler = f
+	pc.mu.Unlock()
 }
 
 func (pc *PeerConnection) onICEConnectionStateChange(cs ICEConnectionState) {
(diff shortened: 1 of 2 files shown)


peerconnection.go
```diff
@@ -1242,9 +1243,9 @@ func (pc *PeerConnection) GetReceivers() []*RTPReceiver {
 // GetTransceivers returns the RTCRtpTransceiver that are currently attached to this RTCPeerConnection
 func (pc *PeerConnection) GetTransceivers() []*RTPTransceiver {
 	pc.mu.Lock()
```
ZizhengTai: Can this be an `RLock()`?

Sean-Der:
I had no idea defer was so expensive!

Sean-Der:
@ZizhengTai are you getting this via profiling? Or some other tooling, do you see anything else stick out performance wise

ZizhengTai:
> @ZizhengTai are you getting this via profiling? Or some other tooling, do you see anything else stick out performance wise

Here are some references:
https://go.googlesource.com/proposal/+/refs/heads/master/design/34481-opencoded-defers.md
https://medium.com/i0exception/runtime-overhead-of-using-defer-in-go-7140d5c40e32

To be honest the impact on this particular codebase could be (and probably is) negligible. My reasoning is just that it's usually good practice to stick to the least powerful / most performant tools _when they don't make the code less readable_. It's a good thing anyway if we can shave off a few hundred nanoseconds here and there while keeping the codebase clean. What do you think?

hugoArregui:
I think we need to consider this carefully, maybe you can share the use cases you are measuring perf against? I totally agree sometimes `defer` impacts perf in a significant way but I think most of the cases `defer` is not important because of the context, for example: `OnDataChannel` is likely going to be called just once for the whole app lifecycle, it's a very infrequent operation so IMHO for readability and security reasons I would prefer to keep the `defer` there. That been said I think you would have a strong argument against `defer` for things like `ICEConnectionState`, since we cannot assume much about its usage.

ZizhengTai:
Hey @hugoArregui, you are absolutely right that in some cases the perf impact is not important at all. I'm not making this changes because of my own benchmark; it's just that given that `defer` is objectively slower than a manual unlock, I see no reason not to do it if it doesn't harm readability. And about specific examples such as `OnDataChannel`, personally I think

```go
pc.mu.Lock()
pc.onDataChannelHandler = f
pc.mu.Unlock()
```
conveys the intent clearer than
```go
pc.mu.Lock()
defer pc.mu.Unlock()
pc.onDataChannelHandler = f
```
since the lines are executed in the order they appear. I see no real advantage of using `defer` in this case.

hugoArregui:
Yes, your point is well taken, I guess this is very subjective, I felt the same way first time I saw `defer`. I think the readability is relative to the point of view of the reader.

- If you are reading the code just to understand it's behavior I think you are right, the no-defer version is at the very least as readable as the other one, maybe even more readable (although I would argue that the defer pattern is so deeply ingrained into gophers brains that this is actually not true :-D).
- If you are reading the code trying to debug a locking problem, I think having a `defer` there is useful to see how no matter what may happen in the code at the end of the call the mutex will be unlocked.

Those are my 2 cents though, I'm not against the change.

Sean-Der:
Thinking this over I don't think we should land this change. I don't disagree with the change at all (I like that it removes some uncertainty), but we can't make changes without data.

@ZizhengTai if you are able to write a linter (or even better get something into `golangci-lint`!) I would love to make this change. My concern is that someone in the future will undo this work, we have nothing forcing a style either way.

Sean-Der:
I really appreciate what you are doing. This is super valuable work, and I encourage you to rewrite/examine everything.

Anything that can cause bugs, hurts performance or that is just hard to read should go! Just need to make sure it stays gone :)

ZizhengTai:
That makes sense, since the impact of this should be negligible anyway, I'll close this PR then.

Sean-Der:
If you ever feel strongly about anything I always would love to discuss further.

Your long term involvement in the project is more important to me then a single change!

PR #844, opened 2019-09-25
Add InterfaceFilter to SettingEngine

### LINKED ISSUE #843: Add interface filtering to SettingEngine
Some users don't want certain interfaces to be used for ICE.

They may know ahead of time certain interfaces won't work. One users has a large number of interfaces, and spinning up that amount of routines to listen causes issues.

Others have reported security concerns from exposing internal topologies via candidates.

Sean-Der:
This allows users to filter interfaces from ICE gathering

Resolves #843


enobufs:
FYI. NAT1To1IPs option incoming... ;)

Sean-Der:
> FYI. NAT1To1IPs option incoming... ;)

Woot, that's fantastic. Hopefully we can push some of these options into the RFC soon!

Issue #778, opened 2019-08-08
Add vnet to SettingEngine


Sean-Der:
We should add the SettingEngine to the public API. This will allow users to simulate different network types, and it would be great for learning.

we should also make a bucket of examples to show people how it works!

enobufs:
Consider fixing also #872

PR #845, opened 2019-09-25
Support 1:1 NAT option

### LINKED ISSUE #835: Support 1:1 NAT option
## Motivation
Pion could be used in  C/S topology, one end is a client and the other end is a server (just like Janus G/W does). But in this case, the server must have a routable (public) IP address. Many server environments may not have access to public IP address to bind to the local interface.

Janus provides the following option. Pion could provide something equivalent.
```
    -1, --nat-1-1=ip              Public IP to put in all host candidates,
                                  assuming a 1:1 NAT is in place (e.g., Amazon
                                  EC2 instances, default=none)
```

Use STUN server for the server-side nodes.

enobufs:
Resolves #835

Added these parameters configured via SettingEngine
* NAT1To1IPs []string - an array external IP addresses
* NAT1To1IPCandidate - target candidate the external IP address is used for ("host" or "srflx")

See #835 for details.

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -62,7 +62,7 @@ func NewICEGatherer(
 	networkTypes []NetworkType,
 	interfaceFilter func(string) bool,
 	nat1To1IPs []string,
-	nat1To1IPCandidate string,
+	nat1To1IPCandidateType ICECandidateType,
 	opts ICEGatherOptions,
 ) (*ICEGatherer, error) {
 	var validatedServers []*ice.URL
@@ -83,14 +83,14 @@ func NewICEGatherer(
 		candidateTypes = append(candidateTypes, ice.CandidateTypeRelay)
 	}
 
-	var nat1To1IPCandidateType ice.CandidateType
-	switch nat1To1IPCandidate {
-	case "host":
-		nat1To1IPCandidateType = ice.CandidateTypeHost
-	case "srflx":
-		nat1To1IPCandidateType = ice.CandidateTypeServerReflexive
+	var nat1To1CandiTyp ice.CandidateType
+	switch nat1To1IPCandidateType {
+	case ICECandidateTypeHost:
+		nat1To1CandiTyp = ice.CandidateTypeHost
+	case ICECandidateTypeSrflx:
+		nat1To1CandiTyp = ice.CandidateTypeServerReflexive
 	default:
-		nat1To1IPCandidateType = ice.CandidateTypeUnspecified
+		nat1To1CandiTyp = ice.CandidateTypeUnspecified
 	}
 
 	return &ICEGatherer{
@@ -113,7 +113,7 @@ func NewICEGatherer(
 		relayAcceptanceMinWait:    relayAcceptanceMinWait,
 		interfaceFilter:           interfaceFilter,
 		nat1To1IPs:                nat1To1IPs,
-		nat1To1IPCandidateType:    nat1To1IPCandidateType,
+		nat1To1IPCandidateType:    nat1To1CandiTyp,
 	}, nil
 }
 
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -38,6 +38,8 @@ type ICEGatherer struct {
 	log                       logging.LeveledLogger
 	networkTypes              []NetworkType
 	interfaceFilter           func(string) bool
+	nat1To1IPs                []string
+	nat1To1IPCandidateType    ice.CandidateType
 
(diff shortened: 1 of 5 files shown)


enobufs:
I will do https://github.com/pion/webrtc/issues/778 after this next!

icegatherer.go
```diff
@@ -79,6 +83,16 @@ func NewICEGatherer(
 		candidateTypes = append(candidateTypes, ice.CandidateTypeRelay)
 	}
 
+	var nat1To1IPCandidateType ice.CandidateType
+	switch nat1To1IPCandidate {
+	case "host":
+		nat1To1IPCandidateType = ice.CandidateTypeHost
+	case "srflx":
```
Sean-Der: `iceCandidateTypeSrflxStr`

icegatherer.go
```diff
@@ -79,6 +83,16 @@ func NewICEGatherer(
 		candidateTypes = append(candidateTypes, ice.CandidateTypeRelay)
 	}
 
+	var nat1To1IPCandidateType ice.CandidateType
+	switch nat1To1IPCandidate {
+	case "host":
```
Sean-Der: `iceCandidateTypeHostStr`

settingengine.go
```diff
@@ -114,3 +116,28 @@ func (e *SettingEngine) SetNetworkTypes(candidateTypes []NetworkType) {
...
+// Two types of candidates are supported:
+// - "host": The public IP address will be used for the host candidate in the SDP.
+// - "srflx": A server reflexive candidate with the given public IP address will be added
+// to the SDP. If you choose "host", then the private IP address won't be advertised with
+// the peer. Also, this option cannot be used along with mDNS.
+// If you choose "srflx", it simply adds a server reflexive candidate with the public IP.
+// The host candidate is still available along with mDNS capabilities unaffected.
+// Please note that you cannot give STUN server URL at the same time. It will result in
+// an error otherwise.
+func (e *SettingEngine) SetNAT1To1IPCandidate(candidate string) {
```
Sean-Der: I would use `ICECandidateType` and then assert that only `ICECandidateTypeSrflx` and `ICECandidateTypeHost` are accepted

enobufs:
> is there any reason not to use one call with both parameters in settingsengine?

No! I TOTALLY forgot our original plan! Thank you for noticing it! I will fix it!

settingengine.go
enobufs: I had thought for JavaScript support (via WASM), use of string was better. But ICECandidateTypeXxx has already been used, I will switch to it as suggested!
Thanks!

icegatherer.go
enobufs: I will fix it.

icegatherer.go
enobufs: I will fix it.

enobufs: Fixed at 70babf2

icegatherer.go
enobufs: Fixed at 70babf2

settingengine.go
enobufs: Fixed at 70babf2

enobufs:
@seppo0010 Now it looks like this:
```
func (e *SettingEngine) SetNAT1To1IPs(ips []string, candidateType ICECandidateType)
```

settingengine.go
```diff
@@ -114,3 +116,31 @@ func (e *SettingEngine) SetNetworkTypes(candidateTypes []NetworkType) {
...
+// to the SDP.
+//
+// Please note that if you choose ICECandidateTypeHost, then the private IP address
+// won't be advertised with the peer. Also, this option cannot be used along with mDNS.
+//
+// If you choose ICECandidateTypeSrflx, it simply adds a server reflexive candidate
+// with the public IP. The host candidate is still available along with mDNS
+// capabilities unaffected. Also, you cannot give STUN server URL at the same time.
+// It will result in an error otherwise.
+func (e *SettingEngine) SetNAT1To1IPs(ips []string, candidateType ICECandidateType) {
```
Sean-Der: I would rename this to `SetNAT1To1` maybe? Since it takes IPs and candidateType. I don't feel strongly though!

seppo0010:
@enobufs I think the `createAgent` is not honoring the settings

something like this around line 143 is missing:
```
NAT1To1IPs: g.nat1To1IPs,
NAT1To1IPCandidateType: g.nat1To1IPCandidateType,
```

Issue #748, opened 2019-07-22
Finish DataChannelInit


Sean-Der:
We have a few options that haven't been implemented yet.

* [ ] Protocol
* [ ] Negotiated
* [x] ~Priority~ -- Removed from WebRTC

Sean-Der:
Priority was removed from webrtc-pc so I deleted it.

Will be implementing `Negotiated` and `Protocol` now!

PR #848, opened 2019-09-26
Remove RTCDataChannel priority

Sean-Der:
This has been removed from webrtc-pc[0]

Relates to #748

[0] https://github.com/w3c/webrtc-pc/issues/2258


enobufs:
Marked as "feature at risk" and actual "removal" are two different things... Though this isn't implemented, are you sure you want to make a breaking change?

Sean-Der:
@enobufs I double checked and we are the only implementors of this.

I don't feel strongly, but I do think we are in danger of misleading users. They could write code that depends on this feature, and then realize it doesn't exist anywhere else.

 I think that is more of a frustration for users than depending on it in Pion right now.

PR #852, opened 2019-09-26
Honor nat 1to1 config when creating agent

seppo0010:
Forgotten LOCs in https://github.com/pion/webrtc/pull/845


enobufs:
@Sean-Der we should reconsider the commit message policy as this has been getting in the way for people who are trying to contribute to pion out of their precious time. Any thoughts?

PR #909, opened 2019-11-14
Properly generate DataChannel streamId

### LINKED ISSUE #908: DataChannel ID generation doesn't take Client/Server role into consideration.
Currently it is possible to get collisions on StreamIDs if both sides create a stream at once. 

https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13
```
When one side wants to open a channel using out-of-band negotiation,
   it picks a stream.  Unless otherwise defined or negotiated, the
   streams are picked based on the DTLS role (the client picks even
   stream identifiers, the server odd stream identifiers).  However, the
   application is responsible for avoiding collisions with existing
   streams.  If it attempts to re-use a stream which is part of an
   existing data channel, the addition MUST fail.  In addition to
   choosing a stream, the application SHOULD also determine the options
   to use for sending messages.  The application MUST ensure in an
   application-specific manner that the application at the peer will
   also know the selected stream to be used, and the options for sending
   data from that side.
```

Sean-Der:
Before we computed DataChannel IDs before signaling, this
is incorrect because IDs must take into account if we are
running an DTLS Client or Server.

This updates the DataChannel ID generation code to take this
into account before generating a streamId.

Resolves #908

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -50,12 +50,6 @@ type PeerConnection struct {
 
 	rtpTransceivers []*RTPTransceiver
 
-	// DataChannels
-	dataChannels          map[uint16]*DataChannel
-	dataChannelsOpened    uint32
-	dataChannelsRequested uint32
-	dataChannelsAccepted  uint32
-
 	onSignalingStateChangeHandler     func(SignalingState)
 	onICEConnectionStateChangeHandler func(ICEConnectionState)
 	onConnectionStateChangeHandler    func(PeerConnectionState)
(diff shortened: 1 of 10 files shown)


Sean-Der:
@hugoArregui How do you feel about Stats+DataChannels testing? Anything I should manually test!

@epes the example you gave works now!

If either/both of you feel good about this feel free to approve+merge. Also happy to add more testing, but things were actually really well covered before! Just that the call to `pc.generateDataChannelID` was hardcoded is the only thing that killed us.

peerconnection.go
```diff
@@ -1867,11 +1816,12 @@ func (pc *PeerConnection) GetStats() StatsReport {
 		Timestamp:             statsTimestampNow(),
 		Type:                  StatsTypePeerConnection,
 		ID:                    pc.statsID,
-		DataChannelsOpened:    pc.dataChannelsOpened,
+		DataChannelsOpened:    pc.sctpTransport.dataChannelsOpened,
```
hugoArregui: I think sctpTransport may be nil

Sean-Der: you rock, thanks for catching that :)

Issue #873, opened 2019-10-10
Deadlock while peerConnection.GetStats()


Filatoff73:
### Your environment.
  - Version: *2.1.6*
  - windows 10
  
I am try to get GetStats() from peerConnection. My project was based on the webrtc/examples/save-to-disk. But the function GetStats() freezes. During debugging, I found that this part of the code hangs:
if pc.sctpTransport! = nil {
pc.sctpTransport.collectStats (statsCollector)
}
The mutex in "func (r *SCTPTransport) collectStats" r.lock.Lock () is trying to lock, but this does not work, since it was previously locked in the "func (r * SCTPTransport) Start (remoteCaps SCTPCapabilities) error" method.

Sean-Der:
Good catch!  I will check this out now and see what I can figure out :)

Thanks for using Pion!

Sean-Der:
@Filatoff73 as a temporary fix I would create a single data channel on the offering side.

The issue is that we are starting the SCTP subsystem even if the other side doesn't support, and we hang forever. I will be able to fix this in the next 72 hours, thanks for reporting!

Filatoff73:
@Sean-Der thanks for the help!

Sean-Der:
Fixed! This was fixed by ce4071b96c6c67e55a65ef23f878d571c4fe1216

Sorry for the delay and really appreciate you reporting this @Filatoff73 if anything else comes up always here to help!

PR #911, opened 2019-11-21
Match W3C when closing

Sean-Der:
PeerConnection.Close() doesn't match the W3C RFC on closing,
re-order the transports so that they close in the proper order.

In the future we will use the DTLS Close alert and RTCP Bye to
better communicate shutdown

----

I just threw recently active people on the review, no worries if you don't have the bandwidth! I am fixing issues to help with using `croc`


Sean-Der:
I need to make a change to `pion/dtls` to notify us when it is closed. Not sure how I will do that yet, but when that feature lands I will create another PR that covers that as well!

Sean-Der:
This also will fix https://github.com/pion/webrtc/issues/873

PR #912, opened 2019-11-21
Only create ICEGather via API

### LINKED ISSUE #872: introduce webrtc.ICEGathererConfig
Now, webrtc.NewICEGatherer() takes **17 arguments** - lol
Every time we introduce a parameter for ice.Agent:
* We break API (it is an exposed function!) every time!
* Has to modify the function signature many places
* Code is unreadable (hard to tell what we are passing to the method)

Let's use a struct for it.

Sean-Der:
Use values directly from SettingValues instead of just copying
when calling NewICEGatherer. This greatly reduces the LoC and makes
the public API a little cleaner.

Resolves #872

### DIFF
--- a/icegatherer_test.go
+++ b/icegatherer_test.go
@@ -6,8 +6,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/pion/ice"
-	"github.com/pion/logging"
 	"github.com/pion/transport/test"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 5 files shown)


icegatherer_test.go
```diff
@@ -114,92 +111,3 @@ func TestICEGather_LocalCandidateOrder(t *testing.T) {
 
 	assert.NoError(t, gatherer.Close())
 }
-
-func TestNewICEGatherer_NAT1To1IP(t *testing.T) {
```
Sean-Der: These tests just assert that values are properly copied from `SettingEngine` -> `ICEGatherer`. I don't think we need them anymore with the indirection removed!

PR #932, opened 2019-12-09
Add option to detect partition head in SampleBuilder

at-wat:
Add functional option to set PartitionHeadChecker to detect the beginning of the frame partition.
It avoids dropping the first packet. (which is often a keyframe, so, loosing the first packet sometimes causes large delay to start.)

requires https://github.com/pion/rtp/pull/26 (ready)

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -25,11 +25,18 @@ type SampleBuilder struct {
 	isContiguous     bool
 	lastPopSeq       uint16
 	lastPopTimestamp uint32
+
+	// Interface that checks whether the packet is the first fragment of the frame or not
+	partitionHeadChecker rtp.PartitionHeadChecker
 }
 
 // New constructs a new SampleBuilder
-func New(maxLate uint16, depacketizer rtp.Depacketizer) *SampleBuilder {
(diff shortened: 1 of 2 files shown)


pkg/media/samplebuilder/samplebuilder.go
```diff
@@ -24,11 +24,18 @@ type SampleBuilder struct {
 	isContiguous     bool
 	lastPopSeq       uint16
 	lastPopTimestamp uint32
+
+	// Interface that checks whether the packet is keyframe or not
```
at-wat:
```suggestion
	// Interface that checks whether the packet is the first fragment of the frame or not
```

at-wat:
@Sean-Der @kc5nra please take a look when you have time.
I think it would be nice that rtpengine will have similar logic to avoid packet drop as possible.

Sean-Der:
@at-wat LGTM!

Really great improvement. In general the 'Media API' needs to dramatically get better, we have no congestion control or loss handling :(

I am trying to close out some easy issues, cut a release and then we can start moving towards making the breaking changes needed for this. Are you interested in being involved/working together on it?

at-wat:
@Sean-Der Yes! I'm sometimes working on codes to handle media data in my product app. Making well-reusable library will make it more easy and happy.

Issue #838, opened 2019-09-19
SettingsEngine.DetachDataChannels() makes attached data channels unusable.


AeroNotix:
I'm not sure if this is intended or not but when you set `.DetachDataChannels()` in the SettingsEngine you _must_ detach the data channel for it to be usable at all.

### What did you expect?

`webrtc.SettingsEngine{}.DetachDataChannels()` would still allow attached data channels to function

### What happened?

`webrtc.SettingsEngine{}.DetachDataChannels()` means that you _must_ detach the datachannel to interact with it.

Sean-Der:
This is by design, but we should probably add some logging/assertion if we don't detach!

```
DetachDataChannels enables detaching data channels. When enabled data channels have to be detached in the OnOpen callback using the DataChannel.Detach method.
```

After the `OnOpen` we should check if it was properly detached and throw an error message if it wasn't. Super easy PR if someone wants to grab that!

PR #936, opened 2019-12-10
Fix race in checkDetachAfterOpen

Sean-Der:
checkDetachAfterOpen incorrectly accesses detachCalled without holding the lock

Relates to #838

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -200,8 +200,11 @@ func (d *DataChannel) Transport() *SCTPTransport {
 // After onOpen is complete check that the user called detach
 // and provide an error message if the call was missed
 func (d *DataChannel) checkDetachAfterOpen() {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
 	if d.api.settingEngine.detach.DataChannels && !d.detachCalled {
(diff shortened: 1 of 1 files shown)


langep:
As far as I understand, this change should fix the data race but the error log might still show nondeterministically if Detach() is called outside OnOpen or in a goroutine inside OnOpen that runs longer than the OnOpen handler itself. Do you think that is a concern?

Sean-Der:
@langep that's a good point!

I do think we need to block until `onOpen` is complete, I think that is where standard users of detach will start things.

We can add more checks in other places, maybe when we recv a message and Detach hasn't been called? As long as it doesn't cause any performance issues.

langep:
Maybe we can make it a warning message. If the user forgets to call Detach() they are warned that they need to call Detach() to get a functional data channel. More advanced users who somehow detach after OnOpen still get the warning (sometimes) but can ignore it.

Sean-Der:
@langep updated! I think we should push users to call detach in the `OnOpen` though.

I don't use the feature myself, but is there any value to using it elsewhere?

langep:
@at-wat I assume you mean the test should be updated to show the intended usage of Detach() inside OnOpen?

PR #943, opened 2019-12-14
Investigating flaky tests (DON'T MERGE)

at-wat:
Testing fix on dtls. (https://github.com/pion/dtls/pull/160)


at-wat:
### run 1
no error

### run 2
```
--- FAIL: TestDataChannel_Open (0.43s)
    --- FAIL: TestDataChannel_Open/handler_should_be_called_once (0.43s)
        datachannel_test.go:23: 
            	Error Trace:	datachannel_test.go:23
            	            				datachannel_test.go:89
            	Error:      	Received unexpected error:
            	            	EOF
            	Test:       	TestDataChannel_Open/handler_should_be_called_once
```
(no routine leak)

### run 3

at-wat:
## with #945
### 1st run
no error
### 2nd run
no error
### 3rd run

Sean-Der:
I wonder if 36b02632d17e9b8ce67d3205ae89d5bfd754517f could have caused the SCTP issues to start. 

Before we stopped ICE first, and then closed Transports. Now we stop ICE last, this order of operations could have caused some behavior change! I haven't investigated fully.

Also we might able to ignore the `EOF` error? If we call `Close` and it returns an `EOF` that doesn't seem fatal, maybe we shouldn't surface it to the user? I am not fully sure yet, thank you so much for all this work it makes a HUGE difference :)

at-wat:
@Sean-Der The `EOF` seems to be came from `dtls.Conn.Close()` under `sctp.Association`.

`dtls.Conn` stores error from `Read` and return it on `Close()`.
https://github.com/pion/dtls/blob/master/conn.go#L529-L531
And, `c.stopWithError(err)` from `inboundLoop` and from `Close` is concurrently called like:
1. `c.stopWithError(ErrConnClosed)` called from `dtls.Conn.Close` registers `ErrConnClosed`
2. `dtls.Conn.nextConn` is closed
3. `dtls.Conn.nextConn.Read` returns `EOF`
4. `c.stopWithError(err)` called from `dtls.Conn.inboundLoop` registers `EOF`
5. `c.getConnErr()` called from `dtls.Conn.Close` returns `EOF`
6. `dtls.Conn.Close` returns `EOF`

I think `dtls.Conn` should ignore any error after `Close`.

at-wat:
All fixes are now in master🎊

PR #951, opened 2019-12-17
Fix seqnum distance in SampleBuilder

at-wat:
Fix overflow and underflow handling.


Sean-Der:
Sorry missed this! You are on fire with fixing stuff lately, thank you so much @at-wat

PR #960, opened 2019-12-28
Fix codecov config

at-wat:
- Rename config file since .codecov.yml is ignored on codecov server.
- Remove unused field.

related: #944

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,10 +2,10 @@ coverage:
   status:
     project:
       default:
-        threshold: 2
+        # Allow decreasing 2% of total coverage to avoid noise.
+        threshold: 2%
     patch:
       default:
-        against: parent
         target: 70%
 
 ignore:


Sean-Der:
@at-wat (or anyone else) have you seen tools to synchronize this stuff across repos?

Would be nice if we can get it so all this stuff is kept in sync across repos. I haven't investigated enough though.

at-wat:
I don't know unfortunately.
codecov config should be on the root of the repo, so git submodule don't help.
Putting such files in a new repo and opening PR on each repo using GitHub Actions might be not very hard.

at-wat:
@Sean-Der I made a GitHub Action to create PR on multiple repos to update such files.
https://github.com/at-wat/assets-sync-action
Cross-repository PR requires a custom GitHub personal access token instead of GitHub Actions one. Making a bot account for pion organization would be good to automate this.

PR #965, opened 2020-01-05
Sdp support transport-cc

adwpc:
Sdp support transport-cc

#### Reference issue
Fixes #...

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -210,18 +210,14 @@ func NewRTPVP8Codec(payloadType uint8, clockrate uint32) *RTPCodec {
 }
 
 // NewRTPVP8CodecExt is a helper to create an VP8 codec
-func NewRTPVP8CodecExt(payloadType uint8, clockrate uint32) *RTPCodec {
+func NewRTPVP8CodecExt(payloadType uint8, clockrate uint32, rtcpfb []RTCPFeedback) *RTPCodec {
 	c := NewRTPCodecExt(RTPCodecTypeVideo,
 		VP8,
 		clockrate,
 		0,
 		"",
 		payloadType,
-		[]RTCPFeedback{
-			RTCPFeedback{
-				Type: "transport-cc",
-			},
-		},
+		rtcpfb,
 		&codecs.VP8Payloader{})
 	return c
 }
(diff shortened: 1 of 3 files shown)


mediaengine.go
```diff
@@ -209,6 +209,23 @@ func NewRTPVP8Codec(payloadType uint8, clockrate uint32) *RTPCodec {
...
 
+// NewRTPVP8CodecExt is a helper to create an VP8 codec
+func NewRTPVP8CodecExt(payloadType uint8, clockrate uint32) *RTPCodec {
+	c := NewRTPCodecExt(RTPCodecTypeVideo,
+		VP8,
+		clockrate,
+		0,
+		"",
+		payloadType,
+		[]RTCPFeedback{
```
Sean-Der: Can you have `RTCPFeedback` be passed as a function argument? In case we want to support other congestion control algorithms.

mediaengine.go
```diff
@@ -233,6 +250,23 @@ func NewRTPH264Codec(payloadType uint8, clockrate uint32) *RTPCodec {
 	return c
 }
 
+// NewRTPH264CodecExt is a helper to create an H264 codec
+func NewRTPH264CodecExt(payloadType uint8, clockrate uint32) *RTPCodec {
+	c := NewRTPCodecExt(RTPCodecTypeVideo,
+		H264,
+		clockrate,
+		0,
+		"level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f",
```
Sean-Der: Can you make this a `const` we shouldn't hardcode this, but I can fix that later!

peerconnection.go
```diff
@@ -1611,6 +1611,7 @@ func (pc *PeerConnection) addTransceiverSDP(d *sdp.SessionDescription, midValue
 	media := sdp.NewJSEPMediaDescription(t.kind.String(), []string{}).
 		WithValueAttribute(sdp.AttrKeyConnectionSetup, dtlsRole.String()).
 		WithValueAttribute(sdp.AttrKeyMID, midValue).
+		WithTransportCCExtMap().
```
Sean-Der: Can you have this be pulled from the `MediaEngine` instead? So you insert a codec (and it has transport-cc on it) and then this pulls the value.

We don't want to enable `transport-cc` if the user doesn't support it.

mediaengine.go
adwpc: OK

mediaengine.go
adwpc: Maybe you fix that better

peerconnection.go
adwpc: fixed ,will add 'a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01' when user add codec that support 'a=rtcp-fb:100 transport-cc'

mediaengine.go
Sean-Der: agree! I will break this in `/v3` I need to revisit how things are done with setting outbound tracks.

PR #967, opened 2020-01-05
Expose an API to get a list of registered RTPCodec from PeerConnection

### LINKED ISSUE #966: Expose an API to get a list of registered RTPCodec from PeerConnection
## Description
As of writing this, there's no way to get a list of registered `RTPCodec` from outside the package. This is because the constructed `mediaEngine` is stored within `api`, and `api` is a private member of `PeerConnection`.

So, instead of exposing the whole underlying `api` or `mediaEngine`, we should expose a new method under `PeerConnection` to get the list of registered `RTPCodec`.

## Why?
This is really useful for people who want to develop a library on top of pion. A very concrete example is [https://github.com/pion/mediadevices](pion/mediadevices) use case. When users want to specify what codec should be used, the `mediadevices` library should be able to detect if the codec is registered and retrieve the `RTPCodecType` from the registered codec to create a track. Below shows an example of the use case:

```
s, err := mediaDevices.GetUserMedia(mediadevices.MediaStreamConstraints{
		Video: mediadevices.VideoTrackConstraints{
			Enabled: true,
			Width:   800, // Optional. This is just an ideal value.
			Height:  480, // Optional. This is just an ideal value.
			Codec:   webrtc.H264, <-- I need to check if this codec is registered and get its type for creating a track
		},
})
```

lherman-cs:
Add a new public method, GetRegisteredRTPCodecs, under PeerConnection to allow access to get the registered list of RTPCodec from the underlying MediaEngine for developers who want to build libraries on top of pion.

Resolves #966

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Ilya Mayorov](https://github.com/faroyam)
 * [Patrick Lange](https://github.com/langep)
 * [cyannuk](https://github.com/cyannuk)
+* [Lukas Herman](https://github.com/lherman-cs)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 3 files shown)


Sean-Der:
@lherman-cs you can take this right to master if you want!

I was only thinking `/v3` for breaking changes, but since this won't effect anyone I have zero qualms :)

lherman-cs:
@Sean-Der Sounds good! That makes sense. I've just updated the base branch to be master.

PR #969, opened 2020-01-06
Expose an API to get Packetizer of the Track.

kchugalinskiy:
#### Description
Add a new public method, Packetizer, under Track to allow access
to get corresponding Packetizer to allow developers configure it.
At the current moment Packetizer allows user to use abs time as a timestamp, as [described here](https://webrtc.org/experiments/rtp-hdrext/abs-send-time), but there is no way in pion webrtc to use this feature.

#### Reference issue
No issues found

### DIFF
--- a/track.go
+++ b/track.go
@@ -77,6 +77,13 @@ func (t *Track) Codec() *RTPCodec {
 	return t.codec
 }
 
+// Packetizer gets the Packetizer of the track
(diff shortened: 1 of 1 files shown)


Sean-Der:
Merged with df4e38f4c27c9aad34a6c3c85411e722b14098c8

Thank you so much for the contribution, and thank you for using Pion! If there is anything we can do better I would love to hear :)

PR #973, opened 2020-01-11
Use wasmbrowsertest

at-wat:
Instead of Node.js + wrtc.


at-wat:
`panic: JavaScript error: Failed to execute 'getRandomValues' on 'Crypto': The ArrayBufferView's byte length (1048576) exceeds the number of bytes of entropy available via this API (65536).`

https://github.com/pion/transport/pull/46

at-wat:
browser looks unable to output coverage report as file.

PR #984, opened 2020-01-21
mediaengine: Case-insensitive media subtype check

encounter:
As defined in: https://tools.ietf.org/html/rfc4855#section-3

```
Media subtype names are commonly shown in lower case.
These names are case-insensitive in both places.
```

As well as: https://tools.ietf.org/html/rfc4288#section-4.2

```
Both type and subtype names are case-insensitive.
```


Sean-Der:
Fantastic, thanks @encounter!

Will merge when CI passes :)

Sean-Der:
Fantastic! Thank you so much for fixing those @encounter 

If you are interested we also have a pretty active [Slack](https://pion.ly/slack) channel if you want to talk more/help debugging.

Issue #991, opened 2020-01-30
Codec problem in the OnTrack function


varadely:
Hi, 
I wanted to make an Update of your librairy (my old version :  June 2019). And while doing the update I have a codec problem when I build for windows only: 
output from pion/example-webrtc-applications/gstreamer-receive : 
On windows: 
```
Connection State has changed connected 
Track has started, of type 0: PCMU 
Track has started, of type 96: VP8 
```

on Linux : 
```
Connection State has changed connected 
Track has started, of type 111: OPUS 
Track has started, of type 96: VP8 
```
Any idea ?

Sean-Der:
Hi @varadely 

What SHA1 are you using now? We did fix a bug around this with 6a29bdf3040f4227ad0e785da9cf479600cabd3c

Sean-Der:
Could you also try before 5a752da6831a61700a22df00ae117ca2ec24c7d9 this was when G711 support was added

varadely:
I use the this https://github.com/pion/webrtc/commit/537c841073600e9fb176d1677ee1b3c41b8baca7
Is this the correct SHA1 ? Or if not, where do I find it?

varadely:
> Hi @varadely
> 
> What SHA1 are you using now? We did fix a bug around this with [6a29bdf](https://github.com/pion/webrtc/commit/6a29bdf3040f4227ad0e785da9cf479600cabd3c)

I've this correction in my code : 
```
	m.RegisterCodec(NewRTPPCMUCodec(DefaultPayloadTypePCMU, 8000))
	m.RegisterCodec(NewRTPPCMACodec(DefaultPayloadTypePCMA, 8000))
	m.RegisterCodec(NewRTPOpusCodec(DefaultPayloadTypeOpus, 48000))
```

varadely:
> Could you also try before [5a752da](https://github.com/pion/webrtc/commit/5a752da6831a61700a22df00ae117ca2ec24c7d9) this was when G711 support was added

How I can install an older version ? 
When I wanted to build, I had problems with dependencies on outdated packages (src\github.com\pion\webrtc\dtlstransport.go:245:3: unknown field 'Certificate' in struct literal of type dtls.Config) .

varadely:
I overwrite the PayloadType and it's work. I don't know why it's 0, in the SDP, the codec is OPUS.
In track.go : 
```
func (t *Track) determinePayloadType() error {
	r, err := t.ReadRTP()
	if err != nil {
		return err
	}

[...]

PR #992, opened 2020-01-30
Prefer Opus over G711

Sean-Der:
Relates to #991


Sean-Der:
@at-wat When I added G711 support I just added it to the top, but I think this is probably not want people want in the default case.

I moved Opus to the top for now, so if we are offering the other side will prefer it.

PR #1003, opened 2020-02-05
Implement re-negotiation

Sean-Der:
Relates to #207


Sean-Der:
@at-wat this is a brutally large PR (sorry about that) we had lots of code that worked with SDPs that would never work with re-negotiation :(  

I can break this into smaller PRs, or add more test coverage to make this easier to review! I also will be writing another `example` that shows `AddTrack`/`RemoveTrack` usage with people joining/leaving a conference call. I can put that in this PR as well, but will take me another day though.

peerconnection.go
```diff
@@ -1878,72 +1675,36 @@ func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, re
...
-	var openedDCCount uint32
-	for _, d := range dataChannels {
-		if d.ReadyState() == DataChannelStateConnecting {
-			err := d.open(pc.sctpTransport)
-			if err != nil {
-				pc.log.Warnf("failed to open data channel: %s", err)
-				continue
-			}
-			openedDCCount++
+		receiver, err := pc.api.NewRTPReceiver(t.Receiver().kind, pc.dtlsTransport)
```
at-wat: In my understanding, this is called if a transceiver `t` in the current list of transceivers is removed from remote description.
I'm not sure why new receiver is set to the transceiver which is removed.
I think `t` should be deleted from `pc.rtpTransceivers` like:
```go
func (pc *PeerConnection) startRenegotation(currentTransceivers []*RTPTransceiver) {
  trackDetails := trackDetailsFromSDP(pc.log, pc.RemoteDescription().parsed)
  var newTransceivers []*RTPTransceiver
  pc.mu.Lock()
  for _, t := range currentTransceivers {
    if t.Receiver() == nil || t.Receiver().Track() == nil {
      if t.Sender() != nil {
        newTransceivers = append(newTransceivers, t)
      } else {
        t.setDirection(RTPTransceiverDirectionInactive)
      }
      continue
    } else if _, ok := trackDetails[t.Receiver().Track().ssrc]; ok {
      newTransceivers = append(newTransceivers, t)
      continue
    }
    
    if err := t.Receiver().Stop(); err != nil {
      pc.log.Warnf("Failed to stop RtpReceiver: %s", err)
      continue
    }
    t.setReceiver(nil)
    if t.Sender() != nil {
      newTransceivers = append(newTransceivers, t)
    } else {
      t.setDirection(RTPTransceiverDirectionInactive)
    } 
  }
  pc.rtpTransceivers = newTransceivers
  pc.mu.Unlock()
  
  pc.startRTPSenders(currentTransceivers)
  pc.startRTPReceivers(trackDetails, currentTransceivers)
}
```

Sean-Der: (Let me preface this with I am still learning the Transceiver API, so this might be wrong)

* trackDetailsFromSDP just looks up the incoming SSRCes. So a Transceiver could just have had its SSRC flipped (remote did a TrackRemove -> TrackAdd). It isn't really transceiver aware. 
* Transceivers can never deleted/removed. They only have their directions changed.

We should probably put this code in a dedicated function like 'pc.stopRTPReceivers`, its purpose is just to loop through and stop incoming stuff.

Later we will stop our `Senders` if we are answering. We loop over the remotes transceivers and if they provide none that satisfy the local we don't tell the remote about one of our senders.

Sean-Der:
I am not sure about `AddTrack`/`RemoveTrack` during negotiation. I will add a test!

`startRenegotation` and `startTransports` take a copied list of transceivers. We should probably make all the PeerConnection APIs take a global lock, there is no reason to be doing this stuff concurrently. I don't see a performance benefit vs the possible bugs it could cause.

But not really sure! Haven't thought about it deeply.

peerconnection.go
at-wat: Thanks, I finally got it. My understanding around Track and Transceiver had been confused.

PR #1004, opened 2020-02-05
Upgrade dtls v2

at-wat:
Follow dtls module structure update.

### Reference issue
Ref: https://github.com/pion/dtls/pull/185


certificate.go
```diff
@@ -94,17 +94,21 @@ func (c Certificate) Expires() time.Time {
 // GetFingerprints returns the list of certificate fingerprints, one of which
 // is computed with the digest algorithm used in the certificate signature.
 func (c Certificate) GetFingerprints() ([]DTLSFingerprint, error) {
-	fingerprintAlgorithms := []dtls.HashAlgorithm{dtls.HashAlgorithmSHA256}
+	fingerprintAlgorithms := []string{"sha-256"}
```
Sean-Der: This not being an enum anymore is kind of unfortunate :/ I could see mistakes coming from this.

Sean-Der: But then exporting bloats the public API. I don't know which tradeoff is worse.

at-wat: We could add inverse func of https://github.com/pion/dtls/blob/master/pkg/crypto/fingerprint/hash.go. Having it would be good.

at-wat: https://github.com/pion/dtls/pull/186

Sean-Der: Change looks fantastic. I left my comments on other PR, but all approves from me :)

Sean-Der:
@at-wat This is perfect! Thanks again for refactoring/cleaning stuff up. Not always the most fun work :(

Issue #42, opened 2018-07-18
testing: integration testing


backkem:
Automate integration tests between pions and web browsers. Example setup: Use a headless browser in combination with a simple web server to serve a test web page and exchange the SDP offer/answer. Validate the success using a browser API like the [Google Chrome Remote API](https://github.com/wirepair/gcd). Finally make the test part of the [CI](https://docs.travis-ci.com/user/gui-and-headless-browsers/).

Sean-Der:
I moved this to our [Project Ideas](https://github.com/pions/webrtc/wiki/Contributing#project-ideas)

PR #1007, opened 2020-02-08
Add fundamental browser E2E test

at-wat:
Add connection establish and RTP packet count test using headless
Chromium. Run the test as GitHub Action.

ref: #42


e2e/e2e_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"github.com/pion/webrtc/v2"
+	"github.com/pion/webrtc/v2/pkg/media"
+)
+
+var silentOpusFrame = []byte{0xf8, 0xff, 0xfe} // 20ms, 8kHz, mono
+
+func TestE2E(t *testing.T) {
+	drivers := map[string]*agouti.WebDriver{
+		"Chrome": agouti.ChromeDriver(
```
at-wat: would have Firefox/GeckoDriver in future.

Sean-Der:
:fire: this fantastic!

I have a bunch of things I want to test with this! We should be able to get WAY more media coverage now, this is really great :)

Sean-Der:
Do you like the workflows? Should we move everything over to that?

It has been on my TODO list. The travis stuff is sometimes flakey

at-wat:
There are still some cons in github actions at now:
- coverage report requires token; can't be triggered by PRs from forked repositories
- not fully stable yet (I met some stucks which is not able to stop from the web interface)

But, repository specific jobs like this is easy to be handled by simple add a new workflow yaml, even if the main jobs are shared between repos.

PR #1018, opened 2020-02-11
Add test for multi-track PlanB signaling

### LINKED ISSUE #1014: Add E2E tests for PlanB generation
CodeCov shows that some of our SDP code isn't executed for PlanB

Sean-Der:
SDP parsing only supported one ssrc per media section.
Now we properly can handle multiple.

Resolves #1014

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -423,8 +423,13 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
 			}
 		}
 
-		mediaSections = append(mediaSections, mediaSection{id: "video", transceivers: video})
-		mediaSections = append(mediaSections, mediaSection{id: "audio", transceivers: audio})
+		if len(video) > 1 {
+			mediaSections = append(mediaSections, mediaSection{id: "video", transceivers: video})
+		}
+
+		if len(audio) > 1 {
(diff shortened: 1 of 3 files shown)


peerconnection.go
```diff
@@ -423,8 +423,13 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
 			}
 		}
 
-		mediaSections = append(mediaSections, mediaSection{id: "video", transceivers: video})
-		mediaSections = append(mediaSections, mediaSection{id: "audio", transceivers: audio})
+		if len(video) > 1 {
```
Sean-Der: I made a mistake when adding re-negotiation. Was caught by new test!

sdp.go
```diff
@@ -46,9 +46,6 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 				}
 
 				incomingTracks[uint32(ssrc)] = trackDetails{codecType, trackLabel, trackID, uint32(ssrc)}
-				if trackID != "" && trackLabel != "" {
```
Sean-Der: This never worked. We had code in other places to accept multi-track PlanB SDPs, but this was always filtering it before it could arrive. Not sure when it regressed :(

Sean-Der: 85a16026984dd4151b096cc5ffd28e3dc4fd0c03

Issue #1023, opened 2020-02-13
Support DTLS fingerprints declared at the media section or global level


Sean-Der:
We should look for fingerprints at both the media and global level. After parsing we should assert that they all match, and then move ahead.

Currently we only scan at the media level.

Sean-Der:
This is actually implemented properly! We just need to add some tests

Sean-Der:
Resolved by 6bb48a2a9243577e09c37a86bd146cd9cdb60697

PR #1024, opened 2020-02-13
Split ICE and DTLS related SDP parsing out

Sean-Der:
Move this stuff out of SetRemoteDescription so it will be easier to test

Relates to #1023


Sean-Der:
This is WIP! Will be covering this all with tests

PR #1031, opened 2020-02-14
Allow formatting parameters for all codecs

hendrikhofstadt:
#### Description

Formatting parameters are not only allowed on video codecs but also on most audio codecs.

Without this change, custom parameters of audio codecs would be dropped when parsing SDPs (e.g. opus parameters)


Sean-Der:
Merged! Thank you so much @hendrikhofstadt 

Hopefully you found the code easy to read/debug, if you have any suggestions on how to make Pion better I would love to hear :)

thanks!

PR #1035, opened 2020-02-16
Ivf reader live stream

ClaytonMcCray:
#### Description
Add a receiver to `*IVFReader` that allows caller to reset the internal io.Reader stream for the reader. This can be used, for instance, to reset the stream if one is encoding raw data live and the previous call to `ivf.ParseNextFrame()` got EOF.


Sean-Der:
Merged+Squashed with 09a6722eec5cc7be4a8573868d71bd42d3833bea

Thank you so much for your contribution @ClaytonMcCray

PR #1038, opened 2020-02-18
RTPSender: Expose the 'track' attribute as per spec

lawl:
The webrtc spec in https://www.w3.org/TR/webrtc/#dom-rtcrtpsender sais an RtpSender object has a read-only attribute track. We expose this attribute via a 'Track()' method.

[`RTPReceiver`](https://github.com/pion/webrtc/blob/v2.2.0/rtpreceiver.go#L54) already exposes this attribute in the same way, so not including it in `RTPSender` was likely just an oversight.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Luke S](https://github.com/encounter)
 * [Hendrik Hofstadt](https://github.com/hendrikhofstadt)
 * [Clayton McCray](https://github.com/ClaytonMcCray)
+* [lawl](https://github.com/lawl)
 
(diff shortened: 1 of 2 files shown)


Sean-Der:
Awesome work @lawl thanks for the contribution.

If you are interested I would love to add you to the Pion organization. You will be able to review other peoples PRs, and merge your own after a PR is approved. If you are interested in actively contributing is nice!

thanks

lawl:
hey @Sean-Der i'm flattered, thanks for the offer! I'm doing this in my spare time at the moment and pretty much on a 'i feel like working on this today' basis. I may not feel like it and stop working with webrtc for a year tomorrow, so I don't think that would be fair to everyone else.

PR #1042, opened 2020-02-20
Fix race in PeerConnection.ConnectionState

gqgs:
#### Description

The data race detector detects a race if PeerConnection.ConnectionState is called while PeerConnection.updateConnectionState is changing the connection state.


Sean-Der:
:fire: thank you @gqgs

PR #1052, opened 2020-02-26
Adding OnICEGatheringFinish.

### LINKED ISSUE #1051: Adding PeerConnection.OnICEGatheringFinish event.
## Summary

An event called once ICE gathering is finished.

## Motivation

The idea to be able to safely close the stream used to trickle when there no more ICE candidate to expect (my protocol wait for both nodes to finishing exchange, not only one).

Registering OnICEGatheringStateChange and checking if state is completed, but that doesn't works or is not safe due to syncronisation issue.
The problem is OnICEGatheringStateChange is executed before OnICECandidate, so the best way I've found is creating a sync.Cond waiting in OnICEGatheringStateChange and broadcasting in OnICECandidate, but Cond are forgetable so if due to lack of enough cpu or context switch at the wrong time OnICECandidate may get executed before OnICEGatheringStateChange.

Its also not possible Lock or channels because Unlock an unlocked locks or closing a close channel panics and the number of OnICECandidate is unkown.

Jorropo:
# Description

It would also be possible to swap the order of event execution but, but that require breaking change to the execution flow maybe breaking some apps. (First setting state completed then executing onLocalCandidateHdlr and then onStateChangeHdlr but that still breaking change to the execution flow).
This is way safer.

# Reference issue
Fixes #1051.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Hendrik Hofstadt](https://github.com/hendrikhofstadt)
 * [Clayton McCray](https://github.com/ClaytonMcCray)
 * [lawl](https://github.com/lawl)
+* [Jorropo](https://github.com/Jorropo)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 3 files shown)


Sean-Der:
Fantastic, thank you so much @Jorropo :)

Jorropo:
@Sean-Der I saw that and yes that works but here is 3 problems with that :
- That not Intuitive (ok that not objective at all but still that maters for me).
- That unsafe, what told me

PR #1053, opened 2020-02-27
Adding documentation about OnICECandidate

### LINKED ISSUE #1051: Adding PeerConnection.OnICEGatheringFinish event.

Jorropo:
#### Description

Follow up to #1051.

#### Reference issue
Fixes #1051, may also see #1052.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Hendrik Hofstadt](https://github.com/hendrikhofstadt)
 * [Clayton McCray](https://github.com/ClaytonMcCray)
 * [lawl](https://github.com/lawl)
+* [Jorropo](https://github.com/Jorropo)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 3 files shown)


Sean-Der:
Glad that worked @Jorropo really appreciate the doc improvement :)

---

If you have any suggestions on how to make Pion WebRTC easier to use (big or small) I would love to hear them! Sorry for the rough edges :(

Jorropo:
@Sean-Der :D no problem. The examples are great enough, in fact I didn't used the doc often, mostly copy paste (the true developer powers !)

PR #1077, opened 2020-03-11
Fix race condition on DTLSTransport init

### LINKED ISSUE #1039: TestPeerConnection_Renegotation_AddTrack failing sometimes
### Your environment.
  - Version: a4353b3668055610a40d50d311cf5c6ddd01f486
  - Browser: n/a
  - Other Information - https://travis-ci.org/pion/webrtc/jobs/652037403

### What did you do?

When creating PR #1038 a the test-suite on travis failed and it had seemingly nothing to do with my change (add a method for users that's never used by pion itself).

After a force push with no changes, to trigger a rebuild, the test suite was green.

### What did you expect?

Test suite being green.

### What happened?

The `TestPeerConnection_Renegotation_AddTrack` test seems to sometimes fail. I have no idea how to reproduce it. See link to the travis build above.

at-wat:
`startSRTP()` was called before establishing DTLS connection if renegotiation is triggered immediately after start.
Fix lock scope and cancel connection handshake by `context.Context` on `Stop()`.

### Reference issue
Fixes #1039

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -3,6 +3,7 @@
 package webrtc
 
 import (
+	"context"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
@@ -12,6 +13,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pion/dtls/v2"
@@ -37,7 +39,8 @@ type DTLSTransport struct {
 
 	onStateChangeHdlr func(DTLSTransportState)
 
-	conn *dtls.Conn
+	conn                 *dtls.Conn
+	cancelInitialization atomic.Value // func()
 
 	srtpSession   *srtp.SessionSRTP
 	srtcpSession  *srtp.SessionSRTCP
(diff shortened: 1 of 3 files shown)


Sean-Der:
@at-wat I think we just need to queue `SetRemoteDescription` calls instead. 

* The first call should never be interrupted (unless Close is called). If the DTLS handshake is cancelled there is no re-doing it.
* We should then process the second call (after the first succeeded) 

We probably should return a `Context` and then users can check-in on the state of each `SetRemoteDescription` call?

Sean-Der:
You could also fix this today if you just have the test wait until `PeerConnectionState` is connected! Then you don't have to worry about the second call to `SetRemoteDescription` being too fast.

at-wat:
> We probably should return a `Context` and then users can check-in on the state of each `SetRemoteDescription` call?

I'm not sure how the `Context` is returned.
`Context` should be passed by the user.

Sean-Der:
Oh yes sorry got it reversed. We need a promise so people can check in on the status of them (or block until we are done)

at-wat:
> You could also fix this today if you just have the test wait until PeerConnectionState is connected! 

I think it's a problem in the library, not in the test.
In real application, duration to establish the connection depends on the environment, so the library should not expect a sleep at there.

Sean-Der:
Agree 100% the way we do re-negotiations is busted! I just don't want to block you from doing other work.

I still think the proper fix is having a global queue of `SetRemoteDescription` values that get processed one at a time. Then we don't have to worry about race conditions, and should make the code much cleaner!

at-wat:
> I think we just need to queue SetRemoteDescription calls instead.

If we select queuing, all other functions internally depending on DTLS connection should also be queued, shouldn't it?

Sean-Der:
Yep!

So [this](https://github.com/pion/webrtc/blob/be9fde1fd3cfad91450f1d090b84afe119405c76/peerconnection.go#L1539-L1563) is run on the first `SetRemoteDescription` 

We shouldn't allow any subsequent ones to happen until ^ has completed.

The issue is that [the next](https://github.com/pion/webrtc/blob/be9fde1fd3cfad91450f1d090b84afe119405c76/peerconnection.go#L729-L730) call skips all this.

If we queue everything the subsequent calls we wait until the first call is completed.

Sean-Der:
I don't feel strongly how we implement it, but the root issue is we can't process any `SetRemoteDescription`s until the first one is finished. We only get one shot at the DTLS/SCTP handshake, so I don't think we should interrupt any of that.

at-wat:
Basically, I think `SetRemoteDescription` should block until finishing and be able to be canceled by `Context`, but it's not v2 API. https://github.com/pion/webrtc/wiki/PlanningV3#pass-contextcontext-to-blocking-functions

PR #1080, opened 2020-03-13
Don't upcase opus when adding to SDP

### LINKED ISSUE #1078: pion offer contains "a=rtpmap:111 OPUS" not "a=rtpmap:111 opus"
### Your environment.
  - Version: github.com/pion/webrtc/v2 v2.2.3
  - Browser: none
  - Other Information -

### What did you do?
Create an offer in pion

### What did you expect?
Offer to contain `a=rtpmap:111 opus/48000/2`. This was the case in pion-v2.1.19.

### What happened?
Offer had line `a=rtpmap:111 OPUS`.
I'm not sure if using uppercase OPUS is a real bug / incompatible with the specification, but it seems like an odd change, and   [RFC7587](https://tools.ietf.org/html/rfc7587) specifies lowercase:  `The media subtype ("opus") goes in SDP "a=rtpmap" as the encoding name.`

Sean-Der:
When making MediaEngine case-insensitive in 20f2d18
Opus was upcased when adding to SDP. This returns it
to what it was before

Resolves #1078

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -78,30 +78,27 @@ func (m *MediaEngine) PopulateFromSDP(sd SessionDescription) error {
 			}
 
 			var codec *RTPCodec
-			clockRate := payloadCodec.ClockRate
-			parameters := payloadCodec.Fmtp
-			codecName := strings.ToUpper(payloadCodec.Name)
-			switch codecName {
-			case PCMU:
-				codec = NewRTPPCMUCodec(payloadType, clockRate)
-			case PCMA:
-				codec = NewRTPPCMACodec(payloadType, clockRate)
-			case G722:
-				codec = NewRTPG722Codec(payloadType, clockRate)
-			case Opus:
-				codec = NewRTPOpusCodec(payloadType, clockRate)
-			case VP8:
-				codec = NewRTPVP8Codec(payloadType, clockRate)
-			case VP9:
-				codec = NewRTPVP9Codec(payloadType, clockRate)
-			case H264:
-				codec = NewRTPH264Codec(payloadType, clockRate)
+			switch {
+			case strings.EqualFold(payloadCodec.Name, PCMA):
+				codec = NewRTPPCMACodec(payloadType, payloadCodec.ClockRate)
+			case strings.EqualFold(payloadCodec.Name, PCMU):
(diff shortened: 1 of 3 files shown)


Sean-Der:
@at-wat not a necessary change, but it might cause unexpected issues for people so just doing to be safe!

mediaengine.go
```diff
@@ -78,30 +78,25 @@ func (m *MediaEngine) PopulateFromSDP(sd SessionDescription) error {
...
-				codec = NewRTPOpusCodec(payloadType, clockRate)
-			case VP8:
-				codec = NewRTPVP8Codec(payloadType, clockRate)
-			case VP9:
-				codec = NewRTPVP9Codec(payloadType, clockRate)
-			case H264:
-				codec = NewRTPH264Codec(payloadType, clockRate)
+			switch {
+			case strings.EqualFold(payloadCodec.Name, PCMU) || strings.EqualFold(payloadCodec.Name, PCMA):
+				codec = NewRTPPCMUCodec(payloadType, payloadCodec.ClockRate)
```
at-wat: https://github.com/pion/webrtc/blob/f4bbe5d76f46142f0a5563d96cec21374fe29dc9/mediaengine.go#L155-L177
`RTPCodec.Name` and `RTPCodec.RTPCodecCapability.MimeType` fields are different between PCMU and PCMA.

mediaengine_test.go
```diff
@@ -92,3 +92,18 @@ a=sctpmap:5000 webrtc-datachannel 1024
...
+	pc, err := NewPeerConnection(Configuration{})
+	assert.NoError(t, err)
+
+	_, err = pc.AddTransceiverFromKind(RTPCodecTypeAudio)
+	assert.NoError(t, err)
+
+	offer, err := pc.CreateOffer(nil)
+	assert.NoError(t, err)
+
+	assert.Contains(t, offer.SDP, "opus/48000")
```
at-wat: I think it should be bit more strict match something like `^a=rtpmap:[0-9]\+ opus/48000\([/ ].*\)?$`

Sean-Der: added!

mediaengine.go
Sean-Der: Oops my mistake, nice catch :/

PR #1087, opened 2020-03-20
Make the IVF reader able to read from a network connection

### LINKED ISSUE #1086: IVFReader doesn't work when reading from e.g. a network connection
`IVFReader` doesn't work when it reads from an `io.Reader` which is not an `os.File`, because `Read` calls aren't guaranteed to fill all the supplied buffer and the data can arrive in chunks.

opennota:
Use `io.ReadFull` to read from the `io.Reader`.

Fixes #1086


opennota:
Done.

PR #1096, opened 2020-03-24
Add detection for removed remote tracks

longsleep:
This PR adds support for `recvonly` remote senders by cleaning up existing local receivers and by preventing start of new local receivers for tracks which are not sending, to make it possible to detect removed remote tracks.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1587,8 +1587,11 @@ func (pc *PeerConnection) startRenegotation(currentTransceivers []*RTPTransceive
 		for _, t := range currentTransceivers {
 			if t.Receiver() == nil || t.Receiver().Track() == nil {
 				continue
-			} else if _, ok := trackDetails[t.Receiver().Track().ssrc]; ok {
-				continue
+			} else if trackDetail, ok := trackDetails[t.Receiver().Track().ssrc]; ok {
+				if trackDetail.direction == RTPTransceiverDirectionSendrecv || trackDetail.direction == RTPTransceiverDirectionSendonly {
+					// Keep the receiver only, when the remote track is actually sending
+					continue
+				}
 			}
 
 			if err := t.Receiver().Stop(); err != nil {
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -830,6 +830,10 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
 	}
 
 	for ssrc, incoming := range incomingTracks {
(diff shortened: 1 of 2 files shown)


Sean-Der:
:fire: thank you @longsleep 

Would you be interested in being added to the Pion organization? I am always looking to get more people involved. You would be able to merge your own PRs (after you get one approval) and you can approve other peoples PRs.

No hard responsibility, but the project always needs more smart people :)

longsleep:
> Would you be interested in being added to the Pion organization?

Sounds good :+1: - Pion is now part of two WebRTC projects i am working on so i expect more pull requests to come and i am willing to get involved and make it better. Thank you!

Sean-Der:
Hey @longsleep

Sorry I took so long on this, I had to do some digging myself! It looks like the idiomatic way to do this is to add support for `MediaStream` and then you should listen for [onremovetrack](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/onremovetrack)

longsleep:
> Sorry I took so long on this, I had to do some digging myself! It looks like the idiomatic way to do this is to add support for `MediaStream` and then you should listen for [onremovetrack](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/onremovetrack)

The MediaStream APi is not very well used in browsers and while the `onremovetrack` way might be a way to detect removal of tracks from a stream, it might be complicated to add (generally in my opinion the MediaStream API in general complicates things as is not really needed in WebRTC).

> 
> I don't fully understand your diff, does this fire any callback/events to the user?

Thus this change set goes the easy way and only extends what is currently there without adding anything more to the API, leveraging the existing RTPReceiver API by closing its reader when a remote sender stops sending (the RTPReceiver is stopped). This close can be detected by any consumer code, making it possible to know when a Receiver ended and by this making it possible to implement logic which detects that a remote track has stopped. The intention of this PR is to be as minimal as possible.

The API of tracks could be extended in a future PR by implementing https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/onended which is what is mostly done right now to detect remote track removals in a browser context. Implementation of this even in pion/webrtc can leverage the same detection and be triggered if an receiver is stopped.

jeremija:
I believe this can be closed now since #1159 was merged? I didn't see it when I submitted my PR

longsleep:
> I believe this can be closed now since #1159 was merged? I didn't see it when I submitted my PR

Confirmed. Thanks!

PR #1114, opened 2020-03-30
Append media data section after transceivers

### LINKED ISSUE #1113: Microsoft Edge Prechromium not working
### Your environment.
  - Version: v2.2.3
  - Browser: Microsoft Edge (Win 10 build 1903)

### What did you do?
Pion server (CreateOffer) to Edge client.

### What did you expect?
Connects

### What happened?
Fails to connect : Error in edge js scripts =>
0: Unable to get property 'state' of undefined or null reference rtcpeerconnection.js (1016,1)

nindolabs:
For some reason edge prechromium does not like media data section before video/audio in the sdp.

It generates the following error:

> 0: Unable to get property 'state' of undefined or null reference    rtcpeerconnection.js (1016,1)

Fixes #1113

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1655,7 +1655,6 @@ func (pc *PeerConnection) generateUnmatchedSDP(useIdentity bool) (*sdp.SessionDe
 		}
 		mediaSections = append(mediaSections, mediaSection{id: "data", data: true})
 	} else {
-
 		for _, t := range pc.GetTransceivers() {
 			mediaSections = append(mediaSections, mediaSection{id: strconv.Itoa(len(mediaSections)), transceivers: []*RTPTransceiver{t}})
 		}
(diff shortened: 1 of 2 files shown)


at-wat:
CI status is not reported to GitHub at now.
Result for 21c2b63e361f5d5641aa800681c0b4172c9cbe83: https://travis-ci.org/github/pion/webrtc/builds/668782044

nindolabs:
this PR conflicts with the following test.

> // When creating an offer the first media section MUST be SCTP
// This is to make renegotation easier, instead of having to remember where it
// is was interleaved
func TestPeerConnection_Renegotation_ApplicationFirst(t *testing.T) {

at https://github.com/pion/webrtc/blob/master/peerconnection_renegotation_test.go#L134

Should we switch the test to ApplicationLast ?

Sean-Der:
@nindolabs I don't even think we need that test anymore!

When I was in the process of implementing re-negotiation I thought I could make my life easier by putting ApplicationData first. However we now use `pc.generateMatchedSDP` so this doesn't even matter anymore.

I would move the application data last, and then remove the test. It is just testing stuff that doesn't really matter!

PR #1115, opened 2020-04-01
Fix RTPReceiver Stop race

longsleep:
I saw panics like this, when an RTPTransceiver is stopped.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9479b2]
goroutine 37117 [running]:
github.com/pion/srtp.(*ReadStreamSRTCP).Close(0x0, 0x0, 0x0)
        github.com/pion/srtp@v1.2.7/stream_srtcp.go:66 +0x42
github.com/pion/webrtc/v2.(*RTPReceiver).Stop(0xc011d4f3e0, 0x0, 0x0)
        github.com/pion/webrtc/v2@v2.2.4-0.20200310073244-be9fde1fd3cf/rtpreceiver.go:139 +0xfb
github.com/pion/webrtc/v2.(*RTPTransceiver).Stop(0xc000ced300, 0x1137798, 0x10)
        github.com/pion/webrtc/v2@v2.2.4-0.20200310073244-be9fde1fd3cf/rtptransceiver.go:55 +0x15a
github.com/pion/webrtc/v2.(*PeerConnection).Close(0xc0395a1d40, 0xc04a5cbbc0, 0xc7f200)
```

If a RTPReceiver is stopped before it ever receivers, no streams are
attached to it which can be closed. This change does adds a check
to ensure that the RTPReceivers's streams are not nil before trying to
close them.

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -66,7 +66,7 @@ func (r *RTPReceiver) Receive(parameters RTPReceiveParameters) error {
 		return fmt.Errorf("Receive has already been called")
 	default:
 	}
-	close(r.received)
+	defer close(r.received)
 
 	r.track = &Track{
 		kind:     r.kind,
@@ -136,11 +136,15 @@ func (r *RTPReceiver) Stop() error {
 
 	select {
 	case <-r.received:
-		if err := r.rtcpReadStream.Close(); err != nil {
-			return err
+		if r.rtcpReadStream != nil {
+			if err := r.rtcpReadStream.Close(); err != nil {
+				return err
+			}
 		}
-		if err := r.rtpReadStream.Close(); err != nil {
-			return err
+		if r.rtpReadStream != nil {
+			if err := r.rtpReadStream.Close(); err != nil {
+				return err
+			}
 		}
 	default:
 	}


longsleep:
@Sean-Der The commit message linter is too strict imo. It is often nice to include stack traces in commit messages for the purpose of people searching for the same/similar stack traces and it is impossible to format them in a way which would make the linter ok.

Sean-Der:
oh fantastic catch!

Sean-Der:
I am totally in support of moving over towards https://www.conventionalcommits.org! We have talked about it a few times, just haven't put the efforts in. You would only have to make the change in one place also, [this](https://github.com/pion/.goassets/blob/master/ci/.github/lint-commit-message.sh) file is deployed.

In the short term I would put a bracktrace in a new github issue.

I think the root issue is that we close received to early. Instead of [here](https://github.com/pion/webrtc/blob/4edde982826260bfffd96dcfd8cbe332eb71f8c2/rtpreceiver.go#L69) should we do it at the end of the function? We also don't have to worry about stuff piling up either because we take the lock at the start of the function.

longsleep:
> I think the root issue is that we close received to early. Instead of [here](https://github.com/pion/webrtc/blob/4edde982826260bfffd96dcfd8cbe332eb71f8c2/rtpreceiver.go#L69) should we do it at the end of the function? We also don't have to worry about stuff piling up either because we take the lock at the start of the function.

Oh yeah that might prevent the issue - will try and update this PR accordingly.

longsleep:
> Oh yeah that might prevent the issue - will try and update this PR accordingly.

@Sean-Der added a `defer` - the checks for nil are still needed, since the first open could have succeeded while only the second has failed (So technically only the check for the second stream is needed, but i guess its better to check for both to prevent bugs for potential future changes).

Sean-Der:
@longsleep fantastic work :) Yea that is perfect.

go ahead and hit the 'Rebase and merge' button when you are happy!

PR #1116, opened 2020-04-01
Add helpers to disable replay protection

longsleep:
Sometimes the replay protection of SRTP and SRTCP needs to be disabled.
This change adds disableSRTPReplayProtection and
DisableSRTCPReplayProtection helpers to SettingEngine.

### DIFF
--- a/settingengine.go
+++ b/settingengine.go
@@ -50,6 +50,8 @@ type SettingEngine struct {
 	}
 	answeringDTLSRole                         DTLSRole
 	disableCertificateFingerprintVerification bool
+	disableSRTPReplayProtection               bool
+	disableSRTCPReplayProtection              bool
 	vnet                                      *vnet.Net
 	LoggerFactory                             logging.LoggerFactory
 }
@@ -218,10 +220,22 @@ func (e *SettingEngine) SetDTLSReplayProtectionWindow(n uint) {
 
 // SetSRTPReplayProtectionWindow sets a replay attack protection window size of SRTP session.
 func (e *SettingEngine) SetSRTPReplayProtectionWindow(n uint) {
+	e.disableSRTPReplayProtection = false
 	e.replayProtection.SRTP = &n
 }
 
 // SetSRTCPReplayProtectionWindow sets a replay attack protection window size of SRTCP session.
 func (e *SettingEngine) SetSRTCPReplayProtectionWindow(n uint) {
+	e.disableSRTCPReplayProtection = false
 	e.replayProtection.SRTCP = &n
+}
+
+// DisableSRTPReplayProtection disables SRTP replay protection.
+func (e *SettingEngine) DisableSRTPReplayProtection(isDisabled bool) {
+	e.disableSRTPReplayProtection = isDisabled
+}
+
+// DisableSRTCPReplayProtection disables SRTCP replay protection.
+func (e *SettingEngine) DisableSRTCPReplayProtection(isDisabled bool) {
+	e.disableSRTCPReplayProtection = isDisabled
 }
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -162,10 +162,25 @@ func (t *DTLSTransport) startSRTP() error {
 			srtp.SRTPReplayProtection(*t.api.settingEngine.replayProtection.SRTP),
 		)
 	}
+
+	if t.api.settingEngine.disableSRTPReplayProtection {
+		srtpConfig.RemoteOptions = append(
+			srtpConfig.RemoteOptions,
+			srtp.SRTPNoReplayProtection(),
+		)
+	}
+
 	if t.api.settingEngine.replayProtection.SRTCP != nil {
 		srtpConfig.RemoteOptions = append(
 			srtpConfig.RemoteOptions,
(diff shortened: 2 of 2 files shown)


longsleep:
Mhm i do not see how this change can be responsible for the test failure
```
=== RUN   TestPeerConnection_GetStats
    TestPeerConnection_GetStats: stats_go_test.go:295: 
        	Error Trace:	stats_go_test.go:295
        	Error:      	"240" is not greater than or equal to "276"
        	Test:       	TestPeerConnection_GetStats
```

Sean-Der:
By default replay protection is disabled, so I don't think we explicitly need this flag. Are you thinking if someone enabled it (and then wanted to disable it latter?)

Also I wonder if we could have a significant value for the existing functions (like if you passed 0) that would disable. Just would like to avoid having overlapping functions.

longsleep:
> By default replay protection is disabled, so I don't think we explicitly need this flag. Are you thinking if someone enabled it (and then wanted to disable it latter?)

No it is not afaict - https://github.com/pion/srtp/blob/master/session_srtcp.go#L40-L45

I was getting a lot of replay messages (not investigated yet if those are valid), so i had to add a knob to disable. Figured it would be nice to have.

If it were disabled by default, then yeah this is not needed

Sean-Der:
Oh nice catch, my mistake :)

What happens if you pass one of the max values from [here](https://golang.org/pkg/math/#pkg-constants) ?

longsleep:
> What happens if you pass one of the max values from [here](https://golang.org/pkg/math/#pkg-constants) ?

Not sure what you mean exactly, but if i set `s.SetSRTPReplayProtectionWindow(1<<64 - 1)` then all packets are counted as duplicate and no audio/video data is decoded. With 1<<8 - 1 it logs duplicates in a similar amountsthan with lower values and streams seem to work fine. 

```
INFO[0065] duplicated packet
```

Sean-Der:
We should probably on print `duplicated packet` when it does the drop, otherwise people might find that confusing.

I think this is the right thing to do though, in the comment we should tell people `If you wish to disable ReplayProtection you can pass math.MaxUint8 (or  1<<8 - 1) `

Sean-Der:
@at-wat what do you think of ^ maybe we just need to make it a verbose log?

longsleep:
> We should probably on print `duplicated packet` when it does the drop, otherwise people might find that confusing.

Isn't that the case? https://github.com/pion/srtp/blob/master/session.go#L138-L140 is what logs `duplicated packet` which in turn means that the package was not decrypted. So maybe i am already confused :)



> I think this is the right thing to do though, in the comment we should tell people `If you wish to disable ReplayProtection you can pass math.MaxUint8 (or 1<<8 - 1)`

What do you mean by that? In the unit test you added `s.SetSRTPReplayProtectionWindow(64)`. 64 is also the default in https://github.com/pion/srtp/blob/master/session_srtcp.go#L12 - so i do not understand.

Sean-Der:
@longsleep ah crap I see what you are dealing with, you want the same packet to be received multiple times (not a very large window).

I think your patch as is right. Mind waiting until @at-wat can review also?

longsleep:
> @longsleep ah crap I see what you are dealing with, you want the same packet to be received multiple times (not a very large window).

I am not sure why i would see the same packet multiple times. None of the senders has any loss or re transmits so at the moment i am not sure if really the same packet is received multiple times. I have not tried to verify yet.


> I think your patch as is right. Mind waiting until @at-wat can review also?

Yes, no rush. I want the stuff flowing back upstream to be solid and well understood.

Sean-Der:
I REALLY appreciate you pushing everything upstream, this is really fantastic :)

settingengine.go
```diff
@@ -225,3 +227,13 @@ func (e *SettingEngine) SetSRTPReplayProtectionWindow(n uint) {
 func (e *SettingEngine) SetSRTCPReplayProtectionWindow(n uint) {
 	e.replayProtection.SRTCP = &n
 }
+
+// DisableSRTPReplayProtection disables SRTP replay protection.
+func (e *SettingEngine) DisableSRTPReplayProtection(isDisabled bool) {
+	e.disableSRTPReplayProtection = isDisabled
```
at-wat, line 233: I think
```suggestion
func (e *SettingEngine) DisableSRTPReplayProtection() {
	e.disableSRTPReplayProtection = true
```
and adding `e.disableSRTPReplayProtection = false` to `SetSRTPReplayProtectionWindow()` simplifies API.

at-wat:
I agree with adding option to disable it.

I'm curious why you get (probably) duplicated packets.
Have you tried dumping network packets?

> `s.SetSRTPReplayProtectionWindow(1<<64 - 1)`

It requires 288 peta-bytes of sliding window if it is really allocated.
I guess overflow or something happens at somewhere and not working for this value.

at-wat:
> None of the senders has any loss or re transmits so at the moment i am not sure if really the same packet is received multiple times.

It is checked using sequence number in RTP header. So, duplicate doesn't always mean the duplication of the whole packet. (Sequence number of each SSRC must be started from random number and incremented for each packet without duplication.)

Sean-Der:
@longsleep sorry forgot to approve! mind rebasing and then should be good to merge :) thanks for dealing with the extra review.

PR #1118, opened 2020-04-02
Remove duplicated ICEGatherer.Gather call

at-wat:
Calling Gather() on NewPeerConnection causes nothing.


at-wat:
`Gather` calls `createAgent` and immediately ends if `ICETrickle == false`.
Looks not needed.

PR #1119, opened 2020-04-02
Update how to install in example projects

tsuu32:
#### Description
I found `go get github.com/pion/webrtc/examples/data-channels` fails.
Currently download instructions should contain 'v2'.
#### Reference issue
None.


Sean-Der:
Fantastic, thank you so much @tsuu32 !

PR #1120, opened 2020-04-02
Close stream in oggwriter

tarrencev:
Update to properly close an oggwriter when writing to a stream.


Sean-Der:
@tarrencev LGTM! Mind adding a body to your commit message, then you can merge yourself!

tarrencev:
@Sean-Der seems like a lint task is failing since im not in readme contribs?

PR #1123, opened 2020-04-03
Don't call OnTrack for same SSRC multiple times

### LINKED ISSUE #1122: OnTrack is incorrectly fired for the same ssrc multiple times
I am opening a PR for this now, but a user reported that called re-negotiation multiple times with the same SSRC would cause duplicated calls to OnTrack to happen.

Sean-Der:
During re-negotiation we didn't properly SSRCes we had already handled.
This makes sure to loop the existing transceivers and assert we don't
try to handle the same SSRC again.

Resolves #1122

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -829,6 +829,17 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
 		remoteIsPlanB = descriptionIsPlanB(pc.RemoteDescription())
 	}
 
+	// Ensure we haven't already started a transceiver for this ssrc
+	for ssrc := range incomingTracks {
+		for i := range localTransceivers {
+			if t := localTransceivers[i]; (t.Receiver()) == nil || t.Receiver().Track() == nil || t.Receiver().Track().ssrc != ssrc {
+				continue
+			}
+
+			delete(incomingTracks, ssrc)
+		}
+	}
+
 	for ssrc, incoming := range incomingTracks {
 		for i := range localTransceivers {
 			t := localTransceivers[i]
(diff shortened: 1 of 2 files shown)


peerconnection.go
```diff
@@ -829,6 +829,17 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
 		remoteIsPlanB = descriptionIsPlanB(pc.RemoteDescription())
 	}
 
+	// Ensure we haven't already started a transcevier for this ssrc
```
audiolion, line 832:
```suggestion
	// Ensure we haven't already started a transceiver for this ssrc
```

Sean-Der: Fixed!

Sean-Der: @audiolion would you be interested in joining the Pion org! You would be be able to approve PRs, and then would be able to merge your own (once approved)

No real obligation, but always looking to grow the community :)

Sean-Der:
> LGTM
> 
> I was not sure what is the correct behavior in the specification.
> As a result of [a short experiment on browsers](https://jsfiddle.net/xengd3oa/), this seems working as same as browser-to-browser renegotiation.
> (It would be helpful if there is a detatiled specification of the renegotiation.)
> 
> Another notable thing is that the browser (tested with Firefox) fires `onicecandidate(null)` on each renegotiation.

Oh weird I had no idea about `onicecandidate` firing null multiple times! The spec isn't hard to follow, but I always miss little things :/

 I am going to get all the media APIs into WASM, and we will be able to test everything *soon* just been distracted :)

peerconnection.go
audiolion: @Sean-Der sure that would be cool! I will try to help out in any way I can.

PR #1126, opened 2020-04-04
Fix candidate population in SDP

### LINKED ISSUE #1121: SessionDescription candidates are duplicated when calling LocalDescription() / CurrentLocalDescription() / PendingLocalDescription()
### Your environment.
  - Version: v2.2.4

### What did you do?
Reproducible example:
```go
func TestDoesntDuplicateSDPCandidates(t *testing.T) {
                pc, _ := webrtc.NewPeerConnection(webrtc.Configuration{})
		offer, _ := pc.CreateOffer(nil)
		pc.SetLocalDescription(offer)

		pendingLocalDesc := pc.PendingLocalDescription()
		pendingLocalDesc2 := pc.PendingLocalDescription()
		if !reflect.DeepEqual(pendingLocalDesc, pendingLocalDesc2) {
			t.Fatalf("expected %v, got %v", pendingLocalDesc, pendingLocalDesc2)
		}
}
```

`LocalDescription()` / `PendingLocalDescription()` /  `CurrentLocalDescription()` all call the [same code path](https://github.com/pion/webrtc/blob/31e3a1a6ac7ca66a3b4a1a741be79611fe7135e0/peerconnection.go#L1431) so a fix there should resolve them all.

### What did you expect?

SDP offers to match

### What happened?

Offers did not match, candidates were duplicated:

image

### LINKED ISSUE #1122: OnTrack is incorrectly fired for the same ssrc multiple times

Sean-Der:
* Don't add end-of-candidates until gathering complete
* Don't double add candidates

Resolves #1121


peerconnection_go_test.go
```diff
@@ -794,3 +794,45 @@ func TestPeerConnectionTrickle(t *testing.T) {
...
+				gatherCompleteCancel()
+			}
+		})
+
+		offer, err := pc.CreateOffer(nil)
+		assert.NoError(t, err)
+		assert.NotContains(t, offer.SDP, "a=candidate")
+		assert.NotContains(t, offer.SDP, "a=end-of-candidates")
+
+		assert.NoError(t, pc.SetLocalDescription(offer))
```
at-wat, line 831: Don't you need to validate
```go
    assert.NotContains(t, pc.PendingLocalDescription().SDP, "a=candidate")
    assert.NotContains(t, pc.PendingLocalDescription().SDP, "a=end-of-candidates")
```

audiolion:
Didn't realize you had made an issue as well. This then also closes #1121 ?

Sean-Der:
@audiolion I messed up the commit message, actually should be #1121 ! thanks for catching that.

peerconnection_go_test.go
Sean-Der: I am just asserting those entries don't exist if Gathering hasn't started yet. Sorry I might be missing what you want me to validate :/

at-wat: I thought that SDP in `PendingLocalDescription()` (not only in `offer`) before ICE gathering complete should not contain them, but it might be bit difficult to check certainly since ICE gathering is running asynchronously.

Sean-Der: So I *think* we are safe here.

The first check is done before ICE Gathering has even started. It is kicked off by SetLocalDescription

The second check we block until `OnGatheringStateChange` reaches complete.

PR #1128, opened 2020-04-04
Update examples to use AddTransceiverFromKind

bonprosoft:
#### Description
I found some example projects use `AddTransceiver`, which is a deprecated method.
So I updated those examples to use `AddTransceiverFromKind`.

#### Reference issue
None

---

I also found the method is used in `peerconnection` and some tests.
If there is no reason to use `AddTransceiver` in those codes, I am happy to fix them as well.


bonprosoft:
Thank you for your approval!

> I agree that the all code should use the new function instead of the deprecated one.

I am happy to hear that. I'll send a PR to fix them.

PR #1131, opened 2020-04-08
Fix potential issue of misinterpreting format string

egonelbre:
Fixes potential issue with calling non-formatting functions with strings that maybe interpreted as a format string. 

As an example `log.Trace("%v")n` would end up logging "%!v(MISSING)" instead.


egonelbre:
Done.

PR #1132, opened 2020-04-09
Fix #1001 The first RTP packet  is missing

### LINKED ISSUE #1001: The first RTP packet that triggers OnTrack wouldn't be received (trivial)
### Your environment.
  - Version: v2.1.18

### What did you do?

### What did you expect?

### What happened?

illuspas:
#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -784,10 +784,11 @@ func (pc *PeerConnection) startReceiver(incoming trackDetails, receiver *RTPRece
 		return
 	}
 
-	if err = receiver.Track().determinePayloadType(); err != nil {
-		pc.log.Warnf("Could not determine PayloadType for SSRC %d", receiver.Track().SSRC())
-		return
-	}
+	//This step results in the loss of the first RTP package
+// 	if err = receiver.Track().determinePayloadType(); err != nil {
+// 		pc.log.Warnf("Could not determine PayloadType for SSRC %d", receiver.Track().SSRC())
+// 		return
+// 	}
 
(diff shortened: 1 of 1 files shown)


peerconnection.go
```diff
@@ -797,17 +798,17 @@ func (pc *PeerConnection) startReceiver(incoming trackDetails, receiver *RTPRece
 		return
 	}
 
-	codec, err := pc.api.mediaEngine.getCodec(receiver.Track().PayloadType())
-	if err != nil {
-		pc.log.Warnf("no codec could be found for payloadType %d", receiver.Track().PayloadType())
+	codecs := pc.api.mediaEngine.GetCodecsByKind(incoming.kind)
```
at-wat, line 801: This may cause misselection of the codec when multiple codecs are registered for same kind. By default, 4 audio codecs and 3 video codecs are registered, and an other peer may select any ones from them. For example, if the other peer selects the second codec, this logic still select the first one.
https://github.com/pion/webrtc/blob/4220088d782ea9446907aae655c58822e35e9d2c/mediaengine.go#L43-L52

I think storing the first raw RTP packet fetched by `determinePayloadType()` to `Track` and return from `Track.Read()` is better. (It would be better to have a unit test if possible.)

illuspas:
Yes, this code does not completely solve the problem.
Just to demonstrate that the loss of the first package is caused by this reason.
Maybe we can cache the first package in the determinePayloadType method,
When the ReadRTP method is called, if the cache package exists, it is returned and cleared.

PR #1140, opened 2020-04-15
Filter out non-sending incoming tracks

jeremija:
This modification will only keep incoming tracks from media sections with `a=sendrecv` or `a=sendonly` attributes. This modification is not necessary for pion-to-pion peer connections, but for browser-to-pion connections using Unified Plan.

#### Background:

When a `RTCPeerConnection#removeTrack()` is called in the browser and the connection is renegotiated, the browser (tested in Firefox 75.0) will change the media section attribute to `a=recvonly` from `a=sendrecv`, or `a=inactive` from `a=sendonly`, but will keep the same
`a=ssrc:<ssrc>` associated with the same media (mid) section.

Previously, when a remote track obtained with Pion's `OnTrack()` handler was being read from, and this remote track was later removed in the browser, the track's `Read()` method would never return `io.EOF` (unless the peer connection was closed).

With this commit, the `trackDetailsFromSDP` function will only include remote track details whose media section has `a=sendonly` or `a=sendrecv` attributes, and previously existing logic will ensure that a receiver for that specific ssrc is stopped, so that `webrtc.Track#Read()` method returns `io.EOF`.

### DIFF
--- a/sdp.go
+++ b/sdp.go
@@ -25,6 +25,9 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 	rtxRepairFlows := map[uint32]bool{}
 
 	for _, media := range s.MediaDescriptions {
+		// Plan B can have multiple tracks in a signle media section
+		tracksInCurrentMediaSection := map[uint32]struct{}{}
+		remoteIsSending := false
 		trackID := ""
 		trackLabel := ""
 		for _, attr := range media.Attributes {
@@ -34,6 +37,10 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 			}
 
 			switch attr.Key {
+			case sdp.AttrKeySendRecv:
+				remoteIsSending = true
+			case sdp.AttrKeySendOnly:
+				remoteIsSending = true
 			case sdp.AttrKeySSRCGroup:
 				split := strings.Split(attr.Value, " ")
 				if split[0] == sdp.SemanticTokenFlowIdentification {
@@ -86,7 +93,17 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 					trackID = split[2]
 				}
 
+				// Plan B might send multiple a=ssrc lines under a single m= section. This is also why a single trackDetails{}
+				// is not defined at the top of the loop over s.MediaDescriptions.
 				incomingTracks[uint32(ssrc)] = trackDetails{codecType, trackLabel, trackID, uint32(ssrc)}
+				tracksInCurrentMediaSection[uint32(ssrc)] = struct{}{}
+			}
+		}
+
+		if !remoteIsSending {
+			for ssrc := range tracksInCurrentMediaSection {
+				// Only include a=sendrecv or a=sendonly track details. In other words, filter out a=recvonly and a=inactive.
+				delete(incomingTracks, ssrc)
 			}
 		}
 	}
--- a/sdp_test.go
+++ b/sdp_test.go
@@ -96,6 +96,7 @@ func TestTrackDetailsFromSDP(t *testing.T) {
 						Media: "foobar",
 					},
 					Attributes: []sdp.Attribute{
+						{Key: "sendrecv"},
 						{Key: "ssrc", Value: "1000 msid:unknown_trk_label unknown_trk_guid"},
 					},
 				},
@@ -104,6 +105,7 @@ func TestTrackDetailsFromSDP(t *testing.T) {
 						Media: "audio",
 					},
 					Attributes: []sdp.Attribute{
+						{Key: "sendrecv"},
 						{Key: "ssrc", Value: "2000 msid:audio_trk_label audio_trk_guid"},
 					},
 				},
@@ -112,6 +114,7 @@ func TestTrackDetailsFromSDP(t *testing.T) {
 						Media: "video",
 					},
 					Attributes: []sdp.Attribute{
+						{Key: "sendrecv"},
 						{Key: "ssrc-group", Value: "FID 3000 4000"},
 						{Key: "ssrc", Value: "3000 msid:video_trk_label video_trk_guid"},
 						{Key: "ssrc", Value: "4000 msid:rtx_trk_label rtx_trck_guid"},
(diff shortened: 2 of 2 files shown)


jeremija:
Should I modify `pion/sdp` to include `AttrKeySendRecv` and `AttrKeySendonly` and use those constants instead of string literals `"sendrecv"` and `"sendonly"`, like I did for [`a=msid`](https://github.com/pion/sdp/pull/47)?

sdp.go
```diff
@@ -25,6 +25,8 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 	rtxRepairFlows := map[uint32]bool{}
 
 	for _, media := range s.MediaDescriptions {
+		tmpIncomingTracks := map[uint32]trackDetails{}
```
Sean-Der, line 28: Instead of `tmpIncomingTracks` would it be possible to filter+delete? Just to remove the extra variable declare, and so you don't need to do things like

```
						delete(incomingTracks, uint32(rtxRepairFlow))
￼						delete(tmpIncomingTracks, uint32(rtxRepairFlow))
```

Sean-Der:
> Should I modify `pion/sdp` to include `AttrKeySendRecv` and `AttrKeySendOnly` and use those constants instead of string literals `"sendrecv"` and `"sendonly"`, like I did for [`a=msid`](https://github.com/pion/sdp/pull/47)?

Adding more constants would be amazing!

sdp.go
jeremija: Hm, originally this code looked a lot nicer and there was no temporary map, but it broke one test for Plan B. If I remove the `tmpIncomingTracks`, I don't think I'd be able to filter out tracks for Plan B like:

```
m=video
a=ssrc:1 msid:label1 id1
a=ssrc:2 msid:label2 id2
a=recvonly
```

I could keep a set of track SSRCs to be kept, like `rtxRepairFlows` and then delete the rest from `incomingTracks` at the end. Would this work?

jeremija: This is what I was thinking about:

```diff
diff --git a/sdp.go b/sdp.go
index 071435c..807a2b5 100644
--- a/sdp.go
+++ b/sdp.go
@@ -25,7 +25,8 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 	rtxRepairFlows := map[uint32]bool{}
 
 	for _, media := range s.MediaDescriptions {
-		tmpIncomingTracks := map[uint32]trackDetails{}
+		// Plan B can have multiple tracks in a signle media section
+		tracksInCurrentMediaSection := map[uint32]struct{}{}
 		remoteIsSending := false
 		trackID := ""
 		trackLabel := ""
@@ -60,7 +61,6 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 						}
 						rtxRepairFlows[uint32(rtxRepairFlow)] = true
 						delete(incomingTracks, uint32(rtxRepairFlow)) // Remove if rtx was added as track before
-						delete(tmpIncomingTracks, uint32(rtxRepairFlow))
 					}
 				}
 
@@ -84,9 +84,6 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 				if rtxRepairFlow := rtxRepairFlows[uint32(ssrc)]; rtxRepairFlow {
 					continue // This ssrc is a RTX repair flow, ignore
 				}
-				if existingValues, ok := tmpIncomingTracks[uint32(ssrc)]; ok && existingValues.label != "" && existingValues.id != "" {
-					continue // This ssrc is already fully defined
-				}
 
 				if len(split) == 3 && strings.HasPrefix(split[1], "msid:") {
 					trackLabel = split[1][len("msid:"):]
@@ -95,14 +92,15 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) m
 
 				// Plan B might send multiple a=ssrc lines under a single m= section. This is also why a single trackDetails{}
 				// is not defined at the top of the loop over s.MediaDescriptions.
-				tmpIncomingTracks[uint32(ssrc)] = trackDetails{codecType, trackLabel, trackID, uint32(ssrc)}
+				incomingTracks[uint32(ssrc)] = trackDetails{codecType, trackLabel, trackID, uint32(ssrc)}
+				tracksInCurrentMediaSection[uint32(ssrc)] = struct{}{}
 			}
 		}
 
-		for ssrc, incoming := range tmpIncomingTracks {
-			if remoteIsSending {
+		if !remoteIsSending {
+			for ssrc, _ := range tracksInCurrentMediaSection {
 				// Only include a=sendrecv or a=sendonly track details. In other words, filter out a=recvonly and a=inactive.
-				incomingTracks[ssrc] = incoming
+				delete(incomingTracks, ssrc)
 			}
 		}
 	}
```

Sean-Der:
@jeremija This LGTM!

We should probably be doing the same then? Currently when removing tracks we do everything wrong it seems.

jeremija:
Thanks @Sean-Der!

> We should probably be doing the same then? 

I'm not sure what you're referring to, could you please clarify?

Sean-Der:
> Thanks @Sean-Der!
> 
> > We should probably be doing the same then?
> 
> I'm not sure what you're referring to, could you please clarify?

We completely remote the transceivers (and don't pluck out the SSRC value)

jeremija:
I started working on this as a workaround from not being able to detect when a remote track was removed. However, according to the [Unified Plan Transition Guide (JavaScript)](https://docs.google.com/document/d/1-ZfikoUtoJa9k-GZG1daN0BU3IjIanQ_JSscHxQesvU/edit#heading=h.aak3wsw2d7k1), tracks are never actually removed, only muted. I think the most correct solution would be to detect track `mute` and `unmute` events.

Unfortunately, this seems quite complicated because of reasons I've already described at the bottom of issue #1155. Copying here again for reference:

From the [WebRTC 1.0 spec](https://w3c.github.io/webrtc-pc/#mediastreamtrack-network-use):

> Whenever an RTCRtpReceiver receives data on an RTP source whose corresponding MediaStreamTrack is muted, but not ended, and the [[Receptive]] slot of the RTCRtpTransceiver object the RTCRtpReceiver is a member of is true, it MUST queue a task to set the muted state of the corresponding MediaStreamTrack to false.
> 
> When one of the SSRCs for RTP source media streams received by an RTCRtpReceiver is removed either due to reception of a BYE or via timeout, it MUST queue a task to set the muted state of the corresponding MediaStreamTrack to true. Note that setRemoteDescription can also lead to the setting of the muted state of the track to the value true.

Here's where it gets complicated - the [spec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-track) mandates that:

> **track** of type **MediaStreamTrack**, _readonly_, _nullable_
> 
> The track attribute is the track that is associated with this RTCRtpSender object. If track is ended, or if the track's output is disabled, i.e. the track is disabled and/or muted, the RTCRtpSender MUST send black frames (video) and MUST NOT send (audio). In the case of video, the RTCRtpSender SHOULD send one black frame per second. If track is null then the RTCRtpSender does not send. On getting, the attribute MUST return the value of the [[SenderTrack]] slot.

Does this mean that in order to support the `mute` and `unmute` events, we'd need be able to encode video in any format supported by `MediaEngine`?

Sean-Der:
Moved to https://github.com/pion/webrtc/pull/1159 (no changes just need write access to branch)

PR #1141, opened 2020-04-15
Populate network types from config

r-novel:
Added function for setting network types from external side;

#### Description

This patch allows to populate network types from external user's side. It might be data from yaml config or simple string slice.
Since function for creating network type from string is private we can't move this function to user side.


Sean-Der:
Hey @r-novel thank you so much for the patch.

What do you think of exporting the function to do `string` -> `NetworkType` instead? Since we have `SetNetworkTypes` already this will cause some duplication. Pushback if you disagree, totally subjective :)

PR #1142, opened 2020-04-16
Issue 494 followup

### LINKED ISSUE #494: Support For Creating an Answer with Passive Negotiation

jbert:
#### Description
Set default dtlsRole for answer client rather than server.

This is done to change the sdp ConnectionRole to  'active' rather than 'passive' when answering, as is recommended by the RFC (details in comments).

#### Reference issue
Fixes #494 follow up issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Egon Elbre](https://github.com/egonelbre)
 * [Jerko Steiner](https://github.com/jeremija)
 * [Roman Romanenko](https://github.com/r-novel)
+* [John Berthels](https://github.com/jbert)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/dtlsrole.go
+++ b/dtlsrole.go
@@ -21,8 +21,22 @@ const (
 )
 
 const (
-	defaultDtlsRoleAnswer = DTLSRoleServer
-	defaultDtlsRoleOffer  = DTLSRoleAuto
+	// https://tools.ietf.org/html/rfc5763
+	/*
+		The answerer MUST use either a
+		setup attribute value of setup:active or setup:passive.  Note that
+		if the answerer uses setup:passive, then the DTLS handshake will
+		not begin until the answerer is received, which adds additional
+		latency. setup:active allows the answer and the DTLS handshake to
+		occur in parallel.  Thus, setup:active is RECOMMENDED.
+	*/
+	defaultDtlsRoleAnswer = DTLSRoleClient
+	/*
+		The endpoint that is the offerer MUST use the setup attribute
+		value of setup:actpass and be prepared to receive a client_hello
+		before it receives the answer.
+	*/
+	defaultDtlsRoleOffer = DTLSRoleAuto
 )
 
 func (r DTLSRole) String() string {


jbert:
Tests failing - I'll close this until I can work out what's going on. Sorry for noise.

Sean-Der:
@jbert I am happy to fix the tests! My guess is that they are just flakes, this really shouldn't effect anything.

If we have tests that do depend on the role we can force behavior via the SettingEngine for testing.

jbert:
Hi @Sean-Der  You're right, altering the DTLS role via the setting engine recreates the test failure:

`
diff --git a/e2e/e2e_test.go b/e2e/e2e_test.go
index 366bee4..a9f1f9f 100644
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -345,7 +345,14 @@ func createTrack(offer webrtc.SessionDescription) (*webrtc.PeerConnection, *webr
        if payloadType == 0 {
                return nil, nil, nil, errors.New("Remote peer does not support VP8")
        }
-       api := webrtc.NewAPI(webrtc.WithMediaEngine(mediaEngine))
+
+       settingEngine := webrtc.SettingEngine{}
+       settingEngine.SetAnsweringDTLSRole(webrtc.DTLSRoleClient)
+
+       api := webrtc.NewAPI(
+               webrtc.WithMediaEngine(mediaEngine),
+               webrtc.WithSettingEngine(settingEngine),
+       )
        pc, errPc := api.NewPeerConnection(webrtc.Configuration{})
        if errPc != nil {
                return nil, nil, nil, errPc
`
`
$ time go test -tags=e2e -v ./e2e/
...
=== RUN   TestE2E_Audio/Chrome
    TestE2E_Audio/Chrome: e2e_test.go:147: Number of OPUS packets is expected to be: 50/second, got: 0/second
--- FAIL: TestE2E_Audio (6.44s)
`

Do you have any suggestions on the best way to progress?

I could adapt the TestE2E_Audio test to run in both DTLSClient and DTLSServer mode to expose the problem (I think this would add another 6-7s to test runtime though) - that would give us a branch with broken E2E tests. Then I guess it's either a test bug, a chrome bug or a @jbert 's understanding of the RFC bug?

Sean-Der:
Hey @jbert 

This turned out to be a bug in `pion/dtls` [here](https://github.com/pion/dtls/pull/265)! I fixed it and will be updating the default state with the RFC you cited.

Sorry for not following up on this sooner. My backlog keeps growing by the day :)

jbert:
That is fantastic. Thank you!

Thank you for not minding that I didn't push this to completion. I got as far as the failing interop tests with headless chromium which shook my confidence that this change was actually correct.

One thing which might be interesting is that - in principle - this reduces the time-to-first-frame for a client calling pions. We don't need to wait for the SDP answer to traverse the signalling pathway before starting the setup.

I don't know if we have any measurements on this, but it would be cool to know if this was the case in practice. (Speedup should be of the order or "time to return SDP offer").

Sean-Der:
@jbert thanks for using Pion and contributing, it is super motivating to have people use it!

> would be cool to know if this was the case i
Yea that would be amazing to know! I personally am only using Pion for testing/one-off things so I don't have that data :(

If there is anything I can do to make Pion better I would love to hear! Always looking to improve things.

PR #1150, opened 2020-04-18
Fix empty response bug in pion-to-pion/answer example if pion-to-pion/offer is run from Chrome

### LINKED ISSUE #1149: examples/pion-to-pion: Running "offer" example in the browser doesn't work
### Your environment.
  - Version: 208e06832be2a59ec6e4de5d5bf323b85086b812
  - Browser: Chrome (Latest), Brave (Latest)
  - Other Information Noted how I'll fix this below

### What did you do?
1) Run "answer" server with `go build && ./answer` in Bash on Windows
2) Build "offer" as WASM and run in browser.

### What did you expect?
I expected "offer" to work and not error. Instead the server got an "EOF" error and "offer" got an error from the fetch request.

### What happened?
The server got an "EOF" error and "offer" got an error from the fetch request.

### How to fix this
Add something like the following code to the HandleFunc callback so that CORS works and that the server gets the request correctly.
```go
if r.Body == nil {
	http.Error(w, "Please send a request body", 400)
	return
}
w.Header().Set("Access-Control-Allow-Origin", "*")
w.Header().Set("Access-Control-Allow-Methods", http.MethodPost)
w.Header().Set("Access-Control-Allow-Headers", "Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization")
if r.Method == http.MethodOptions {
	return
}
if r.Method != http.MethodPost {
	http.Error(w, "Please send a "+http.MethodPost+" request", 400)
	return
}
```

silbinarywolf:
#### Description

Add CORS code to the answer example code
Prefix default "address" flag with "127.0.0.1" so that http.Post works in the browser

#### Reference issue
Fixes #1149

### DIFF
--- a/examples/pion-to-pion/offer/main.go
+++ b/examples/pion-to-pion/offer/main.go
@@ -14,7 +14,7 @@ import (
 )
 
 func main() {
-	addr := flag.String("address", ":50000", "Address that the HTTP server is hosted on.")
+	addr := flag.String("address", "127.0.0.1:50000", "Address that the HTTP server is hosted on.")
 	flag.Parse()
 
 	// Everything below is the Pion WebRTC API! Thanks for using it ❤️.
--- a/examples/pion-to-pion/answer/main.go
+++ b/examples/pion-to-pion/answer/main.go
@@ -101,6 +101,21 @@ func mustSignalViaHTTP(address string) (offerOut chan webrtc.SessionDescription,
 	answerIn = make(chan webrtc.SessionDescription)
 
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		if r.Body == nil {
+			http.Error(w, "Please send a request body", 400)
+			return
+		}
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+		w.Header().Set("Access-Control-Allow-Methods", http.MethodPost)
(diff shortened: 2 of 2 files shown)


examples/pion-to-pion/offer/main.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func main() {
-	addr := flag.String("address", ":50000", "Address that the HTTP server is hosted on.")
+	addr := flag.String("address", "127.0.0.1:50000", "Address that the HTTP server is hosted on.")
```
Sean-Der, line 17: Mind leaving it as `:50000`? Beyond that PR looks perfect to me! Thank you so much for the contribution :)

silbinarywolf: I could, but then the `fetch()` will still fail if this is run in a browser.
Technically I guess its a problem with `http.Post` but it might be worth working around here I think.

Thoughts?

Sean-Der: Aye! That makes sense to me, will merge now :)

Thanks for fixing

Sean-Der:
Closed with 5a702b6c3173db62ad90180929854ea2f701601b

Thank you so much @silbinarywolf I just fixed lint error (longer functions can't do naked returns)

Issue #1148, opened 2020-04-18
Order of Transceiver / Track association sometimes changes after track is removed


jeremija:
Could be related to my changes in https://github.com/pion/webrtc/pull/1140, since this happens after I remove a track.

### Your environment.
  - Version: d5dbe1f
  - Browser: N/A

### What did you do?

I called `peerConnection.RemoveTrack()` and renegotiated.

### What did you expect?

I expected `a=mid`/`a=ssrc` associations to stay the same.

### What happened?
Looks like the order of `a=mid`/`a=ssrc` changed.

### Before SDP

I've removed the `a=candidate` and `a=rtpmap` lines, and added spaces between `m=` sections to make the SDPs easier to read. 

An offer before the track `sfu_{82cc6157-3738-3447-9952-93bcf6adbe32}` (ssrc: `197955246`) was removed:

```
2020-04-18T08:48:26.569854+02:00 [            sdp] [VDYSr00cAfqrnx02zsxjA6] Local signal.type: offer, signal.sdp: v=0
o=- 842954722 1587192506 IN IP4 0.0.0.0
s=-
t=0 0
a=fingerprint:sha-256 86:CF:28:50:97:06:CB:AB:06:4B:33:20:57:85:E3:9E:DF:81:B2:B3:A0:94:E4:C9:67:41:2E:12:D7:32:BF:7A
a=group:BUNDLE 0 1 2 3 4 5

m=video 9 UDP/TLS/RTP/SAVPF 96 98 102
c=IN IP4 0.0.0.0
a=setup:actpass
a=mid:0
a=ice-ufrag:zQrcqINzoemkGBQB
a=ice-pwd:vUWlevctbIMZJLaJjVBbccKjVVXwyZDj
a=rtcp-mux
a=rtcp-rsize
a=fmtp:102 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f
a=ssrc:2596996162 cname:UaSxtEzjIuBvsjNK
a=ssrc:2596996162 msid:UaSxtEzjIuBvsjNK ZaYuzSRctwZRCnEN
a=ssrc:2596996162 mslabel:UaSxtEzjIuBvsjNK
a=ssrc:2596996162 label:ZaYuzSRctwZRCnEN
a=msid:UaSxtEzjIuBvsjNK ZaYuzSRctwZRCnEN
a=sendrecv
a=end-of-candidates

m=audio 9 UDP/TLS/RTP/SAVPF 111 0 8 9
[...]

PR #1156, opened 2020-04-19
Ensure existing transceivers are reused

### LINKED ISSUE #1155: Transceivers are never reused for sending
### Your environment.
  - Version: 234a170
  - Browser: N/A

### What did you do?

```golang
package webrtc
import (
	"math/rand"
	"testing"
	"github.com/stretchr/testify/assert"
)
func TestAddTrans(t *testing.T) {
	webrtcConfig := Configuration{}
	settingEngine := SettingEngine{}
	settingEngine.SetTrickle(false)
	mediaEngine := MediaEngine{}
	mediaEngine.RegisterDefaultCodecs()
	api := NewAPI(
		WithMediaEngine(mediaEngine),
		WithSettingEngine(settingEngine),
	)
	pcOffer, err := api.NewPeerConnection(webrtcConfig)
	assert.NoError(t, err)
	tr, err := pcOffer.AddTransceiverFromKind(
		RTPCodecTypeVideo,
		RtpTransceiverInit{Direction: RTPTransceiverDirectionRecvonly},
	)
	assert.NoError(t, err)
	assert.Equal(t, []*RTPTransceiver{tr}, pcOffer.GetTransceivers())
	vp8Track, err := pcOffer.NewTrack(DefaultPayloadTypeVP8, rand.Uint32(), "foo", "bar")
	assert.NoError(t, err)
	_, err = pcOffer.AddTrack(vp8Track)
	assert.NoError(t, err)

	assert.Equal(t, 1, len(pcOffer.GetTransceivers()))
}
```

### What did you expect?

I expected the existing transceiver to be used for this track

### What happened?

The logic in `peerConnection.AddTrack` only reuses a transceiver when a `Sender() !=nil`. The `sender` will be `nil` in a `Recvonly` transceiver. It is not clear to me why a check for `t.Receiver()` is being made [here](https://github.com/pion/webrtc/blob/234a170fb43eef5edbd89f5f77b9c5d812724dcb/peerconnection.go#L1105-L1107).

This might be more complicated. A fix could change the Track ssrc and ID/Labels after a new track is added and connection renegotiatiated, even though there is a bug report that this already happens: #1148.

Also, the Unified Plan mandates that:

> transceiver _always_ consists of both a sender and a receiver

But the current code-base removes a sender after a track is removed. Please see below for more details.

---

From the [Unified Plan Transition Guide](https://docs.google.com/document/d/1-ZfikoUtoJa9k-GZG1daN0BU3IjIanQ_JSscHxQesvU/edit#):

> # The RTCRtpTransceiver APIs and Changes in Behavior
> A transceiver represents an m= section in SDP. Because an m= section can be used for both sending and receiving, a transceiver always consists of both a sender and a receiver. The direction can change between “inactive”, “sendonly”, “recvonly” and “sendrecv”. A track can be attached to the sender, it may be replaced (with replaceTrack()) or the sender can have no track attached.
> 
> When a track is added to a peer connection with addTrack() or addTransceiver(), a transceiver is created (or reused) and the track is attached to the sender. Because a transceiver may be used for receiving, a receiving track is always created and attached to the receiver. Unlike the sender, the receiver’s track is always there and cannot be replaced, but it is muted by default.
> 
> setLocalDescription() and setRemoteDescription() permanently ties transceivers to m= sections by setting the value of transceiver.mid to the m=section’s mid. This is a string identifier that can be used by both endpoints to correlate transceivers and thus which track is which. Applying remote SDP may cause transceivers to be created or updated and RTCPeerConnection.ontrack to fire for each remote track that should now receive. The track in question is a receiver’s track, which may or may not have been created at an earlier stage due to reusing transceivers, such as one created at addTrack(). The track will unmute; if a subsequent setRemoteDescription() changes the transceiver’s direction not to receive anymore, the track is muted again.
> 
> ## Local and Remote Track IDs Do Not Match
> In Plan B, a sender is created for each local track that is added and a receiver is created for each remote track that is negotiated, the track IDs of local and remote endpoints match.
> 
> In Unified Plan, senders and receivers are created in pairs of transceivers, and a transceiver.receiver.track may have been created long before remote SDP offers a track. As such, the track that is fired in RTCPeerConnection.ontrack is no longer guaranteed to have an ID matching the sender-side version of the track. Furthermore, because of addTransceiver() and replaceTrack(), a track may be sent multiple times. Track IDs are misleading, and shouldn’t be assumed to match. Instead, transceiver.mid should be used to correlate local and remote tracks. You may have to “setLocalDescription(answer)” before the mid is known.
> 
> ## Remote Tracks Are Never Truly “Removed”
> In Plan B, RTCPeerConnection.removeTrack() removes the sender and track, and during negotiation the other endpoint removes the corresponding receiver and track.
> 
> In Unified Plan, removeTrack() changes the direction of the transceiver and nulls the sender’s track. During negotiation, the other endpoint changes the corresponding transceiver’s direction and mutes the track but it is never removed, and it may be reused in the future. See code snippet.
> MediaStreamTrack.onended No Longer Fires
> Because remote tracks may be muted but are never removed, MediaStreamTrack.onended no longer fires. Instead, rely on MediaStreamTrack.onmute to discover if a track is removed. See code snippet.

---

I'm not sure if we'd have to add the handling of Track `mute` and `unmute` events to fully support reuse of transceivers. From the [WebRTC 1.0 spec](https://w3c.github.io/webrtc-pc/#mediastreamtrack-network-use):

> Whenever an RTCRtpReceiver receives data on an RTP source whose corresponding MediaStreamTrack is muted, but not ended, and the [[Receptive]] slot of the RTCRtpTransceiver object the RTCRtpReceiver is a member of is true, it MUST queue a task to set the muted state of the corresponding MediaStreamTrack to false.
> 
> When one of the SSRCs for RTP source media streams received by an RTCRtpReceiver is removed either due to reception of a BYE or via timeout, it MUST queue a task to set the muted state of the corresponding MediaStreamTrack to true. Note that setRemoteDescription can also lead to the setting of the muted state of the track to the value true.

Here's where it gets complicated - the [spec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-track) mandates that:

> **track** of type **MediaStreamTrack**, _readonly_, _nullable_
> 
> The track attribute is the track that is associated with this RTCRtpSender object. If track is ended, or if the track's output is disabled, i.e. the track is disabled and/or muted, the RTCRtpSender MUST send black frames (video) and MUST NOT send (audio). In the case of video, the RTCRtpSender SHOULD send one black frame per second. If track is null then the RTCRtpSender does not send. On getting, the attribute MUST return the value of the [[SenderTrack]] slot.

Does this mean that in order to support the `mute` and `unmute` events, we'd need be able to encode video in any format supported by `MediaEngine`?

jeremija:
#### Description

This might cause problems for existing users who have gotten used to the fact that a new transceiver will always be created when `AddTrack` is called.

I'm not sure if we can simply merge it to master and tag as a patch/minor release. Would greatly appreciate feedback!

#### Reference issue
Fixes #1155

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1100,20 +1100,22 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 
 	var transceiver *RTPTransceiver
 	for _, t := range pc.GetTransceivers() {
-		if !t.stopped && t.Sender() != nil &&
-			!t.Sender().hasSent() &&
-			t.Receiver() != nil &&
-			t.Receiver().Track() != nil &&
-			t.Receiver().Track().Kind() == track.Kind() {
+		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil {
 			transceiver = t
 			break
 		}
 	}
 	if transceiver != nil {
+		sender, err := pc.api.NewRTPSender(track, pc.dtlsTransport)
+		if err != nil {
+			return nil, err
+		}
+		transceiver.setSender(sender)
+		// we still need to call setSendingTrack to ensure direction has changed
 		if err := transceiver.setSendingTrack(track); err != nil {
 			return nil, err
 		}
-		return transceiver.Sender(), nil
+		return sender, nil
 	}
 
 	transceiver, err := pc.AddTransceiverFromTrack(track)
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/pion/transport/test"
 	"github.com/pion/webrtc/v2/pkg/media"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func offerMediaHasDirection(offer SessionDescription, kind RTPCodecType, direction RTPTransceiverDirection) bool {
@@ -227,7 +228,7 @@ func TestPeerConnection_Media_Shutdown(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	_, err = pcOffer.AddTransceiverFromKind(RTPCodecTypeVideo, RtpTransceiverInit{Direction: RTPTransceiverDirectionRecvonly})
+	_, err = pcOffer.AddTransceiverFromKind(RTPCodecTypeAudio, RtpTransceiverInit{Direction: RTPTransceiverDirectionRecvonly})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -279,7 +280,7 @@ func TestPeerConnection_Media_Shutdown(t *testing.T) {
 	<-iceCompleteAnswer
 	<-iceCompleteOffer
 
-	// Each PeerConnection should have one sender, one receiver and two transceivers
+	// Each PeerConnection should have one sender, one receiver and one transceiver
 	for _, pc := range []*PeerConnection{pcOffer, pcAnswer} {
 		senders := pc.GetSenders()
 		if len(senders) != 1 {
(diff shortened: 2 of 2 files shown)


Sean-Der:
Really sorry this didn't get reviewed @jeremija today :( I got sucked into work stuff all Sunday.

Promise I will get to this ASAP!

jeremija:
@Sean-Der No worries, it's no rush! I actually did some more coding yesterday and today and have implemented `Transceiver.Mid()` and I think I'm finally not getting any transceiver reorders in SDP (hopefully it fixes #1148). 

The branch I'm working on has both the code from this PR (fixes #1155), and #1140.

If you like, I can submit another PR and close the other PRs I made.

Issue #1020, opened 2020-02-11
Support RTPSender.ReplaceTrack()


at-wat:
In Web API, `Track` connected to `RTPSender` can be replaced without renegotiation.
For example, it is used to switch between front/rear cameras.
https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/replaceTrack

In the currently implementation, `Track` which is not bound to `RTPSender` returns error.
https://github.com/pion/webrtc/blob/2b212f8a7c4feed4621272bf8309637827e695cd/track.go#L155-L157

- Add `ReplaceTrack()` to `RTPSender` which removes itself from the currently bound `Track` and add to the new `Track`
- Allow `Track.WriteRTP()` to be called on `Track` without `RTPSender`

(It will be useful when using with https://github.com/pion/mediadevices)

Sean-Der:
* 👍 from me on ReplaceTrack, that would be really helpful
* Can we just have the caller ignore `io.ErrClosedPipe`? In the past users were confused why video wasn't showing up. They would create a `Track` and just expect it to work. By returning an error it helped users debugging a lot.

I do think we need to revisit the Media API entirely. This could be a good kicking off point. Stuff like we shouldn't be story the PayloadType/SSRC on tracks at all.

at-wat:
https://github.com/pion/webrtc/issues/1020#issue-563147171 is a kind of a workaround to do it on v2 API.
https://github.com/pion/webrtc/issues/1025 may be a better solution to implement `ReplaceTrack` easily.

cretz:
Are there any plans to implement this in v2 (or is v3 coming anytime soon)? I am considering adding it myself and have not dug into the specs, but any idea what `ReplaceTrack` look like? I see [the spec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack) but I fear details like checking whether renegotiation is needed and what "switch seamlessly" entails.

at-wat:
@cretz  I don't have specific plan yet. It's very helpful if you can dig into this.

> checking whether renegotiation is needed

MDN said:
https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/replaceTrack
> The new track must be of the same media kind (audio, video, etc) and switching the track should not require negotiation.

cretz:
[...]

PR #1166, opened 2020-04-25
Swap tracks example

cretz:
Example showing how to simulate `RTPSender.ReplaceTrack` as mentioned in #1020

### DIFF
--- a/examples/swap-tracks/main.go
+++ b/examples/swap-tracks/main.go
@@ -1,0 +1,196 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"math/rand"
+	"time"
+
+	"github.com/pion/rtcp"
+	"github.com/pion/rtp"
+	"github.com/pion/webrtc/v2"
+	"github.com/pion/webrtc/v2/examples/internal/signal"
+)
+
+func main() {
+	// Everything below is the Pion WebRTC API! Thanks for using it ❤️.
+
+	// Wait for the offer to be pasted
+	offer := webrtc.SessionDescription{}
+	signal.Decode(signal.MustReadStdin(), &offer)
+
+	// We make our own mediaEngine so we can place the sender's codecs in it. Since we are echoing their RTP packet
+	// back to them we are actually codec agnostic - we can accept all their codecs. This also ensures that we use the
+	// dynamic media type from the sender in our answer.
+	mediaEngine := webrtc.MediaEngine{}
+
+	// Add codecs to the mediaEngine. Note that even though we are only going to echo back the sender's video we also
+	// add audio codecs. This is because createAnswer will create an audioTransceiver and associated SDP and we currently
+	// cannot tell it not to. The audio SDP must match the sender's codecs too...
+	err := mediaEngine.PopulateFromSDP(offer)
+	if err != nil {
+		panic(err)
+	}
+
+	videoCodecs := mediaEngine.GetCodecsByKind(webrtc.RTPCodecTypeVideo)
+	if len(videoCodecs) == 0 {
+		panic("Offer contained no video codecs")
+	}
+
+	api := webrtc.NewAPI(webrtc.WithMediaEngine(mediaEngine))
+
+	// Prepare the configuration
+	config := webrtc.Configuration{
+		ICEServers: []webrtc.ICEServer{
+			{
+				URLs: []string{"stun:stun.l.google.com:19302"},
+			},
+		},
+	}
+	// Create a new RTCPeerConnection
+	peerConnection, err := api.NewPeerConnection(config)
+	if err != nil {
+		panic(err)
+	}
+
+	// Create Track that we send video back to browser on
+	outputTrack, err := peerConnection.NewTrack(videoCodecs[0].PayloadType, rand.Uint32(), "video", "pion")
+	if err != nil {
+		panic(err)
+	}
+
+	// Add this newly created track to the PeerConnection
+	if _, err = peerConnection.AddTrack(outputTrack); err != nil {
+		panic(err)
+	}
+
+	// In addition to the implicit transceiver added by the track, we add two more
+	// for the other tracks
+	_, err = peerConnection.AddTransceiverFromKind(webrtc.RTPCodecTypeVideo,
+		webrtc.RtpTransceiverInit{Direction: webrtc.RTPTransceiverDirectionRecvonly})
+	if err != nil {
+		panic(err)
+	}
+	_, err = peerConnection.AddTransceiverFromKind(webrtc.RTPCodecTypeVideo,
+		webrtc.RtpTransceiverInit{Direction: webrtc.RTPTransceiverDirectionRecvonly})
+	if err != nil {
+		panic(err)
+	}
+
+	// Set the remote SessionDescription
+	err = peerConnection.SetRemoteDescription(offer)
+	if err != nil {
+		panic(err)
+	}
+
+	// Which track is currently being handled
+	currTrack := 0
+	// The total number of tracks
+	trackCount := 0
+	// The channel of packets with a bit of buffer
+	packets := make(chan *rtp.Packet, 60)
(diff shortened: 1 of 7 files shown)


Sean-Der:
This is fantastic! Thank you so much @cretz :)

Merging now

examples/swap-tracks/main.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// Output the answer in base64 so we can paste it in browser
+	fmt.Printf("Paste below base64 in browser:\n%v\n", signal.Encode(answer))
+
+	// Asynchronously take all packets in the channel and write them out to our
+	// track
+	go func() {
```
Sean-Der, line 165: What do you think of dropping the channel communication and doing the input/output in the OnTrack read loop? The indirection might confuse people new to Go (who don't know channels yet)

Conceptually I think I see two code blocks

* Firing something every 5 seconds 
* Reading the packets, updating the SSRC/SequenceNumber and pushing

examples/swap-tracks/main.go
```diff
@@ -0,0 +1,196 @@
...
+				onMe = false
+			}
+		}
+	})
+	// Set the handler for ICE connection state and update chan if connected
+	connected := make(chan struct{}, 1)
+	peerConnection.OnICEConnectionStateChange(func(connectionState webrtc.ICEConnectionState) {
+		fmt.Printf("Connection State has changed %s \n", connectionState.String())
+		// Non-blocking just in case already sent
+		if connectionState == webrtc.ICEConnectionStateConnected {
```
Sean-Der, line 140: I would do a context, I think you can call done as many times as you want?

Sean-Der:
This is really amazing @cretz 

I don't very often see demos in WebRTC that makes me go 'Damn that is cool', but this does :) the animation is so fluid, and the switching is so perfect. Really fantastic work.

I am excited to share this on Twitter, you have a lot of cool concepts going on here that people should see. Would you be interested in joining the Pion org? We have a talked a few times, but seeing this really makes me want to get you more involved :)

examples/swap-tracks/main.go
cretz: This is what I had originally, but keeping the shared timestamp and sequence numbers intact w/ atomic updates from either side became a bit confusing and not as guaranteed to keep packets (or I'd have to lock the entire loop iteration). But I can change it back to that if I must,.

In this use case, I think the benefit using the channel to combine multiple sources outweighs the cost of somehow a Go dev not knowing about them.

examples/swap-tracks/main.go
cretz: There are a ton of ways to skin this cat. Could use a context cancel func, could use an atomic value, could use `sync.Once` with simple chan with `Do(func() { close(doneCh) })`, could use `sync.Once` with `sync.WaitGroup` and `Do(wg.Done)`, could use `sync.Cond`, etc, etc.

If it's much clearer with context cancellation, I can use that (I tend not to use it as a notifier as it does read as clear as context's common uses).

cretz:
> Would you be interested in joining the Pion org? We have a talked a few times, but seeing this really makes me want to get you more involved :)

I am in awe of your dedication, but I fear I don't share it :-) I come back to this project every so often, make a toy, move on, heh. I fear me joining the org would be doing it a disservice as I can just go radio silent here and there. Offer appreciated though.

Here is some code I hacked together for a project I'm working on (untested yet) that lets you switch an inbound stream if it helps anyone:

```go

import (
	"context"
	"io"
	"sync/atomic"

	"github.com/pion/rtp"
	"github.com/pion/webrtc/v2"
)

// TrackSink pipes an incoming track to an outgoing one and allows the incoming
// track to be changed.
type TrackSink struct {
	out            interface{ WriteRTP(*rtp.Packet) error }
	outSSRC        uint32
	inTrack        atomic.Value
	inTrackUpdated chan struct{}
}

// NewTrackSink creates a new TrackSink for the given out track (given as a
// writer and a SSRC value).
func NewTrackSink(out interface{ WriteRTP(*rtp.Packet) error }, outSSRC uint32) *TrackSink {
	return &TrackSink{
		out:     out,
		outSSRC: outSSRC,
		// Buffer one
		inTrackUpdated: make(chan struct{}, 1),
	}
}

type inTrack struct {
	track         *webrtc.Track
	lastTimestamp uint32
}

// SetInTrack changes the current incoming track to the given one. If a nil
// track is given, the outbound writes are stopped until a non-nil track is
// given.
func (t *TrackSink) SetInTrack(track *webrtc.Track) {
	// Change and signal non-blocking
	t.inTrack.Store(&inTrack{track: track})
	select {
	case t.inTrackUpdated <- struct{}{}:
	default:
	}
}

// Run continuously writes to the outbound track as needed until the context is
// closed or an error occurs. If an error occurs on read of a track, onReadError
// is invoked if it's non-nil. Only if the result of onReadError is non-nil will
// the process be stopped and that error returned. Otherwise the read error is
// swalled. All write errors stop this process and are returned. If the context
// is closed, the process is stopped and that error is returned. This always
// returns a non-nil error.
func (t *TrackSink) Run(ctx context.Context, onReadError func(error) error) error {
	// Keep track of the timestamp
	var currTimestamp, lastTimestampDelta uint32
	// The byte buffer and rtp packet to reuse
	buf := make([]byte, 1460)
	var rtpPacket rtp.Packet
	// Keep track of last failed reader to ignore
	var lastFailedIn *inTrack
	// Sequence number to increment each successful packet
	var seqNum uint16
	// Run until context canceled
	for ctx.Err() == nil {
		// Grab in track
		in, _ := t.inTrack.Load().(*inTrack)
		// If the track is not there, wait for signal and try again
		if in == nil || in.track == nil || in == lastFailedIn {
			select {
			case <-ctx.Done():
				return ctx.Err()
			case <-t.inTrackUpdated:
				continue
			}
		}
		// Read info byte buf
		// TODO: What about deadlines?
		n, err := in.track.Read(buf)
		// If no error, unmarshal into packet. We confirmed only the following two
		// header values need to be cleared.
		if err == nil {
			rtpPacket.ExtensionProfile, rtpPacket.ExtensionPayload = 0, nil
			err = rtpPacket.Unmarshal(buf[:n])
		}
		// If there's an error, notify, mark failed, continue
		if err != nil {
			// Send read error
			if onReadError != nil {
				if err = onReadError(err); err != nil {
					return err
				}
			}
			// Mark as last failed and continue
			lastFailedIn = in
			continue
		}
		// Set the ssrc and sequence number
		rtpPacket.SSRC = t.outSSRC
		seqNum++
		rtpPacket.SequenceNumber = seqNum
		// Apply last timestamp delta if there is a prev timestamp. If there's not,
		// it means it's the first time we've seen this track, so we use the last
		// delta from the previous stream.
		if in.lastTimestamp != 0 {
			lastTimestampDelta = rtpPacket.Timestamp - in.lastTimestamp
		}
		// Apply the last timestamp
		currTimestamp += lastTimestampDelta
		in.lastTimestamp = rtpPacket.Timestamp
		rtpPacket.Timestamp = currTimestamp
		// Attempt write. We ignore ErrClosedPipe. We have confirmed that the packet
		// is not referenced beyond the call.
		if err := t.out.WriteRTP(&rtpPacket); err != nil && err != io.ErrClosedPipe {
			return err
		}
	}
	return ctx.Err()
}
```

Sean-Der:
Merged with 97c8f5290c0c00bb4fa7cab989f5a4e9631293fb

No worries about joining, when you get the bug though and want to get deep into jitterbuffer stuff I am here :p 

I ended up dropping the channel completely, I just continue if `trackCount == 0`! I ended up leaving the reader goroutine, I think that is good. I might reorganize things, maybe it would feel better if it right next to the `OnTrack`? I tend to overthink things, but these examples are probably the most important thing we have, lots of people launch their projects off of them :)

PR #1170, opened 2020-04-27
Start only negotiated senders

### LINKED ISSUE #1169: AddTrack/Add*Transceiver/RemoveTrack concurrency issues during negotiation.
### Your environment.
  - Version: master eab7996d848f2cc1f16c1ec9aad2eac3ef224f0f
  - Browser: Not related to a browser but reproducible with Chrome at probably any version (tested with 81/82/83/84)

### What did you do?

I noticed this while building an sfu and testing it with chrome (looks like firefox is less susceptible to such issues).

Basically, calling AddTrack/Add*Transceiver/RemoveTrack during a negotiation or, also if the negotiation is ended but the goroutine spawned by `StartTransports`/`StartRenegotiation` is yet running and calling `StartRTPSender`, will create different kind of issues. I'll report two of the possible cases a discovered in my tests:

1. If we're offering and then we call `AddTrack` before receiving an answer/calling `SetRemoteDescription(answer)` the `StartTransports`/`StartRenegotiation` method will start some transceivers not negotiated and so some rtpSenders not provided in the offered sdp. In chrome this will cause an obscure exception when calling `setRemoteDescription` :

```
"Failed to set local answer sdp: Failed to process the bundled m= section"
```

2. If we already offered, received the answered and called SetRemoteDescription(answer)` BUT the `StartTransports`/`StartRenegotiation` spawned goroutines are not finished and we call `AddTrack` that will add a Sender to an existing transport currently without a sender (implemented in bddbfa7e11efd0c0a134ecbd677377a7555c68b1).

Chrome will not report an exception but the stream won't start and in the log we'll find something like: 

```
[3043620:16:0424/144525.744681:ERROR:webrtc_video_engine.cc(3115)] Absent receive stream; ignoring clearing encoded frame sink for ssrc 0
[3043620:16:0424/144525.773075:ERROR:call.cc(1257)] receive_rtp_config_ lookup failed for ssrc 1491190739
[3043620:16:0424/144525.781252:ERROR:rtp_stream_receiver_controller.cc(26)] RtpStreamReceiverController::Receiver::Receiver: Sink could not be added for SSRC=1491190739.
```

Looks like starting a rtpsender before negotiating it is a known issue with chrome: https://bugs.chromium.org/p/webrtc/issues/detail?id=10139
basically starting a rtpsender providing a ssrc not available in the current local sdp will cause issue 1. Issue 2. is more subtle since we don't get any exception but the related track won't be added.

### Possible solution.

**TL;DR**

Only start sender if the both the transceiver and the sender were the one negotiated.

The above issues cannot be avoided with just a "clever" use of the pion/webrtc API by avoid calling AddTrack/Add*Transceiver/RemoveTrack between a negotiation because the `StartTransports`/`StartRenegotiation` goroutines will continue after `SetRemoteDescription` has returned.

sgotti:
#### Description

Only start senders that were negotiated.

Since `startTransports`/`startRenegotiation` start rtp senders in a goroutine,
an user could call AddTrack/Add*Transceiver/RemoveTrack causing a not negotiated
sender track ssrc to be started.

#### Reference issue
Fixes #1169

### DIFF
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -1069,3 +1069,56 @@ func TestPlanBMultiTrack(t *testing.T) {
 	assert.NoError(t, pcOffer.Close())
 	assert.NoError(t, pcAnswer.Close())
 }
+
+// TestPeerConnection_Start_Only_Negotiated_Senders tests that only
+// the current negotiated transceivers senders provided in an
+// offer/answer are started
+func TestPeerConnection_Start_Only_Negotiated_Senders(t *testing.T) {
+	lim := test.TimeOut(time.Second * 30)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	pcOffer, err := NewPeerConnection(Configuration{})
+	assert.NoError(t, err)
+	defer func() { assert.NoError(t, pcOffer.Close()) }()
+
+	pcAnswer, err := NewPeerConnection(Configuration{})
+	assert.NoError(t, err)
+	defer func() { assert.NoError(t, pcAnswer.Close()) }()
+
+	track1, err := pcOffer.NewTrack(DefaultPayloadTypeVP8, rand.Uint32(), "video", "pion1")
+	require.NoError(t, err)
+
+	sender1, err := pcOffer.AddTrack(track1)
+	require.NoError(t, err)
+
+	offer, err := pcOffer.CreateOffer(nil)
+	assert.NoError(t, err)
+
+	assert.NoError(t, pcOffer.SetLocalDescription(offer))
+	assert.NoError(t, pcAnswer.SetRemoteDescription(offer))
+	answer, err := pcAnswer.CreateAnswer(nil)
+	assert.NoError(t, err)
+	assert.NoError(t, pcAnswer.SetLocalDescription(answer))
+
+	// Add a new track between providing the offer and applying the answer
+
+	track2, err := pcOffer.NewTrack(DefaultPayloadTypeVP8, rand.Uint32(), "video", "pion2")
+	require.NoError(t, err)
+
+	sender2, err := pcOffer.AddTrack(track2)
+	require.NoError(t, err)
+
+	// apply answer so we'll test generateMatchedSDP
+	assert.NoError(t, pcOffer.SetRemoteDescription(answer))
+
+	// Wait for senders to be started by startTransports spawned goroutine
+	pcOffer.negotationLock.Lock()
+	defer pcOffer.negotationLock.Unlock()
+
+	// sender1 should be started but sender2 should not be started
+	assert.True(t, sender1.hasSent(), "sender1 is not started but should be started")
+	assert.False(t, sender2.hasSent(), "sender2 is started but should not be started")
+}
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -966,7 +966,8 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
 // startRTPSenders starts all outbound RTP streams
 func (pc *PeerConnection) startRTPSenders(currentTransceivers []*RTPTransceiver) {
 	for _, transceiver := range currentTransceivers {
-		if transceiver.Sender() != nil && !transceiver.Sender().hasSent() {
+		// TODO(sgotti) when in future we'll avoid replacing a transceiver sender just check the transceiver negotiation status
+		if transceiver.Sender() != nil && transceiver.Sender().isNegotiated() && !transceiver.Sender().hasSent() {
 			err := transceiver.Sender().Send(RTPSendParameters{
 				Encodings: RTPEncodingParameters{
 					RTPCodingParameters{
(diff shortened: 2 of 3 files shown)


Sean-Der:
This is fantastic! Thank you for taking this on @sgotti

Sean-Der:
Do you think it would be possible to get a test for this? This seems like something that could easily be regressed if we aren't careful.

The other thing I am curious about is if this is something that is already represented in the W3C! I need to look a little bit, but it would be cool if we were just implementing more (and not adding another field) but not a huge deal.

sgotti:
@Sean-Der Yeah, I just opened the PR a Work in progress to just discuss the implementation idea. I'm working on adding some tests, protecting the `negotiated` field since now it could cause races etc...



> The other thing I am curious about is if this is something that is already represented in the W3C! I need to look a little bit, but it would be cool if we were just implementing more (and not adding another field) but not a huge deal.

IMHO the spec here https://www.w3.org/TR/webrtc looks like a mix of specification and implementation details (very tied to javascript) that make it quite difficult to follow. It doesn't explicitly talk about "starting" a sender but it could be deduced since the spec always talk about a series of steps that will be done in a single thread and "operations" that will be "enqued" in a chain if the peerconnection is not "settled": https://www.w3.org/TR/webrtc/#chain-an-asynchronous-operation

Since pion implementation is quite different in many details (also related to the fact it's written in go and uses go features) and there's no such kind of "operations" queue the current implementation just marks a sender as negotiated (as it could be replaced/removed in the same transceiver or a new transceiver+sender added while we're running startTransports). Another implemention should try to avoid using a goroutint in `startTransports`/`startNegotiation` and implement such kind of "operations" queue but will require a quite big rewrite.

Additionally this chromium issue comment https://bugs.chromium.org/p/webrtc/issues/detail?id=10139#c14 I posted in #1169 has some insights on why chrome behaves in this way and due to the fact that chrome<->firefox works I thinks firefox also starts only negotiated sender but is less sensitive than chrome.

jeremija:
@sgotti This is awesome! I'm finally able to connect four peer connections initiated in Chrome through my SFU without issues any in my app:

image

Previously, I'd get the `Failed to set local answer sdp: Failed to process the bundled m= section` error almost every time the 3rd peer was connected. 

I was secretly hoping this would resolve an issue I have with desktop sharing, but that's probably not related to this :)

jeremija:
@sgotti is right that the spec is very JavaScript-centric. Pion already differs from the spec for Unified plan since senders and receivers of a transceiver can be set to `nil` or reset, whereas the spec mandates [senders and receivers to always be the same](https://www.w3.org/TR/webrtc/#rtp-media-api) for a lifetime of a transceiver:

> An RTCPeerConnection object contains a set of RTCRtpTransceivers, representing the paired senders and receivers with some shared state. This set is initialized to the empty set when the RTCPeerConnection object is created. RTCRtpSenders and RTCRtpReceivers are always created at the same time as an RTCRtpTransceiver, which they will remain attached to for their lifetime.

Refactoring this could be a lot of work, although I agree that it would be nice to be fully spec compliant in the future.

@sgotti I sent you an invite to join the Pion org. Let me know if I can help with the tests for this PR. Also, I just merged PR #1156 so there are some conflicts - sorry!

sgotti:
@jeremija @Sean-Der Added a test and locking around reading/setting sender negotiation field. Unfortunately it increased the cyclomatic complexity of func `(*PeerConnection).generateMatchedSDP` (31 > 30). Do you have any suggestion (remove cyclo check, refactor etc...)?

> An RTCPeerConnection object contains a set of RTCRtpTransceivers, representing the paired senders and receivers with some shared state. This set is initialized to the empty set when the RTCPeerConnection object is created. RTCRtpSenders and RTCRtpReceivers are always created at the same time as an RTCRtpTransceiver, which they will remain attached to for their lifetime.

Also discussed about this offline with @jeremija (need to create a related issue). If in future we'll partially drop some changes to #1156 to no add a new sender to an existing transceiver, we could also drop the sender negotiation field added in this PR and just check that the transceiver has been negotiated.

peerconnection_media_test.go
```diff
@@ -1074,3 +1074,63 @@ func TestPlanBMultiTrack(t *testing.T) {
...
+	select {
+	case <-sender1.sendCalled:
+	default:
+		t.Fatal("sender1 is not started but should be started")
+	}
+	select {
+	case <-sender2.sendCalled:
+		t.Fatal("sender2 is started but should not be started")
+	default:
+	}
```
jeremija, line 1135: I think the end of this test can be simplified a little. Instead of `time.Sleep()` we could use:

```golang
	// apply answer so we'll test generateMatchedSDP
	assert.NoError(t, pcOffer.SetRemoteDescription(answer))

	// Wait for senders to be started by startTransports spawned goroutine
	pcOffer.negotationLock.Lock()
	defer pcOffer.negotationLock.Unlock()
```

Sleeps in test add up quickly and then the tests are slow to run. And instead of select/case we can use:

```
	assert.True(t, sender1.hasSent(), "sender1 is not started but should be started")
	assert.False(t, sender2.hasSent(), "sender2 is started but should not be started")
```

sgotti: @jeremija Definetly better. Updated.

jeremija:
Thanks for updating @sgotti! I see now that I put the `setNegotiated` calls in the wrong place!

PR #1173, opened 2020-04-30
make the RTPSender SendRTP public

cedricfung:
There is no JitterBuffer in Pion yet, and this will make it good to implement one in the application code.


Sean-Der:
Me and @cedricfung discussed this in the channel, and I am gonna merge this!

I am going to update the comment explaining why it exists (and why you should use Track) and why we will be breaking this API eventually. We need it today though, people are blocked in production.

Sean-Der:
Merged with abe3ea1ea86d0daafca4a4f08c2ac6f36e5cddcf

PR #1158, opened 2020-04-20
Add RTPTransceiver.Mid()

jeremija:
#### Description
@Sean-Der - just wanted to share this before I call it a night. I'm not sure if you had other plans regarding issue #1148, so I'd love to hear your thoughts.

#### Reference issue
(Hopefully) Fixes #1148

Issue #1176, opened 2020-04-30
TestPeerConnection_Renegotation_AddTrack_Multiple failing sometimes


at-wat:
```
=== RUN   TestPeerConnection_Renegotation_AddTrack_Multiple
    TestPeerConnection_Renegotation_AddTrack_Multiple: peerconnection_renegotation_test.go:175: 
        	Error Trace:	peerconnection_renegotation_test.go:175
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 1
        	Test:       	TestPeerConnection_Renegotation_AddTrack_Multiple
```

https://travis-ci.org/github/pion/webrtc/jobs/681361005#L513 (restarted)

Issue #1171, opened 2020-04-29
Pion should create a recvonly transceiver when needed.


sgotti:
### Your environment.
  - Version: master currently @ 18dfe7919313be036ec632e47856903596f25722
  - Browser: *include version*
  - Other Information - *stacktraces, related issues, suggestions how to fix, links for us to have context*

### Issue Description

If a remote adds a new track that creates a new remote transceiver, when receiving the offer, pion in `setRemoteDescription` won't find any available recvonly transceiver and no receiver will be started. Later when we'll call `CreateAnswer`, `generateMatchedSDP` will create a new `inactive` transceiver and report it in the answer. The user so won't be notified of any new track (with `OnTrack` handler).

A browser like chrome/firefox instead behave differently and they'll create a recvonly transceiver and the new track will be available.

I think that's the reason why all the pion examples manually create some transceivers calling `AddTransceiver*`.
Additionally this way of calling `AddTransceiver*` could work for the first negotiation, but if the remote start removing and adding tracks, also if we add tons of transceiver with `AddTransceiver` they could be ignored by the remote that will add an additional transceiver.

The pion behavior looks against the spec that at the end of [4.4.1.5 Set the RTCSessionDescriptionSet](https://w3c.github.io/webrtc-pc/#set-the-rtcsessiondescription)  says:

```
If no suitable transceiver was found (transceiver is unset), run the following steps:

Create an RTCRtpSender, sender, from the media description using sendEncodings.

Create an RTCRtpReceiver, receiver, from the media description.

Create an RTCRtpTransceiver with sender, receiver and an RTCRtpTransceiverDirection value of "recvonly", and let transceiver be the result.

Add transceiver to the connection's set of transceivers.
```

### Proposed solution.

Refactor `setRemoteDescription` to parse the media sections (now this is only done later by `CreateAnswer`) and add missing `recvonly` transceivers. In this way the transceiver is available when `setRemoteDescription` will call startTransports/startRenegotiation that will start the related receivers.

PR #1177, opened 2020-04-30
Start the right receiver

sgotti:
#### Description

From PR #1158 during negotiation every required transceiver is assigned
to a specific media section (using the media section mid value).

The same should be done when starting receivers. We should start the
receiver assigned to the media section of that track.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -30,7 +30,7 @@ type PeerConnection struct {
 	mu      sync.RWMutex
 
 	// Negotiations must be processed serially
-	negotationLock sync.Mutex
+	negotiationLock sync.Mutex
 
 	configuration Configuration
 
@@ -935,6 +935,10 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
 		for i := range localTransceivers {
 			t := localTransceivers[i]
 
+			if t.Mid() != incoming.mid {
+				continue
+			}
+
 			if (incomingTracks[ssrc].kind != t.kind) ||
 				(t.Direction() != RTPTransceiverDirectionRecvonly && t.Direction() != RTPTransceiverDirectionSendrecv) ||
 				(t.Receiver()) == nil ||
@@ -1648,10 +1652,10 @@ func (pc *PeerConnection) GetStats() StatsReport {
 
 // Start all transports. PeerConnection now has enough state
 func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, remoteUfrag, remotePwd, fingerprint, fingerprintHash string, currentTransceivers []*RTPTransceiver, incomingTracks map[uint32]trackDetails) {
-	pc.negotationLock.Lock()
+	pc.negotiationLock.Lock()
 
 	go func() {
-		defer pc.negotationLock.Unlock()
+		defer pc.negotiationLock.Unlock()
 
 		// Start the ice transport
 		err := pc.iceTransport.Start(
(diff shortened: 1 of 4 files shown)


jeremija:
@sgotti Nice work! Unfortunately it looks like the build was broken. I am able to reproduce the failing test with Go 1.14 on my machine, which passed on Travis CI for Go 1.14, but failed for 1.13. Looks like an issue in `TestPeerConnection_Renegotation_AddTrack_Multiple`

jeremija:
I think the problem is that the new code skips available transceivers that have never been used.

peerconnection.go
```diff
@@ -854,6 +854,10 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
 		for i := range localTransceivers {
 			t := localTransceivers[i]
 
+			if t.Mid() != incoming.mid {
+				continue
```
jeremija, line 858: I believe this will skip available transceivers that have never been used for sending. That's why the `TestPeerConnection_Renegotation_AddTrack_Multiple` test is failing

peerconnection_media_test.go
```diff
@@ -1074,3 +1074,99 @@ func TestPlanBMultiTrack(t *testing.T) {
...
+				continue
+			}
+			if transceiver.Receiver() == nil {
+				return false, nil
+			}
+			select {
+			case <-transceiver.Receiver().received:
+				return true, nil
+			default:
+				return false, nil
```
jeremija, line 1094: I think you can just use `return transceiver.Receiver() != nil && transceiver.Receiver().haveReceived()` instead of all of these lines.

Perhaps this could be changed to `hasReceived` to keep it consistent with `rtpsender.go` which has a `hasSent` function.

peerconnection.go
jeremija: I think we'll need to do the same thing we did in `generateMatchedSDP` when `includeUnmatched` is set to `true`.

sgotti:
> @sgotti Nice work! Unfortunately it looks like the build was broken. I am able to reproduce the failing test with Go 1.14 on my machine, which passed on Travis CI for Go 1.14, but failed for 1.13. Looks like an issue in `TestPeerConnection_Renegotation_AddTrack_Multiple`

The hang of the test could be related to #1176 ? I can reproduce (sometimes) it also with go 1.14 without this patch

Instead I'll take a look at the warning.

jeremija:
@sgotti Thanks, I was unaware of this issue. I'll also have a deeper look into this!

peerconnection.go
sgotti: @jeremija Yeah, the real issue is that, when receiving an offer we're starting receivers in SetRemoteDescription and assigning transceiver mid in CreateAnswer (so later). We should follow JSEP (https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-26#section-5.10) and create assign transceivers to a media section in SetRemoteDescription

I reported this in #1171 and proposed PR #1179. Perhaps this should go after #1179.

jeremija: @sgotti Yeah I was just thinking about that other PR of yours. I'm having a look at it now.

peerconnection_media_test.go
sgotti: @jeremija Done!

peerconnection.go
sgotti: @jeremija Now that #1179 has been merged this should be ok.

jeremija:
The failing test is most likely related to #1186. Let me try to fix that...

PR #1179, opened 2020-04-30
Move transceiver selection inside SetRemoteDescription and CreateOffer

sgotti:
#### Description

This tries to better match the JSEP spec and will fix at least two
issues:  #1171 and #1178

* Move transceiver selection/creation in SetRemoteDescription and
CreateOffer
* In SetRemoteDescription also create new Transceivers of type recvonly
when no satisfying transceiver is available
* In CreateOffer generate unique mid in number format avoiding possible
collisions with remote provide mids and to also already handle a future
implementation of m= section rejection and reuse
* Now generateMatchedSDP will just find the transceivers with the
required mid since they are already selected previously.

More details in these extract from JSEP:

JSEP 5.10 (Applying a Remote Description) says:

```
For each m= section, the following steps MUST be performed

[...]

If the m= section is not associated with any RtpTransceiver
(possibly because it was dissociated in the previous step),
either find an RtpTransceiver or create one according to the
following steps:
```

JSEP 5.2 (Constructing an Offer) says:

```
[...]

An m= section is generated for each RtpTransceiver
that has been added to the PeerConnection, excluding any stopped
RtpTransceivers;
```

Note that we are currently directly associating a mid to a transceiver
in CreateOffer, instead the spec says to also keep a m= section index
mapping to a transceiver and set the transceiver mid only when applying
the local description. This is needed to support rollback of proposed
offer/answer but currently we don't have support and tests for rollback
situations.

#### Reference issue
Fixes #1171 
Fixes #1178

PR #1192, opened 2020-05-12
Start RTP transports at the end of the signaling flow

### LINKED ISSUE #1157: SetRemoteDescription when renegotiation causes deadlock
### Your environment.
  - Version: *234a170*
  - Browser: *Firefox 75.0*

### What did you do?

1. Firefox send an offer with an audio track
2. Pion: pc = NewPeerConnection => SetRemote(offer) => OnSignalingStateChange(have-remote-offer)
3. Pion: OnICEConnectionStateChange(checking)
4. Pion: CreateAnswer => SetLocal(answer) => OnSignalingStateChange(stable)
5. Pion: new audio track => pc.AddTrack
6. Pion: CreateOffer => SetLocal(offer) => OnSignalingStateChange(have-local-offer)
7. Firefox: SetRemote(offer) => CreateAnswer
8. Pion: SetRemote(answer) => OnSignalingStateChange(stable)

But the SetRemoteDescription blocks now, won't continue.

Then I read the code and find 
```
	if haveRemoteDescription {
		pc.startRenegotation(currentTransceivers)
```

so it must be startRenegotation blocks. I have removed `pc.startSCTP()` to prevent datachannel lock. If I don't remove the SCTP code, then it will always deadlock, and now it only have random deadlocks. 

### What did you expect?

renegotiate successfully

### What happened?

pc.startRenegotation(currentTransceivers) blocks

### LINKED ISSUE #1191: RTP streams should be started from SetLocalDescription when an answer is being set (not from SetRemoteDescription when an offer is being set)
### Your environment.
  - Version: 2.2.9
  - Browser: N/A/

### What did you do?

I tried to add a test for `OnTrack` after renegotiation in my application, but the `OnTrack` callback is sometimes not called because the sender is not added to `track.activeSenders` after renegotiation.

The test case I have in my app is available here: https://github.com/peer-calls/peer-calls/tree/ontrack-test

Link to specific commit in case the branch gets deleted in the future: https://github.com/peer-calls/peer-calls/commit/be3f19159e82f1697b0732b68df00ff1d371900b

Command to run the test:

```sh
git cloone https://github.com/peer-calls/peer-calls
cd peer-calls
PEERCALLS_LOG='-sdp,-pion:*:trace,-pion:*:debug,-pion:*:info,-signaller,negotiator,*' go test ./server -v -run TestSFU_OnTrack -race -count=20
```

The tests fail about 50% of the time, hence the `-count` flag.

**UPDATE: the PR now has a test too**

### What did you expect?

I was expecting the `OnTrack` handler to be called, but sometimes it is not.

### What happened?

In my app this particular peer has to request negotiation from the server after the track is added. The problem occurs when the server sends the offer, and the `pc.startRenegotiation` is called before the `sender.setNegotiated` is called, so the sender is never added to `track.activeSenders`, and thus `track.WriteSample` does not write to the expected sender.

jeremija:
#### Description

This PR contains the following updates:

1. correctly start RTP streams at the end of the signaling flow: `SetLocalDescrption` or `SetRemoteDescription`. See example below:

```go
offer, _ := pcOffer.CreateOffer(nil)
_ := pcOffer.SetLocalDescription(offer)
_ := pcAnswer.SetRemoteDescription(offer)
answer, _ := pcAnswer.CreateAnswer(nil)
_ := pcAnswer.SetLocalDecription(answer)  // rtp streams will be started here for pcAnswer (new behavior)
_ := pcOffer.SetRemoteDescription(answer) // rtp streams will be started here for pcOffer (old behavior)
```
2. Add an operations queue for all pending transports/streams operations. This should also stop blocking the calls to `SetRemoteDescription` described in issue #1157.
3. Refactor `startRenegotiation` and `startTransports` methods so `negotiationLock` is removed in favor of the aforementioned operations queue. The `startRenegotiation` has been renamed `startRTP`.
4. Remove the `peerconnection_setlocaldescription_test.go` since this test was no longer needed, and the updates from this PR just made it hang indefinitely. Discussed with @Sean-Der on Slack.
5. Fix the ice-candidate related goroutine leaks in `TestPeerConnection_Renegotiation_Trickle` test. Unrelated to this PR - this was a hidden issue on master branch. There were no goroutine leak checks in this test, and it was the last connectivity test that was executed so it went under the radar. This was accidentally uncovered by executing a test which checked for leaked goroutines after this one.
6. Remove unnecessary `if err := ...; err != nil` checks in `peerConnection.Close()` since the `util.FlattenErrs` does this already. This should increase the test coverage a bit.
7. Correct all misspellings of `negotiate`.

#### Reference issues
Fixes #1191
Most likely fixes #1157


jeremija:
@Sean-Der this is the bug you helped me trace today over slack - thanks again!
@sgotti It would be great to hear your opinion about this since this is possibly related to #1179.

sgotti:
@jeremija I already answered in the related issue #1191 since I'm not sure I really understand the issue. 

I think we'll need a test to reproduce this issue to avoid future regressions (ans so I can also better understand it :smile: ).

Basically I'm not sure why we should start renegotiation when applying the local description with an answer. If it's an answer we cannot offer anything so there's no need to start any sender. So starting renegotiation in SetRemoteDescription with both an offer on an answer should work in every case (since we have to call it before setLocalDescription when receiving an offer and after SetLocalDescription when receiving an answer)

jeremija:
@sgotti haha it's hard to explain this. Let me try again. The flow for renegotiation is following:

```golang
offer, _ := peer1.CreateOffer(nil)
_ := peer1.SetLocalDescription(offer)
_ := peer2.SetRemoteDescription(offer)
answer, _ := peer2.CreateAnswer(nil)
_ := peer2.SetLocalDecription(answer)
_ := peer1.SetRemoteDescription(answer)
```

The current code works if `peer1` adds a new track and renegotiates. But it might fail if `peer2` adds a track and asks `peer1` to create a new offer.

The problem is that in the current codebase the call to `peer2.SetRemoteDescription` will call `startRenegotiation` before it marked the sender as negotiated using `sender.setNegotiated` - this happens later in `peer2.CreateAnswer`.

My update will just call `startRenegotiation` from `peer2.SetLocalDescription` (after the answer is created), but for `peer` the `startRenegotiation` will be called from `peer1.SetRemoteDescription`, like it was before.

I just pushed a test for this too!

sgotti:
> The current code works if `peer1` adds a new track and renegotiates. But it will fail when `peer2` adds a track and asks `peer1` to add a transceiver and create a new offer (renegotiate).
> 
> The problem is that the call to `peer2.SetRemoteDescription` will trigger `startRenegotiation` before it marked the sender as negotiated using `sender.setNegotiated` - this happens later in `peer2.CreateAnswer`.
> 
> My update will just call `startRenegotiation` from `peer2.SetLocalDescription` (after the answer is created), but `startRenegotiation` will still be triggered from `peer1.SetRemoteDescription`, like it was before.
> 
> I just pushed a test for this too!

I saw your test and your description but I think it doesn't work because it's wrong. Per the webrtc spec if peer2 adds a track then it has to create an offer (start a negotiation itself). It's wrong to offer a new track answering to a remote offer.

jeremija:
@sgotti I don't think there's anything inherently wrong if a remote peer asks the other one to add a transceiver and start the negotiaton. I did not invent this - the `simple-peer` JS library uses this method and it works perfectly in browser-to-browser peer connections:

https://github.com/feross/simple-peer/blob/eb602e71ee532ed6bf2e4bbeda0f55fb04980f52/index.js#L263

The `peer2` will not add this track to the answer unless `peer1` has an available transceiver (receiver), hence I mentioned the request to add a transceiver.

sgotti:
> I don't think there's anything inherently wrong if a remote peer asks the other one to add a transceiver and start the negotiaton.

Ah ok, sorry. I haven't considered that in your case there was an available recvonly/sendrecv transceiver in the offer.

So, basically, the change will start renegotiation (stop/start receivers and start senders) only at the end of the signaling flow, right? (if we received an offer we start renegotiation when applying our local description with our answer, if we're offering we start renegotiation when applying the remote description with remote answer).

Should this change also cover the first negotiation? Currently the "first" signaling/negotiation uses startTransports (that will start receivers and senders only when the transports are ready) and this is done when applying the remote description in both cases (if it's an offer or an answer).

jeremija:
@sgotti I was just thinking about the first negotiation. Perhaps the whole bottom part of `SetRemoteDescription` could be extracted into a new function and then called as needed from `SetRemoteDescription` or `SetLocalDescription`.

sgotti:
> I was just thinking about the first negotiation. Perhaps the whole bottom part of `SetRemoteDescription` could be extracted into a new function and then called as needed from `SetRemoteDescription` or `SetLocalDescription`.

Yeah, I think that we should do some refactor removing from start transport all the start of senders/receivers since it's duplicated code already done in `startRenegotiation`.

Then call `startTransports` only on first negotiation and let `startRenegotiation` be called when needed regardless of first or successive negotiation with the unique change that `startRenegotiation` needs to wait for `startTransports` to be finished. 
There's also the need to move `drainSRTP` outside `startTransports` and start `drainSRTP` only when the first `startRenegotiation` has been called or it'll start draining read streams not already owned by any receiver.

I'll also rename `startRenegotiation` since its name is imho misleading.

jeremija:
@sgotti I killed travis with all the force pushes. Started getting error `429 Too Many Requests` from GitHub. Now seems to have finally settled and the tests have passed.

This PR turned out to be much larger than I initially expected. I've updated the main description and I'd appreciate if you could have a look!

jeremija:
> @jeremija Just an initial review, I'll need to test it. I have a doubt:
> 
>     * This isn't related to https://w3c.github.io/webrtc-pc/#chain-an-asynchronous-operation right? In the spec this operation chain represent the operation to do for asynchronously creating an offer/answer while in this case we're chaining operations to start transport/receivers/transceivers. Plus the operation cannot fail

No, even though this has served as an inspiration. Perhaps in the future the `type operations struct` can be used for chains described in the spec as well. Implementing the operations chain described in the spec would be a lot more work.

I needed something better than a `sync.Mutex` to ensure the operations will be executed in the same order they were added, but without blocking.

The previous version had this workflow:

```go 
mutex.Lock()  // this would block
go func() {
  defer mutex.Unlock()
  // do async work
}
```

An alternative was to put the `Lock` call inside the goroutine:

```go 
// this won't block
go func() {
  mutex.Lock() // but this might be called out of order
  defer mutex.Unlock()
  // do async work
}
```

but then we'd lose the guarantee that the two goroutines will be started in the same order.

operations.go
```diff
@@ -0,0 +1,66 @@
...
+// Done will return a channel that will be closed as soon as all currently
+// enqueued opearations are finished.
+func (o *operations) Done() <-chan struct{} {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if o.done != nil {
+		return o.done
+	}
+	done := make(chan struct{})
+	close(done)
```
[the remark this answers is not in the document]
Sean-Der: maybe you can initialize `type operations struct` with a closed chan? Then you can always do ` return o.done`

You will still need to guard with a mutex, but might read nicer!

operations.go
```diff
@@ -0,0 +1,67 @@
...
+		return o.done
+	}
+	// no queued operations, return the closed channel so reading from it  does not block
+	done := make(chan struct{})
+	close(done)
+	return done
+}
+
+func (o *operations) start() {
+	for {
```
Sean-Der, line 52: What do you think of having a local function that does the popping? It can hold the lock as well. This code is good now, but I am afraid a bug fix/modification could break it in the future :(

```
popOperation := func() func() {
    o.mu.Lock()
    defer o.mu.Unlock()
    
    if len(o.ops) == 0 {
        return nil
    }
 
   fn := o.ops[0]
   o.ops = o.ops[1:]
   return fn
}

```

Sean-Der:
@jeremija This is amazing! Nice work, we should have done this a long time ago :)

In the future we should also return these operations, then it will feel equivalent to promises which would really help.

operations.go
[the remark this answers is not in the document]
jeremija: @Sean-Der That'll work. I was just trying to avoid adding a constructor, but I'll add it shortly!

operations.go
```diff
@@ -0,0 +1,67 @@
...
+	done    chan struct{}
+}
+
+// Enqueue adds a new action to be executed. If there are no actions scheduled,
+// the execution will start immediately in a new goroutine.
+func (o *operations) Enqueue(op operation) {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	o.ops = append(o.ops, op)
+	if len(o.ops) == 1 {
```
Sean-Der, line 24: I don't think we have a race, but just double checking.

* start drains the last operation.
* Enqueue is called, `len(o.ops) == 1`, we just added a new operation
* `start` will start pulling the new operation, meaning we don't exit.

This will cause us to have two `o.done` in flight. We should probably pop `o.ops` and return under the same lock. Otherwise someone could sneak in another operation.

operations.go
jeremija: You're right. I didn't like doing the unlock on two different places. I came up with something a little simpler based on your other comment:

```
func (o *operations) pop() (fn func(), isLast bool) {
	o.mu.Lock()
	defer o.mu.Unlock()
	fn = o.ops[0]
	o.ops = o.ops[1:]
	return fn, len(o.ops) == 0
}
```

operations.go
jeremija: I hadn't thought about adding an `isLast` / `hasMore` flag before. I just updated the code to be:

```golang
func (o *operations) pop() (fn func(), isLast bool) {
	o.mu.Lock()
	defer o.mu.Unlock()
	fn = o.ops[0]
	o.ops = o.ops[1:]
	return fn, len(o.ops) == 0
}

func (o *operations) start() {
	var fn func()
	isLast := false
	for !isLast {
		fn, isLast = o.pop()
		fn()
	}
}
```

I think this should take care of it?

operations.go
```diff
@@ -0,0 +1,75 @@
...
+	// operations.
+	closed chan struct{}
+	done   chan struct{}
+}
+
+func newOperations() *operations {
+	closed := make(chan struct{})
+	close(closed)
+	return &operations{
+		closed: closed,
```
Sean-Der, line 25: Sorry I mislead you in my last comment, what about `&operations{ done : closed}`? Then you can drop the nil check!

jeremija: Haha that makes perfect sense - done! Only one `if` clause in the whole implementation now 😎

operations.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Operation is a function
+type operation func()
+
+// Operations is a task executor.
+type operations struct {
+	ops     []operation
+	mu      sync.Mutex
+	startMu sync.Mutex
+	// closed channel is alway closed and is returned when there are no queued
```
Sean-Der, line 15: Just kill this comment and LGTM!

jeremija: Thank you - done!

PR #1193, opened 2020-05-13
Fix handling of Empty DataChannel messages

normanr:
This is the second half of https://github.com/pion/datachannel/pull/60:

> Empty messages should be sent as one zero byte. When receiving an Empty
> message, the contents should be ignored.
> 
> Currently pion/webrtc converts zero-byte messages into single null-byte messages and datachannels transmits it as a regular message. This fixes datachannels to send the correct payload and strip the payload for Empty messages.
> 
> I'll send a PR for webrtc to remove the conversion from webrtc too.

Now that datachannel handles Empty messages correctly, webrtc should pass them to datachannel at they are.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Chad Retz](https://github.com/cretz)
 * [Simone Gotti](https://github.com/sgotti)
 * [Cedric Fung](https://github.com/cedricfung)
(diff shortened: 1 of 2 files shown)


normanr:
I'm open to suggestions on if this needs tests like the datachannel PR. It doesn't look like webrtc/datachannel_test currently check the contents of received messages as thoroughly as the datachannel tests do.

Sean-Der:
@normanr no LGTM! I think we will end up just adding more tests, but not adding any value :)

I will tag a new version of `pion/webrtc` with all of your work, thank you so much for doing this.

Issue #1016, opened 2020-02-10
Add support for Simulcast/SVC.


cloudwebrtc:
## Summary

Add multiple resolution/rate support for the video track.

## Motivation

In a video conference or live broadcast scenario, multiple resolutions need to be generated at the same time on the same PC, and different devices can subscribe to the resolution they want.

SVC can adapt to network transmission by dynamically adjusting resolution and bit rate during the SFU Relay process.

## Describe alternatives you've considered

[mediasoup](https://github.com/versatica/mediasoup) already implements similar functionality, but it uses C ++

Sean-Der:
@cloudwebrtc that would be an amazing feature! Also would be really nice to just have it be on the `Track`

I will need to think about a little bit though. I need to do some research/API thinking and will come back on this!

thanks for filing, this could really open up some cool things.

sgotti:
@Sean-Der I started studying how adding simulcast will work and doing an initial implementation. Since it'll break various pion packages api (primarily webrtc and srtp) (like also discussed in #1025 for track api) I'd like to propose a design doc to cover and explain the required changes and discuss to possible options. Is the wiki the right place (a page like https://github.com/pion/webrtc/wiki/PlanningV3) to add and discuss such design doc or do you have something better in mind?

Sean-Der:
@sgotti Yes the wiki is the best place, and please edit away! 

I don't think it will be a substantial amount of work, but I am just timid to change the API and not do it right. The `/v2` API was dramatically different then the original one, and introduced a whole new set of different issues :(

sgotti:
@Sean-Der 

> I don't think it will be a substantial amount of work, but I am just timid to change the API and not do it right. The `/v2` API was dramatically different then the original one, and introduced a whole new set of different issues :(

I'll start writing something here the keep you updated, I'll try to summarize my work until now and explain why we need a different pion `Track` API.

I have an initial working implementation of simulcast handling and, to be honest, is a quite big amount of work. I'll clean it up and split in multiple patches and then publish it in the next days so you can take a look at it.

It's a big amount of work because it has to change the way tracks are handled. We have to handle mid, rid rtp/rtcp extensions, sdp extmap, rid and simulcast attributes.

[...]

PR #1195, opened 2020-05-13
Implement RTPSender.ReplaceTrack

### LINKED ISSUE #1020: Support RTPSender.ReplaceTrack()

kyteague:
#### Description
This PR implements RTPSender.ReplaceTrack, allowing callers to atomically swap tracks even as the stream is in progress. See https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/replaceTrack

It doesn't do a full check for renegotiation, but I think it's good enough for most people's use cases.

Let me know if there is an appropriate place to add a test.

#### Reference issue
Fixes #1020

### DIFF
--- a/rtpsender.go
+++ b/rtpsender.go
@@ -32,26 +32,23 @@ type RTPSender struct {
 
 // NewRTPSender constructs a new RTPSender
 func (api *API) NewRTPSender(track *Track, transport *DTLSTransport) (*RTPSender, error) {
-	if track == nil {
-		return nil, fmt.Errorf("Track must not be nil")
-	} else if transport == nil {
+	if transport == nil {
 		return nil, fmt.Errorf("DTLSTransport must not be nil")
 	}
 
-	track.mu.Lock()
-	defer track.mu.Unlock()
-	if track.receiver != nil {
-		return nil, fmt.Errorf("RTPSender can not be constructed with remote track")
-	}
-	track.totalSenderCount++
-
-	return &RTPSender{
-		track:      track,
+	r := &RTPSender{
 		transport:  transport,
 		api:        api,
 		sendCalled: make(chan interface{}),
 		stopCalled: make(chan interface{}),
-	}, nil
+	}
+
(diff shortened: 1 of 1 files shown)


sgotti:
@kyteague Thanks for the PR!

I'm not sure if this will work with the current Track semantic: replaceTrack in webrtc spec will replace a mediaTrack, this means that the encoder will continue sending the same rtp stream with the same ssrc and consequent rtp sequences, timestamps and codec changes (for example vp8 pictureID and TL0PICIDX).

Instead current pion/webrtc track is just a way to read/write rtp packets.

So I don't think a swap track will be useful with the current pion track semantics but should be done from the callers of pion/webrtc. As a small examples see the swap-tracks https://github.com/pion/webrtc/blob/master/examples/swap-tracks/main.go (that just replaces the rtp ssrc, seq and timestamp).


I have a work in progress sfu core that I'm going to release that just does this with all the advanced logic (related to simulcast implementation I'm working on #1016)

jech:
@sgotti, what if we check for SSID's and codecs being identical?

Sean-Der:
Resolving for #1527 

Thank you so much for this contribution @kyteague, I am sorry we didn't get to merge this. Our RTP infrastructure just went through a massive refactor, and now we can properly assert the MimeType/Format Parameters.

PR #1200, opened 2020-05-15
[WIP] Implement simulcast

sgotti:
#### Description

---
#### NOTE
This should be considered a patch set. As a starting point I opened it as an unique PR so you can see all the work split in multiple patches (for an easier review) and the final outcome. The last patch is a simulcast example so everyone can test it with both chrome and firefox.

@Sean-Der @jeremija If you prefer that I create multiple PR for every commit let me know, opening them right now will make it difficult to follow the reason of every single PR without the global view.

---

Here's a list of the various patches

The first Patches are initial hacks to improve transceiver selection since simulcast requires sendonly sender and the transceiver must be read only, something that currently is not automatic without these patches.

The 3 patches:
* `Make trackDetails handle multi rtp streams tracks`
* `Make track multistream`
* `Make rtpreceiver multi stream`
Introduce the concept of multi rtp stream tracks. Every patch should have a quite complete explanation.

* `Negotiate extmaps` Adds extmaps negotiation (see commit description) 

* `Implement simulcast (receive)` Implement simulcast on the receiver side. The sending side is not yet implemented since the primary use case is to receive simulcast tracks from a browser (and browsers only implement simulcast sending).

* `Add simulast example` Adds a simulcast example that work with both chrome and firefox.

### TODO

[*] Add many tests. Without the sending part it's difficult to create e2e test pion<->pion. It'll also be good to implement tests using the browser e2e tests.

#### Reference issue
#1016

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -810,7 +810,26 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 			}
 
 			t, localTransceivers = findByMid(midValue, localTransceivers)
-			if t == nil {
+			if t != nil {
+				transceiveDirection := t.Direction()
+
+				switch direction {
+				case RTPTransceiverDirectionRecvonly:
+					if transceiveDirection == RTPTransceiverDirectionRecvonly {
+						t.setDirection(RTPTransceiverDirectionInactive)
+					}
+					if transceiveDirection == RTPTransceiverDirectionSendrecv {
+						t.setDirection(RTPTransceiverDirectionSendonly)
+					}
+				case RTPTransceiverDirectionSendonly:
+					if transceiveDirection == RTPTransceiverDirectionSendonly {
+						t.setDirection(RTPTransceiverDirectionInactive)
+					}
+					if transceiveDirection == RTPTransceiverDirectionSendrecv {
+						t.setDirection(RTPTransceiverDirectionRecvonly)
+					}
+				}
+			} else {
 				t, localTransceivers = satisfyTypeAndDirection(kind, direction, localTransceivers)
 			}
 			if t == nil {
@@ -823,6 +842,7 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 			if t.Mid() == "" {
 				_ = t.setMid(midValue)
 			}
+			t.setRemoteDirection(direction)
 		}
 	}
 
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1198,7 +1198,12 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 
 	var transceiver *RTPTransceiver
 	for _, t := range pc.GetTransceivers() {
-		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil {
+		// use existing transceiver if:
+		// * not stopped
+		// * same kind
+		// * without a sender
+		// * the transceiver has not been negotiated yet or negotiated and remote direction can receive
+		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil && (t.getRemoteDirection() == RTPTransceiverDirection(Unknown) || t.getRemoteDirection() == RTPTransceiverDirectionSendrecv || t.getRemoteDirection() == RTPTransceiverDirectionRecvonly) {
 			transceiver = t
 			break
 		}
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -9,6 +9,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	mathRand "math/rand"
+	"net/url"
 	"strconv"
 	"strings"
 	"sync"
@@ -59,6 +60,10 @@ type PeerConnection struct {
 	// should be defined (see JSEP 3.4.1).
 	greaterMid int
 
+	// extmaps
+	remoteExtMaps map[int]*sdp.ExtMap
+	extMaps       map[int]*sdp.ExtMap
+
 	rtpTransceivers []*RTPTransceiver
 
 	onSignalingStateChangeHandler     func(SignalingState)
@@ -108,12 +113,40 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
 		lastOffer:                    "",
 		lastAnswer:                   "",
 		greaterMid:                   -1,
+		remoteExtMaps:                make(map[int]*sdp.ExtMap),
 		signalingState:               SignalingStateStable,
 		iceConnectionState:           ICEConnectionStateNew,
 		connectionState:              PeerConnectionStateNew,
 
 		api: api,
 		log: api.settingEngine.LoggerFactory.NewLogger("pc"),
+	}
+
+	transportCCURL, _ := url.Parse(TransportCCURI)
+	sdesMidURL, _ := url.Parse(SDESMidURI)
+	sdesRTPStreamIDURL, _ := url.Parse(SDESRTPStreamIDURI)
+
+	// populate with locally supported extmaps
+	pc.extMaps = map[int]*sdp.ExtMap{
+		// use the default value for transportcc
+		extMapValueTransportCC: {
+			Value: extMapValueTransportCC,
+			URI:   transportCCURL,
+			// support both send and recv direction
+			Direction: 0,
+		},
+		extMapValueSDESMid: {
+			Value: extMapValueSDESMid,
+			URI:   sdesMidURL,
+			// support both send and recv direction
+			Direction: 0,
+		},
+		extMapValueSDESRTPStreamID: {
+			Value: extMapValueSDESRTPStreamID,
+			URI:   sdesRTPStreamIDURL,
+			// support both send and recv direction
+			Direction: 0,
+		},
 	}
 
 	var err error
@@ -791,6 +824,10 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 	var t *RTPTransceiver
 	localTransceivers := append([]*RTPTransceiver{}, pc.GetTransceivers()...)
 	detectedPlanB := descriptionIsPlanB(pc.RemoteDescription())
+
+	if err := pc.updateExtMaps(weOffer); err != nil {
+		return err
+	}
 
 	if !weOffer && !detectedPlanB {
 		for _, media := range pc.RemoteDescription().parsed.MediaDescriptions {
@@ -1970,4 +2007,71 @@ func (pc *PeerConnection) generateMatchedSDP(useIdentity bool, includeUnmatched 
 	}
 
 	return populateSDP(d, detectedPlanB, pc.api.settingEngine.candidates.ICELite, pc.api.mediaEngine, connectionRole, candidates, iceParams, mediaSections, pc.ICEGatheringState())
+}
+
+func (pc *PeerConnection) updateExtMaps(weOffer bool) error {
+	remoteExtMaps := map[int]*sdp.ExtMap{}
+
+	// populate the extmaps from the current remote description
+	for _, media := range pc.RemoteDescription().parsed.MediaDescriptions {
+		// populate known remote extmap and handle conflicts.
+		for _, attr := range media.Attributes {
+			if attr.Key != "extmap" {
+				continue
+			}
+			em := &sdp.ExtMap{}
+			if err := em.Unmarshal("extmap:" + attr.Value); err != nil {
+				pc.log.Warnf("Failed to parse ExtMap: %v", err)
+				continue
+			}
+			if remoteExtMap, ok := remoteExtMaps[em.Value]; ok {
+				if remoteExtMap.Value != em.Value {
+					return fmt.Errorf("RemoteDescription changed some extmaps values")
+				}
+			} else {
+				remoteExtMaps[em.Value] = em
+			}
+		}
+	}
+
+	// update our know remote extmaps from new extmaps from the current remote description
+	for _, curRemoteExtMap := range remoteExtMaps {
+		if remoteExtMap, ok := pc.remoteExtMaps[curRemoteExtMap.Value]; ok {
+			// check that the remote extmaps haven't changed some already known values
+			if remoteExtMap.Value != curRemoteExtMap.Value {
+				return fmt.Errorf("RemoteDescription changed some extmaps values")
+			}
+		} else {
+			// add new extmap
+			pc.remoteExtMaps[curRemoteExtMap.Value] = curRemoteExtMap
+		}
+	}
+
+	// If we are receiving an offer update our extMaps adapting to the remote values
+	if !weOffer {
+		for _, remoteExtMap := range pc.remoteExtMaps {
+			// update existing locally provided value
+			for _, extMap := range pc.extMaps {
+				if extMap.URI.String() == remoteExtMap.URI.String() {
+					extMap.Value = remoteExtMap.Value
+					break
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+// GetExtMapByURI return a copy of the extmap matching the provided
+// URI. Note that the extmap value will change if not yet negotiated
+func (pc *PeerConnection) GetExtMapByURI(uri string) *sdp.ExtMap {
+	pc.mu.RLock()
+	defer pc.mu.RUnlock()
+	for _, extMap := range pc.extMaps {
+		if extMap.URI.String() == uri {
+			return extMap
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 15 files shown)


jeremija:
Wow great work @sgotti! This will take a long time to process 😃 

I think a single PR works better because it will be easier to reason about the changes - if we have more related PRs it might become hard to keep them in sync.

But perhaps @Sean-Der has a different view?

peerconnection.go
```diff
@@ -932,49 +1005,64 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks map[uint32]trackDetai
...
-			t, err := pc.AddTransceiverFromKind(incoming.kind, RtpTransceiverInit{
-				Direction: RTPTransceiverDirectionSendrecv,
-			})
-			if err != nil {
-				pc.log.Warnf("Could not add transceiver for remote SSRC %d: %s", ssrc, err)
-				continue
+		for _, incoming := range incomingTracks {
+			for ssrc := range incoming.ssrcStreams {
+				t, err := pc.AddTransceiverFromKind(incoming.kind, RtpTransceiverInit{
+					Direction: RTPTransceiverDirectionSendrecv,
```
cretz, line 1058: What if it's recv only in the remote desc (pardon my ignorance towards plan b vs unified)?

peerconnection.go
```diff
@@ -1079,15 +1171,119 @@ func (pc *PeerConnection) drainSRTP() {
...
 
-			if !handleUndeclaredSSRC(ssrc) {
-				pc.log.Warnf("Incoming unhandled RTP ssrc(%d), OnTrack will not be fired", ssrc)
-			}
+			pc.mu.Lock()
+			pc.pendingReadStreamsSRTP[ssrc] = r
+			pc.mu.Unlock()
+
+			go func() {
+				// test first N (10) packets, if no mid and rid are found give up
```
cretz, line 1185: Is this the common way other impls determine whether a stream has a known rid? Is there nothing about the track that can be used to determine the rid without reading its packets first? So I may lose the first 10 packets?

peerconnection.go
```diff
@@ -1079,15 +1171,119 @@ func (pc *PeerConnection) drainSRTP() {
...
+				c := 0
+
+				// read incoming packet until we can populate mid and rid from packet
+				// extensions (not all packet will provide such information) so continue reading until we'll find both
+				var mid, rid string
+				for {
+					b := make([]byte, receiveMTU)
```
cretz, line 1192: This and the `rtp.Packet` below should be created outside the loop and reused (you may have to clear a value or two our of the rtp packet before unmarshalling each time...the lib should be using a buf pool IMO, but that's a different topic)

peerconnection.go
```diff
@@ -1079,15 +1171,119 @@ func (pc *PeerConnection) drainSRTP() {
...
+					var sdesMidExtMap *sdp.ExtMap
+					var sdesStreamIDExtMap *sdp.ExtMap
+					for _, extMap := range pc.extMaps {
+						switch extMap.URI.String() {
+						case sdesMidURI.String():
+							sdesMidExtMap = extMap
+						case sdesRTPStreamIDURI.String():
+							sdesStreamIDExtMap = extMap
+						}
+					}
```
cretz, line 1229: IMO, this type of thing should be precalculated at the PC level

peerconnection.go
```diff
@@ -1079,15 +1171,119 @@ func (pc *PeerConnection) drainSRTP() {
...
+
+					if payload := rp.GetExtension(uint8(sdesMidExtMap.Value)); payload != nil {
+						curMid := string(payload)
+						pc.log.Infof("curMid: %q", curMid)
```
cretz, line 1233: Arguably some of this is a bit noisy for info level

peerconnection.go
```diff
@@ -1685,24 +1895,61 @@ func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, re
...
+			if incoming, ok := trackDetails[trackID]; !ok {
+				// no matching track id
+				// check if there's a track with the same ssrc but different track id
+				for _, incoming := range trackDetails {
+					for ssrc := range incoming.ssrcStreams {
+						if t.Receiver().Track().streams[0].ssrc == ssrc {
+							matched = true
+							// update receiver track id and label
+							t.Receiver().Track().id = incoming.mstid
+							t.Receiver().Track().label = incoming.msid
```
cretz, line 1930: Break out both loops after this?

peerconnection.go
```diff
@@ -1867,7 +2134,41 @@ func (pc *PeerConnection) generateMatchedSDP(useIdentity bool, includeUnmatched
...
-			mediaSections = append(mediaSections, mediaSection{id: midValue, transceivers: mediaTransceivers})
+
+			extMaps := []*sdp.ExtMap{}
+
+			if hasSimulcast {
+				// Add mid and streamID extmaps
+				for _, extMap := range pc.extMaps {
+					switch extMap.URI.String() {
+					case sdesMidURI.String():
+						fallthrough
```
cretz, line 2145: Alternatively, `case sdesMidURI.String(), sdesRTPStreamIDURI.String():`

peerconnection.go
```diff
@@ -1887,3 +2191,57 @@ func (pc *PeerConnection) generateMatchedSDP(useIdentity bool, includeUnmatched
...
+				continue
+			}
+			em := &sdp.ExtMap{}
+			if err := em.Unmarshal("extmap:" + attr.Value); err != nil {
+				pc.log.Warnf("Failed to parse ExtMap: %v", err)
+				continue
+			}
+			if remoteExtMap, ok := remoteExtMaps[em.Value]; ok {
+				if remoteExtMap.Value != em.Value {
+					return fmt.Errorf("RemoteDescription changed some extmaps values")
```
cretz, line 2212: I haven't dug into the spec, how common is it that possibly different media descriptions can have the same extmap URI with a different ID?

sdp.go
```diff
@@ -12,23 +13,52 @@ import (
 	"github.com/pion/sdp/v2"
 )
 
+const (
+	// TODO(sgotti) add these to pion/sdp
+	extMapValueSDESMid         = 4
+	extMapValueSDESRTPStreamID = 5
```
cretz, line 19: Is this the value that Chrome just happens to use for these extensions or is fixed in spec?

cretz, line 19: Is this the value that Chrome just happens to use for these extensions or is fixed in spec?

cretz, line 19: Is this the value that Chrome just happens to use for these extensions or is fixed in spec?

sdp.go
```diff
@@ -247,7 +351,23 @@ func addTransceiverSDP(d *sdp.SessionDescription, isPlanB bool, mediaEngine *Med
 	for _, mt := range transceivers {
 		if mt.Sender() != nil && mt.Sender().track != nil {
 			track := mt.Sender().track
-			media = media.WithMediaSource(track.SSRC(), track.Label() /* cname */, track.Label() /* streamLabel */, track.ID())
+			if len(track.streams) > 1 {
+				//if track.simulcast {
```
cretz, line 355: Commented out code?

sdp.go
```diff
@@ -315,6 +438,37 @@ func getMidValue(media *sdp.MediaDescription) string {
...
+		if attr.Key == "simulcast" {
+			return true
+		}
+	}
+	return false
+}
+
+func hasExtension(extMaps map[int]*sdp.ExtMap, uri fmt.Stringer) bool {
+	for _, extMap := range extMaps {
+		if extMap.URI.String() == uri.String() {
```
cretz, line 465: I haven't really nitpicked on perf on this review, but in several cases like this, accepting the string URI is clearer IMO

stream.go
```diff
@@ -0,0 +1,185 @@
...
+// received.
+func (s *TrackRTPStream) PayloadType() uint8 {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.payloadType
+}
+
+// SSRC gets the current SSRC of the stream. If a stream is RID based,
+// this is the currently known ssrc (may be 0 until a packet is
+// received) for the stream rid and may change
```
cretz, line 65: (pedantic) Inconsistent throughout on whether punctuation used at end of doc sentences

stream.go
```diff
@@ -0,0 +1,185 @@
...
+// this is useful if we are dealing with a remote stream and we can't announce it to the user until we know the payloadType
+func (s *TrackRTPStream) determinePayloadType() error {
+	r, err := s.ReadRTP()
+	if err != nil {
+		return err
+	}
+
+	s.mu.Lock()
+	s.payloadType = r.PayloadType
+	defer s.mu.Unlock()
```
cretz, line 182: Unnecessary `defer`

track.go
```diff
@@ -63,56 +64,48 @@ func (t *Track) Label() string {
...
 }
 
-// SSRC gets the SSRC of the track
-func (t *Track) SSRC() uint32 {
+// Streams return the track streams
+func (t *Track) Streams() []*TrackRTPStream {
 	t.mu.RLock()
 	defer t.mu.RUnlock()
-	return t.ssrc
+	return append([]*TrackRTPStream{}, t.streams...)
```
cretz, line 71: Pedantic perf... costs a few more lines of code, but a `make([]*TrackRTPStream{}, len(t.streams))` and a `copy` are clearer about capacity

Sean-Der:
Fantastic work @sgotti! I finally got around to typing up [MediaAPI](https://github.com/pion/webrtc/wiki/MediaAPI) when you get a chance mind looking?

I am sure it has lots of issue, but what do you think of taking things in this direction? I think the best thing we can do is allow people to express Simulcast (and SVC) using ORTC primitives, but I have never actually done anything around this! Would love your opinion.

peerconnection.go
sgotti: Chrome and firefox send the mid/streamID header extensions in every rtcp packet so we will skip the first packet. Note that the first packet was also previously skipped in single stream tracks: https://github.com/pion/webrtc/blob/v2.2.12/peerconnection.go#L896

peerconnection.go
sgotti: I just kept the same logic as before adapting to the new trackDetails format. So it'll work as before. IMHO I'll just remote planb...

peerconnection.go
sgotti: Just a debugging refuse. Removed

peerconnection.go
sgotti: Chrome and Firefox uses totally different values.

sdp.go
sgotti: I used the values like the one chrome used, but they could just be random. The spec doesn't provide default values.

stream.go
sgotti: Just copied the current implementation in Track. I'll remove the defer.

track.go
sgotti: I agree. Updated.

peerconnection.go
sgotti: I created a function to retrieve and extmaps and a check that all the required ones are negotiated to execute this goroutine.

sgotti:
@cretz Thanks a lot for your detailed review. I hope to have answered all of your questions and addressed all your suggestions.

sdp.go
cretz: Right, they should be determined from the offer per media description. Implementations I've seen start the extmap values at 1 and count up for each of the extensions they offer.

peerconnection.go
cretz: What I'm saying is, if I send two video tracks in the same offer (maybe with different codecs), can you guarantee that the `extmap` for `urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id` has the same ID on both of them (i.e. a common value across the entire offer)? Or might the `extmap` IDs for the same extension on different media descriptions be different even in the same offer?

cretz:
@Sean-Der Unsure if this is the best place for feedback on the wiki, but for my use case, I'd really like to still have raw read and write abilities.  I need to control my byte buffers and I don't want to create an `rtp.Packet` for _every_ read for example. I'll accept a `Next(*rtp.Packet)` or something that lets me pass in a packet (and hopefully reuses byte slices underneath instead of creating a `[]byte` everytime like today's impl). I know it may not seem like much, but it adds up. Also, I use the SSRC and the label and what not in the track, so I think visibility still has value.

sdp.go
sgotti: We could just do it, it really doesn't matter. We could also use random numbers every time...

Currently I'm using pion/sdp  ExtMapValueTransportCC that has a value of 3 (don't know why and probably I started with 4 for that reason).

I'll just propose to remove `sdp.WithTransportCCExtmap` function that's not used anymore and IMHO wrong.

sgotti:
> @Sean-Der Unsure if this is the best place for feedback on the wiki, but for my use case, I'd really like to still have raw read and write abilities. I need to control my byte buffers and I don't want to create an `rtp.Packet` for _every_ read for example. I'll accept a `Next(*rtp.Packet)` or something that lets me pass in a packet (and hopefully reuses byte slices underneath instead of creating a `[]byte` everytime like today's impl). I know it may not seem like much, but it adds up. Also, I use the SSRC and the label and what not in the track, so I think visibility still has value.

This is quite off-topic :smile: and we should continue on Slack. But I have the same needs and that's why I made my own proposal: https://github.com/pion/webrtc/wiki/MediaAPI#sgotti-proposal

peerconnection.go
sgotti: @cretz The spec https://tools.ietf.org/html/rfc5285 says:
```
The mapping may be provided per media stream (in the media-level
   section(s) of SDP, i.e., after an "m=" line) or globally for all
   streams (i.e., before the first "m=" line, at session level).  The
   definitions MUST be either all session level or all media level; it
   is not permitted to mix the two styles.  In addition, as noted above,
   the IDs used MUST be unique for each stream type for a given media,
   or for the session for session-level declarations.
```
So the last phrase: **In addition, as noted above, the IDs used MUST be unique for each stream type for a given media, or for the session for session-level declarations.** make it intends that ID (extmap values) may be different for different media streams.

The conflict is that the sdes mid extension header value cannot be different since it's used for determining the media section...
Currently chrome and firefox put extmaps per media section (not globally) and use the same IDs for all the media sections.

Let me know if you prefer that I split extmaps for media sections with an additional check to ensure that the sdes-mid extension value is the same for all the media section.

sgotti: @cretz I'll change this patch to store and negotiate them per transceiver level since it's cleaner and useful for future improvements.

tarrencev:
Is this still active? Excited for simulcast support :D

Sean-Der:
Hey 100% we need to get this in @tarrencev! The delay is my fault, there is a lot of really good stuff to go through here :)

I am hoping to spend some time this weekend reviewing this and reflecting on [webrtc-v3-api-examples](https://github.com/pion/webrtc-v3-api-examples)

jbrady42:
I'm taking a stab at cherry-picking out extmap support with the existing api. Its needed for https://github.com/pion/ion/pull/253

LHRchina:
@Sean-Der @sgotti will this PR merge soon? Some PR have already been merged into master like https://github.com/pion/webrtc/pull/1309 contain the impl of rid what conflict with this PR

Sean-Der:
Yes! I am shooting for ~Wednesday @LHRchina 

@jbrady42 is doing all the work, but the 'annoying subjective' part is done so the rest will be much easier :)

peerconnection.go
```diff
@@ -1685,24 +1907,66 @@ func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, re
...
 
-			t.Receiver().Track().mu.Lock()
-			ssrc := t.Receiver().Track().ssrc
-			if _, ok := trackDetails[ssrc]; ok {
-				incoming := trackDetails[ssrc]
-				t.Receiver().Track().id = incoming.id
-				t.Receiver().Track().label = incoming.label
-				t.Receiver().Track().mu.Unlock()
+			// skip receiver using rids instead of ssrcs
+			if t.Receiver().useRid {
```
LHRchina, line 1920: if invoke t.Receiver() return nil , this line will make panic

Sean-Der:
This has landed in master with 6ee528d349a5642fd43e90fec67a287cf31c01a9 @LHRchina please try it out when you get a chance

Thanks you so much for all the work on this @sgotti, sorry it took so long to land :(

LHRchina:
> This has landed in master with [6ee528d](https://github.com/pion/webrtc/commit/6ee528d349a5642fd43e90fec67a287cf31c01a9) @LHRchina please try it out when you get a chance
> 
> Thanks you so much for all the work on this @sgotti, sorry it took so long to land :(

Thank You , I will try it soon

PR #1215, opened 2020-05-24
fixed #1001

### LINKED ISSUE #1001: The first RTP packet that triggers OnTrack wouldn't be received (trivial)

illuspas:
The determinePayloadType function read the first packet of RTP to get the PayloadType.
Cache this packet and return it on the next read.

### DIFF
--- a/track.go
+++ b/track.go
@@ -33,6 +33,8 @@ type Track struct {
 	receiver         *RTPReceiver
 	activeSenders    []*RTPSender
 	totalSenderCount int // count of all senders (accounts for senders that have not been started yet)
+
+	cachePacket *rtp.Packet
 }
 
 // ID gets the ID of the track
@@ -99,6 +101,12 @@ func (t *Track) Read(b []byte) (n int, err error) {
 
 // ReadRTP is a convenience method that wraps Read and unmarshals for you
 func (t *Track) ReadRTP() (*rtp.Packet, error) {
+	if t.cachePacket != nil {
+		r := t.cachePacket
+		t.cachePacket = nil
+		return r, nil
+	}
+
 	b := make([]byte, receiveMTU)
 	i, err := t.Read(b)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Sean-Der:
Fantastic! This LGTM @illuspas 

Would you mind writing a simple test? Just asserting that if I send two packets both get received?

track.go
```diff
@@ -99,6 +101,12 @@ func (t *Track) Read(b []byte) (n int, err error) {
 
 // ReadRTP is a convenience method that wraps Read and unmarshals for you
 func (t *Track) ReadRTP() (*rtp.Packet, error) {
+	if t.cachePacket != nil {
```
Sean-Der, line 104: Can you move this into `Read`, then should be good!

illuspas: What needs to be cached here is the rtp.Packet structure, so why move to Read?

illuspas:
It does work well in my project.
However I also feel that it is not elegant to fix the problem in this way.
It is better if there are other ways to get the PayloadType in 
```
func (pc *PeerConnection) startReceiver()
```
I haven't found it yet.

track.go
Sean-Der: This fix only works if the user calls `ReadRTP` if a user calls `Read` they will not get the first packet.

If we move the caching into `Read` then `Read` and `ReadRTP` will be fixed!

illuspas: Oh, I see. So it seems that this is not a good way.

Sean-Der:
Hey @illuspas 

I think your solution makes sense. We need to implement this caching somewhere, why not in `pion/webrtc`?

We could add a `Peek` method to `pion/srtp`, but we are just pushing the complexity into another package. IMO I would prefer any custom behavior (like this) to stay in `pion/webrtc`

PR #1216, opened 2020-05-24
Document Configuration and MediaEngine re-use

josharian:
I did a quick pass through to check that these claims are true (and I want them to be true!), but I am new to this codebase, so I could be wrong. Please review skeptically.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -26,34 +26,46 @@ const (
 	mediaNameVideo = "video"
 )
 
-// MediaEngine defines the codecs supported by a PeerConnection
+// A MediaEngine defines the codecs supported by a PeerConnection.
+// MediaEngines populated using RegisterCodec (and RegisterDefaultCodecs)
+// may be set up once and reused, including concurrently,
+// as long as no other codecs are added subsequently.
+// MediaEngines populated using PopulateFromSDP should be used
+// only for that session.
 type MediaEngine struct {
 	codecs []*RTPCodec
 }
 
-// RegisterCodec registers a codec to a media engine
+// RegisterCodec adds codec to m.
+// RegisterCodec is not safe for concurrent use.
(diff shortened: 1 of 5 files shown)


mediaengine.go
```diff
@@ -26,34 +26,40 @@ const (
...
 	m.RegisterCodec(NewRTPVP8Codec(DefaultPayloadTypeVP8, 90000))
 	m.RegisterCodec(NewRTPVP9Codec(DefaultPayloadTypeVP9, 90000))
 	m.RegisterCodec(NewRTPH264Codec(DefaultPayloadTypeH264, 90000))
 }
 
-// PopulateFromSDP finds all codecs in a session description and adds them to a MediaEngine, using dynamic
-// payload types and parameters from the sdp.
+// PopulateFromSDP finds all codecs in sd and adds them to m, using the dynamic
+// payload types and parameters from sd.
+// PopulateFromSDP is not safe for concurrent use.
```
Sean-Der, line 62: Not sure how to word this properly, but `PopulateFromSDP` is meant to be used when you are answering. The offerer controls the PayloadTypes for the call, and this allows an answerer to match the PayloadTypes properly of the offerer.

You should also probably only call `PopulateFromSDP` once on a MediaEngine, and then use it only for that single session.

josharian: Thanks. I took a stab and wordsmithing. Tell me what you think. It's slightly verbose, but I think it's better to be a bit too verbose than a bit too terse when it comes to docs.

josharian:
Just like https://github.com/pion/sdp/pull/52, Travis is complaining that:

> it failed 'Separate subject from body with a blank line'

But I don't see any commit messages that match that description. I really want to want to contribute to pion, but the constant battles with the linter are pretty frustrating and time consuming.

Sean-Der:
Merged!

The issue was that `Fix broken link to issue in comment` didn't have a body, just added one and we are all good :)

PR #1219, opened 2020-05-24
WASM: Fallback to parsing SDP line for OnICECandidate

saljam:
On Firefox, the RTCIceCandidate interface appears to just be an RTCIceCandidateInit in disguise. That is, it does not have properties for each individual component of the candidate line. It only has the raw SDP string in the candidate property.

This change falls back to parsing the candidate line if some expected property is missing when preparing the candidate for the callback OnICECandidate.

https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate

https://caniuse.com/#feat=mdn-api_rtcicecandidate_priority

Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Cedric Fung](https://github.com/cedricfung)
 * [Norman Rasmussen](https://github.com/normanr) - *Fix Empty DataChannel messages*
 * [Josh Bleecher Snyder](https://github.com/josharian)
+* [salmān aljammāz](https://github.com/saljam)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/ice_go.go
+++ b/ice_go.go
@@ -2,33 +2,9 @@
 
 package webrtc
 
-import "github.com/pion/sdp/v2"
-
 // NewICETransport creates a new NewICETransport.
 // This constructor is part of the ORTC API. It is not
 // meant to be used together with the basic WebRTC API.
 func (api *API) NewICETransport(gatherer *ICEGatherer) *ICETransport {
 	return NewICETransport(gatherer, api.settingEngine.LoggerFactory)
-}
-
-func newICECandidateFromSDP(c sdp.ICECandidate) (ICECandidate, error) {
-	typ, err := NewICECandidateType(c.Typ)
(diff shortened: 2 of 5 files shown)


Sean-Der:
Nice catch!

This changes makes sense to me. Would you be able to add a simple JS test? If you don't have the bandwidth I am happy to take it :)

This also failed some of our linters, mind taking a look at them? If you don't have the time happy to cherry-pick and fix myself :)
```
2.04s$ bash .github/assert-contributors.sh
Please add the following contributors to the README
salman aljammaz
salmān aljammāz
```

```
0.02s$ bash .github/lint-commit-message.sh
WASM: Fallback to parsing SDP line for OnICECandidate
On Firefox, the RTCIceCandidate interface appears to just be an
RTCIceCandidateInit in disguise. That is, it does not have properties
for each individual component of the candidate line. It only has
the raw SDP string in the candidate property.
This change falls back to parsing the candidate line if some expected
property is missing when preparing the candidate for the callback
OnICECandidate.
https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate
https://caniuse.com/#feat=mdn-api_rtcicecandidate_priority
-------------------------------------------------
The preceding commit message is invalid
it failed 'Limit the subject line to 50 characters' of the following checks
* Separate subject from body with a blank line
* Limit the subject line to 50 characters
* Capitalize the subject line
* Do not end the subject line with a period
* Wrap the body at 72 characters
```

saljam:
Sounds good, I'll address lint later today.

A side note: would it be more portable if OnICECandidate took ICECandidateInit instead of ICECandidate as argument? It would save us decoding the string only to re-encode again, and it would mirror AddICECandidate.

Obviously it would break the API. Maybe something to consider for v3.

saljam:
Added a simple test for both webkit and firefox styles.

Sean-Der:
@saljam :fire: so excited to see you contributing :)

Would it be helpful if I tagged a new version? Are you interested in getting involved in Pion? Would love to add you to the Pion org. Then you can review other peoples PRs (and create branches and tag yourself)

No really commitment, just looking to get more people involved. Would love your helping making all the WASM stuff better!

saljam:
Glad I can help! Hopefully there's more, time and bandwidth permitting. I'm still finding my way around the codebase.

The changes I have that use this aren't going to be ready for a while, so no rush for tagging a release.

PR #1220, opened 2020-05-25
WASM: allow PeerConnection to be initialised from a JS object

saljam:
This allows us to make a new PeerConnection from an RTCPeerConnection js.Value.

#### Description

My use case for this is a Go package that handles the signalling parts, but leaves the rest to the application. A JS application can pass its own PeerConnection initialised with whatever datachannels/streams then wait for ICE/signalling to finish.

### DIFF
--- a/peerconnection_js.go
+++ b/peerconnection_js.go
@@ -51,6 +51,10 @@ func (api *API) NewPeerConnection(configuration Configuration) (_ *PeerConnectio
 	}, nil
 }
 
+func (pc *PeerConnection) JSValue() js.Value {
+	return pc.underlying
+}
+
 // OnSignalingStateChange sets an event handler which is invoked when the
 // peer connection's signaling state changes
 func (pc *PeerConnection) OnSignalingStateChange(f func(SignalingState)) {


peerconnection_js.go
```diff
@@ -50,6 +50,14 @@ func (api *API) NewPeerConnection(configuration Configuration) (_ *PeerConnectio
 	}, nil
 }
 
+// NewPeerConnectionFromJS creates a new PeerConnection with the provided underlying JS RTCPeerConnectoin object.
+func NewPeerConnectionFromJS(pc js.Value) *PeerConnection {
```
saljam, line 54: I don't know how you feel about enlarging the API surface with this, or the divergence between wasm and non-wasm APIs.

One alternative would be to make `underlying` public, maybe embedded. This allows:

```
pc := webrtc.PeerConnection{Value: jsPC}
```

(We'd also have to deal with `api` being nil.)

Sean-Der:
I am not super familiar with `syscall/js`, but what do you think of adding somethng like `ValueOf`? You would need to create the PeerConnection in Go still. You quickly get the escape hatch to be fully in JS land though.

That looks like a pattern we do, but please push back if I am wrong!

saljam:
Yeah that should work.

My approach was to, as much as possible, defer any interaction with PeerConnection to the application, including calling its constructor. But having just had another look at the RTCConfiguration I don't think there's much value leaving that to the application.

saljam:
[js.ValueOf()](https://golang.org/pkg/syscall/js/#ValueOf) is for primitive values or anything that implements js.Wrapper.

Changed this commit to implement [js.Wrapper](https://tip.golang.org/pkg/syscall/js/#Wrapper)'s JSValue(), which makes it work with ValueOf.

Let me know if you had something else on mind.

saljam:
Rebased. My feeling is that there's no such thing as idiomatic Go/WASM style yet! But I think this is pretty close.

PR #1222, opened 2020-05-28
fix bug: rtpsender.Read may block infinite when rtpsender stopped bef…

cnderrauber:
rtpsender.Read return err when rtpsender stopped

#### Description
rtpsender.Read may block infinite when rtpsender stopped before Send been called
#### Reference issue
Fixes #...


Sean-Der:
Fantastic change! Thank you so much @cnderrauber. Would you like me to cut another release with your change?

cnderrauber:
@Sean-Der Does another release means next release? That's ok.  Receiver and Test has been added.

Sean-Der:
@cnderrauber merged, thank you so much for your contribution :)

jech:
@cnderrauber, this fixes an issue I've been trying to debug for weeks, under the assumption that the goroutine leak is in my code.

PR #1224, opened 2020-05-29
use exported errors for reads/writes to stopped rtp i/o

josharian:
This makes checking for these errors much more robust
than doing a string comparison.

This is a follow-up to #1222 (which I was delighted to see). There are other possible errors we could use (io.ErrUnexpectedEOF, io.EOF, a new exported webrtc error var), but this one seemed appropriate.

Hopefully we could merge this or something like it before tagging a new release, to avoid backwards compatibillty problems.

### DIFF
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -827,11 +827,11 @@ func TestRtpSenderReceiver_ReadClose_Error(t *testing.T) {
 	sender, receiver := tr.Sender(), tr.Receiver()
 	assert.NoError(t, sender.Stop())
 	_, err = sender.Read(make([]byte, 0, 1400))
-	assert.Error(t, err, "RtpSender has been stopped")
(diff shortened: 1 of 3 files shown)


Sean-Der:
Oh yes 100% agree with this @josharian! I approved, if you don't mind fixing the linter error you can merge :) if you are AFK happy to push and fix myself

```
bash .github/lint-commit-message.sh
use exported errors for reads/writes to stopped rtp i/o

This makes checking for these errors much more robust
than doing a string comparison.

-------------------------------------------------
The preceding commit message is invalid
it failed 'Limit the subject line to 50 characters' of the following checks

* Separate subject from body with a blank line
* Limit the subject line to 50 characters
* Capitalize the subject line
* Do not end the subject line with a period
* Wrap the body at 72 characters
```

josharian:
Done.

PR #1225, opened 2020-05-29
diagnose calling SetRemoteDescription with offer instead of answer

### LINKED ISSUE #1210: better error message when offer/answer get confused
I accidentally typed `conn.SetRemoteDescription(offer)` when I should have written `conn.SetRemoteDescription(answer)`. This resulted in a nil pointer panic. It took me a while digging around in the code to find out what went wrong. It would be nice if pion could detect this mistake and return a helpful error message instead of panic.

josharian:
This same test used to cause a panic.

Test courtesy of @jech.

Fixes #1210


josharian:
This is not the right fix; see https://github.com/pion/webrtc/issues/1210#issuecomment-636436078.

PR #1229, opened 2020-05-31
Simplify operations

jech:
This simplifies the operations queue.  While the new code should be slightly faster and more robust, the main benefit is to reduce the amount of noise in the blocking profile.

### DIFF
--- a/operations.go
+++ b/operations.go
@@ -12,15 +12,10 @@ type operations struct {
 	ops     []operation
 	mu      sync.Mutex
 	startMu sync.Mutex
-	done    chan struct{}
 }
 
 func newOperations() *operations {
-	closed := make(chan struct{})
-	close(closed)
-	return &operations{
-		done: closed,
-	}
+	return &operations{}
 }
 
 // Enqueue adds a new action to be executed. If there are no actions scheduled,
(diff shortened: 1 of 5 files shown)


operations.go
```diff
@@ -9,60 +9,73 @@ type operation func()
...
-	mu      sync.Mutex
-	startMu sync.Mutex
-	done    chan struct{}
+	mu   sync.Mutex
+	busy bool
+	ops  []operation
 }
 
 func newOperations() *operations {
 	closed := make(chan struct{})
```
Sean-Der, line 18: Can you delete this and the following line, thanks!

operations.go
```diff
@@ -9,60 +9,73 @@ type operation func()
...
 // Enqueue adds a new action to be executed. If there are no actions scheduled,
 // the execution will start immediately in a new goroutine.
 func (o *operations) Enqueue(op operation) {
+	if op == nil {
+		return
+	}
+
 	o.mu.Lock()
-	defer o.mu.Unlock()
+	busy := o.busy
```
Sean-Der, line 31: What do you think of something like `queueInProgress`? Major nit, but for some reason I kept having to re-read because `busy` is on the struct and on stack.

operations.go
```diff
@@ -9,60 +9,73 @@ type operation func()
...
 	}
 }
 
-// Done will return a channel that will be closed as soon as all currently
-// enqueued operations are finished.
-func (o *operations) Done() <-chan struct{} {
-	o.mu.Lock()
-	defer o.mu.Unlock()
-	return o.done
+// Done blocks until all currently enqueued operations are finished executing.
```
Sean-Der, line 41: Mind returning a `context.Context`, then the caller can give up waiting if they want. In the future might be useful for timeouts/measuring stuff.

Sean-Der:
This fantastic! Thank you so much @jech, if you disagree I am happy to just push things forward. 

I am just paranoid about adding a bug to the operations queue, seem like the best place for a deadlock to pop up

jech:
> Can you delete this and the following line, thanks!

Done, sorry.

> What do you think of something like queueInProgress? Major nit, but for
> some reason I kept having to re-read because busy is on the struct and
> on stack.

Renamed to `running`.  Better?

> Mind returning a context.Context

I disagree, for at least two reasons.  First, any more complex
synchronisation can be easily built by the user; I've added an example in
operations_test.go.  Second, many people dislike Context, and I don't
think we should be imposing its use.

> I am just paranoid about adding a bug to the operations queue, seem like
> the best place for a deadlock to pop up

I agree, we should get rid of it altogether.

-- Juliusz

Sean-Der:
LGTM! Merging, thank you so much @jech

Sean-Der:
Merged with 9bbffdd5aab2f271629aef6b912190d0de358d6e thank you!

PR #1230, opened 2020-05-31
Extmap Support

jbrady42:
This is a pass at pulling out the extmap functionality from https://github.com/pion/webrtc/pull/1200. Thanks to @sgotti for your great work.

Currently only transport-cc and abs-send are enabled.
Looking into adding some ext specific tests.
 
Any feedback is appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [salmān aljammāz](https://github.com/saljam)
 * [cnderrauber](https://github.com/cnderrauber)
 * [Juliusz Chroboczek](https://github.com/jech)
+* [Jason Brady](https://github.com/jbrady42)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -9,6 +9,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	mathRand "math/rand"
+	"net/url"
 	"strconv"
 	"strings"
 	"sync"
@@ -59,6 +60,10 @@ type PeerConnection struct {
 	// should be defined (see JSEP 3.4.1).
 	greaterMid int
 
+	// extmaps
+	remoteExtMaps map[int]*sdp.ExtMap
+	extMaps       map[int]*sdp.ExtMap
+
 	rtpTransceivers []*RTPTransceiver
 
 	onSignalingStateChangeHandler     func(SignalingState)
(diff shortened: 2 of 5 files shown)


Sean-Der:
Thank you so much for cherry-picking this @jbrady42! This might be a little painful, but we can get this in together :) If you don't have the bandwidth or you prefer I can do these changes myself! Mind just making a branch directly on the repo?

* Can you remove any mention of `recvSimulcast`. Sorry to nit, but just want to get one concept in at a time.
* Should the `extmap` be empty by default? Since we don't by default support anything we probably shouldn't set it.
* It would be nice if we could put the logic around `media type` when setting. It seems like I should as a user be able to set an extmap for only audio, video or datachannels (or all).
* `updateExtMaps` should probably be moved into `sdp.go`. Would love for it to not be aware of the `PeerConnection` at all and keep the logic smaller. In general I would like to see if we can pull this off without adding anything extra to the `PeerConnection`

I also need to confirm myself the logic around extmap handling. It sounds like when answering we MUST not include an extension the remote doesn't support? I think we could make the logic simpler, we just need to filter extensions if the remote list is set.

jech:
> Additionally these value must be the same for every media section since, for example, the mid extension extmap is used for getting the related mid section of the rtp packet.

Looking at offers generated by Chrome, the global extmap and the extmap associated with each section are all distinct, although they use the same values.  I have yet to grok the RFC fully, but that makes a lot of sense to me.  This is the set of extensions negotiated globally:
```
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
```
This is the set of extensions negotiated for video tracks:
```
a=extmap:14 urn:ietf:params:rtp-hdrext:toffset
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:13 urn:3gpp:video-orientation
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:12 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:11 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:8 http://tools.ietf.org/html/draft-ietf-avtext-framemarking-07
a=extmap:9 http://www.webrtc.org/experiments/rtp-hdrext/color-space
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
```
No extensions are negotiated specifically for audio tracks.

Sean-Der:
@jech what do you think of the checklist over on https://github.com/pion/webrtc/pull/1233 ?

Would love your input, I totally missed they need to be at the global level as well!

Issue #889, opened 2019-10-23
Implement H.264 writer


gqgs:
## Summary

Implement H.264 writer.

## Motivation

Apparently the only video writer currently available is the IVF VP8 writer.
https://github.com/pion/webrtc/blob/758476212459d4faa882aa82934bf7099fe6a86e/pkg/media/ivfwriter/ivfwriter.go#L72

It would be useful to have a writer for H.264 as well.
Since the support for both codecs is required by the WebRTC specification the user would be able to use and save to disk whichever format best fits his/her use case.

Sean-Der:
Agree! If there is a simple container format I am all for adding it to `pion/webrtc`

For now I would use webm, there is an example here https://github.com/pion/example-webrtc-applications/tree/master/save-to-webm

Maybe we should add a comment to `save-to-disk` to use `webm` if you want multiple tracks in one file or to support other codecs! I think that would be a great PR, could really help others in the future :)

PR #1232, opened 2020-05-31
Add H264 writer

gqgs:
Currently it only supports non-interleaved mode
Therefore, only 1-23, 24 (STAP-A), 28 (FU-A) NAL types are allowed

#### Reference issue
Related #889

### DIFF
--- a/pkg/media/h264writer/h264writer.go
+++ b/pkg/media/h264writer/h264writer.go
@@ -1,0 +1,90 @@
+// Package h264writer implements H264 media container writer
+package h264writer
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


Sean-Der:
@gqgs nice work! thank you so much for doing this.

I am looking now, but I think your code is a superset of https://github.com/pion/rtp/blob/master/codecs/h264_packet.go#L151 what do you think of porting it over there?

gqgs:
@Sean-Der Redundant code removed.

Sean-Der:
Thank you so much for the contribution :)

PR #1233, opened 2020-05-31
Extmap Support

jbrady42:
#### Description

Continuation from https://github.com/pion/webrtc/pull/1230

Comments from @Sean-Der 
- Should the extmap be empty by default? Since we don't by default support anything we probably shouldn't set it.
- It would be nice if we could put the logic around media type when setting. It seems like I should as a user be able to set an extmap for only audio, video or datachannels (or all).
- `updateExtMaps` should probably be moved into sdp.go. Would love for it to not be aware of the PeerConnection at all and keep the logic smaller. In general I would like to see if we can pull this off without adding anything extra to the PeerConnection
I also need to confirm myself the logic around extmap handling. It sounds like when answering we MUST not include an extension the remote doesn't support? I think we could make the logic simpler, we just need to filter extensions if the remote list is set.

TODO
- [ ] No extensions by default
- [ ] Set extensions conditionally (global, audio, video, application)
- [ ] Move SDP specific logic into `sdp.go` and cover with unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [salmān aljammāz](https://github.com/saljam)
 * [cnderrauber](https://github.com/cnderrauber)
 * [Juliusz Chroboczek](https://github.com/jech)
+* [Jason Brady](https://github.com/jbrady42)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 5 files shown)


Sean-Der:
@jbrady42 I am happy to help with this! Would you like me to make changes/write tests myself?

This is a learning experience for me as well, so I would double check my assumptions if you get a chance :)

jbrady42:
I was going to take another pass today and those first points then will pass it over your way.

Where are you thinking the exts will be set if its per media type? Is that done when constructing the pc or adding tracks/transceivers?

PR #1238, opened 2020-06-05
Allow SDP munging

jech:
I'm submitting this for discussion.  I'm not happy with this, but I don't see a different way to make it possible to implement in client code features that are not natively supported by the library.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -610,8 +610,6 @@ func (pc *PeerConnection) setDescription(sd *SessionDescription, op stateChangeO
 		cur := pc.signalingState
 		setLocal := stateChangeOpSetLocal
 		setRemote := stateChangeOpSetRemote
-		newSDPDoesNotMatchOffer := &rtcerr.InvalidModificationError{Err: fmt.Errorf("new sdp does not match previous offer")}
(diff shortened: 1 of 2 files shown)


Sean-Der:
I don't have a problem with this! It might be nice to store a SDP, and then be able to call `GetLocalDescription` later with ICE candidates populated.

----

So one thing to note is that we don't drive anything via the `SetLocalDescription`. All 'custom behavior' is drive by the `SettingEngine`. This differs from Google's implementation, maybe the wrong decision but this was my motivation.

* API Stability - No way for users to know when features are turned on/off with SDP munging
* API Discoverability - Users have no way to figure out what APIs we actually provide
* Makes Pion code way simpler - Reacting to `SettingEngine` is much easier then SDP parsing.

I don't know the full story behind SDP munging. My guess is that Google doesn't want to run everything through the IETF/W3C. A goal with ORTC was to get away from munging, so I think we are making the right decision! I try to implement everything via ORTC/PeerConnection, and then last resort is `SettingEngine`

jech:
So are you saying that I might as well munge the SDP after `SetLocalDescription`, since munging it before doesn't achieve anything different?

Sean-Der:
@jech correct! The only thing you are losing is ICE candidates.

Issue #1242, opened 2020-06-06
improve samplebuilder docs


josharian:
There's no indication of the units for maxLate.

It's not clear what an rtp.Depacketizer is or what a standard one might look like. What is the byte  slice arg and what byte slice is it returning? The tests only show a `fakeDepacketizer` that is a passthrough. In my testing, a passthrough depacketizer works fine. (Should depacketizer by an option? Should it have type func instead of type interface?)

Are Push and Pop safe for concurrent use? If not, what is the intended use? (If you only Pop after Pushing, then it seems like you might lose packets unnecessarily if there is a delay in inbound packets, but maybe I misunderstand.)

tarrencev:
@josharian can we close this with your PR? or are further improvements necessary?

josharian:
I’m no longer using pion; I’m happy to let you decide about the best path forward for this issue.

Sean-Der:
The SampleBuilder has gone through lots of changes since this issue, don't believe these apply anymore.

PR #1245, opened 2020-06-07
Overhaul the samplebuilder docs

josharian:
There's one more piece I want to document, but I need some input.

What is the intended usage pattern? The obvious one is:

```go
// create SampleBuilder sb
for  {
  //  get packet p
  sb.Push(p)
  sample := sb.Pop()
  if sample == nil {
    continue
  }
  // use sample
}
```

But it seems like it is theoretically possible for multiple samples to be available as a result of a single new packet (e.g. a single packet was missing a long time and then shows up). So maybe correct use is like this?

```go
// create SampleBuilder sb
for  {
  //  get packet p
  sb.Push(p)
  for {
    sample := sb.Pop()
    if sample == nil {
      break
    }
    // use sample
  }
}
```

If you advise, I will document.

Also, I think the docs would benefit from mentioning typical maxLate sizes. What would you normally suggest? (Presumably different for audio and video?)

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -1,4 +1,4 @@
-// Package samplebuilder provides functionality to reconstruct media frame from RTP packets
+// Package samplebuilder provides functionality to reconstruct media frames from RTP packets.
 package samplebuilder
 
 import (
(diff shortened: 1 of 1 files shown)


josharian:
This will fix #1242.

Sean-Der:
@josharian 
Yes I think the second is a better example! Nice catch

> Also, I think the docs would benefit from mentioning typical maxLate sizes. What would you normally suggest? (Presumably different for audio and video?)

That is a tough one, it depends on
* Latency you are willing to tolerate
* Sample size (large video frames you will want more latency)
* Encoding properties of the video. If you are sending lots of keyframes I wouldn't wait at all.

I did [this](https://github.com/dialup-inc/ascii/blob/master/conn.go#L118) for ascii. I just came to that number via trial and error. I think what we really need to do is adjust the latency on the fly to find a good spot.

PR #1249, opened 2020-06-10
Implement NewRTPVP9CodecExt

jech:
This was omitted for some reason.

Note that the `Ext` variants are missing for all audio codecs.  I'm refraining from adding them right now, since the Unnamed SFU doesn't use any RTCP feedback for audio tracks, and hence I'm not able to test them easily.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -248,6 +248,19 @@ func NewRTPVP9Codec(payloadType uint8, clockrate uint32) *RTPCodec {
 	return c
 }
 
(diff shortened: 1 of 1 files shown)


jech:
> Do you want me to tag a new version with this change?

Ot's not particularly urgent, it's for simulcasting only, and I'm not sure yet if I'm going to do codec agility (it has some serious drawbacks, due to needing to renegotiate when switching layers).  The change is just so I can experiment.

PR #1273, opened 2020-06-25
Start /v3

Sean-Der:
See #9 for the features we have planned, and the breaking changes that
may occur.


Sean-Der:
@wdouglass excited to have you back involved!

PR #1285, opened 2020-06-29
Media Level Fingerprints Configuration Option

wdouglass:
This commit adds an option to put DTLS Fingerprint specification
at the media description level rather then the session description
level. This enhances compatibility with some 3rd party webrtc
implementations (notably Kurento).

Note: This will require pion/sdp PR 57 to get merged. I'm marking this PR as a draft until then.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1732,9 +1732,6 @@ func (pc *PeerConnection) GetRegisteredRTPCodecs(kind RTPCodecType) []*RTPCodec 
 // This is used for the initial call for CreateOffer
 func (pc *PeerConnection) generateUnmatchedSDP(useIdentity bool) (*sdp.SessionDescription, error) {
 	d := sdp.NewJSEPSessionDescription(useIdentity)
-	if err := addFingerprints(d, pc.configuration.Certificates[0]); err != nil {
-		return nil, err
-	}
 
 	iceParams, err := pc.iceGatherer.GetLocalParameters()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


peerconnection.go
```diff
@@ -1787,14 +1795,20 @@ func (pc *PeerConnection) generateUnmatchedSDP(useIdentity bool) (*sdp.SessionDe
...
+	
+	return populateSDP(d, isPlanB, mediaCert, pc.api.settingEngine.candidates.ICELite, pc.api.mediaEngine, connectionRoleFromDtlsRole(defaultDtlsRoleOffer), candidates, iceParams, mediaSections, pc.ICEGatheringState())
 }
 
 // generateMatchedSDP generates a SDP and takes the remote state into account
 // this is used everytime we have a RemoteDescription
 func (pc *PeerConnection) generateMatchedSDP(useIdentity bool, includeUnmatched bool, connectionRole sdp.ConnectionRole) (*sdp.SessionDescription, error) {
 	d := sdp.NewJSEPSessionDescription(useIdentity)
-	if err := addFingerprints(d, pc.configuration.Certificates[0]); err != nil {
+	if err := addSessionFingerprints(d, pc.configuration.Certificates[0]); err != nil {
```
Sean-Der, line 1811: What do you think of keeping `addFingerprints` but then have a third argument `SDPMediaLevelFingerprints`. Will make testing a lot easier I think.

I am just worried that having multiple places check `SDPMediaLevelFingerprints` will cause people to mess things up.

wdouglass: I agree, centralizing is probably a good idea, i'll take another pass

wdouglass: I've actually factored out the `addFingerprints` function entirely, the decision gets made in `populateSDP`. It's not safe to call `addFingerprints` more then once anyway, so this is a bit safer. I've messed something up in my refactor, and some tests are failing. I'll push a new commit soon once i figure that out. I think this is a definite improvement.

PR #1291, opened 2020-06-30
Get a list of transceivers before starting RTP

### LINKED ISSUE #1290: Enqueue logic should take copy of PeerConnection
I am seeing some races in `TestPeerConnection_Renegotiation_AddTrack` where we `AddTrack` after a `SetRemoteDescription`. The test `TestPeerConnection_Renegotiation_AddTrack` is flaking in master a bit. I am able to reproduce by doing `go test -run 'TestPeerConnection_Renegotiation_AddTrack' -count 50 -v`

The `SetRemoteDescription is then processed with the Track that was added BEFORE the call to `SetRemoteDescription`.

jeremija:
#### Description

@Sean-Der - does this fix the flakiness?

#### Reference issue
Hopefully Fixes #1290

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -727,11 +727,13 @@ func (pc *PeerConnection) SetLocalDescription(desc SessionDescription) error {
 		return err
 	}
 
+	currentTransceivers := append([]*RTPTransceiver{}, pc.GetTransceivers()...)
+
 	weAnswer := desc.Type == SDPTypeAnswer
 	remoteDesc := pc.RemoteDescription()
 	if weAnswer && remoteDesc != nil {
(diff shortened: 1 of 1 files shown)


Sean-Der:
Yes, this is amazing thank you so much @jeremija :fire:

I ran for 250 runs and zero failures (couldn't make it past 10 before)

jeremija:
Glad I was able to help @Sean-Der! :)

Thinking about this some more, perhaps a better way would be to lock the mutex first thing in `SetRemoteDescription` and then simply make a copy of the transceivers by using the private field. Because otherwise another goroutine might still modify the tracks after `SetRemoteDescription` was called, but before it `GetTransceivers` was called.

Issue #1277, opened 2020-06-27
QUIC tests are disabled


Sean-Der:
We only run the QUIC tests when `-tags quic` are enabled. We should add this functionality back.

cryptix:
Is there more to it then removing the build tag?

cryptix:
Apart from making the tests pass, of course. 

I found some TODOs in the pion/quic wrapper around end-of-stream stuff, currently trying to resolve that.

at-wat:
`pion/quic` is depending on the quite old version of `lucas-clemente/quic-go`.
I think it's better to resolve https://github.com/pion/quic/pull/8 before enabling it by default.

cryptix:
@at-wat yup, I already did that in pion/quic#31 together with some related changes.

at-wat:
oh, that's great!

Sean-Der:
Fixed by 940d6aae86fa13c993460900da7a7f21455bd806

PR #1294, opened 2020-07-01
Fix quic transport

cryptix:
* [x] the e2e test passes 
* [x] merge pion/ice#222
* [x] uses pion/quic@v0.1.2

~~Should I also remove the `quic` build tags from all the files or is there another task to #1277 ?~~ `quic` build tag will stay in place but i think we can close that issue.

closes #1323


cryptix:
I converted this to a draft because I hit a snag when testing this across NATed devices. It seems like the server looses the connection after a couple of packets but I didn't find the time yet to explore why this is and document it with packet captures and log output.

Sean-Der:
@cryptix I think we want to leave the `quic` build tags, but update [.travis.yml](https://github.com/pion/webrtc/blob/master/.travis.yml#L55) to specify.

I think in [this file](https://github.com/pion/webrtc/blob/master/.github/.ci.conf) you might need to just do `TEST_EXTRA_ARGS=--build-tags quic`

We also need to release a new version of ci-strings that pass https://github.com/pion/webrtc/blob/master/.travis.yml#L43 argument.

64eecd6d39f90dd9f447904fb62643a28434ea2a is the original commit. The issue is that QUIC brings in a lot of stuff, and for users on smaller devices it was causing issues :(

cryptix:
I think I got it. golangci-lint needs `--build-tags quic` after run while `TEST_EXTRA_ARGS` is passed to `go test` as is (so just `-tags quic` there).

cryptix:
This needs pion/quic#36 now to fix the panic of the missing logger in config.

.travis.yml
```diff
@@ -40,7 +40,7 @@ _lint_job: &lint_job
     - bash .github/lint-disallowed-functions-in-library.sh
     - bash .github/lint-commit-message.sh
     - bash .github/lint-filename.sh
-    - golangci-lint run ./...
+    - golangci-lint run --build-tags quic ./...
```
Sean-Der, line 43: Mind removing this line? We use [this](https://github.com/pion/.goassets) repo to keep all our configs in sync. Maybe we can just use `TEST_EXTRA_ARGS` here, or we need to add a new `GOLANGCI_LINT_EXRA_ARGS`

But this can be a follow up :)

Sean-Der:
amazing work @cryptix !

I approved, mind just remove the `.travis.yml` change and you can go ahead and do the `Rebase and merge` button :)

Sean-Der:
This really bloats our package tree, I think we need to find a way to reduce this :(

 Before this PR running `go mod tidy` takes ~3 seconds, after it goes to minutes also my `pkg` folder goes from `143M` -> `1.1G`

I will investigate and see what I can find!

cryptix:
> my pkg folder goes from 143M -> 1.1G

yikes.. i guess that’s because the go.mod doesn’t care about the build tags, right?

PR #1298, opened 2020-07-05
oggreader always fails if the input stream doesn't have data ready. 

kittuov:
#### Description
When tried to map a Reader which is not a file (stdout of subprocess) The oggreader always fails because it is expecting to read from a buffered input. 

In order to fix this, io.ReadFull is used like in ivfreader.


Sean-Der:
Thanks for fixing this @kittuov, merging!

Sean-Der:
Merged with 90121499393d122ee136d7c02774553043439b7d

Thank you so much for the contribution!

at-wat:
@Sean-Der assert-contributors.sh seems failing on master

Sean-Der:
@at-wat 👍 sorry my mistake, fixing in PR for remove CandidateSelectionTimeout!

PR #1313, opened 2020-07-13
Assert SDPType in setDescription

### LINKED ISSUE #1297: SetRemoteDescription should return an error if SDPType is not set
Currently we silently fail to process if unset

Sean-Der:
Before users could pass a SessionDescription with a 0 type. This would
be discarded silently. Now we return an error that matches browser
behavior.

Resolves #1297

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -592,8 +592,11 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
 
 // 4.4.1.6 Set the SessionDescription
 func (pc *PeerConnection) setDescription(sd *SessionDescription, op stateChangeOp) error {
-	if pc.isClosed.get() {
+	switch {
+	case pc.isClosed.get():
 		return &rtcerr.InvalidStateError{Err: ErrConnectionClosed}
+	case newSDPType(sd.Type.String()) == SDPType(Unknown):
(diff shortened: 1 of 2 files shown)


cryptix:
LGTM!

peerconnection.go
```diff
@@ -592,8 +592,11 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
 
 // 4.4.1.6 Set the SessionDescription
 func (pc *PeerConnection) setDescription(sd *SessionDescription, op stateChangeOp) error {
-	if pc.isClosed.get() {
+	switch {
+	case pc.isClosed.get():
 		return &rtcerr.InvalidStateError{Err: ErrConnectionClosed}
+	case sd.Type.String() == ErrUnknownType.Error():
```
at-wat, line 598:
```suggestion
	case newSDPType(sd.Type.String()) == SDPType(Unknown):
```
might be little bit better at semantics

peerconnection.go
```diff
@@ -592,8 +592,11 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
...
+		return &rtcerr.TypeError{Err: fmt.Errorf("The provided value '%d' is not a valid enum value of type SDPType", sd.Type)}
```
at-wat, line 599: stylecheck: error strings should not be capitalized

PR #1322, opened 2020-07-14
Implement OnNegotiationNeeded

### LINKED ISSUE #817: Implement OnNegotiationNeeded
This should be added to support the Tor folks!

tekig:
#### Description
Simple implementation of OnNegotiationNeeded
#### Reference issue
Fixes #817


Sean-Der:
Fantastic work @MrDell61 excited to have you involved :)

I think we will need some tests to be sure of this, but curious about what happens when.

* If you add multiple tracks, is `OnNegotiationNeeded` fired multiple times?
* Do we also want to fire when creating a DataChannel?
* Does this work across re-negotiations

jech:
> If you add multiple tracks, is OnNegotiationNeeded fired multiple times?

Since Javascript is synchronous, this is not a problem in Javascript — you just add multiple tracks synchronously, and the callback is invoked just once.  What is the right approach in Pion, where callbacks are fired in a different goroutne?

tarrencev:
@MrDell61 @Sean-Der anything i can help with to help land this? happy to write tests etc

MrDell61:
@Sean-Der,
> * If you add multiple tracks, is `OnNegotiationNeeded` fired multiple times?

He will try to wait for all the tracks to be added. Only then will it fired the event.

> * Do we also want to fire when creating a DataChannel?

Yes, this it is working

> * Does this work across re-negotiations

I think yes, you can try to test

It works for me, but maybe I'm missing something.

MrDell61:
@tarrencev 
I have doubts, false positives may occur. I've tested with `time.Sleep`, but don't know how to test correctly. It will be great if you test this automatically.

MrDell61:
```go
localMedia := make(map[string]*sdp.MediaDescription)
for _, m := range localDesc.parsed.MediaDescriptions {
	if mid, ok := m.Attribute(sdp.AttrKeyMID); ok {
		localMedia[mid] = m
	}
}

remotMedia := make(map[string]*sdp.MediaDescription)
for _, m := range remotDesc.parsed.MediaDescriptions {
	if mid, ok := m.Attribute(sdp.AttrKeyMID); ok {
		remotMedia[mid] = m
	}
}
```
I don't think this is a good idea, any suggestions?

MrDell61:
Need to implement variable `stopped`. In the next two weeks, I can't do this

Sean-Der:
Absolutely amazing work @tekig 

Sorry this took so long to review, I just moved this to https://github.com/pion/webrtc/pull/1359

Me and @tarrencev are going to get this merged today. The commit will still have you as the author, but we will work on anything else that needs to be done!

jech:
The change to `operations.start` is not correct — it will not do the right thing if an operation panics.

jech:
I'm not sure the ad-hoc `ops.done` thing is the right thing to do.  Instead of
```
ops.done = f
```
wouldn't it be possible to do a simple recursion like
```
g := func() {
    if !ops.isEmpty() {
        ops.Enqueue(g)
        return
    }
    f()
}
ops.Enqueue(g)
```

tarrencev:
@jech might be best to move your feedback to #1359

PR #1328, opened 2020-07-15
Add experimental support for ICE TCP

### LINKED ISSUE #1327: Add support for passive TCP candidates
# Summary

See https://github.com/pion/ice/issues/196 and the PR https://github.com/pion/ice/pull/226

jeremija:
See pion/ice issue and PR:

- https://github.com/pion/ice/issues/196
- https://github.com/pion/ice/pull/226

#### Description

This PR bumps `pion/ice` to `v2.0.0-rc6` and adds experimental support for passive ICE candidates over TCP. An example `examples/ice-tcp` has been added. It only enables UDP connectivity so there is no need to configure your firewall to disable UDP packets.

To enable ICE TCP the following needs to be done:

```
		settingEngine := webrtc.SettingEngine{}

		settingEngine.SetNetworkTypes([]webrtc.NetworkType{
			webrtc.NetworkTypeUDP4,
			webrtc.NetworkTypeUDP6,
			webrtc.NetworkTypeTCP4,
			webrtc.NetworkTypeTCP6,
		})
		settingEngine.SetICETCPPort(8443)
```

By default, only UDP is allowed as this is previous behavior. Currently `pion/ice` supports only passive TCP candidates. As this feature is experimental, the current API is likely to change in the future.

See https://tools.ietf.org/html/rfc6544 for more info.

#### Reference issue

Closes #1327


examples/ice-restart/main.go
```diff
@@ -15,7 +15,25 @@ func doSignaling(w http.ResponseWriter, r *http.Request) {
 	var err error
 
 	if peerConnection == nil {
-		if peerConnection, err = webrtc.NewPeerConnection(webrtc.Configuration{}); err != nil {
+		m := webrtc.MediaEngine{}
```
Sean-Der, line 18: Maybe better to do this in another example (and that can come in another PR!)

I just don't want people to have to deal with multiple concepts at once :/ but I think duplication is fine!

jeremija: Makes total sense. I added another example to `examples/ice-tcp` and reverted this one. Now users don't need to write rules for blocking UDP packets!

PR #1332, opened 2020-07-16
Simulcast2

jbrady42:


jech:
@jbrady42, FWIW, I've got a sort-of-working prototype of simulcast in

    git clone -b simulcast https://www.irif.fr/~jch/software/sfu.git

My approach requires no protocol changes, it just sends multiple tracks from the browser to the server; in other words, layers are identified by mid, with no need to introduce rids.  Since there are no protocol changes, the approach works across browsers.

This doesn't mean that rids should not be implemented in pion, especially if the browsers converge on the rid-based approach, but I think we should be aware of the tradeoffs involved and make sure that any rid work doesn't break the mid-based approach.

Sean-Der:
@jbrady42 I have the day off work tomorrow, mind if I take it from here? It looks like everything is here!

I just think the back and forth will take more time then me just fixing lint/small issues :) The commit owner will be you, the diff will just be different.

thanks

Sean-Der:
@jech 100% I don't want to regress anything that works today. Behavior should stay exactly the same as behavior (unless we see RID details in the SDP)

I still don't fully understand the value of RIDs (but I haven't looked that hard) but looks like everyone is diverging on them!

jech:
> I still don't fully understand the value of RIDs (but I haven't looked that
> hard)

Same here.  Since track labels are not preserved across associations, for
any non-trivial application you're going to carry a mapping between mids
and labels within your signalling protocol.  If you're going to do that,
then it looks to me like you might as well implement simulcast at the mid level.

>  but looks like everyone is [converging] on them!

Are they?  I was under the impression that they're only implemented in
a single browser implementation (I don't recall which).

PR #1333, opened 2020-07-16
Fix a typo in DESIGN.md

JonghunBok:
I believe this would read more naturally.

#### Description

The second sentence is not correct in grammar. 
I combined the first two sentences so that it reads more naturally.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,7 +1,7 @@
 <h1 align="center">
   Design
 </h1>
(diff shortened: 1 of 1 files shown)


Sean-Der:
Merged with 04d3f05a4a5f2c1751f1dc3a451bf6bff3df2ade

Thank you so much @JonghunBok and thanks for doing reviews @longsleep, they are starting to pile up on me :(

PR #1337, opened 2020-07-17
Datachannel: Don't set state open before datachannel valid

nerd2:
Previously it was possible for datachannel writers to see
"DataChannelStateOpen" without "dataChannel" non-nil and
crash

#### Reference issue
Fixes #...


Sean-Der:
Merged with 87e23ced1006f7ca814056b9f51e34c2ac68a773

Nice catch @nerd2 !

PR #1344, opened 2020-07-22
Update godoc link to pkg.go.dev

tsuu32:
#### Description
Use links and a badge from https://pkg.go.dev.
https://pkg.go.dev/badge/ supports creating a badge now.
#### Reference issue
~Fixes #...~


Sean-Der:
Thank you so much @tsuu32 

If you have the time it would be super helpful if you can make PRs for the other Pion repos! No rush though. If you don't have the time I can do it myself :)

thanks!

PR #1352, opened 2020-08-01
Upgrade pion/ice to v2.0.0-rc.9

jeremija:
This adds support for enabling ICE TCP by providing `ice.TCPMux` to `settingEngine`. The global state in `pion/ice` was removed and users can now specify their own `net.Listener` when creating a `ice.TCPMux`. A convenience function `NewICETCPMux` was added that returns a new instance of `ice.TCPMuxDefault`.

See https://github.com/pion/ice/pull/254 for more details.

### DIFF
--- a/examples/ice-tcp/index.html
+++ b/examples/ice-tcp/index.html
@@ -4,8 +4,7 @@
   </head>
 
   <body>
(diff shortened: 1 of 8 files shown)


jeremija:
Thanks both!

PR #1199, opened 2020-05-14
Answer with sendonly when remote is recvonly

longsleep:
When receiving a remote offer with a transceiver which is recvonly, the
answer must not set the transceiver to a receiving mode to avoid Firefox
from complaining and failing to set the remote description with the
following error:

```
Answer tried to set recv when offer did not set send
```

While Chrome is more forgiving and happily sets remote transceiver which
include receive even when the offer did not request ist, this needs to
be changed to get interop with Firefox.

The change is to create the implicitly created receivers introduced with
#1179 as inactive if the remote description does offer only to
receive. This will later on switch the transceiver to sendonly as soon
as a matching track is added.

Example of wrong behavior:

Pion receives the following offer
```
v=0
o=mozilla...THIS_IS_SDPARTA-77.0 954045388085230531 1 IN IP4 0.0.0.0
s=-
t=0 0
a=sendrecv
a=fingerprint:sha-256 B3:0E:89:A3:47:8C:A7:92:6E:3B:45:7A:C3:FF:58:F1:4A:36:2A:15:1E:A9:A0:A8:9C:3B:36:92:89:E4:F3:A9
a=group:BUNDLE 0 1
a=ice-options:trickle
a=msid-semantic:WMS *
m=application 36396 DTLS/SCTP 5000
c=IN IP4 172.17.0.1
a=candidate:0 1 UDP 2121269503 192.168.1.164 51738 typ host
a=candidate:3 1 UDP 2122055935 10.98.91.1 55680 typ host
a=candidate:6 1 UDP 2122121471 172.18.0.1 37378 typ host
a=candidate:9 1 UDP 2122187007 172.17.0.1 36396 typ host
a=candidate:12 1 UDP 2121793791 172.31.0.1 52583 typ host
a=candidate:15 1 UDP 2121728255 172.21.0.1 56611 typ host
a=candidate:18 1 UDP 2121662719 172.22.0.1 35638 typ host
a=candidate:21 1 UDP 2121597183 172.30.0.1 58337 typ host
a=candidate:24 1 UDP 2121924863 172.29.0.1 53127 typ host
a=candidate:27 1 UDP 2121859327 172.28.0.1 60922 typ host
a=candidate:30 1 UDP 2121531647 172.19.0.1 35106 typ host
a=candidate:33 1 UDP 2121466111 172.27.0.1 45025 typ host
a=candidate:36 1 UDP 2121335039 fd00:1:1:1:c191:bec9:270b:d79f 37412 typ host
[...]

Issue #717, opened 2019-06-19
Answer tried to set recv when offer did not set send


thetooth:
### Your environment.
  - Version: 2838b1a83649
  - Browser: Firefox 67.0.2

### What did you do?
When setting up a one way broadcast with the client creating the offer and server answering with a video and audio media channel, the SDP exchange fails with "InvalidSessionDescriptionError: Answer tried to set recv when offer did not set send".

I went and read the RFC to double check a change I noticed in #657. The RFC says "If a media stream is listed as recvonly in the offer, the answer MUST be marked as sendonly or inactive in the answer."

I think I understand the reason it was changed to sendrecv, as we can now share tracks between peer connections, this change would make that easy if this attribute is stored in the track itself.

Perhaps we could compare the offer when creating the answer and erroneously report it as sendonly for compatibility, that way we retain the easy track sharing API.

wezzle:
Unfortunately I ran into the same problem. I tried adding a receiver manually with sendonly but this is not supported and triggers the error `AddTransceiver currently only suports recvonly and sendrecv`.

> Side note: there is a typo in supports

When changing the [PeerConnection's AddTransceiver](https://github.com/pion/webrtc/blob/master/peerconnection.go#L1383) method to include the following case I got it to work:
```go
case RTPTransceiverDirectionSendonly:
	payloadType := DefaultPayloadTypeOpus
	if trackOrKind == RTPCodecTypeVideo {
		payloadType = DefaultPayloadTypeVP8
	}

	track, err := pc.NewTrack(uint8(payloadType), mathRand.Uint32(), util.RandSeq(trackDefaultIDLength), util.RandSeq(trackDefaultLabelLength))
	if err != nil {
		return nil, err
	}

	sender, err := pc.api.NewRTPSender(track, pc.dtlsTransport)
	if err != nil {
		return nil, err
	}

	return pc.newRTPTransceiver(
		nil,
		sender,
		RTPTransceiverDirectionSendonly,
		trackOrKind,
	), nil
```

You can then call:

```go
peerConnection.AddTransceiver(webrtc.RTPCodecTypeVideo, webrtc.RtpTransceiverInit{
	Direction: webrtc.RTPTransceiverDirectionSendonly,
})
```

Before adding a track and creating the answer. The answer will then have `a=sendonly` as expected by Firefox and the RFC.

As I'm unfamiliar with the codebase I have no idea if this is the way to go but I'm happy to create a pull request out of it if one of the contributors can confirm.

[...]

PR #1359, opened 2020-08-10
Implement OnNegotiationNeeded

### LINKED ISSUE #817: Implement OnNegotiationNeeded

Sean-Der:
This event is fired when a change has occurred
which requires session negotiation.

Resolves #817

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -46,8 +46,9 @@ type PeerConnection struct {
 
 	idpLoginURL *string
 
-	isClosed          *atomicBool
-	negotiationNeeded bool
+	isClosed               *atomicBool
+	negotiationNeeded      bool
+	negotiationNeededState negotiationNeededState
 
 	lastOffer  string
 	lastAnswer string
@@ -65,6 +66,7 @@ type PeerConnection struct {
 	onConnectionStateChangeHandler    func(PeerConnectionState)
 	onTrackHandler                    func(*Track, *RTPReceiver)
 	onDataChannelHandler              func(*DataChannel)
+	onNegotiationNeededHandler        func()
 
 	iceGatherer   *ICEGatherer
 	iceTransport  *ICETransport
@@ -92,7 +94,6 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
 	// allow better readability to understand what is happening.
 	pc := &PeerConnection{
 		statsID: fmt.Sprintf("PeerConnection-%d", time.Now().UnixNano()),
-		ops:     newOperations(),
 		configuration: Configuration{
 			ICEServers:           []ICEServer{},
 			ICETransportPolicy:   ICETransportPolicyAll,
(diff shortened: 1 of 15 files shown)


jech:
The change to operations.start is not correct.  The original code was carefully written to continue the operations queue if an operation panics; the new code will stop the operations queue if an operation panics.  I'm afraid that stopping the queue will lead to difficult to debug deadlocks.

The `ops.done` mechanism feels very ad-hoc to me, and I'm not sure whether it's needed.  Instead of special-casing `done`:
```
ops.done = f
```
couldn't we simply enqueue an operation that recursively reenqueues itself as long as the queue is not empty?
```
g := func() {
    if !ops.isEmpty() {
        ops.Enqueue(g)
        return
    }
    f()
}
ops.Enqueue(g)
```

Finally, I'm not sure I understand how the code ensures that `OnNegotiationNeeded` is eventually triggered.  If a client keeps enqueuing new ops, then the queue will never drain.  Are there guarantees that this kind of livelock never happens?

tekig:
left to do in pull #1322

1. Maybe, the renegotiation does not understand when the stream is closed. Will always call him.
pion, does not close the connection, but makes it inactive? need to check

2. I'm sure [this code](https://github.com/pion/webrtc/pull/1322#issuecomment-663878012) can be written better

3. Chrome closes the transceivers as the spec says, but pion makes it inactive. Or you need to remove the [check for closed](https://github.com/pion/webrtc/pull/1322/commits/a06e8a27e06e3bfb24716636a3f5d868935ede7e#diff-d2ee665bb8c62c914fee5e9c743be25cR332-R334) transceivers in peon or implement transceivers closure, as in the specification. 

4. More tests)

It might have been faster to write code than to write this comment

Sean-Der:
@tekig i added you to the Pion GitHub organization so you can push to the branch directly! Sorry about cherry-picking I though I would have more time

Sean-Der:
https://github.com/pion/webrtc/commit/4aa50a36be648181f5512e0b551141a0e166b30f I did take this and squashed.

Sean-Der:
@tekig please take w/e you can and I can help tomorrow. This is my #1 priority for the next few days!

tekig:
With an example, it will be easier to test with a browser. Didn't use websocket to not include dependencies.
I hope for you, I will help if possible

tekig:
#1361

tekig:
Done, test with an example. Especially in the process of deleting a track, it differs from the specification and some steps are skipped due to the peculiarities of the pion code
Sometimes an error appears in the browser, but it is not related to OnNegotiationNeeded
Chrome
`DOMException: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': Failed to set remote answer sdp: Failed to apply the description for 0: Failed to set SSL role for the transport.`
Firefox
`DOMException: Answer tried to set recv when offer did not set sen`

Sean-Der:
> `DOMException: Answer tried to set recv when offer did not set send`

We should fix this now. It is really easy actually! https://github.com/pion/webrtc/issues/717 and https://github.com/pion/webrtc/pull/1199 

I will take a look at that tomorrow.

> `Failed to set SSL role for the transport`

My guess is we are doing the wrong value for `setup:` during renegotation. I don't know the answer around this. It will just be a signaling/text change though!

peerconnection.go
```diff
@@ -242,6 +245,154 @@ func (pc *PeerConnection) OnDataChannel(f func(*DataChannel)) {
...
+
+	if localDesc == nil {
+		return true
+	}
+
+	if pc.sctpTransport == nil {
+		return true
+	}
+
+	if len(pc.sctpTransport.dataChannels) != 0 {
```
Sean-Der, line 335: Is this just asserting that you have requested a DataChannel, but the Transport hasn't been started?

Can you use the `SCTPTransport.State()` instead?

Sean-Der: Never mind I see this is clearly defined in the RFC.

Can you make this a helper function in `sdp.go`? Then we can re-use it. We scan a SDP for `Does it have a datachannel a few places`

Sean-Der:
@tekig this looks really great to me (for merging into master) Last few things.

* Can you fix the failures in CI? 
    - `go test -race`
    - `golangci-lint`
* Can you do a rebase and squash everything into one commit?

@jech @jeremija you see anything that really stands out? I think after we land I think we can simplify the examples a bit. They are really great, just a bit verbose. Also maybe revisit `negotiationNeededState`.

Really great work on this again. This is a huge undertaking for first contribution.

tekig:
> Can you fix the failures in CI?
> * `go test -race`
> * `golangci-lint`

`golangci-lint`  [reports](https://travis-ci.org/github/pion/webrtc/jobs/717978296#L356-L364) the error in only one place. I decided to make the method more general, so that it will be used more often, if necessary.

PR #1370, opened 2020-08-14
Adding Stream type

tarrencev:
Implementing a `Stream` type and updating `onTrack` to return `Stream`s similar to the browser apis.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -60,11 +60,12 @@ type PeerConnection struct {
 	greaterMid int
 
 	rtpTransceivers []*RTPTransceiver
+	streams         map[string]*Stream
 
 	onSignalingStateChangeHandler     func(SignalingState)
 	onICEConnectionStateChangeHandler func(ICEConnectionState)
 	onConnectionStateChangeHandler    func(PeerConnectionState)
-	onTrackHandler                    func(*Track, *RTPReceiver)
+	onTrackHandler                    func(*Track, *RTPReceiver, []*Stream)
 	onDataChannelHandler              func(*DataChannel)
 	onNegotiationNeededHandler        func()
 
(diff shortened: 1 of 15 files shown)


peerconnection.go
```diff
@@ -926,7 +938,7 @@ func (pc *PeerConnection) startReceiver(incoming trackDetails, receiver *RTPRece
 		receiver.Track().mu.Unlock()
 
 		if pc.onTrackHandler != nil {
-			pc.onTrack(receiver.Track(), receiver)
+			pc.onTrack(receiver.Track(), []*Stream{stream})
```
tarrencev, line 941: This should return all the `Stream`s the track belongs to. But in pion's case, a `Track` can only have a single `label`. Does that mean it can only ever belong to a single `Stream`?

peerconnection.go
```diff
@@ -1768,13 +1790,17 @@ func (pc *PeerConnection) startRTP(isRenegotiation bool, remoteDesc *SessionDesc
...
+			// might get a track later when `Receive()` is called on the receiver.
 			if t.Receiver() == nil || t.Receiver().Track() == nil {
 				continue
 			}
 
 			t.Receiver().Track().mu.Lock()
 			ssrc := t.Receiver().Track().ssrc
 
+			// If receiver already exists for track ssrc, set id/label.
+			// Why do we do this? Is there a case where the track id/label changes?
```
tarrencev, line 1803: In this case, if `t.Receiver().Track().label != details.label`, should we remove the track from `stream[t.Receiver().Track().label]` and add it to the stream with new label?

tarrencev: I'm not sure 100%, but it seems a SSRC can be reused with a different id/label. Is that true?

PR #1371, opened 2020-08-16
Write to all of a track's active senders even if some fail.

s4y:
#### Description

Currently, if one of a track's senders is in a bad state, `WriteRTP()` stops at that sender and returns an error. This change continues writing to the other senders and returns the first error, if any.

#### Reference issue
Fixes #...


Sean-Der:
Fixed by 6cca2786c9f46f5c22d1baf3433b1102fbaa24b7

PR #1377, opened 2020-08-19
Always return sdpMid & sdpMLineIndex if value is null

DevRockstarZ:
#### Description
sdpMid & sdpMLineIndex ToJSON description : (https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-tojson)
I thought ICECandidateInit should return sdpMid & sdpMLineIndex if value is null.
I'm doing some stuffs using react-native-webrtc, I suffered issue in this [line](https://github.com/react-native-webrtc/react-native-webrtc/blob/9e627e07102f4b83c72d0ca1dcaa9f549d84c6c7/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L873).
So I thought sdpMind & sdpMLineIndex should not be null, therefore I request PR.
Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Sam Lancia](https://github.com/nerd2)
 * [Henry](https://github.com/cryptix)
 * [Jeff Tchang](https://github.com/tachang)
-* [JooYoung](https://github.com/DevRockstarZ)
+* [JooYoung Lim](https://github.com/DevRockstarZ)
 * [Sidney San Martín](https://github.com/s4y)
 * [soolaugust](https://github.com/soolaugust)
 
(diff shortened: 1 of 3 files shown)


DevRockstarZ:
In my candidate
`candidate: "candidate:3460887983 1 udp 2122260223 192.168.1.64 50367 typ host generation 0 ufrag 8kI+ network-id 1 network-cost 10"
`
It dosen't have mid.
So it's sdpMid, sdpMLineIndex value should be zero or null?
Integer value should not be zero... I'm so confused by this...
[sdpMLineIndex](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-sdpmlineindex) says it can be nullable.
It means zero? or null?

Sean-Der:
Thanks @DevRockstarZ 

This looks right to me. The w3c docs seem to imply the key should always exist.

DevRockstarZ:
@Sean-Der I also changed usernamefragment it also should be default to null.
I changed readme.md also because I added my last name! 
Thanks for your comment!

Sean-Der:
Merged 6d3633e58923227e00e96ed80ae46e6fa25a39cc !

Thanks for all these great commits and using Pion :) love to see this stuff fixed

tarrencev:
I'm running into an error with this change:

```
Failed to execute 'addIceCandidate' on 'RTCPeerConnection': Candidate missing values for both sdpMid and sdpMLineIndex
```

MDN has this note:
```
Note: Attempting to add a candidate (using addIceCandidate()) that has a value of null for either sdpMid or sdpMLineIndex will throw a TypeError exception.
```
https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateInit/sdpMLineIndex

So im not sure the key should exist with a null value

PR #1379, opened 2020-08-19
Failing OnNegotiationNeeded test

tarrencev:
Seems `OnNegotiationNeeded` doesn't perform as expected when multiple tracks are added

@tekig what do you think?


tarrencev:
actually i think it is working correctly in this case 😁

PR #1381, opened 2020-08-20
GetReceiversFromKind method

r-novel:
I suggest to add a getter for RTPReceivers depended on codec type; It might be useful for Simulcast cases.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1399,32 +1399,14 @@ func (pc *PeerConnection) GetSenders() []*RTPSender {
 	return result
 }
 
-// GetReceiversFromKind returns the RTPReceivers depended on codec type that are currently attached to this PeerConnection
-func (pc *PeerConnection) GetReceiversFromKind(kind RTPCodecType) (result []*RTPReceiver) {
-	pc.mu.Lock()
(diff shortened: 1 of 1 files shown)


Sean-Der:
Hey @r-novel thank you so much for the PR!

I am torn on adding this. I disagree because we are making a non-standard extension to the WebRTC API. Since the mediaengine is already non-standard I think it is ok there.

I do like the change to GetReceivers though! Maybe call it `receivers` because it will be in the GoDoc.

If you feel my argument for not doing this is bad I am happy to be overridden!

r-novel:
Hello, @Sean-Der!
Thank you so much for the response. 
You were right about the non-standard extension. I agree with you GetReceiversFromKind should be left to the discretion of Pion users and get it off the PeerConnection API.

>Maybe call it receivers because it will be in the GoDoc.
Do you mean to rename output slice `result` to `receivers`?

I will be appreciated for your response!

Best regards, Roman.

Sean-Der:
Merged with c0b9c12b1112b138106ce87f3c49d6382a065ccc

Really fantastic PR @r-novel I love iterative improvements like this. Makes the code base better to read for everyone!

PR #1394, opened 2020-09-01
Revert "Don't use omitempty for ICECandidateInit"

tarrencev:
This reverts commit 6d3633e58923227e00e96ed80ae46e6fa25a39cc from pr https://github.com/pion/webrtc/pull/1377

I'm running into an error with this change:

```
Failed to execute 'addIceCandidate' on 'RTCPeerConnection': Candidate missing values for both sdpMid and sdpMLineIndex
```

MDN has this note:
```
Note: Attempting to add a candidate (using addIceCandidate()) that has a value of null for either sdpMid or sdpMLineIndex will throw a TypeError exception.
```
https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateInit/sdpMLineIndex

So im not sure the key should exist with a null value

@DevRockstarZ

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Sam Lancia](https://github.com/nerd2)
 * [Henry](https://github.com/cryptix)
 * [Jeff Tchang](https://github.com/tachang)
-* [JooYoung Lim](https://github.com/DevRockstarZ)
+* [JooYoung](https://github.com/DevRockstarZ)
 * [Sidney San Martín](https://github.com/s4y)
 * [soolaugust](https://github.com/soolaugust)
 * [Kuzmin Vladimir](https://github.com/tekig)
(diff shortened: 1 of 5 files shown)


Sean-Der:
Screen Shot 2020-09-01 at 08 02 31

The W3C makes it sound like NULL is a valid value. Reverting this will break other software, so I am not sure what to do :(

If you do '' or 0 does it still work?

tarrencev:
I feel releasing this will cause more issues if chrome/firefox are throwing with a null value set. Even though the spec does seem to suggest null is valid. I'm not sure we can set `0` since it is an index.

DevRockstarZ:
@tarrencev If null value cause type error, I think it should be zero or empty string.
It can also cause errors?
Thanks for reviewing my code!

Sean-Der:
@DevRockstarZ no worries this is all fixed! I also filed an issues against [webrtc-pc](https://github.com/w3c/webrtc-pc/issues/2571)

DevRockstarZ:
@Sean-Der Thanks! Have a nice day :D

PR #1397, opened 2020-09-04
Refactor ICE Candidate logic into pion/ice

korymiller1489:
Refactor ICE Candidate logic into `pion/ice`
Reduce ICE logic dependency on `pion/sdp`


korymiller1489:
Need to update with new version of pion/ice after https://github.com/pion/ice/pull/270 is pulled in

Sean-Der:
Sorry I mislead you :(

Can we keep the public API `ICECanidateInit` and `ICECandidate`. They will mirror the structures in `pion/ice`

We are going to have some duplication, but I think that is important to not break users. I also want to not make users dig across multiple packages as much as possible.

You are SUPER close to being done though, thank you so much for all the hard work you are doing.

korymiller1489:
If we are maintaining the public API, does anything actually have to change in `pion/webrtc` then?

Sean-Der:
Implemented with 43e9cd84d49e5b56d5c7f75907476aa4fa466a19

Thanks for taking this on @korymiller1489 you should be able to use `pion/ice` directly now and get all the candidate logic you need!

PR #1403, opened 2020-09-08
ci(workflow): test action

tarrencev:
working on moving ci to github actions. integration is nicer and tests run faster (no queueing)

@Sean-Der @at-wat let me know what you think. i will migrate others if you are fine with it


Sean-Der:
:fire: I am all for this!

* We can do this one piece at a time. I am up for pushing this to https://github.com/pion/.goassets and migrating more stuff later. Don't think it needs to be perfect for us to start moving :)
* Do you know if `TEST_EXTRA_ARGS` is still getting picked up?

It would be worth looking at other Go projects in github actions, and see if they need the following. I feel like we are doing something wrong to need all these extra steps.

```
      - name: set test packages
        run: echo ::set-env name=TEST_PACKAGES::$(go list ./... | grep -v examples)

      - name: set cover packages
        run: echo ::set-env name=COVER_PACKAGES::$(echo "${TEST_PACKAGES}" | paste -s -d ',')
```

tarrencev:
It doesnt look like `TEST_EXTRA_ARGS` is being used. afaict the reason for `TEST_PACKAGES` is to filter `examples` from the test packages

PR #1410, opened 2020-09-10
Fix error checking

q191201771:
Wrong error variable was checked and redundant parentheses in
play-from-disk example.


tarrencev:
thanks!

PR #1412, opened 2020-09-11
fix(negotiation): stress test + race fixes

tarrencev:
i've uncovered quite a few race conditions with this test. a few more still exist.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -310,7 +310,9 @@ func (pc *PeerConnection) negotiationNeededOp() {
 	}
 
 	// Step 2.6
+	pc.mu.Lock()
 	pc.negotiationNeeded = true
+	pc.mu.Unlock()
 
 	// Step 2.7
 	if pc.onNegotiationNeededHandler != nil {
@@ -322,8 +324,10 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool {
 	// To check if negotiation is needed for connection, perform the following checks:
 	// Skip 1, 2 steps
 	// Step 3
+	pc.mu.RLock()
 	localDesc := pc.currentLocalDescription
 	remoteDesc := pc.currentRemoteDescription
+	pc.mu.RUnlock()
 
 	if localDesc == nil {
 		return true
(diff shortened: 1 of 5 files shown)


peerconnection.go
```diff
@@ -356,7 +360,7 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool {
 			// Step 5.3.1
 			if t.Direction() == RTPTransceiverDirectionSendrecv || t.Direction() == RTPTransceiverDirectionSendonly {
 				descMsid, okMsid := m.Attribute(sdp.AttrKeyMsid)
-				if !okMsid || descMsid != t.Sender().Track().Msid() {
+				if !okMsid || t.Sender() == nil || descMsid != t.Sender().Track().Msid() {
```
tarrencev, line 363: this is not enough. `t.Sender()` can become nil between the nil check and accessing `Track().Msid()`

peerconnection.go
```diff
@@ -1160,16 +1167,23 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks []trackDetails, curre
 func (pc *PeerConnection) startRTPSenders(currentTransceivers []*RTPTransceiver) {
 	for _, transceiver := range currentTransceivers {
 		// TODO(sgotti) when in future we'll avoid replacing a transceiver sender just check the transceiver negotiation status
-		if transceiver.Sender() != nil && transceiver.Sender().isNegotiated() && !transceiver.Sender().hasSent() {
```
tarrencev, line 1163: again, `Sender()` can be mutated during execution here. the fix i made is messy. maybe there is a better way?

peerconnection_renegotiation_test.go
```diff
@@ -899,3 +899,87 @@ func TestNegotiationNeededRemoveTrack(t *testing.T) {
...
+	pcB.OnNegotiationNeeded(func() {
+		mu.Lock()
+		assert.NoError(t, signalPair(pcB, pcA))
+		mu.Unlock()
+	})
+
+	done := make(chan bool)
+	go func() {
+		for {
+			time.Sleep(time.Duration(rand.Intn(100))*time.Millisecond + 10)
```
tarrencev, line 933: you can expose more errors by reducing the `10` and `50` lower bounds here and below

rtptransceiver.go
```diff
@@ -92,6 +101,10 @@ func (t *RTPTransceiver) setReceiver(r *RTPReceiver) {
 }
 
 func (t *RTPTransceiver) setDirection(d RTPTransceiverDirection) {
+	// https://www.w3.org/TR/webrtc/#ref-for-dfn-stopping-4
+	// if stopping {
```
tarrencev, line 105: should return an error here according to spec but requires quite a lot of changes to the code base. not sure if we should update?

sdp.go
```diff
@@ -324,13 +324,19 @@ func addTransceiverSDP(d *sdp.SessionDescription, isPlanB bool, dtlsFingerprints
 	}
 
 	for _, mt := range transceivers {
-		if mt.Sender() != nil && mt.Sender().Track() != nil {
```
tarrencev, line 327: another case where `Sender()` can be mutated during execution

PR #1413, opened 2020-09-11
Fix: negotiation state race conditions

tarrencev:
Adds a few locks to solve a few race conditions with renegotiations. This are the simplest fixes so pulled them out. Hopefully it will make them easier to land


Sean-Der (approved):

Sean-Der:
It's a shame we have to do this :(

We for sure should rethink all this and how we can make it better.

Issue #1401, opened 2020-09-08
Fix Flakey test TestNegotiationNeededRemoveTrack


Sean-Der:
```
=== RUN   TestNegotiationNeededRemoveTrack
    TestNegotiationNeededRemoveTrack: peerconnection_renegotiation_test.go:866:
                Error Trace:    peerconnection_renegotiation_test.go:866
                                                        peerconnection.go:317
                                                        operations.go:84
                                                        asm_amd64.s:1373
                Error:          Received unexpected error:
                                InvalidStateError: connection closed
                Test:           TestNegotiationNeededRemoveTrack
    TestNegotiationNeededRemoveTrack: peerconnection_renegotiation_test.go:869:
                Error Trace:    peerconnection_renegotiation_test.go:869
                                                        peerconnection.go:317
                                                        operations.go:84
                                                        asm_amd64.s:1373
                Error:          Received unexpected error:
                                InvalidStateError: connection closed
                Test:           TestNegotiationNeededRemoveTrack
    TestNegotiationNeededRemoveTrack: peerconnection_renegotiation_test.go:872:
                Error Trace:    peerconnection_renegotiation_test.go:872
                                                        peerconnection.go:317
                                                        operations.go:84
                                                        asm_amd64.s:1373
                Error:          Received unexpected error:
                                InvalidStateError: connection closed
                Test:           TestNegotiationNeededRemoveTrack
    TestNegotiationNeededRemoveTrack: peerconnection_renegotiation_test.go:875:
                Error Trace:    peerconnection_renegotiation_test.go:875
                                                        peerconnection.go:317
                                                        operations.go:84
                                                        asm_amd64.s:1373
                Error:          Received unexpected error:
                                InvalidStateError: connection closed
                Test:           TestNegotiationNeededRemoveTrack
    TestNegotiationNeededRemoveTrack: peerconnection_renegotiation_test.go:878:
                Error Trace:    peerconnection_renegotiation_test.go:878
                                                        peerconnection.go:317
                                                        operations.go:84
                                                        asm_amd64.s:1373
                Error:          Received unexpected error:
                                InvalidStateError: connection closed
                Test:           TestNegotiationNeededRemoveTrack
    TestNegotiationNeededRemoveTrack: peerconnection_renegotiation_test.go:881:
                Error Trace:    peerconnection_renegotiation_test.go:881
                                                        peerconnection.go:317
                                                        operations.go:84
                                                        asm_amd64.s:1373
[...]

PR #1414, opened 2020-09-11
Fix: CreateOffer race condition

tarrencev:
Fixes a race condition that occurs when the local media (transceivers) are mutated during the offer generation process.

Resolves an issue where an offer is created with no `mid` causing the error below:
image

The test is not great, since it depends on the performance of the executor. I was able to consistently reproduce locally with it though.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -348,7 +348,7 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool {
 	for _, t := range pc.GetTransceivers() {
 		// https://www.w3.org/TR/webrtc/#dfn-update-the-negotiation-needed-flag
 		// Step 5.1
-		// if t.stoping && !t.stopped {
+		// if t.stopping && !t.stopped {
 		// 	return true
 		// }
 		m := getByMid(t.Mid(), localDesc)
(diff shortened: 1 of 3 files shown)


peerconnection_renegotiation_test.go
```diff
@@ -899,3 +899,39 @@ func TestNegotiationNeededRemoveTrack(t *testing.T) {
...
+
+		_, err = pcA.AddTrack(track)
+		assert.NoError(t, err)
+
+		err = track.WriteSample(media.Sample{Data: []byte{0x00}, Samples: 1})
+		assert.NoError(t, err)
+	}
+
+	// Wait for async to complete otherwise dangling renegotation
+	// will occur on closed connection and throw InvalidStateError: connection closed
```
tarrencev, line 935: This is #1401. I will fix it in a separate PR.

peerconnection.go
```diff
@@ -557,67 +574,89 @@ func (pc *PeerConnection) CreateOffer(options *OfferOptions) (SessionDescription
...
+	// inspection of video system resources.
+	for {
+		// We cache current transceivers to ensure they aren't
+		// mutated during offer generation. We later check if they have
+		// been mutated and recompute the offer if necessary.
+		currentTransceivers := pc.GetTransceivers()
+
+		// in-parallel steps to create an offer
+		// https://w3c.github.io/webrtc-pc/#dfn-in-parallel-steps-to-create-an-offer
+		isPlanB := pc.configuration.SDPSemantics == SDPSemanticsPlanB
```
tarrencev, line 595: the only modification between here and line `653` is to use `currentTransceivers` rather than `pc.GetTransceivers()`

peerconnection.go
```diff
@@ -348,7 +348,7 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool {
 	for _, t := range pc.GetTransceivers() {
 		// https://www.w3.org/TR/webrtc/#dfn-update-the-negotiation-needed-flag
 		// Step 5.1
-		// if t.stoping && !t.stopped {
```
Sean-Der, line 351: Should these be commented out? They were in the original PR, maybe we should just delete?

Sean-Der (approved):

peerconnection.go
tarrencev: according to the spec we should have a `stopping` property on a transceiver but we dont right now. so this was added prematurely. i'll leave for now and we can remove it separately

PR #1417, opened 2020-09-12
Move candidate parsing to pion/ice

Sean-Der:
Instead of having logic in pion/webrtc and pion/sdp move all logic to
pion/ice. Users of pion/ice don't want to import these packages to use
basic pion/ice functionality.


Sean-Der:
@korymiller1489 `pion/ice` is ready for you to use!

Issue #1211, opened 2020-05-22
inconsistent errors when sending on closed track vs closed channel


josharian:
It's possible I'm misunderstanding something, but...

When you send media to a track with no peers, you get `io.ErrClosedPipe` back. Many of the examples use this to suppress unimportant error messages.

When you send data to a data channel that is closed, you get `data channel not open`.

It would be nice if data channels also returned `io.ErrClosedPipe` in this situation, for consistency.

I guess this is a breaking API change, so maybe it should be considered for v3?

Sean-Der:
:+1: 100% agree with this.

Yea probably `V3` is probably best, great catch :)

PR #1424, opened 2020-09-14
Fix issue 1211

obasajujoshua31:
return io.ErrClosedPipe instead of default error coming from ensureOpened

wrote test for attempting to send when close

#### Reference issue
Fixes #...https://github.com/pion/webrtc/issues/1211


tarrencev:
@obasajujoshua31 you can just delete `ErrDataChannelNotOpen` in `errors.go` and update the return value in `datachannel.go` line 357 to `err. ErrClosedPipe`

Issue #1212, opened 2020-05-22
pc.AddIceCandidate complains when candidate.Candidate is the empty string


jech:
[Section 4.8.2 of the spec](https://www.w3.org/TR/webrtc/#rtcpeerconnectioniceevent) has the following to say:

> An RTCIceTransport has finished gathering a generation of candidates, and is providing an end-of-candidates indication as defined by Section 8.2 of [TRICKLE-ICE]. This is indicated by candidate.candidate being set to an empty string. The candidate object should be signaled to the remote peer and passed into addIceCandidate like a typical ICE candidate, in order to provide the end-of-candidates indication to the remote peer.

However, if I pass such a candidate to `pc.AddIceCandidate`, it complains with the error `attribute not long enough to be ICE candidate (0)`.

marverix:
Is there any workaround for now?

Sean-Der:
Not right now, but this would only take ~hour to fix if you are interested!

At this line https://github.com/pion/webrtc/blob/master/peerconnection.go#L1364

```
if candidateValue == "" {
  return nil
}
``

Then you just need to add a test that asserts it doesn't return an error!

marverix:
Thanks for your quick answer. So just from curiosity: So how the library "knows" that there is end-of-candidates?

Sean-Der:
Pion doesn't care about `end-of-candidates` It takes the first connected candidate pair it can get. It waits before taking more expensive candidate pairs though (like TURN and STUN).

You can configure the amount of time you are willing to wait with the [SettingEngine](https://pkg.go.dev/github.com/pion/webrtc/v3?tab=doc#SettingEngine.SetPrflxAcceptanceMinWait)!

Sean-Der:
https://github.com/pion/ice/blob/master/agent_config.go#L23-L33 these are the defaults

marverix:
Thanks @Sean-Der !

Sean-Der:
@jech This happens [here](https://github.com/pion/ice/blob/master/agent.go#L487-L497)

Right now it just does disconnected+failed time.  After 30 seconds of sending connectivity checks it goes to failed. This is a really easy change we could make!

In general `pion/ice` has been really naieve. I wish I could put more time into it, but I feel like we are always missing big parts :(

jech:
I see, thanks.  I've summarised this in https://github.com/pion/ice/issues/271.

PR #1428, opened 2020-09-17
Fix issue #1212 - Candidate with empty string

DevRockstarZ:
#### Description
I made the first test code by myself, it can be not perfect 😢
I tested on my local terminal, it passes all the cases.
I created test `PeerConnection` to use `AddICECandidate`, and I tested 3 test cases.
Thanks for reviewing!

#### Reference issue
Fixes [issue #1212](https://github.com/pion/webrtc/issues/1212)


Sean-Der:
Hey @DevRockstarZ this PR looks really good to me!

If you want to reopen I would love to merge this :) I will just squash your work into one commit. Really great work!

DevRockstarZ:
@Sean-Der Thanks for your reply! I closed this PR due to you're working on this in Pion-V3.
I thought this PR will be bother you 😢
If re-open is fine, I will re-open it! Thanks!

jsmouret:
Isn't it hiding the real issue?
https://github.com/pion/ice/issues/271

DevRockstarZ:
@jsmouret 
I also think real issue is not cleared. 
So I did temporary fix to do not print errors from empty candidate.
I think real issue will be cleared when pion releases version 3.
Thanks!

Sean-Der:
You are right @jsmouret thanks for keeping me honest :) 

So I think this is a two step process

* Update `pion/ice` to accept a `nil` candidate. [AddRemoteCandidate](https://github.com/pion/ice/blob/master/agent.go#L704) when called with `nil` should then set a field on the agent.

in the [contact](https://github.com/pion/ice/blob/master/agent.go#L494) loop it shouldn't go to failed unless this field has been set.

[restart](https://github.com/pion/ice/blob/master/agent.go#L1187) should update this field as well.

* then `pion/webrtc` should be updated to pass a `nil` candidate in.

This is a good amount of work, but I think there is a clear path forward! If anyone is interested in grabbing I can help :) the work done on the tests here is really great, I think we can use as is :)

scorpionknifes:
Hey, @Sean-Der I'm reading into this and trying to clarify some stuff. (1st step ICE)

I'm reading into [5.6.3 RTCIceGathererState](https://www.w3.org/TR/webrtc/#rtcicegathererstate) and

complete | The RTCIceTransport has completed gathering and the end-of-candidates indication for this transport has been sent. It will not gather candidates again until an ICE restart causes it to restart.
--: | :--

I'm assuming the field on the agent that should be set is GatheringState by [agent.setGatheringState](https://github.com/pion/ice/blob/302290a7846159720a0feaa9d0d696f62ee7b50f/agent.go#L1214)

Since GatheringState is already reset in [agent.Restart](https://github.com/pion/ice/blob/302290a7846159720a0feaa9d0d696f62ee7b50f/agent.go#L1194) i think we don't need to update this field.

I'm also not too sure where to check this in contact loop.

Sean-Der:
`ICEGatherState` is good, I don't believe we need to update that!

What we need to do is 

* Update the ICE Agent to not go to failed until it sees a `end-of-candidates` that check would go [here]( https://github.com/pion/ice/blob/master/agent.go#L498) 
* Update `pion/webrtc` to accept `nil` candidates, and send a `end-of-candidates` notification into `pion/ice`

scorpionknifes:
@Sean-Der, need some advice

how do I store `end-of-candidate` attribute in ICE Agent?

since we sending nil to AddICECandidate do we want to pass nil to ICE or should we create a new func?

PR #1434, opened 2020-09-19
Fix issue #1425 - track event doesn't emit

### LINKED ISSUE #1425: daedlock in PeerConnection.startReceiver
### Your environment.
  - Version: *v2.2.21*

### What did you do?
I'm developing a SFU based on pion and have a test with 6 peers. That test is very flaky.

### What did you expect?
I expect that test should be passed.

### What happened?
I tried to figure out what happened, and it turns out that some *OnTrack* event is not fired. Long story short, there is a scenario that might cause deadlock in webrtc.PeerConnection.

- PeerConnection.startReceiver <- hold a reader lock: pc.mu
- PeerConnection.onTrack <- hold the same reader lock 2nd time

If another routine calls PeerConnection.GetTransceivers, it will try to hold a writer lock, then the routine in `startReceiver` would hang there. Refer to here to [here](https://stackoverflow.com/questions/30547916/goroutine-blocks-when-calling-rwmutex-rlock-twice-after-an-rwmutex-unlock) for the story behind hold reader-lock twice in a routine.

I wonder if:
- we can replace the writer-lock in PeerConnection.GetTransceivers with reader-lock.
- acquire a reader-lock twice should be avoided, thus we have to refine the code in PeerConnection.startReceiver

mission-liao:
#### Description
Sometimes track events may not emit.

After tracing this issue, it a deadlock caused by two consecutive reader lock acquisition then a writer lock acquisition in another routine. (see [here](https://stackoverflow.com/a/30549188))

- to fix it, remove the unnecessary reader lock acquisition
- add a test to reproduce it with high probability

#### Reference issue
Fixes #1425

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Ivan Egorov](https://github.com/vany-egorov)
 * [krishna chiatanya](https://github.com/kittuov)
 * [JacobZwang](https://github.com/JacobZwang)
+* [Mission Liao](https://github.com/mission-liao)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -277,8 +277,12 @@ func (pc *PeerConnection) onTrack(t *Track, r *RTPReceiver) {
 	pc.mu.RUnlock()
 
 	pc.log.Debugf("got new track: %+v", t)
-	if hdlr != nil && t != nil {
-		go hdlr(t, r)
+	if t != nil {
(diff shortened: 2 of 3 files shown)


tarrencev:
@mission-liao would you mind creating this against the master branch and we can backport to v2?

Sean-Der:
Thanks you so much for the contribution @mission-liao!

This is cherry-picked as 

* f98e08446a823b2b0cb537763f19b17845e9d1f2 in master
* 0d1700cbcb2126b9894adad32f26abbdeca088a1 in v2

I also tagged `v2.2.26` thanks for contributing. I really hope you stay involved, you are doing some great stuff in such a short amount of time :)

PR #1437, opened 2020-09-21
Implements GetStats: RTCCodecStats 

obasajujoshua31:
implemented getstats api for rpccodec

#### Reference issue
https://github.com/pion/webrtc/issues/610
Fixes #...

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pion/rtp"
 	"github.com/pion/rtp/codecs"
@@ -340,6 +341,7 @@ type RTPCodec struct {
 	Name        string
 	PayloadType uint8
 	Payloader   rtp.Payloader
+	statsID     string
 }
 
 // NewRTPCodec is used to define a new codec
@@ -363,6 +365,7 @@ func NewRTPCodec(
 		Payloader:   payloader,
 		Type:        codecType,
 		Name:        name,
+		statsID:     fmt.Sprintf("RTPCodec-%d", time.Now().UnixNano()),
 	}
 }
 
@@ -410,4 +413,23 @@ type RTPHeaderExtensionCapability struct {
 type RTPCapabilities struct {
 	Codecs           []RTPCodecCapability
 	HeaderExtensions []RTPHeaderExtensionCapability
+}
+
+func (m *MediaEngine) collectStats(collector *statsReportCollector) {
+
+	for _, codec := range m.codecs {
+		collector.Collecting()
+		stats := CodecStats{
+			Timestamp:   statsTimestampFrom(time.Now()),
+			Type:        StatsTypeCodec,
+			ID:          codec.statsID,
+			PayloadType: codec.PayloadType,
+			MimeType:    codec.MimeType,
+			ClockRate:   codec.ClockRate,
+			Channels:    uint8(codec.Channels),
+			SDPFmtpLine: codec.SDPFmtpLine,
+		}
+
+		collector.Collect(stats.ID, stats)
+	}
 }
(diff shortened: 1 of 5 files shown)


stats_go.go
```diff
@@ -61,3 +61,18 @@ func (r StatsReport) GetICECandidatePairStats(c *ICECandidatePair) (ICECandidate
 	}
 	return candidateStats, true
 }
+
+// GetCodecStats is a helper method to return the associated stats for a given Codec
+func (r StatsReport) GetCodecStats(c *RTPCodec) (CodecStats, bool) {
```
Sean-Der, line 66: `GetMediaStats` is probably better. I need to look at `pc-stats` and see if a better name comes out.

Sean-Der:
Look pretty good @obasajujoshua31 I am still reviewing but would you mind changing the commit message to something like this.

```
Add GetMediaStats to StatsReport

Add an API to get basic stats around media. 

Relates to #610 
```

If I am reading `git log` the current commit message makes it really hard to understand what is going on. thanks!

mediaengine.go
```diff
@@ -411,3 +414,21 @@ type RTPCapabilities struct {
...
 }
+
+func (r *RTPCodec) collectStats(collector *statsReportCollector, codecType CodecType) {
+	collector.Collecting()
+
+	stats := CodecStats{
+		Timestamp:   statsTimestampFrom(time.Now()),
+		Type:        StatsTypeCodec,
+		ID:          "codec",
+		PayloadType: uint32(r.PayloadType),
```
tarrencev, line 425: This type can be modified to `uint8` in the `CodecStats` struct definition

mediaengine.go
```diff
@@ -411,3 +414,21 @@ type RTPCapabilities struct {
...
+		MimeType:    r.MimeType,
+		ClockRate:   r.ClockRate,
+		Channels:    uint32(r.Channels),
```
tarrencev, line 428: This is defined as a `unsigned short` in the spec but we can use `uint8` in go. Please update this in the `CodecStats` struct too.

mediaengine.go
```diff
@@ -411,3 +414,21 @@ type RTPCapabilities struct {
...
+		SDPFmtpLine: r.SDPFmtpLine,
+		CodecType:   codecType,
```
tarrencev, line 430: > "encode" or "decode", depending on whether this object represents a media format that the implementation is prepared to encode or decode.

@Sean-Der does this make sense in the context of pion?

mediaengine.go
```diff
@@ -411,3 +414,21 @@ type RTPCapabilities struct {
 	Codecs           []RTPCodecCapability
 	HeaderExtensions []RTPHeaderExtensionCapability
 }
+
+func (r *RTPCodec) collectStats(collector *statsReportCollector, codecType CodecType) {
```
tarrencev, line 418: `codecType` shouldn't be passed in here

tarrencev (changes requested):

peerconnection.go
```diff
@@ -1950,6 +1950,34 @@ func (pc *PeerConnection) GetStats() StatsReport {
 	pc.mu.Unlock()
 
 	statsCollector.Collect(stats.ID, stats)
+
+	if pc.api.mediaEngine != nil {
+		for _, sender := range pc.GetSenders() {
```
tarrencev, line 1955: You can use `GetTransceivers` to get both senders and receivers

peerconnection.go
```diff
@@ -1950,6 +1950,34 @@ func (pc *PeerConnection) GetStats() StatsReport {
 	pc.mu.Unlock()
 
 	statsCollector.Collect(stats.ID, stats)
+
+	if pc.api.mediaEngine != nil {
```
tarrencev, line 1954: I dont think this can be `nil`

peerconnection.go
tarrencev: Actually, I dont think this should be derived from the transceivers. It should just be generated based on the current `MediaEngine`. You can define a method `MediaEngine.collectStats` that aggregates codec stats then is called here with `pc.api.mediaEngine.collectStats(...)`

PR #1441, opened 2020-09-22
Add GetMediaStats to StatsReport

obasajujoshua31:
Add an API to get basic stats around rtpsender.

Relates to #610

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1950,13 +1950,25 @@ func (pc *PeerConnection) GetStats() StatsReport {
 
 	statsCollector.Collect(stats.ID, stats)
 
+	pc.mu.Unlock()
+
+	transceivers := pc.GetTransceivers()
+	for _, transceiver := range transceivers {
+		if sender := transceiver.Sender(); sender != nil {
+			sender.collectStats(statsCollector)
+		}
+
+		if receiver := transceiver.Receiver(); receiver != nil {
+			receiver.collectStats(statsCollector)
+		}
+	}
+
 	certificates := pc.configuration.Certificates
 	for _, certificate := range certificates {
 		if err := certificate.collectStats(statsCollector); err != nil {
 			continue
 		}
 	}
-	pc.mu.Unlock()
 
 	pc.api.mediaEngine.collectStats(statsCollector)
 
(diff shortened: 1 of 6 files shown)


peerconnection.go
```diff
@@ -1947,8 +1947,14 @@ func (pc *PeerConnection) GetStats() StatsReport {
 		DataChannelsOpened:    dataChannelsOpened,
 		DataChannelsRequested: dataChannelsRequested,
 	}
+
 	pc.mu.Unlock()
 
+	senders := pc.GetSenders()
```
tarrencev, line 1953: It would be better to include both `Sender` and `Receiver` stats in this PR. Since they can be computed with a single pass over `pc.rtpTransceivers`

obasajujoshua31: alright cool.

track.go
```diff
@@ -32,9 +33,11 @@ type Track struct {
 
 	packetizer rtp.Packetizer
 
-	receiver         *RTPReceiver
-	activeSenders    []*RTPSender
-	totalSenderCount int // count of all senders (accounts for senders that have not been started yet)
+	receiver             *RTPReceiver
+	activeSenders        []*RTPSender
+	totalSenderCount     int // count of all senders (accounts for senders that have not been started yet)
+	totalSamplesSent     uint32
```
tarrencev, line 39: lets use atomic operations for these `total*` properties rather than holding the mutex

https://golang.org/pkg/sync/atomic/

rtpsender.go
```diff
@@ -201,3 +204,48 @@ func (r *RTPSender) hasSent() bool {
...
+			TotalSamplesSent:     uint64(totalSamplesSent),
+		}
+
+		report.Collect(stats.ID, stats)
+	case RTPCodecTypeVideo:
+		stats := VideoSenderStats{
+			Timestamp:      statsTimestampFrom(time.Now()),
+			Type:           StatsTypeSender,
+			ID:             r.statsID,
+			FramesCaptured: 0, // Todo Frames Captured
```
tarrencev, line 242: Can you do `// TODO ...` for consistency. also, lets make sure to track that these are missing in the

PR #1446, opened 2020-09-24
Remove media stream struct for stats

obasajujoshua31:
Relates to #610

Description
Reference issue
Fixes #...

#### Reference issue
Fixes #...


tarrencev:
Looks like this stats object is now obsolete https://www.w3.org/TR/webrtc-stats/#obsolete-rtcmediastreamstats-members

I think we can remove it completely

PR #1464, opened 2020-10-05
Don't drop a packet at the beginning of a new track.

### LINKED ISSUE #1001: The first RTP packet that triggers OnTrack wouldn't be received (trivial)

jech:
Dropping a packet at the beginning of every track is bad, since
it corrupts the first keyframe of each video track.  This works
by adding a peek method to Track, and using it in NewTrack.

The peeked data is protected by the RWLock already associated
with the track.  We check for its presence with the reader lock
taken, which avoids taking the writer lock (or doing an atomic read)
in the common case (no peeked data).

#### Reference issue
Fixes #1001

### DIFF
--- a/track.go
+++ b/track.go
@@ -34,6 +34,7 @@ type Track struct {
 	receiver         *RTPReceiver
 	activeSenders    []*RTPSender
 	totalSenderCount int // count of all senders (accounts for senders that have not been started yet)
+	peeked           []byte
 }
 
 // ID gets the ID of the track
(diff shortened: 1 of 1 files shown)


Sean-Der:
Merged!

This became a952829951b9c741b31315c446f8e6876534b5db I need to fix one test around this.

PR #1467, opened 2020-10-07
Add SCTPTransport getter to PeerConnection

### LINKED ISSUE #713: OnSelectedCandidatePairChange doesn't work
### Your environment.
  - Version: v2.0.22
  - Browser: n/a

### What did you do?
Attempted to set callback handler via OnSelectedCandidatePairChange() in order to learn the resulting selected-pair for diagnostic purpose.

My app is DC-only.  iceTranport is obtained via dc => sctpTransport => dtlsTransprot => iceTransport...

### What did you expect?
The callback would be made to report the current selected-pair.

#### Offerer:
* Not possible to set the callback because dc.sctpTransport is nil
* When OnOpen is called, sctpTransport was non-nil, but callback was already made (too late)
#### Answerer:
* dc.sctpTransport is never set

Sean-Der:
This is needed so users can use OnSelectedCandidatePairChange
for DataChannel use cases.

Resolves #713

### DIFF
--- a/icetransport_test.go
+++ b/icetransport_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/pion/randutil"
 	"github.com/pion/transport/test"
 	"github.com/stretchr/testify/assert"
 )
@@ -19,18 +18,8 @@ func TestICETransport_OnSelectedCandidatePairChange(t *testing.T) {
 	lim := test.TimeOut(time.Second * 30)
 	defer lim.Stop()
 
-	api := NewAPI()
-	api.mediaEngine.RegisterDefaultCodecs()
-	pcOffer, pcAnswer, err := api.newPair(Configuration{})
+	pcOffer, pcAnswer, err := newPair()
 	if err != nil {
-		t.Fatal(err)
-	}
-
-	opusTrack, err := pcOffer.NewTrack(DefaultPayloadTypeOpus, randutil.NewMathRandomGenerator().Uint32(), "audio", "pion1")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if _, err = pcOffer.AddTrack(opusTrack); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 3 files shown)


icetransport_test.go
```diff
@@ -43,17 +32,9 @@ func TestICETransport_OnSelectedCandidatePairChange(t *testing.T) {
 	})
 
 	senderCalledCandidateChange := int32(0)
-	for _, sender := range pcOffer.GetSenders() {
```
Sean-Der, line 46: Just to make the test shorter

peerconnection.go
```diff
@@ -330,10 +330,6 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool { //nolint:gocognit
 		return true
 	}
 
-	if pc.sctpTransport == nil {
```
Sean-Der, line 333: `sctpTransport` is never nil. If something is depending on this behavior I would like to know

peerconnection.go
```diff
@@ -1773,13 +1769,11 @@ func (pc *PeerConnection) Close() error {
 	}
 
 	// https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close (step #5)
-	if pc.sctpTransport != nil {
```
Sean-Der, line 1776: `sctpTransport` is never nil. If something is depending on this behavior I would like to know

peerconnection.go
```diff
@@ -1916,24 +1910,22 @@ func (pc *PeerConnection) GetStats() StatsReport {
 		pc.iceTransport.collectStats(statsCollector)
 	}
 
-	if pc.sctpTransport != nil {
```
Sean-Der, line 1919: `sctpTransport` is never nil. If something is depending on this behavior I would like to know

peerconnection_renegotiation_test.go
```diff
@@ -689,15 +687,26 @@ func TestPeerConnection_Renegotiation_NoApplication(t *testing.T) {
...
+		}
+	})
+
 	_, err = pcOffer.AddTransceiverFromKind(RTPCodecTypeVideo, RtpTransceiverInit{Direction: RTPTransceiverDirectionSendrecv})
 	assert.NoError(t, err)
 
 	_, err = pcAnswer.AddTransceiverFromKind(RTPCodecTypeVideo, RtpTransceiverInit{Direction: RTPTransceiverDirectionSendrecv})
 	assert.NoError(t, err)
 
-	// Setting SCTPTransport to nil ensures any interaction with it will cause a segafault
```
Sean-Der, line 698: This is a really stupid way to test this. The test didn't even work because we had nil guards.

Instead I wait until Connected and check the SCTPTransport status

tarrencev (approved):

PR #1474, opened 2020-10-09
Fix simulcast tracks not stopping on remove

OrlandoCo:
#### Description

This commit fix the case where simulcast tracks are removed, and on the negotiation the receivers are not being stopped.

The transceivers receivers are stopped in the *PeerConnection.startRTP(l:2010) function but currently only supports non-simulcast tracks because it checks that the tracks has been removed by SSRC in SDP, since Simulcast does not use SSRC the only way to check if a track has been removed is to check mid direction.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -992,7 +992,6 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 			}
 
 			t, localTransceivers = findByMid(midValue, localTransceivers)
-			// Fix simulcast tracks removing
 			if t != nil && t.Receiver() != nil && len(t.Receiver().Tracks()) > 1 {
 				switch direction {
 				case RTPTransceiverDirectionInactive:
@@ -1000,10 +999,8 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 						return err
 					}
 				case RTPTransceiverDirectionRecvonly:
-					if recv := t.Receiver(); recv != nil {
-						if err := recv.Stop(); err != nil {
-							return err
-						}
+					if err := t.Receiver().Stop(); err != nil {
+						return err
 					}
 				case RTPTransceiverDirectionSendonly, RTPTransceiverDirectionSendrecv:
 					t.setDirection(RTPTransceiverDirectionRecvonly)
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -992,6 +992,29 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 			}
 
 			t, localTransceivers = findByMid(midValue, localTransceivers)
+			// Fix simulcast tracks removing
+			if t != nil && t.Receiver() != nil && len(t.Receiver().Tracks()) > 1 {
+				switch direction {
+				case RTPTransceiverDirectionInactive:
+					if err := t.Stop(); err != nil {
+						return err
+					}
+				case RTPTransceiverDirectionRecvonly:
+					if recv := t.Receiver(); recv != nil {
+						if err := recv.Stop(); err != nil {
+							return err
+						}
+					}
+				case RTPTransceiverDirectionSendonly, RTPTransceiverDirectionSendrecv:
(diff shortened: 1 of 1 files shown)


peerconnection.go
```diff
@@ -992,6 +992,29 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+					if err != nil {
+						return err
+					}
+					t.setReceiver(receiver)
+				}
+				continue
+			}
 			if t == nil {
 				t, localTransceivers = satisfyTypeAndDirection(kind, direction, localTransceivers)
 			} else if direction == RTPTransceiverDirectionInactive {
```
billylindeman, line 1020: is this still necessary?

Sean-Der: I don't think so! if the test still passes we are all good :)

peerconnection.go
```diff
@@ -992,6 +992,29 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+					if err := t.Stop(); err != nil {
+						return err
+					}
+				case RTPTransceiverDirectionRecvonly:
+					if recv := t.Receiver(); recv != nil {
+						if err := recv.Stop(); err != nil {
+							return err
+						}
+					}
+				case RTPTransceiverDirectionSendonly, RTPTransceiverDirectionSendrecv:
```
Sean-Der, line 1008: Do we need `RTPTransceiverDirectionSendrecv` here?

peerconnection.go
```diff
@@ -992,6 +992,29 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 			}
 
 			t, localTransceivers = findByMid(midValue, localTransceivers)
+			// Fix simulcast tracks removing
```
Sean-Der, line 995: I would remove the comment

peerconnection.go
```diff
@@ -992,6 +992,29 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+			if t != nil && t.Receiver() != nil && len(t.Receiver().Tracks()) > 1 {
```
Sean-Der, line 996: This all LGTM! We do need test coverage, I don't think this behavior is Simulcast specific right? We are just reacting to the remotes directions changing?

peerconnection.go
OrlandoCo: In my tests, after the track is re attached in `sendrecv` transceiver it changes to `sendonly` so it may not be required just let it there as precaution, but it can be removed.

peerconnection.go
OrlandoCo: This behaviour is specific to Simulcast since I explained in the commit message this is handled in `PeerConnection.startRTP(l:2010)` but doesn't support Simulcast tracks.

This behaviour can be seen in many renog tests like TestPeerConnection_Renegotiation_AddTrack and TestPeerConnection_Renegotiation_RemoveTrack, but with a simulcast track fails.

peerconnection.go
OrlandoCo: If this is removed this test fails `TestPeerConnection_Renegotiation_DisableTrack`

peerconnection.go
billylindeman: Can you explain the length check? If it was removed I think this would cover the case from below right?

OrlandoCo: It's to check if current transceiver have a simulcast receiver, If I remove it, it would affect normal tracks, but that is already supported in `PeerConnection.startRTP(l:2010)`

billylindeman: It seems like startRTP should be modified to call .Tracks() and loop them instead of .Track() ? https://github.com/pion/webrtc/blob/master/peerconnection.go#L1995. This function will always continue on a transceiver with more than 1 track because .Track() returns nil. It does an == not a >= check https://github.com/pion/webrtc/blob/6e7249351760f23cbc88eaab4f032ee05adad1a8/rtpreceiver.go#L65. Not sure of the implications of any of this @Sean-Der

Sean-Der:
Closing for https://github.com/pion/webrtc/pull/1942

PR #1479, opened 2020-10-14
Found JS Errors when running JS fiddle code examples in safari

jamiegood:
Logs: TypeError: pc.addStream is not a function. (In 'pc.addStream(displayVideo(stream))', 'pc.addStream' is undefined)
Fixed js error when using safari due to deprecated RTCPeerConnection.addStream() method.

Updated js fiddle examples to use the addTrack() method instead.
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addStream

#### Reference issue
Fixes #...

### DIFF
--- a/examples/reflect/README.md
+++ b/examples/reflect/README.md
@@ -9,7 +9,7 @@ go get github.com/pion/webrtc/v3/examples/reflect
 ```
 
 ### Open reflect example page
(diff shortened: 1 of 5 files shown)


Sean-Der:
Merged, thank you so much for your contributions @jamiegood ! Hope you stay involved, we need all the help we can get :)

63112a4ed409faa923a67c62b8cc2c75dc571590

PR #1488, opened 2020-10-23
expose pc signaling state

tarrencev:
signaling state is necessary for implementing perfect negotiation according to https://w3c.github.io/webrtc-pc/#perfect-negotiation-example


tarrencev:
hmm this already exists not sure how i missed it 😵

PR #1492, opened 2020-10-24
Fix usernameFragment string pointer to value

### LINKED ISSUE #1484: recovered with non-error value: (string) ValueOf: invalid value
Caused by this line:

https://github.com/pion/webrtc/blob/7fc6f5435bfb651abb5d609a84e8763579da6105/peerconnection_js.go#L655

should be passed through `stringPointerToValue` before being put into the map

assadobaid:
#### Description
Use the value of string pointer.
#### Reference issue
Fixes #1484


Sean-Der:
Merged, thank you so much @assadobaid :)

PR #1496, opened 2020-10-27
fix: nil sender for send transceiver

### LINKED ISSUE #1494: panic on checkNegotiationNeeded
Seems there is a case where `Track` is nil while executing `checkNegotiationNeeded`. I don't think this should be possible, since the Senders direction is `Send*` but it doesn't have a track set. I think all senders in `Send*` state should have a track defined.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x15caf3c]
goroutine 4980 [running]:
github.com/pion/webrtc/v3.(*RTPSender).Track(0x0, 0x0)
	/Users/tvanas/go/pkg/mod/github.com/pion/webrtc/v3@v3.0.0-beta.10/rtpsender.go:80 +0x3c
github.com/pion/webrtc/v3.(*PeerConnection).checkNegotiationNeeded(0xc00011eb60, 0x15dc500)
	/Users/tvanas/go/pkg/mod/github.com/pion/webrtc/v3@v3.0.0-beta.10/peerconnection.go:359 +0x91a
github.com/pion/webrtc/v3.(*PeerConnection).negotiationNeededOp(0xc00011eb60)
	/Users/tvanas/go/pkg/mod/github.com/pion/webrtc/v3@v3.0.0-beta.10/peerconnection.go:301 +0x90
github.com/pion/webrtc/v3.(*operations).start(0xc0009aa240)
	/Users/tvanas/go/pkg/mod/github.com/pion/webrtc/v3@v3.0.0-beta.10/operations.go:84 +0x61
created by github.com/pion/webrtc/v3.(*operations).Enqueue
	/Users/tvanas/go/pkg/mod/github.com/pion/webrtc/v3@v3.0.0-beta.10/operations.go:35 +0xcb
exit status 2
```

tarrencev:
there is a race condition where a `Send*` transceiver can be mutated during renegotiation to no longer have a sender. In this case, pion will throw. This pr attempts to resolve this. The assumption is that the mutation that triggered the race condition has queued it's own `onNegotiationNeeded`, so we can just end the current one early and let that one proceed

resolves #1494

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -355,6 +355,13 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool { //nolint:gocognit
 		if !t.stopped && m != nil {
 			// Step 5.3.1
 			if t.Direction() == RTPTransceiverDirectionSendrecv || t.Direction() == RTPTransceiverDirectionSendonly {
+				// non-cannon:
+				// if sender track is nil, transceivers were modified during
+				// current negotiation dont proceed
+				if t.Sender() == nil {
+					return false
+				}
+
 				descMsid, okMsid := m.Attribute(sdp.AttrKeyMsid)
 				if !okMsid || descMsid != t.Sender().Track().Msid() {
 					return true


tekig:
@tarrencev hi!

Transceiver direction is set `Sendrecv` or `Sendonly`. I think that `Sender` should not be `nil`. If a race condition occurs, then nothing prevents on the next line from set `Sender` to `nil`.

Your thought about the race gave me an idea. When [removing a track,](https://github.com/pion/webrtc/blob/fix/nilsender/peerconnection.go#L1542) a state may arise when the direction is still set, but the sender is nil. Since the [setSendingTrack](https://github.com/pion/webrtc/blob/fix/nilsender/rtptransceiver.go#L99-L114) method is not thread safe. I think if make a piece of code thread-safe, the problem will be solved.

Please correct me if I am wrong

tarrencev:
@tekig hi!

i think you are right. it should never be the case that a `Send*` track has a nil Sender. it should ideally be made thread safe so that can not occur. i think that would be the most correct solution

Sean-Der:
@tekig @tarrencev anything we should do here still? thanks!

tekig:
@Sean-Der the transceiver is not thread safe, I think need to rewrite the transceiver to thread safe code instead.
As a solution to the problem, it may work if you copy the address to Sender and work with it. Are there any better ideas?

```go
if t.Direction() == RTPTransceiverDirectionSendrecv || t.Direction() == RTPTransceiverDirectionSendonly {
	// non-cannon:
	// if sender track is nil, transceivers were modified during
	// current negotiation dont proceed
	sender := t.Sender()
	if sender == nil {
		return false
	}

	descMsid, okMsid := m.Attribute(sdp.AttrKeyMsid)
	if !okMsid || descMsid != sender.Track().Msid() {
		return true
	}
}
```

tarrencev:
@Sean-Der @tekig we can close this PR and create an issue for the thread safe improvement

PR #976, opened 2020-01-13
Add examples/play-h264-from-disk

chertov:
#### Description
It doesn't work in Chrome. Only Firefox has been tested successfully.

PR #1502, opened 2020-10-30
Add  h264reader and test for it

ashellunts:
Continuing work started by @chertov in https://github.com/pion/webrtc/pull/976

Moved h264reader from examples to inside library. And adapted little bit code to look like ivfreader.
For a moment it reads all frames in 1 call. It should be improved in later PRs.

### DIFF
--- a/pkg/media/h264reader/h264reader.go
+++ b/pkg/media/h264reader/h264reader.go
@@ -1,0 +1,166 @@
+package h264reader
+
+import (
+	"bytes"
+	"errors"
+	"io"
+)
+
+// H264Reader reads data from stream and constructs h264 nal units
+type H264Reader struct {
+	stream                      io.Reader
+	nalBuffer                   []byte
+	countOfConsecutiveZeroBytes int
+	nalPrefixParsed             bool
+}
+
+var (
+	errNilReader           = errors.New("stream is nil")
+	errReadData            = errors.New("error on data read")
+	errDataIsNotH264Stream = errors.New("data is not a H264 bitstream")
+	errNotEnoughData       = errors.New("not enough data")
+)
+
+// NewReader creates new H264Reader
+func NewReader(in io.Reader) (*H264Reader, error) {
+	if in == nil {
+		return nil, errNilReader
+	}
+
+	reader := &H264Reader{
+		stream:          in,
+		nalBuffer:       make([]byte, 0),
+		nalPrefixParsed: false,
+	}
+
+	return reader, nil
+}
+
+// NAL H.264 Network Abstraction Layer
+type NAL struct {
+	PictureOrderCount uint32
+
+	// NAL header
+	ForbiddenZeroBit bool
+	RefIdc           uint8
+	UnitType         NalUnitType
+
+	Data []byte // header byte + rbsp
+}
+
+func (reader *H264Reader) bitStreamStartsWithH264Prefix() (prefixLength int, e error) {
+	nalPrefix3Bytes := []byte{0, 0, 1}
+	nalPrefix4Bytes := []byte{0, 0, 0, 1}
+
+	prefixBuffer := make([]byte, 4)
+
+	n, err := reader.stream.Read(prefixBuffer)
+
+	if err != nil || n == 0 {
+		return 0, errReadData
+	}
+
+	if n < 3 {
+		return 0, errDataIsNotH264Stream
+	}
+
+	nalPrefix3BytesFound := bytes.Equal(nalPrefix3Bytes, prefixBuffer[:3])
+	if n == 3 {
+		if nalPrefix3BytesFound {
+			return 0, errNotEnoughData
+		}
+		return 0, errDataIsNotH264Stream
+	}
+
+	// n == 4
+	if nalPrefix3BytesFound {
+		reader.nalBuffer = append(reader.nalBuffer, prefixBuffer[3])
+		return 3, nil
+	}
+
+	nalPrefix4BytesFound := bytes.Equal(nalPrefix4Bytes, prefixBuffer)
+	if nalPrefix4BytesFound {
+		return 4, nil
+	}
+	return 0, errDataIsNotH264Stream
+}
+
+func (reader *H264Reader) NextNAL() (*NAL, error) {
+	if !reader.nalPrefixParsed {
+		_, err := reader.bitStreamStartsWithH264Prefix()
+		if err != nil {
+			return nil, err
+		}
+
+		reader.nalPrefixParsed = true
+	}
+
+	for {
+		buffer := make([]byte, 1)
+		n, err := reader.stream.Read(buffer)
+
+		if err != nil || n != 1 {
+			break
+		}
+		readByte := buffer[0]
+		nalFound := reader.processByte(readByte)
+		if nalFound {
+			nal := newNal(reader.nalBuffer)
+			nal.parseHeader()
+			if nal.UnitType == NalUnitTypeSEI {
+				reader.nalBuffer = nil
+				continue
+			} else {
+				break
+			}
+		}
+
+		reader.nalBuffer = append(reader.nalBuffer, readByte)
+	}
+
+	if len(reader.nalBuffer) == 0 {
+		return nil, errNotEnoughData
+	}
+
+	nal := newNal(reader.nalBuffer)
+	reader.nalBuffer = nil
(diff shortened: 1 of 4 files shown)


Sean-Der:
@ashellunts absolutely amazing work! Nice job :D 

I will get this merged this weekend, looking quickly this is perfect. Thank you so much. This has been hanging over us for months.

Sean-Der:
Oh you mind adding yourself to the README?

Hopefully you enjoyed yourself and start to feel more comfortable with Go :)

ashellunts:
Thank you @Sean-Der. Yep, I will update README :)

pkg/media/h264reader/h264reader.go
```diff
@@ -0,0 +1,272 @@
...
+	endOfStream              nalUnitType = 11 // End of stream
+	filler                   nalUnitType = 12 // filler data
+	spsExt                   nalUnitType = 13 // Sequence parameter set extension
+	// 14..18           // Reserved
+	nalUnitTypeCodedSliceAux nalUnitType = 19 // Coded slice of an auxiliary coded picture without partitioning
+	// 20..23           // Reserved
+	// 24..31           // unspecified
+)
+
+func nalUnitTypeStr(v nalUnitType) string {
```
Sean-Der, line 192: Mind changing this to be

```
func (n *nalUnitType) String() string { }
```

pkg/media/h264reader/h264reader.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		h.LastNullCount = 0
+	}
+	return nals
+}
+
+type nalUnitType uint8
+
+const ( //   Table 7-1 NAL unit type codes
+	unspecified              nalUnitType = 0  // unspecified
```
Sean-Der, line 172: I think those are ok to be exported! This is probably good information for users to have.

I think it is also good form to include the type name `NALUnitTypeUnspecified`. I need to check if we do that consistently though :)

pkg/media/h264reader/h264reader.go
```diff
@@ -0,0 +1,272 @@
...
+	// NAL header
+	ForbiddenZeroBit bool
+	RefIdc           uint8
+	UnitType         nalUnitType
+
+	Data []byte // header byte + rbsp
+}
+
+// ReadFrames reads all data from stream and returns array of all parsed nal units
+func (reader *H264Reader) ReadFrames() []NAL {
```
Sean-Der, line 43: I was playing with this, and I think we need to re-think this :( I wanted to do `ffmpeg stdio -> pion` and reading all up front doesn't work. All of our other Reader APIs do a frame at a time.

I think we need to refactor to read 1 byte at a time, and then emit when we read a full frame.

* First run assert it starts with `0x00, 0x00, 0x01`
* Read until you hit `0x00, 0x00, 0x01` again, return the frame.

Then all the next reads just go until they hit `0x00, 0x00, 0x01` and we just keep some local state (already seen NAL header)

sorry if this is a pain, if you want we can push a V1 and then do a refactor if you are out of time on this :)

pkg/media/h264reader/h264reader.go
```diff
@@ -0,0 +1,272 @@
...
+	buf           []byte
+}
+
+func newFindNalState() findNalState {
+	return findNalState{PrefixCount: 0, LastNullCount: 0, buf: make([]byte, 0)}
+}
+
+func (h *findNalState) NalScan(data []byte) [][]byte {
+	if len(h.buf) > 1024*1024 {
+		panic("findNalState buf len panic")
```
Sean-Der, line 94: We don't want do `panic` inside the library. Instead return `[][]byte, error` but since we need to do a major refactor this might not mater!

pkg/media/h264reader/h264reader_test.go
```diff
@@ -0,0 +1,34 @@
+package h264reader
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestH264Reader(t *testing.T) {
```
Sean-Der, line 10: I wouldn't even use real H264. I would just do something like this. 

```
testH264 := []byte{0x00, 0x00, 0x00, 0x01, 0xAB, 0xAB}
```

Then we can do tests for
* Does it handle the different start codes
* How does it handle invalid data

pkg/media/h264reader/h264reader.go
```diff
@@ -0,0 +1,272 @@
...
+	nalUnitTypeCodedSliceAux nalUnitType = 19 // Coded slice of an auxiliary coded picture without partitioning
+	// 20..23           // Reserved
+	// 24..31           // unspecified
+)
+
+func nalUnitTypeStr(v nalUnitType) string {
+	str := "Unknown"
+	switch v {
+	case 0:
+		{
```
Sean-Der, line 196: I would drop the `{ }` on each, you don't need a new scope!

Sean-Der:
Left some comments @ashellunts sorry this took so long!

I think the big issue is we need to be able to read as a stream, and not just a pre-made file from disk.

The [ogg tests](https://github.com/pion/webrtc/blob/master/pkg/media/oggreader/oggreader_test.go#L26) are really good example of this.

If you don't have the bandwidth I am happy to help out as well! Will leave you as the author on the commit, you have done the majority of the work.

pkg/media/h264reader/h264reader.go
ashellunts: I think about removing this function. It is not used in parser.

pkg/media/h264reader/h264reader.go
ashellunts: Thanks for comments. I will refactor as you suggest.

pkg/media/h264reader/h264reader.go
ashellunts: done

pkg/media/h264reader/h264reader.go
ashellunts: done

pkg/media/h264reader/h264reader.go
ashellunts: done

pkg/media/h264reader/h264reader.go
ashellunts: ok

pkg/media/h264reader/h264reader.go
ashellunts: done

ashellunts:
Something is wrong with codecov/path check. It shows that code coverage is decreased in sctptransport.go though this PR didn't change that file.

Sean-Der:
This works perfectly @ashellunts this is amazing! Merging now. I am also going to drop the instructions on how to use it :) 

2020-11-04-232017_808x558_scrot

If you are interested I would love to make a little `ffmpeg -> pion` demo. The application will just take ffmpeg arguments.  Then we can easily show how to do desktop capture and other cool stuff! 

This is going to let people build some really cool things. By making ffmpeg and WebRTC easier you are unlocking so many interesting use cases.

Sean-Der:
To play with this first encode a video `ffmpeg -i doom.mkv -g 30  -preset fast -profile:v baseline -bsf h264_mp4toannexb doom.h264`

Then replace [play-from-disk](https://github.com/pion/webrtc/blob/master/examples/play-from-disk/main.go) with this modified [main.go](https://gist.github.com/Sean-Der/69e2f5a13557de7bdadb2c7a9b0adc73)

Sean-Der (approved):

ashellunts:
Thank you @Sean-Der! I am very glad to contribute to pion.

I can do ffmpeg example, it is interesting. What do you think also about updating play-from-disk to support h264?

Sean-Der:
@ashellunts I am all for updating `play-from-disk`! If you put `H264` and `VP8` in self contained functions I think it will help a lot. There is a careful balance. If the example does too much it will be overwhelming, not enough then it is anemic :(

I am really excited to see a `ffpmeg` example (under your personal Github for a few reasons)
* It gives you the credit you deserve. I like seeing people own their stuff, and then improve Pion to serve what they need. It is more motivating, and it seems to have bigger career impact!

* It builds a bigger community (we will link to it in `awesome-pion` and we can publish on Twitter etc...)

* It keeps `pion/webrtc` simple. People already find the code base challenging to work with, so anything I can do to keep it slim is a :+1:

PR #1504, opened 2020-11-02
interceptor POC

masterada:
Updated interceptor design based on TrackLocal:

Changes:
1. Interceptor now have Bind methods similar to track, which are called for every track (local/remote) for rtp, and once per peerconnection for rtcp.
2. Added interceptors separately to API, instead of under setting engine. Did this because interceptors must be created per peerconnection, and InterceptorFactory seemed confusing. Also API was already a depedency of Receiver/Sender, so this was was much simpler. Similarly to MediaEngine, Interceptors now must be configured for each PeerConnection and added to API. Default interceptors can later be added in NewAPI method.
3. Instead of using interfaces with single a single function (eg: RTPWriter with a single WriteRTP function), I use function types. This makes implementing an interceptor simpler, because there is no need to create a separate struct for each interface (inline functions can be used instead).

```
type WriteRTP func(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error)
type ReadRTP func() (*rtp.Packet, map[interface{}]interface{}, error)
type WriteRTCP func(pkts []rtcp.Packet, attributes map[interface{}]interface{}) (int, error)
type ReadRTCP func() ([]rtcp.Packet, map[interface{}]interface{}, error)

type Interceptor interface {
	BindReadRTCP(read ReadRTCP) ReadRTCP     // TODO: call this
	BindWriteRTCP(write WriteRTCP) WriteRTCP // TODO: call this

	BindLocalTrack(ctx *TrackLocalContext, write WriteRTP) WriteRTP
	UnbindLocalTrack(ctx *TrackLocalContext)

	BindRemoteTrack(ctx *TrackRemoteContext, read ReadRTP) ReadRTP
	UnbindRemoteTrack(ctx *TrackRemoteContext)

	io.Closer
}
```

Questions:
1. Please see interceptor_peerconnection.go and interceptor_track_remote.go. They are quite ugly. I'm open to any suggestions.
2. ReadRTCP is currently present on both RTPSender and RTPReceiver, while WriteRTCP is on PeerConnection. I think it would make much more sense to move ReadRTCP to PeerConnection as well. The current solution using `DestinationSSRC()` method is very opionated and causes some rtcp packets to be duplicated (in case they belong to multiple ssrc, or they are in a batch with another rtcp packet that belong to a different rtcp, see session_rtcp.go/destinationSSRC). In fact in our project we went to great length to undo this behaviour and get a single ReadRTCP method, without packet duplications. Is there any plan to modify this behaviour? (It would make calling BindReadRTCP in the Interceptor much easier).

Any suggestions are welcome.

@Sean-Der @at-wat please check


Sean-Der:
@masterada amazing work, reviewing now! I am so excited for the base to land, then we can start adding stuff. I am especially excited to implement a congestion controller (for my own learning)

I forced pushed just to fix the WASM build! Just make sure to reset off of that.

api.go
```diff
@@ -53,3 +59,15 @@ func WithSettingEngine(s SettingEngine) func(a *API) {
...
 	}
 }
+
+func WithInterceptor(interceptor Interceptor) func(a *API) {
+	return func(a *API) {
+		a.interceptors = append(a.interceptors, interceptor)
+	}
+}
+
+func ClearInterceptors() func(a *API) {
```
Sean-Der, line 69: What do you think of copying `MediaEngine` design? You get a list of well configured defaults OR you can customize them all.

`RegisterDefaultInterceptors` or w/e we call it will be easy to read so people can understand what is happening!

interceptor.go
```diff
@@ -0,0 +1,54 @@
...
+	BindLocalTrack(ctx *TrackLocalContext, write WriteRTP) WriteRTP
+	UnbindLocalTrack(ctx *TrackLocalContext)
+
+	BindRemoteTrack(ctx *TrackRemoteContext, read ReadRTP) ReadRTP
+	UnbindRemoteTrack(ctx *TrackRemoteContext)
+
+	io.Closer
+}
+
+type DummyInterceptor struct{}
```
Sean-Der, line 30: I would re-name this `NoOpInterceptor` or something like that. Maybe we can move to a test and use it for some simple things?

api.go
```diff
@@ -13,6 +13,8 @@ import (
 type API struct {
 	settingEngine *SettingEngine
 	mediaEngine   *MediaEngine
+	interceptors  []Interceptor
```
Sean-Der, line 16: What do you think of the following?

* Do `settingEngine.interceptors`
* On PeerConnection creation pluck instantiate the chain and do it as `PeerConnection.interceptor`
* If `len(a.interceptors) == 0` use `NoOpInterceptor`. I would like to drop all the `if a.interceptor == nil` checks, eventually we are going to miss one.

interceptor_chain.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (i *interceptorChain) UnbindRemoteTrack(ctx *TrackRemoteContext) {
+	for _, interceptor := range i.interceptors {
+		interceptor.UnbindRemoteTrack(ctx)
+	}
+}
+
+func (i *interceptorChain) Close() error {
+	var err error
```
Sean-Der, line 63: `util.FlattenErrors` is nice here!

interceptor_track_local.go
```diff
@@ -0,0 +1,46 @@
...
+	TrackLocal
+	interceptor Interceptor
+}
+
+type interceptorTrackLocalWriter struct {
+	TrackLocalWriter
+	writeRTP WriteRTP
+}
+
+func WrapTrackLocal(track TrackLocal, interceptor Interceptor) TrackLocal {
```
Sean-Der, line 19: Instead of doing `Wrap...` would it be possible to rewrite stuff to assume an `Interceptor` is in use everywhere? Helper functions are sometimes harder to follow like this. You have to step into things when debugging.

Sean-Der:
@masterada This is really great. I am all for merging this ASAP, we just need to get the basics right.

I think if we move quickly to merge it will also get others to review soonner (rather then later as well), brings more excitement and involvement :)

interceptor.go
masterada: I made it public so it can be embedded in other Interceptors if you do not want to implement all methods.

api.go
masterada: I removed it from settingEngine for two reasons:
a. The Interceptor factory seemed unnecessarily confusing.
b. See: https://github.com/pion/webrtc-v3-design/blob/interceptor/interceptor/transportcc.go#L26 Some interceptors will need to register stuff with the MediaEngine. That means, as MediaEngine is per PeerConnection, you will need to call Configure** methods for every PeerConnection anyway. So putting the RegisterDefaultInterceptors into SettingEngine would be confusing I think. 

> If len(a.interceptors) == 0 use NoOpInterceptor. I would like to drop all the if a.interceptor == nil checks, eventually we are going to miss one.

Good idea, will do that.

interceptor_chain.go
masterada: Thanks!

api.go
masterada: Added a similar usage example to save-to-disk example.

interceptor_track_local.go
masterada: Could do this for TrackLocal, because I could wrap the TrackLocalWriter interface with one that runs the interceptors, then calls the original method. I'm having trouble with RemoteTrack though:

The TrackLocalWriter is implemented by WriteStreamSRTP in srtp package.
Similarly, there could be a TrackRemoteReader interface that could be implemented by ReadStreamSRTP. But TrackRemote does NOT use the ReadRTP function from ReadStreamSRTP, instead it is reimplemented in TrackRemote. I think this is because of peeking mechanism for determining the payload type. Would it be ok to move the payload type determination logic to srtp/ReadStreamSRTP? That would solve this issue, and avoid having the ReadRTP logic duplcated in TrackRemote.

Sean-Der:
@masterada this looks really good! I am gonna fix up some stuff locally and push (commit message, last lint errors) and get this green.

I think we can merge this and then work on filling out the interceptors :) 

@at-wat @lherman-cs @adwpc @OrlandoCo any thoughts/concerns? I am happy to leave it open longer if people need more time

interceptor_track_local.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"sync"
+
+	"github.com/pion/rtp"
+)
+
+type interceptorTrackLocalWriter struct {
+	TrackLocalWriter
+	writeRTP WriteRTP
+	m        sync.RWMutex
```
OrlandoCo, line 14: Can we use atomics here?

Sean-Der: Yea! Will make that happen now

masterada: Good catch, modified to use atomic.

masterada:
Added some tests, and fixed all review comments (I think) + some other stuff I found. Still need to call UnbindLocalTrack and UnbindRemoteTrack (not sure where those would be good to call), but apart from that I think this can be merged.

Sean-Der:
@masterada I agree, I am up for a merge :) Lets wait maybe 2 more days for @at-wat he pushed for this originally, so would love his buy in if possible! He is pretty busy so if we don't hear we can always come back :)

This is some really amazing/foundational work you have done here! Being able to have publically viewable/easy to understand interceptors is going to be so great. It isn't just going to make Pion better, but I hope make RTC more accessiable. 

Lots of people read Pion's code, and then go work on other implementations. So this is really great for everyone

Sean-Der:
@masterada sorry to hit you with another thing :(

The import pattern right now is reversed I think. I don't think we want something in `pkg` importing the base package. I started to look into that, and realize the headaches :(

* We need to move the typedef for `SSRC` and `PayloadType` into `github.com/pion/rtp`
* We might not be able to use `TrackLocalContext` and `TrackRemoteContext` or we should move those into `github.com/pion/webrtc/pkg/context` ?

I am happy to make these changes! Or if you want I can merge them right away. One big goal for me with the interceptors is that people should be able to use them outside of WebRTC. I think for `/v4+` we will even talk about moving Interceptors into their own package. That way we can start to share stuff with https://github.com/aler9/rtsp-simple-server

Sorry for how long this has dragged on. I 100% promise this is the last thing! Right after we fix this we can merge, anything else can wait.

Sean-Der:
@aler9 I would love to hear your opinion on the work we are doing here!

https://github.com/pion/webrtc-v3-design/issues/34 is the design doc I wrote. I saw you are doing work on receiver reports! It would be awesome if we could get something that would work for all of us :)

masterada:
The core parts of the interceptor are in the base package, only the implementations are in pkg/interceptor. I was thinking that if there was a 3rd party interceptor, it would make sense that it would import the webrtc package. That's why I added the specific interceptors into pkg/interceptor package, which use the base package. 

Currenty the dependency would be:
3rd party interceptor -> webrtc
pkg/interceptor -> webrtc
I will think about how we can reverse the dependency to something like: 
webrtc -> pkg/interceptor
3rd party interceptor -> pkg/interceptor

Although I think it will complicate things to be able to modify the sdp from inteceptors. Maybe functions like `modifyCodecs` method on the Interceptor would be needed.

Sean-Der:
@masterada I think we will need to make interceptors SDP unaware. That way non-WebRTC users can find them useful. The in the base package we can put the helpers together?

Maybe over engineering it. I just want to make these useful to people outside Go/Webrtc.

interceptor.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"io"
+
+	"github.com/pion/rtcp"
+	"github.com/pion/rtp"
+)
+
+// WriteRTP is used by Interceptor.BindLocalTrack.
+type WriteRTP func(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error)
```
at-wat, line 13: I think it's better to define Attribute type in the same way as `net/http.Header`.
Does it require to take `interface{}` as a key instead of `string`?

interceptor.go
```diff
@@ -0,0 +1,90 @@
...
+	// change in the future. The returned method will be called once per packet batch.
+	BindReadRTCP(read ReadRTCP) ReadRTCP
+
+	// BindWriteRTCP lets you modify any outgoing RTCP packets. It is called once per PeerConnection. The returned method
+	// will be called once per packet batch.
+	BindWriteRTCP(write WriteRTCP) WriteRTCP
+
+	// BindLocalTrack lets you modify any outgoing RTP packets. It is called once for per LocalTrack. The returned method
+	// will be called once per rtp packet.
+	BindLocalTrack(ctx *TrackLocalContext, write WriteRTP) WriteRTP
```
at-wat, line 38: It looks very confusing with `context.Context`. `TrackLocalContext` may be better to be changed?

interceptor.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// UnbindLocalTrack is called when the Track is removed. It can be used to clean up any data related to that track.
+	UnbindLocalTrack(ctx *TrackLocalContext)
+
+	// BindRemoteTrack lets you modify any incoming RTP packets. It is called once for per RemoteTrack. The returned method
+	// will be called once per rtp packet.
+	BindRemoteTrack(ctx *TrackRemoteContext, read ReadRTP) ReadRTP
```
at-wat, line 45: Don't we want to split `Interceptor` into `InterceptorLocal` and `InterceptorRemote`? I guess most of the interceptor implementation implements only one of local/remote.

interceptor.go
at-wat: Personally, I like to make it Reader/Writer interface as like `io` stdlib.
For example
```go
type WriterRTP interface {
  Write(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error)
}
```
It may have an adapter like
```go
type WriterRTPFunc func(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error)
func (f WriterRTPFunc) Write(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error) {
  return f(p, attribute)
}
```

at-wat (commented):
Sorry for being offline at here. Added some comments.

interceptor.go
Sean-Der: There is a possibility that someone may want both! I was thinking it would be nice if `interceptor.Stats` covered both paths. I am sure we will have more as we build them out.

The nice thing is that a user doesn't need to define everything, they can embed a `interceptor.NoOp` and then only define the things they care about.

interceptor.go
Sean-Der: Oh yea that is a good point, forgot about that! `io.Reader` is nice because it integrates into other things easily.

If we satisfy that interface then we can use a bunch of existing stuff.

interceptor.go
Sean-Der: I think we should define a new type just for interceptors. We might hit snags where we want to add things to Interceptors (and not Track)

interceptor.go
OrlandoCo: That can be implemented right now with the `TrackLocalWriter`

OrlandoCo: For me is not very clear the purpose of the attributes map

lherman-cs: For me personally, I really like http.ResponseWriter approach, add a method to WriterRTP to retrieve the attributes, and the retrieved attributes will have its own type that contains setters/getters:

```golang
type Attributes map[interface{}]interface{}
func (attributes Attributes) Get(k interface{}) interface{} {}
func (attributes Attributes) Set(k, v interface{}) {}
func (attributes Attributes) GetUint32(k interface{}) uint32 {}
func (attributes Attributes) SetUint32(k, v uint32) {}

type WriterRTP interface {
  Write(p *rtp.Packet) (int, error)
  Attributes() Attributes
}
```

I think this approach gives us some benefits:
  1. By having Attributes type, it allows us to change its implementation in the future, as we can operate on the Get/Set methods, not touching the map directly.
  2. Since we can add methods to Attributes, it's possible to add static types setter/getter, e.g. GetInt32/SetInt32, GetFloat32/SetFloat32, etc. Although we still store these values as interface{}, at least we can hide the type casting inside of Attributes, which I think will make the caller simpler and less error prone.

For the adapter, we can have it as:

```golang
type writerRTPAdapter struct {
  Write func(p *rtp.Packet) (int, error)
  Attributes Attributes
}

func (adapter *writerRTPAdapter) Write(p *rtp.Packet) (int, error) {}
func (adapter *writerRTPAdapter) Attributes() Attributes {}
```

interceptor.go
masterada: I think we won't be able to use it anyway once the dependency is reverted to webrtc -> pkg/interceptor

interceptor.go
masterada: I was thinking about this a lot. Even thought most interceptors either only handle one of outgoing rtp/incoming rtp, or can be split easily into two (eg: NackRemote and NackLocal, TransportCCRemote and TransportCCLocal), there are some that need to process both incoming and outhoing rtp streams.

For example a SenderReport can include ReceptionReports. In order to generate this, it will need to monitor both incoming and outgoing rtp streams. Even with Local/Remote separation, this could be solved by passing the same struct that implements both interfaces as both a Remote and Local interceptor, but that would mean the interceptor is registered twice, and all common methods (RTCP handling, Close) would be called twice.

So in short:
1 interface:
- con: lot of interceptors will not implemented certain methods (but this can be easily achived by embedding the NoOpInterceptor)
- pro: certain interceptors that do need all the methods are easier to implement
Separate Local/Remote:
- con: certain interceptor would be very difficult to implement
- pro: slightly smaller interfaces

Based on this I decided to go with a single interface.

Even with a single interface, it's possible to do the separation (eg: NackRemote vs NackLocal), so people can use only one of them if they want to.

interceptor.go
masterada: I'm very happy the attributes handling got so much attention.

## Purpose of the attributes

The purpose of attributes is to pass metadata belonging to the packet between interceptors, and ideally between interceptor and the user (developer). Some idea for the usage:
- attaching arrival time to packets in the first interceptor (because some interceptors, like a jitter buffer will delay packets, so calling time.Now after those will not return the correct arrival time)
- calculating an unwrapped sequence number once, and passing it
- adding a send time time.Time attribute to packets, for audio/video sync (this can be done using the NTP time from incoming rtcp SenderReports) - this would be most useful if the user could read the attributes too, not just the interceptors
- debugging

## Passing attributes with rtp packets

I'm really not happy with how attributes are passed right now. I think it would be best to attach them to the RTP packet somehow, I just haven't found a good solution to that. Some things I considered:
- adding Attributes field to rtp.Packet
    - this would modify the public api
- adding attributes unexported field to rtp.Packet, and adding a GetAttribute(p *Packet, key interface{}) interface{}, and a similar SetAttribute method to the rtp packet
    - this is ugly
- use a separate global map for attributes (eg: key is ssrc+sequence number, value is attributes)
    - it's hard to know when it's safe to remove attributes for a packet
    - also very ugly
- modifying rtp.Packet to be an interface, and creating a wrapper for this in interceptor packet(see below)
   - this would require a LOT of modification
```
type PacketWithAttribute struct {
    rtp.Packet
    attributes map[interface{}]interface{}
}

func GetPacketAttribute(p rtp.Packet, key interface{}) interface{} {
    pa, ok := p.(*PacketWithAttribute)
    if !ok {
        return nil
    }
    return pa.attributes[key]
} 
```

Any idea is welcome for this!

## Attributes type and key type
The key type is interface{} instead of string, because I planned to use attributes with unexported struct keys, similarly to how context.Context usage is recommended (this would also mean generic SetUint32 and similar methods would be unnecesary, as every attribute would have it's own setter/getter):
```
    type fooKey struct{}

    func SetFoo(attributes map[interface{}]interface{}, value string) {
        attributes[fooKey{}] = value
    }

    func GetFoo(attributes map[interface{}]interface{}) string {
        str, _ := attributes[fooKey{}].(string)
        return str
    }
```

I'm all for using a dedicated `type Attribtes map[interface{}]interface{} though. We can simply use string keys as well if you guys think this is unnecessarily complicated.

## Function vs Writer interface
@at-wat 
I initially chose function types vs interfaces so the user don't need to define a lot of structs. But this adapter function is awesome, never saw this pattern before :)

```
type WriterRTPFunc func(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error)
func (f WriterRTPFunc) Write(p *rtp.Packet, attributes map[interface{}]interface{}) (int, error) {
  return f(p, attribute)
}
```

This is a great idea, I will do this!

## http.ResponseWriter approach
@lherman-cs 
The equivalent of a http response is an rtp packet. http.ResponseWriter writes the body and headers of a single http response. That means sadly we can't use the same aproach for writing attributes in RTPWriter as is used for writing headers in ResponseWriter, because RTPWriter writes multiple rtp packets, each with different attributes, while http.ResponseWriter writes a single http response.

masterada:
Added a commit with these modifications:
- dependency is now webrtc->pkg/interceptor
- added Attribute type
- use Reader/Writer interfaces with adapter functions

for now I added this in a separate commit so it's easier to see the changes.

@Sean-Der please check the contents of movetopionrtp package, I need help to determine where these should be moved to?

interceptor.go
Sean-Der: I am in support of the single interface! I think the pros outweigh the conns in this case!

interceptor.go
OrlandoCo: @masterada Now is clear for me the usage of attributes, but would not be better to use `context.Context`? or there is a special reason to use an `interface` map?

Sean-Der:
@masterada I will take care of moving everything to `pion/rtp` now!

Thanks for dealing with the long review. This PR is nothing short of amazing though. I have really enjoyed seeing the patterns you have done (like `Chain` implementing `Interceptor`) the code has been a joy to review. It will be even more fun when we merge it!

Sean-Der:
@masterada I renamed from `Track` -> `Stream` to match RTP RFCs. Just decoupling this from WebRTC a bit, will make name conflicts a little easier I hope!

Fundamentally I think this is really great. I am going to finish cleaning up `deleteme`, but after that I am approving :) I think we should land your hard work.

I want to write a Tweet thanking you and @at-wat for Interceptors. Do either of you prefer I do a `@` or full names? thanks!

Sean-Der (approved):

masterada:
> I think we shouldn't have anything 'SDP Centric' in the design. Instead StreamInfo should use Attributes an interceptor can then Get/Set values it wants. I think that stops SDP concepts from leaking into the interceptor, and lets other people use it who have no concept of them. If you disagree please push back :)

I think these are very important info, and it would be nearly impossible to implement some interceptors without them:
- RTCPFeedback: this is the most important, without it a nack/trasnprotcc interceptor won't know if it should send feedback. Even for the same peerconnection you can't just send the feedback for all streams, because usually audio streams don't need these feedbacks.
- ClockRate: this is needed to be able to understand the meaning of Timestamp attribute of an rtp packet.
- MimeType: some advanced inteceptors may need to determine if a packet is a video keyframe or not, and they need to know the MimeType for it
- Channels/SDPFmtpLine: i can't think of anything right now that would need these, but I'm pretty sure if we left them out they would be needed at some point (altough we can skip these and add them later if needed)
- HeaderExtensions: these are needed to be able to make sense of the rtp packet header section
- SSRC: absolutely needed, necessary to construct feedback rtcp packets
- PayloadType: don't know, probably needed :)

I think these parameters are neccessary for understanding what an rtp packet is, and as such they belong to rtp more than they belong to SDP (sdp is really just a way to signal these). And you can't make interceptors unaware of rtp/rtcp because they handle rtp/rtcp packets.

Pushed a commit with what I think StreamInfo should contain, please check. Also fixed interceptor binding in remote track, previously params/codec was not setup properly when BindRemoteStream was called.

interceptor.go
masterada: There are 3 reasons why I used map instead of context:
1. In my original proposal context.Context was passed in Read/Write functions, so it could be used for context deadiline/context cancel. But the read/write call in sender/receiver do not support these, so I dropped them to avoid confusion.
2. In the read methods, packets and attributes are returned. As context.Context is usually passed as to functions and not returned, I would find it confusing to use them as return values. It would be even more confusing if in the future we would add context parameter to read for deadlin/cancel, because that would mean a context is used as an argument, and another context is retruned, and the two would have very different purpose.
3. Adding a value to the context needs wrapping, and getting it means a recursive call. This feels like this could cause some performance issues in some cases, however I didn't do any benchmarks so this is just a hunch.

aler9:
@Sean-Der in my opinion, when dealing with RTCP generation, there are two approaches:
* generating RTCP together with RTP, since they both share some initial state values, but this would require decoding and re-encoding all incoming RTP packets
* generating RTCP by looking at existing incoming and outgoing RTPs

The second approach is the one i use in gortsplib and the one proposed here. It allows to implement RTCP feedback, and in particular it would allow to automatically handle keyframe generation, that now is done with a [manual PictureLossIndicator routine](https://github.com/lasernav/Alterego/blob/4337d30817f9296a9ea8c9de00fb52517476dd72/src/brsessionrunner/browser-send-client.go#L281) and could be handled directly by the library, by replicating the behavior of browsers.

What i can say about this approach is that it requires additional informations about the stream, like the ones listed by @masterada,. Since most of these are included into the SDP, i would expose the SDP's MediaDescription directly, instead of creating a new struct{} that allows to avoid SDP concepts but practically exposes the same exact informations, like (by using @masterada list)
* ClockRate --> included in SDP
* MimeType --> included in SDP
* Channels --> included in SDP
* PayloadType --> included in SDP

masterada:
@aler9 originally RTPParameters were passed to bind methods:

https://github.com/pion/webrtc/pull/1504/files#diff-8b596789ae952f46740e3dabb4df2a5391953429f8f84da7f58820feb5838ae3R91
```
// RTPParameters is a list of negotiated codecs and header extensions
//
// https://w3c.github.io/webrtc-pc/#dictionary-rtcrtpparameters-members
type RTPParameters struct {
	HeaderExtensions []RTPHeaderExtensionParameter
	Codecs           []RTPCodecParameters
}
```

These basically contain all these info in a nice parsed form, but there was a dependency issue between packages (we didn't want pkg/interceptor to depend on webrtc).

I think passing in the raw SDP text would unnecessarily complicate things, all these info are already parsed and available at the point of calling Bind* functions (so there is no real performance overhead).

Also, passing MediaDescription alone from SDP might not be enough. Attributes can be session-level as well (see: https://tools.ietf.org/html/rfc4566#section-5.13), for example it's possible to signal an extension (extmap attribubte) on the session level, instead of inside a MediaDescription block.

mediaengine.go
```diff
@@ -405,6 +421,39 @@ func (m *MediaEngine) getCodecsByKind(typ RTPCodecType) []RTPCodecParameters {
 	return nil
 }
 
+func (m *MediaEngine) getRTPParametersByKind(typ RTPCodecType) RTPParameters {
```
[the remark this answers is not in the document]
masterada: > @Sean-Der @masterada My PR address this issue, if we do like this, the interceptor can not be viable for a SFU, since the RTPParameters belongs to the transceivers RTCRTPRecivers and RTCRTPSenders and not to the peer connection mediaengine.

Can you explain this a bit more please? I don't understand it. I'm not familiar with SFU/simulcast, so it's possible I'm missing something.

> Also the uses cases I have seen for interecptor to implement congestion control, would be hard to be a global soultion for all use cases, if we want to implement a NACK, RR,SR we require to centralize a buffer to get the the informatio, but also a way to recover and evict those packets. For twcc is more complex since it requires to get the correct information for all the peer incoming tracks.

RR and SR requires no buffer (they store a couple avarage numbers, and update them with every packet, it's some math magic, so they don't need to store the packets). Receiver side NACK requires only a binary buffer (1 if packet arrived, 0 if not), which even if we want nack to work for 8096 packets (which should be sufficient in most cases), thats 1kb memory per incoming track. TWCC on the receiver side is a bit more tricky, it needs to store arrival times for packets, but it's not much data because once you send out the report you can clear the buffer. I don't know much about TWCC on sender side, but I think it only reacts to reports from the receiver and does not need a buffer either.

The only problematic one is NACK on the sender side, which needs to store the whole packets to be able to resend them. This might be a bit tricky to optimize for SFU.

> And maybe I'm not understanding how this works, seems that a very complex solution, for most congestion controls a plugable Buffer to the tranceivers should be more than enough and a plugable TWCC for the PC.

I already have a working solution for receiver/sender NACK, receiver TWCC, jitter buffer (that's another tricky one). Once this is merged I will quickly add the sender/receiver NACK and Jitter buffer (they are the easiest to move out from our code base), and it will hopefuly make it much clearer how these interceptors are supposed to work.

masterada:
Added a draft pr for nack interceptors to show how these interceptors would work (tested receiver nack, didn't check sender nack yet, it might contain bugs but the concept is visible). See: https://github.com/pion/webrtc/pull/1549/files
@OrlandoCo

mediaengine.go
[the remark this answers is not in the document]
OrlandoCo: @masterada I fully undestand what you mean, we have implemented the congestion controls for ion-sfu, but a plugabble buffer could be a more global solution instead a full interceptor chain, these way you can choose to save packets and use it also as a jitter buffer or use it to get packets to other SFU pubs, or just use metadata to forward nacks. All these can live in the same plugable buffer wihtout creating more interceptors.

I sure know the utility of interceptors, but not sure why we should use it internally on Pion codebase where we have access to the write read methods, pc transceivers etc. With these abastractions seems to make it more complex to write the logic.


> Can you explain this a bit more please? I don't understand it. I'm not familiar with SFU/simulcast, so it's possible I'm missing something.

In a SFU you can attach all the remote publishers to a single PC, but all the publishers may not have the same capabilities, so the capabilities should live in every transceivers and not in the global mediaengine. For example chrome have a Vp8 track with payload 96 and a twcc ext map id=4, Firefox may have the vp8 track with a payload 100 with a twcc ext map id=2, and maybe a old Safari does not even support vp8 or twcc.

So the sender and receiver parameters belongs to the transceivers, to get the correct data and not from the mediaengine. The mediaengine is more like the global configured capabilities you want to support.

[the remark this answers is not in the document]
Sean-Der: These are my motivations for pushing the interceptor design. Not without trade offs, but for pion/webrtc to see wider adoption I thought these were important.

* Easy auditing. We need to win devs trust, so if they can quickly jump into these parts it means a lot. RTP implementation is really contentious so it is important to make it accessible.

* Easy iteration. Developers can easily edit and improve. I want to put these in their own repo eventually so they aren’t even tied to pion/webrtc.

* Easy learning. Lots of users are reading Pion and then taking it back to their C/C++ projects. This is having a positive impact, and is getting people into Go/Pion that would have disregarded it before.

* Easy sharing. I am hoping the Go RTSP server can use it and will bring even more eyes/interest into the project!

I think we can get through all the technical hurdles. The long term pay off is huge (I hope)

masterada:
@Sean-Der if you're ok with the last commit (adding more fields to stream info), i will squash the commit and i think we can merge this.

Sean-Der:
I am in support! 

I am sure we will encounter road bumps along the way, but this feels 100% in the right direction to me. I am going to also move the interceptor to ‘GitHub.com/pion/interceptor’ but I can do that tomorrow. It’s a bit tedious.

You will be the author on the commits for the new repo :)

Sean-Der:
Merged :1st_place_medal: 

@masterada @at-wat Thank you so much for everything you have done! This is a major step forward for pion/webrtc. I also think is going to help bring other projects into Go and RTC space. People can easily use `pion/interceptor` for RIST, RTSP etc...

----

@aler9 Are things in the current format useful for `gosrtplib`? I would love to find a way that we could share this stuff. Maybe not possible, but it feels like bring the Go/RTC community together on this stuff could be really great. I think all the Pion devs could learn a lot from the work you have done :)

----

@OrlandoCo Hopefully we can keep iterating to make this work for ion. I would love to get all the great stuff you have done into `pion/interceptor`. Ion really charged ahead and surpassed pion/webrtc in this space. Hopefully we haven't made any decisions that can stop us from getting all the `ion` goodness into every `pion/webrtc` project :)

PR #1505, opened 2020-11-03
Fix: should only add ice candidates on first media section

tarrencev:
The use of a BUNDLE transport allows the usage of a single set of
   Interactive Connectivity Establishment (ICE)
   [I-D.ietf-ice-rfc5245bis] candidates for the whole BUNDLE group.
   A given BUNDLE address:port MUST only be associated with a single
   BUNDLE group.  If an SDP offer or answer contains multiple BUNDLE
   groups, the procedures in this specification apply to each group
   independently.  All RTP-based bundled media associated with a given
   BUNDLE group belong to a single RTP session [RFC3550].


Sean-Der (approved):

Sean-Der:
@tarrencev did this fix ICEConnectionState flapping for you in Chrome?

tarrencev:
i haven't had a chance to verify yet but will report back!

PR #1506, opened 2020-11-03
Add public helper methods

OrlandoCo:
#### Description

This pr add public methods helpers.

### DIFF
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -30,6 +30,12 @@ func (t *RTPTransceiver) Sender() *RTPSender {
 	return nil
 }
 
+// SetSender sets the RTPSender and Track to current transceiver
+func (t *RTPTransceiver) SetSender(s *RTPSender, track *Track) error {
+	t.setSender(s)
+	return t.setSendingTrack(track)
+}
+
 func (t *RTPTransceiver) setSender(s *RTPSender) {
 	t.sender.Store(s)
 }


peerconnection.go
```diff
@@ -2246,3 +2246,9 @@ func (pc *PeerConnection) setGatherCompleteHandler(handler func()) {
 func (pc *PeerConnection) SCTP() *SCTPTransport {
 	return pc.sctpTransport
 }
+
+// DTLS returns the currently-configured *DTLSTransport or nil
+// if one has not yet been configured
+func (pc *PeerConnection) DTLS() *DTLSTransport {
```
Sean-Der, line 2252: I think you can already get this via `DataChannel` or `Transceiver` let me check!

Does this exist in the webrtc-pc API?

rtptransceiver.go
```diff
@@ -30,6 +30,12 @@ func (t *RTPTransceiver) Sender() *RTPSender {
 	return nil
 }
 
+// SetSender sets the RTPSender and Track to current transceiver
+func (t *RTPTransceiver) SetSender(s *RTPSender, track *Track) error {
```
Sean-Der, line 34: I think you want `ReplaceTrack`. This isn't easily possible with the API we have today :(

With the PR I am landing you should have this with-in the week though!

peerconnection.go
Sean-Der: * https://pkg.go.dev/github.com/pion/webrtc/v3#RTPReceiver.Transport
* https://pkg.go.dev/github.com/pion/webrtc/v3#RTPSender.Transport
* https://pkg.go.dev/github.com/pion/webrtc/v3#SCTPTransport.Transport (The PeerConnection has a global SCTPTransport)

rtptransceiver.go
OrlandoCo: @Sean-Der  We are doing this PR to reuse inactive transceivers, since atm in the SFU, when a track is removed the transceiver is set to inactive, and left there, so after some joins and rejoins we end up with a gigantic sdp full of inactive transceivers. These helper is to allow reuse the inactive sender transceiver.

Sean-Der: @OrlandoCo oh nice, thanks makes sense to me!

If we ever come back to this I looked through [AddTrack](https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack) and friends and looks like we can't accomplish this with any standard APIs :(

Sean-Der (approved):

rtptransceiver.go
OrlandoCo: @Sean-Der An option should be not setting the sender to `nil` and replace the track.

tarrencev (approved):

PR #1515, opened 2020-11-11
Add set custom ext map callback

OrlandoCo:
#### Description

This PR will allow to se custom ext map headers per mid for transceivers, this way it support offering ext maps if not previously offer was given, it also allow to set map from different browsers if they are ordered different.

If the callback is not set it will fallback to default behaviour.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2117,7 +2117,14 @@ func (pc *PeerConnection) generateUnmatchedSDP(transceivers []*RTPTransceiver, u
 		return nil, err
 	}
 
-	return populateSDP(d, isPlanB, dtlsFingerprints, pc.api.settingEngine.sdpMediaLevelFingerprints, pc.api.settingEngine.candidates.ICELite, pc.api.mediaEngine, connectionRoleFromDtlsRole(defaultDtlsRoleOffer), candidates, iceParams, mediaSections, pc.ICEGatheringState(), pc.api.settingEngine.getSDPExtensions())
+	var unmatchedSDPMap map[SDPSectionType][]sdp.ExtMap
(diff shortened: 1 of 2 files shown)


Sean-Der:
@OrlandoCo I am happy to merge this if you need this today!

#1517 provides an API that is easy to use for this exact use case `(m *MediaEngine) GetHeaderExtensionID(extension RTPHeaderExtensionCapability) (val int, ok bool)`

Then in your code you can do

```
    midExtensionID, ok := pc.api.mediaEngine.GetHeaderExtensionID(RTPHeaderExtensionCapability{sdp.SDESMidURI})
    if !ok {
        panic("extension not available")
    }
```

OrlandoCo:
> @OrlandoCo I am happy to merge this if you need this today!
> 
> #1517 provides an API that is easy to use for this exact use case `(m *MediaEngine) GetHeaderExtensionID(extension RTPHeaderExtensionCapability) (val int, ok bool)`
> 
> Then in your code you can do
> 
> ```
>     midExtensionID, ok := pc.api.mediaEngine.GetHeaderExtensionID(RTPHeaderExtensionCapability{sdp.SDESMidURI})
>     if !ok {
>         panic("extension not available")
>     }
> ```

@Sean-Der Nice, let me look at it, we can leave this PR open for now, we are using the branch import atm, So if the new api works we can use it instead this workaround.

Sean-Der:
@OrlandoCo mind if I close? I will make sure not to delete the branch!

OrlandoCo:
> @OrlandoCo mind if I close? I will make sure not to delete the branch!

Sure not problem, this branch is still required to ion-sfu, but will removed once new track api is stable

PR #1517, opened 2020-11-11
New Track API

Sean-Der:
The Pion WebRTC API has been dramatically redesigned. The design docs
are located here [0]

You can also read the release notes [1] on how to migrate your
application.

[0] https://github.com/pion/webrtc-v3-design/issues/33
[1] https://github.com/pion/webrtc/wiki/Release-WebRTC@v3.0.0


Sean-Der:
This is still a WIP. I really think I will be done tomorrow, if anyone is interested I would love help!

* [ ] Finish TestMediaEngineRemoteDescription
      - This test reads the remote description and updates our local PayloadTypes to their proper values
* [ ] Finish all TODOs
* [ ] Review examples. Some now allow all codecs instead of just VP8/Opus. MediaEngine needs to come back in them
* [ ] Write `TrackLocal` tests.

mediaengine.go
```diff
@@ -432,3 +293,44 @@ func (m *MediaEngine) collectStats(collector *statsReportCollector) {
...
+			m.negotiatedVideo = true
+		default:
+			continue
+		}
+
+		// * convert media.Attributes in []RTPCodecParameters
+		// * Update media.videoCodecs (or media.audioCodecs)
+		//   - set negotiated=true
+		//   - set the proper payloadtype
+		//   - set the proper fmtp
```
masterada, line 315: i think it should also:
- set proper extension ids, and remove extensions not present in sdp
- remove all feedback not present in the sdp
- remove all codecs not present in sdp

this method is strongly related to addTransceiverSDP, as it will determine whether the returned sdp is correct or not (the answer sdp must contain the same payload type/extension id/feedback mechanisms)

mediaengine.go
```diff
@@ -432,3 +293,44 @@ func (m *MediaEngine) collectStats(collector *statsReportCollector) {
...
+	}
+	return nil
+}
+
+// negotiatedCodecsForType and negotiatedCodecsforType both return a properly
+// filtered view of the MediaEngine.
+// Note if negotiation hasn't taken place yet all values are assumed enabled
+
+func (m *MediaEngine) negotiatedCodecsForType(typ RTPCodecType) []RTPCodecParameters {
```
masterada, line 324: I think this method should be removed, because if MediaEngine can be reused, then it can't store PeerConnection specific information such as negatiation status and PeerConnection specific payload types. Instead updateFromRemoteDescription could return an RTPParameters object, and it could be stored for the peerconnection and passed to RTPSender/etc. when needed.

masterada (commented):
This is looks very promising :) I really like the direction MediaEngine is taking. I wrote together my ideas/suggestions about handling payload type/extension ids/feedbacks in media engine/answer sdp generation/local tracks. Take these with a grain of salt, this is a very big change and there are probably cases I did not consider.

mediaengine.go
```diff
@@ -4,419 +4,280 @@ package webrtc
...
-func (m *MediaEngine) getCodecSDP(sdpCodec sdp.Codec) (*RTPCodec, error) {
-	for _, codec := range m.codecs {
-		if strings.EqualFold(codec.Name, sdpCodec.Name) &&
-			codec.ClockRate == sdpCodec.ClockRate &&
-			(sdpCodec.EncodingParameters == "" ||
-				strconv.Itoa(int(codec.Channels)) == sdpCodec.EncodingParameters) &&
-			codec.SDPFmtpLine == sdpCodec.Fmtp { // pion/webrtc#43
-			return codec, nil
+	// Default Pion Audio Header Extensions
+	for _, extension := range []string{
```
OrlandoCo, line 70: I think we should remove the following headers from default:

- "http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01"
- "urn:ietf:params:rtp-hdrext:sdes:mid"
- "urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id"
- "urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id"

If pion users do not implement the correct behavior in rtp packets for above headers, it may lead to errors that are hard to debug, like twcc will drop the bitrate to zero it's not implemented.

Sean-Der: :+1: nice, my mistake! Will pull those out.

Would it be ok to leave mid/rtp-stream-id/repaired-stream-id? Right now our Simulcast examples does REMB + those headers and seems ok. It would just be nice if users didn't have to touch headers when doing Simulcast.

track_local.go
```diff
@@ -0,0 +1,62 @@
...
+	// WriteRTP encrypts a RTP packet and writes to the connection
+	WriteRTP(header *rtp.Header, payload []byte) (int, error)
+
+	// Write encrypts and writes a full RTP packet
+	Write(b []byte) (int, error)
+}
+
+// TrackLocalContext is the Context passed when a TrackLocal has been Binded/Unbinded from a PeerConnection
+type TrackLocalContext struct {
+	codecs      []RTPCodecParameters
```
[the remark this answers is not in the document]
Sean-Der: :+1: will make sure that happens, nice catch!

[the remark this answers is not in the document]
Sean-Der: I could even just pass the `MediaEngine`? Then give users nice APIs to pull details. Which do you think is better?

mediaengine.go
```diff
@@ -432,3 +293,44 @@ func (m *MediaEngine) collectStats(collector *statsReportCollector) {
 		collector.Collect(stats.ID, stats)
 	}
 }
+
+// TODO Sean-Der, everything below is not finished
+
+// Update the MediaEngine from a remote description
+func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) error {
```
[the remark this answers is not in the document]
Sean-Der: My vote is that `MediaEngine` shouldn't be shared. That comment just hasn't been updated yet.

As a user I think it makes sense to set a MediaEngine per PeerConnection (but share a SettingEngine). Then you query each MediaEngine to get details needed when crafting things.

mediaengine.go
Sean-Der: 👍 hopefully will finish this all tonight. I can't wait to merge this and then your Interceptor PR and just tag `v3.0.0`

We can follow up quickly with `v3.0.1`, but this was just so much more work then I anticipated.

mediaengine.go
OrlandoCo: Simulcast will require the sdes:mid and rtp-stream-id, the problem with those, is that if they are always negotiated and pion is the media producer if users dont add correct sdes:mid in the header they may encounter cryptic erros in Chrome at least.

mid/rtp-stream-id/repaired-stream-id is for simulcast rtx (rrid) as far as I know, I think this should be optional also. Or not sure how rtx will be handled by Pion.

track_local.go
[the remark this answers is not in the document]
masterada: Sounds good either way to me, MediaEngine and RTPParameters contain basically the same data.

mediaengine.go
masterada: This doesn't matter if MediaEngine is not shared.

mediaengine.go
masterada: One more thing that came to mind is renegotiation: what happens, if we support opus codec, but there is no audio track in the original sdp, so we remove it from MediaEngine, but then there is a renegotiation and an opus track is added (is that possbile? I have never done re-negotiation). Maybe instead of setting negotiated: true parameters, you could save the negotiated codecs in a separate slice:
```
type MediaEngine struct {
	videoCodecs           []RTPCodecParameters
	audioCodecs           []RTPCodecParameters
	negotiatedVideoCodecs []RTPCodecParameters
	negotiatedAudioCodecs []RTPCodecParameters
}
```

Sean-Der: That is a much better design! Before I added `negotiated` as an entry for Headers/Codecs.

I am going to instead have a seperate list for negotiated Headers/Codecs. It will be much easier to understand the state of everythinhg.

jeremija:
This looks really nice! From what I can tell this looks good to me, I'll try to find some time to test the new functionality. I'm excited about finally being able to get rid of the need to always start negotiating on the server side.

I tried to help out by writing the tests for `TrackLocal` and I got a little stuck at this bullet point:

```
// * Offer fail when Answerer doesn't support a codec
```

I am not able to get it to fail - I'm getting no errors. The only thing I see is that the SDP does not contain the required `mid` when the codecs don't match. Is this the expected behavior?

track_local_static.go
```diff
@@ -0,0 +1,235 @@
...
+type trackBinding struct {
+	id          string
+	ssrc        SSRC
+	payloadType PayloadType
+	writeStream TrackLocalWriter
+}
+
+// TrackLocalStaticRTP  is a TrackLocal that has a pre-set codec and accepts RTP Packets.
+// If you wish to send a media.Sample use TrackLocalStaticSample
+type TrackLocalStaticRTP struct {
```
at-wat, line 29: Maybe it's better to be a subpackage?

track_remote.go
```diff
@@ -0,0 +1,161 @@
...
+package webrtc
+
+import (
+	"sync"
+
+	"github.com/pion/rtp"
+)
+
+// TrackRemote represents a single inbound source of media
+type TrackRemote struct {
```
at-wat, line 12: I think it's better to define an interface for supporting custom remote track.

Sean-Der: :+1: will do that!

track_local_static.go
Sean-Der: That was really hard because of testing. I had to move a lot of tests to the subpackage, but then I still wanted to touch WebRTC internals.

We can do this if you feel strongly though!

mediaengine.go
```diff
@@ -4,431 +4,400 @@ package webrtc
...
-// A MediaEngine populated by PopulateFromSDP should be used only for a single session.
-func (m *MediaEngine) PopulateFromSDP(sd SessionDescription) error {
-	sdp := sdp.SessionDescription{}
-	if err := sdp.Unmarshal([]byte(sd.SDP)); err != nil {
-		return err
+func (m *MediaEngine) RegisterDefaultCodecs() error {
+	// Default Pion Audio Codecs
+	for _, codec := range []RTPCodecParameters{
+		{
+			RTPCodecCapability: RTPCodecCapability{mimeTypeOpus, 48000, 2, "minptime=10;useinbandfec=1", []RTCPFeedback{{"transport-cc", ""}}},
```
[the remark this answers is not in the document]
Sean-Der: nope! will take all those out

Sean-Der:
> This looks really nice! From what I can tell this looks good to me, I'll try to find some time to test the new functionality. I'm excited about finally being able to get rid of the need to always start negotiating on the server side.
> 
> I tried to help out by writing the tests for `TrackLocal` and I got a little stuck at this bullet point:
> 
> ```
> // * Offer fail when Answerer doesn't support a codec
> ```
> 
> I am not able to get it to fail - I'm getting no errors. The only thing I see is that the SDP does not contain the required `mid` when the codecs don't match. Is this the expected behavior?
> 
> For more details, see the last test in `track_local_static_test.go` in `jeremija/track-local-tests` branch.

That would be so amazing! Thanks for helping :) The issue is the errors are async is the issue, `Bind` is called in a goroutine so they don't get returned by `SetRemoteDescription`

We have two choices
* Add any additional 'validate codec' to a Track
* Call `Bind` earlier?

mediaengine.go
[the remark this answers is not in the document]
Sean-Der: done!

track_remote.go
Sean-Der: Mind if I leave out of this PR? This is a really big change again, I really want to land this weekend. 

I am falling behind on other things :(

rtpcodec.go
```diff
@@ -0,0 +1,82 @@
...
+	negotiated bool
+
+	statsID string
+}
+
+// RTPParameters is a list of supported codecs and header extensions
+//
+// https://w3c.github.io/webrtc-pc/#dom-rtcrtpparameters-codecs
+type RTPParameters struct {
+	HeaderExtensions []RTPHeaderExtensionCapability
```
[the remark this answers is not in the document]
Sean-Der: Fixed!

mediaengine.go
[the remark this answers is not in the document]
Sean-Der: Fixed!

mediaengine.go
Sean-Der: Done!

mediaengine.go
Sean-Der: Pulled everything transport-cc out!

I confirmed and mid/rtp-stream-id doesn't effect anything, so leaving in for now!

Sean-Der:
IT'S GREEN :fire:

I also tested all the examples. I was able to broadcast from Chrome and then view in Chrome/FireFox/Safari (that felt awesome)

Personally I would like to merge this, and then get help from others going forward. I would really love to get @jeremija involved, I have been working on this for two weeks so I feel pretty done with this PR :p

OrlandoCo (approved):

track_local_static.go
at-wat: It sounds ok to be here, but as a general functionality in the main package, parameters like MTU should be able to be configured.

jeremija:
@Sean-Der let me know if you think we should take this discussion elsewhere since I think it's only indirectly related to this PR 🙂 

> We have two choices
>
>  - Add any additional 'validate codec' to a Track
> - Call Bind earlier?

I _think_ we should be able to call `Bind` earlier. I'll try to move it out of the `RTPSender::Send` and see if I find any issues.

Side question. Why do we need arguments to `RTPSender::Send`? The only place it's called from is in `startRTPSenders` and all the parameters passed to it are read from the same instance of `RTPSender`:

```go
			err := transceiver.Sender().Send(RTPSendParameters{
				Encodings: RTPEncodingParameters{
					RTPCodingParameters{
						SSRC:        transceiver.Sender().ssrc,
						PayloadType: transceiver.Sender().payloadType,
					},
				},
			})
```

Why not make it `err := transceiver.Sender().Send()`, since the `Send` method should already have access to `RTPSender::ssrc` and `RTPSender::payloadType`? Is there something I'm missing here?

masterada:
> We can also then rebase @masterada PR and merge that quickly. I would love to release `v3.0.0` before thanksgiving with NACKs (or something similar) and this. It would be a massive improvement and pull a lot of people into Pion I think :)

I want to try to make the interceptor design a bit closed to how TrackLocal works, maybe even use TrackLocalContext somehow, or something similar to that. Will have time for this on Tuesday afternoon (Europe time).

Sean-Der:
@masterada awesome, sounds great to me! Since it is a new API we can merge + iterate without having to worry about breaking users. As long as we design an interface and keep it stable everything else can change internally :)

Really excited that multiple people can be working on different interceptors, it is going to make this go so fast.

Sean-Der:
@jeremija 

Awesome, if we move the `Bind` that solves everything! I think we just need to make a pseudo `rtpWriteStream` that internally throws away packets until SRTP is ready.

It would be nice if the API of `RTPSender` doesn't change, it was split up to match the ORTC API. If not possible that is ok!

PR #1527, opened 2020-11-16
Implement RTPSender.ReplaceTrack

### LINKED ISSUE #1020: Support RTPSender.ReplaceTrack()

Sean-Der:
Resolves #1020

### DIFF
--- a/rtpsender.go
+++ b/rtpsender.go
@@ -13,13 +13,16 @@ import (
 
 // RTPSender allows an application to control how a given Track is encoded and transmitted to a remote peer
 type RTPSender struct {
-	track          TrackLocal
+	track TrackLocal
+
 	rtcpReadStream *srtp.ReadStreamSRTCP
+	rtpWriteStream *srtp.WriteStreamSRTP
 
 	transport *DTLSTransport
 
 	payloadType PayloadType
 	ssrc        SSRC
+	codec       RTPCodecParameters
 
 	// nolint:godox
 	// TODO(sgotti) remove this when in future we'll avoid replacing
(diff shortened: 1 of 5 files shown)


rtpsender.go
```diff
@@ -86,10 +89,35 @@ func (r *RTPSender) Track() TrackLocal {
...
+			id:          r.id,
+			ssrc:        r.ssrc,
+			writeStream: r.rtpWriteStream,
+		}); err != nil {
+			return err
+		}
+	}
+
 	r.track = track
+	if !r.hasSent() || r.track == nil {
```
OrlandoCo, line 110: Should we not implement more constraints?, according to spec we should check that track does not difer from current negotiated, also check special cases like audio not being negotiated as 1 channel and replace with 2 channel

Sean-Der: Yea! Subsequent calls to `Bind` only can use the same codec as the first call https://github.com/pion/webrtc/pull/1527/files#diff-1af62154d0fa7a1013519d5059b0bb1b13f07f25aa4ec4a70bdbc0007e3fd652R100 is a test showing it

rtpsender.go
```diff
@@ -86,10 +89,35 @@ func (r *RTPSender) Track() TrackLocal {
...
+		if err := r.track.Unbind(TrackLocalContext{
+			id:          r.id,
+			ssrc:        r.ssrc,
+			writeStream: r.rtpWriteStream,
+		}); err != nil {
+			return err
+		}
+	}
+
 	r.track = track
```
lherman-cs, line 109: I think r.track should be set when `Bind` returns successfully. Otherwise, the next `ReplaceTrack` will `Unbind` the track that failed to `Bind`.

track_local.go
```diff
@@ -49,7 +49,7 @@ type TrackLocal interface {
 	// Bind should implement the way how the media data flows from the Track to the PeerConnection
 	// This will be called internally after signaling is complete and the list of available
 	// codecs has been determined
-	Bind(TrackLocalContext) error
+	Bind(TrackLocalContext) (RTPCodecParameters, error)
```
lherman-cs, line 52: Since the codec selection is done within SDP negotiation, is it possible to have multiple codecs in the context? When there are multiple codecs, does the order represent a priority?

rtpsender.go
Sean-Der: Fixed!

track_local.go
Sean-Der: This returns the final selection by the `TrackLocal`. This locks in the RTPSender to this selected codec.

You can this in `Send` where `r.codec` is set. Every subsequent call to `Bind` you only get once choice, the codec you selected before.

The first call to `Bind` you can choose any codec you want, you get the entire list. The list will be presented in the order chosen by the remote. I didn't design it that way on purpose though, just was easiest parsing wise.

Sean-Der:
@lherman-cs @OrlandoCo @at-wat If you think this is an un-needed feature we can table it also!

OrlandoCo:
> @lherman-cs @OrlandoCo @at-wat If you think this is an un-needed feature we can table it also!

Well I think it can stay,  maybe add some warning that should be taken in account like it requires to sync the tracks to keep the SN and Timestamp, also waiting for a keyframe if it's video, so dont get many issues.

But with the rtp interceptor it maybe a nice solution for simulcast.

lherman-cs (approved):
LGTM

PR #1534, opened 2020-11-17
Add GetParameters to RTPReceivers and RTPSenders

OrlandoCo:
#### Description

Add GetParameters to RTPReceivers and RTPSenders, allowing setting and getting ext map at transceiver level.


Sean-Der:
@OrlandoCo this LGTM! It is just extending existing W3C structures so I see no problem with it at all :) 

I need to read more to understand the distinction, but would be nice if `updateFromRemoteDescription` and the Plan-B stuff could be merged.

OrlandoCo:
@Sean-Der This PR is ready for review, with this PR we attempt to handle all the media info in transceivers, these way we can handle in single PC peers with different codec capabilities, diferent payloads and ext headers.

Sean-Der:
Moved to #1554

PR #1536, opened 2020-11-17
Fix the default value for previousTimestamp.

### LINKED ISSUE #1537: The playback time of the exported Ogg audio file will be incorrect.
### Your environment.
  - Version: *Release or SHA*
  - Browser: *include version*
  - Other Information - *stacktraces, related issues, suggestions how to fix, links for us to have context*

### What did you expect?
If you see the PR(#1536 ) and there is no problem, please merge.

### What happened?
When trying to record the microphone input of a user connected to webrtc using oggwriter.
An audio file with a playback time longer than the recorded time has been created. It doesn't actually play that time, and the header seems to be broken.

The default value for previousTimestamp is 1 instead of 0. The playback time of the ogg file created by it is supposed to be strange.
As a result, the value of previousGranulePosition starts with a large number, and the playback time becomes several hours or more.

I've created a modified PR so I'd be happy if you could review it for correctness.

BambooTuna:
#### Description
The default value for previousTimestamp is 1 instead of 0. The playback time of the ogg file created by it is supposed to be strange.

The following is where the initial value is substituted
https://github.com/BambooTuna/webrtc/blob/512b2c5fdbd8fa8734ae8f56962c007632e222b9/pkg/media/oggwriter/oggwriter.go#L73

Is this how to issue a modified PR?

#### Reference issue
Fixes #1537

### DIFF
--- a/pkg/media/oggwriter/oggwriter.go
+++ b/pkg/media/oggwriter/oggwriter.go
@@ -182,7 +182,7 @@ func (i *OggWriter) WriteRTP(packet *rtp.Packet) error {
 	payload := opusPacket.Payload[0:]
 
 	// Should be equivalent to sampleRate * duration
-	if i.previousTimestamp != 0 {
+	if i.previousTimestamp != 1 {
 		increment := packet.Timestamp - i.previousTimestamp
 		i.previousGranulePosition += uint64(increment)
 	}


Sean-Der:
Merged! 1ffa87eb0733e7bfc6a50dcd2b2bac7d93c380ad

Thank you so much. If you are able it would be really great if you could write a test also :) thanks!

PR #1540, opened 2020-11-18
Use rtp.Packet instead of *rtp.Packet in APIs

Sean-Der:
Reduce a sharp edge in the public API. Before we would keep a reference
internally. Users would be suprised when values on the passed
*rtp.Packet would change.

TODO write Benchmark

### DIFF
--- a/track_local_static.go
+++ b/track_local_static.go
@@ -107,10 +107,12 @@ func (s *TrackLocalStaticRTP) WriteRTP(p *rtp.Packet) error {
 	defer s.mu.RUnlock()
 
 	writeErrs := []error{}
+	outboundPacket := *p
+
 	for _, b := range s.bindings {
-		p.Header.SSRC = uint32(b.ssrc)
-		p.Header.PayloadType = uint8(b.payloadType)
-		if _, err := b.writeStream.WriteRTP(&p.Header, p.Payload); err != nil {
+		outboundPacket.Header.SSRC = uint32(b.ssrc)
+		outboundPacket.Header.PayloadType = uint8(b.payloadType)
+		if _, err := b.writeStream.WriteRTP(&outboundPacket.Header, outboundPacket.Payload); err != nil {
 			writeErrs = append(writeErrs, err)
 		}
 	}
--- a/track_local_static_test.go
+++ b/track_local_static_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pion/rtp"
 	"github.com/pion/transport/test"
 	"github.com/stretchr/testify/assert"
 )
@@ -161,4 +162,34 @@ func Test_TrackLocalStatic_PayloadType(t *testing.T) {
 
 	assert.NoError(t, offerer.Close())
 	assert.NoError(t, answerer.Close())
+}
+
+// Assert that writing to a Track doesn't modify the input
+// Even though we can pass a pointer we shouldn't modify the incoming value
+func Test_TrackLocalStatic_Mutate_Input(t *testing.T) {
+	lim := test.TimeOut(time.Second * 30)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	pcOffer, pcAnswer, err := newPair()
+	assert.NoError(t, err)
+
+	vp8Writer, err := NewTrackLocalStaticRTP(RTPCodecCapability{MimeType: "video/vp8"}, "video", "pion")
+	assert.NoError(t, err)
+
+	_, err = pcOffer.AddTrack(vp8Writer)
+	assert.NoError(t, err)
+
+	assert.NoError(t, signalPair(pcOffer, pcAnswer))
+
+	pkt := &rtp.Packet{Header: rtp.Header{SSRC: 1, PayloadType: 1}}
+	assert.NoError(t, vp8Writer.WriteRTP(pkt))
+
+	assert.Equal(t, pkt.Header.SSRC, uint32(1))
+	assert.Equal(t, pkt.Header.PayloadType, uint8(1))
+
+	assert.NoError(t, pcOffer.Close())
+	assert.NoError(t, pcAnswer.Close())
 }


OrlandoCo (approved):

Sean-Der:
@at-wat @OrlandoCo @bshi (anyone else who is good with performance stuff?) What do you think of this change, here are my results so far.

The goal here is to reduce sharp edges and some motivation from https://medium.com/@vCabbage/go-are-pointers-a-performance-optimization-a95840d3ef85

```
# Copy Based
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          75387             15420 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76329             14994 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          72397             15292 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          75758             15383 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76332             16396 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          79160             15475 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          62388             19728 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          77582             15672 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76843             17727 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          75664             14963 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76006             15033 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          78680             15036 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          61647             16346 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76102             15309 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76863             15439 ns/op



# Pointer Based
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          53791             19168 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76670             15054 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76576             14956 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76573             15313 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          79381             15019 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          76624             15312 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          75909             14949 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          79003             16044 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          78114             14904 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          77294             15068 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          81478             16868 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          78312             15229 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          78694             14963 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          75019             15080 ns/op
Benchmark_TrackLocalStaticRTP_SendOnly_WriteRTP-4          71244             15425 ns/op
```

Sean-Der:
@OrlandoCo one issue is that `pion/srtp` expects a *rtp.Packet, so we probably should address there also!

I am tempted to say `/v4` I just really want to get `Per Transceiver Configuration` done and this is starting to seem like a larger task

OrlandoCo:
> @OrlandoCo one issue is that `pion/srtp` expects a *rtp.Packet, so we probably should address there also!
> 
> I am tempted to say `/v4` I just really want to get `Per Transceiver Configuration` done and this is starting to seem like a larger task

The main concern was that now by default the `TrackLocalStaticRTP` modify the header, not sure if for some implementations it may cause weird bugs if the depend of the original values of the packets. We can just for now just change the behaviour in the writer to make a copy there, and later make the change. But reduce allocations in hot paths like writing packets would be great!

bshi:
Using https://godoc.org/golang.org/x/perf/cmd/benchstat

```
$ benchstat /tmp/copy.txt /tmp/ptr.txt
name                                      old time/op  new time/op  delta
_TrackLocalStaticRTP_SendOnly_WriteRTP-4  15.6µs ±14%  15.2µs ± 6%  -2.74%  (p=0.044 n=14+13)
```

On the surface seems like the safety improvement comes with a slight improvement in timing benchmarks (just barely statistically significant P<0.05).  It would be useful to see how this impacts allocations via `b.ReportAllocs()` in your benchmark function or `-benchmem` flag.

at-wat (commented):
> Before we would keep a reference internally. Users would be suprised when values on the passed *rtp.Packet would change.

The contents stored as slices are not copied by passing as a value.
https://github.com/pion/rtp/blob/c8b3d3ec7ff956797d8d29c5ab1cfb054c9e4028/packet.go#L17-L38
I think it's more confusing than passing as pointer since it's partially copied and partially shared. (both for users and developers)
IMO, it would be better to pass it as a pointer and clone whole packet including its slice members where copy is needed.

Sean-Der:
Ah I completely missed your point originally @OrlandoCo I simplified this a lot.

I opened https://github.com/pion/webrtc/issues/1560 and we should do a deep dive into performance/API of Pion across the board. We just haven't been at a point of having enough features where I can focus on this. Anymore it seems like performance has been the thing most users have reached out about (after Interceptors)

PR #1543, opened 2020-11-19
Fix race checkNegotiationNeeded

tekig:
#### Description
Deleting a track is not thread-safe.
It is confusing when you can come across a transceiver that has a direction but no sender.

#### Reference issue
Issue #1494

#### Similar pull request
https://github.com/pion/webrtc/pull/1496


OrlandoCo:
I think that the main issue here is setting the track to nil, instead of setting the track to nil, we just stop the sender and consider that is the transceiver is `SENDONLY` set it to `INACTIVE` if it's `SENDRECV` set it to `RECVONLY`

The issue here is that some browsers need to know the SSRC from the inactive track.

tekig:
Sorry for the spam. If I understood correctly. You are suggesting not to set the sender value to zero, but the spec says to do so. I think that this can lead to other errors, I am not an expert in this, so I will write as in the specification. You can create a separate request.

OrlandoCo:
> Sorry for the spam. If I understood correctly. You are suggesting not to set the sender value to zero, but the spec says to do so. I think that this can lead to other errors, I am not an expert in this, so I will write as in the specification. You can create a separate request.

You are confusing terms, the spec says that track should set to `null` and the `RTCRTPSender` should remain, my suggest was to fix an issue with negotiated tracks setting to inactive without ssrc (thought that may be related). Your comment say that you find a transceiver with direction but no sender, it's difficult to see that scenario since when you create a transciever a sender is created, the only way to not get a sender is during a remote negotiation where transceiver is set to `recvonly` so not sender is required.

What are you doing here is just holding the lock more time, but that is not warranty that the transceivers can not being modified somewhere else, since other path can modify transceivers. Thats why during negotiation, the spec says to create a shallow copy and work on it, if another change is done another negotiation is queued.

Sean-Der:
@OrlandoCo @tekig I will be honest I don't know this stuff well enough (all the rules around Sender/Receiver/Transceiver creation)

Do we need to radically re-design how pion/webrtc internally works to make this easier? I have heard that people find it frustrating to work on with the complexity. Maybe we need to convert it to a state machine, and put that in its own package?

You can batch a bunch of requested changes, and they either succeed or fail. We could also decouple SDP from it. Just food for thought, I am in total support of w/e is 'the right way'

tekig:
I merged the branches wrong. The idea is that we copy the link to the sender and track. He will not be able to change the value of this variable and will always be available.

tekig:
@Sean-Der the problem is still relevant, as I see in the code. I don't know how to write testing to cover all changes, but it works if you use `time.Sleep`.

tekig:
The problem came up when removing a track that was [not thread](https://github.com/pion/webrtc/issues/1574) safe that the [PR closed](https://github.com/pion/webrtc/pull/1737). Now sender cannot become unexpectedly nil.

PR #1545, opened 2020-11-22
Compute the track's MSID to compare with the one in the SDP.

s4y:
#### Description
I noticed that, when using the `OnNegotiationNeeded` callback, I got called back continuously, every time a negotiation finished. As far as I can tell, this is because `checkNegotiationNeeded()` was comparing an MSID against `track.StreamID()`, which isn't quite an MSID, so it never matched.

### DIFF
--- a/.github/.ci.conf
+++ b/.github/.ci.conf
@@ -1,4 +1,4 @@
 GO_JS_WASM_EXEC=${PWD}/test-wasm/go_js_wasm_exec
 TEST_EXTRA_ARGS="-tags quic"
 GOLANGCI_LINT_EXRA_ARGS="--build-tags quic"
-EXCLUDED_CONTRIBUTORS=('Josh Bleecher Snyder')
+EXCLUDED_CONTRIBUTORS=('Josh Bleecher Snyder' 'Sidney San Martín')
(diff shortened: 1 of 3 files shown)


s4y:
Oof, I'm fighting with the way `assert-contributors.sh` is parsing my name. Still experimenting.

s4y:
Urm. Sorry for the spam. As far as I can tell, this had to do with how the GitHub action runner is handling combining characters (i.e. "San Martín" vs "San Martín"). I'll fix my commits to use the standard LATIN SMALL LETTER I WITH ACUTE going forward.

Sean-Der (approved):

Sean-Der:
Nice! This looks great to me @s4y Approved.

I think you can use `Msid` instead, beyond that looks perfect.

s4y:
Hm, as far as I can tell, `TrackLocal` doesn't have an `Msid()` method, only `TrackRemote`.

Sean-Der:
Ah my mistake! @s4y merged, thanks for updating to `/v3` and testing out the new API :)

s4y:
Of course, and I'll definitely keep doing so.

Issue #1526, opened 2020-11-15
Track V3 Tasks Remaining


Sean-Der:
These are all the things we need to finish with V3 Track API 

* [x] Re-enable `TestPeerConnection_Media_Closed`
* [x] Rework API so that `Bind` is called in `SetRemoteDescription`
* [x] Write the following tests for new Track API
    - [x] Writing to a un-added Track
    - [x] Write to a track on Closed PeerConnection
    - [x] Write to a track on a disconnected PeerConnection
    - [x] Write to a track with one bad PeerConnection
    - [x] Offer fail when Answerer doesn't support a codec
    - [x] Answer fail when Offerer doesn't support a codec
    - [x] Does sending use the proper Payload type
* [ ] Make `TrackRemote` an interface
* [x] Review Public API for sharp edges.
    - [x] `MimeType` is currently case sensitive. Users will need to do `video/VP8` and `video/vp8`, anything we can do to help?
    - [x] Are we all clear to do all the use cases for Interceptor (NACK, FEC, Congestion Control, Stats)

Sean-Der:
@at-wat @masterada @jeremija ^ following up on the other PR. Add anything else you think is important!

If you agree I would like to merge the other PR and then begin the process of getting this done for `v3.0.0`. We are so close, and really want more people involved :)

at-wat:
> MimeType is currently case sensitive. Users will need to do `video/VP8` and `video/vp8`, anything we can do to help?

https://tools.ietf.org/html/rfc2045#section-2

> All numeric and octet values are given in decimal notation in this set of documents. All media type values, subtype values, and parameter names as defined are case-insensitive.  However, parameter values are case-sensitive unless otherwise specified for the specific parameter.

at-wat:
I think as `TrackLocalStaticRTP` is in the main package, parameters like MTU is better to be configurable by users.
https://github.com/pion/webrtc/pull/1517#discussion_r523729641

Sean-Der:
@at-wat That sounds good to me! If you want to start the `...opt` or doing a configuration struct I am happy with either.

Sean-Der:
@jeremija @OrlandoCo Any interest in splitting/taking on any of these tasks? I am happy to do all of them, just don't want to duplicate work :)

Sean-Der:
@at-wat I tried a few different ways (and failed) to make `TrackRemote` an interface. I might be missing something, but I don't see the value.

[...]

PR #1266, opened 2020-06-20
Add Track.ReadRTPContext and Track.ReadContext

josharian:
To allow easier management of deadlines.

Depends on https://github.com/pion/transport/pull/74 and https://github.com/pion/srtp/pull/80

PR #1546, opened 2020-11-23
Improve TrackLocal error handling

Sean-Der:
* Return error to user when remote doesn't support codecs
* Assert that a unadded track returns an error on Write
* Assert that a added track doesn't error even when disconnected

Relates to #1526

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -45,6 +45,7 @@ type DTLSTransport struct {
 	srtcpSession  atomic.Value
 	srtpEndpoint  *mux.Endpoint
 	srtcpEndpoint *mux.Endpoint
+	srtpReady     chan struct{}
 
 	dtlsMatcher mux.MatchFunc
 
@@ -60,6 +61,7 @@ func (api *API) NewDTLSTransport(transport *ICETransport, certificates []Certifi
 		api:          api,
 		state:        DTLSTransportStateNew,
 		dtlsMatcher:  mux.MatchDTLS,
+		srtpReady:    make(chan struct{}),
 	}
 
 	if len(certificates) > 0 {
@@ -145,15 +147,6 @@ func (t *DTLSTransport) GetRemoteCertificate() []byte {
 }
 
 func (t *DTLSTransport) startSRTP() error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
-
-	if t.srtpSession.Load() != nil && t.srtcpSession.Load() != nil {
-		return nil
-	} else if t.conn == nil {
-		return errDtlsTransportNotStarted
-	}
-
 	srtpConfig := &srtp.Config{
 		Profile:       t.srtpProtectionProfile,
 		LoggerFactory: t.api.settingEngine.LoggerFactory,
@@ -204,30 +197,24 @@ func (t *DTLSTransport) startSRTP() error {
 
 	t.srtpSession.Store(srtpSession)
 	t.srtcpSession.Store(srtcpSession)
+	close(t.srtpReady)
 	return nil
 }
 
 func (t *DTLSTransport) getSRTPSession() (*srtp.SessionSRTP, error) {
-	value := t.srtpSession.Load()
-	if value != nil {
+	if value := t.srtpSession.Load(); value != nil {
 		return value.(*srtp.SessionSRTP), nil
 	}
-	if err := t.startSRTP(); err != nil {
-		return nil, err
-	}
 
-	return t.srtpSession.Load().(*srtp.SessionSRTP), nil
+	return nil, errDtlsTransportNotStarted
 }
 
 func (t *DTLSTransport) getSRTCPSession() (*srtp.SessionSRTCP, error) {
-	value := t.srtcpSession.Load()
-	if value != nil {
+	if value := t.srtcpSession.Load(); value != nil {
 		return value.(*srtp.SessionSRTCP), nil
 	}
-	if err := t.startSRTP(); err != nil {
-		return nil, err
-	}
-	return t.srtcpSession.Load().(*srtp.SessionSRTCP), nil
+
+	return nil, errDtlsTransportNotStarted
 }
 
 func (t *DTLSTransport) role() DTLSRole {
@@ -358,11 +345,12 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error {
 		return err
 	}
 
-	err = t.validateFingerPrint(parsedRemoteCert)
-	if err != nil {
+	if err = t.validateFingerPrint(parsedRemoteCert); err != nil {
 		t.onStateChange(DTLSTransportStateFailed)
+		return err
 	}
-	return err
+
+	return t.startSRTP()
 }
 
 // Stop stops and closes the DTLSTransport object.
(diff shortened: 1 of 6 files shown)


Sean-Der:
@OrlandoCo @at-wat @lherman-cs @jeremija 

With this I feel good calling the 'V3 Track API' good for `v3.0.0` ! I covered all the tests that I colud think of.

I wasn't able to make `TrackRemote` work as an interface. Getting it to fit into `OnTrack` required casting, which felt like a danger to me.

peerconnection_media_test.go
```diff
@@ -8,15 +8,13 @@ import (
 	"errors"
```
Sean-Der, line 8: These tests were just moved, not deleted

dtlstransport.go
```diff
@@ -45,6 +45,7 @@ type DTLSTransport struct {
 	srtcpSession  atomic.Value
 	srtpEndpoint  *mux.Endpoint
 	srtcpEndpoint *mux.Endpoint
+	srtpReady     chan interface{}
```
at-wat, line 48: Seems used only for closing:
```suggestion
	srtpReady     chan struct{}
```

srtp_writer_future.go
```diff
@@ -0,0 +1,95 @@
...
+
+// srtpWriterFuture blocks Read/Write calls until
+// the SRTP Session is available
+type srtpWriterFuture struct {
+	rtpSender      *RTPSender
+	rtcpReadStream atomic.Value // *srtp.ReadStreamSRTCP
+	rtpWriteStream atomic.Value // *srtp.WriteStreamSRTP
+}
+
+func (s *srtpWriterFuture) init() error {
```
at-wat, line 21: I would like to pass `context.Context` to all `Read/WriteRTP/RTCP` methods and handle cancel and timeout when blocking at here, as I proposed at https://github.com/pion/webrtc-v3-design/blob/01b2dee52ae4417a89c73385b8cae69b3877453b/rtpengine/io.go#L10-L35

dtlstransport.go
Sean-Der: Fixed!

srtp_writer_future.go
Sean-Der: I forget the full story, but we attempted this before and it had serious performance issues. With that in mind I need to also confirm that interceptors don't cause any issues!

* https://github.com/pion/transport/pull/74
* https://github.com/pion/webrtc/pull/1266

Sean-Der: I am all for doing it eventually though, API wise it makes sense. I don't think we should do it for `/v3` though.

After this is merged and the Interceptors change I really want to feature freeze. Unless you think we are giving people a really bad experience

at-wat: If it blocks without context, I think it should have `SetDeadline` like `net.Conn` to make it able to be used safely. (`transport.Buffer` is doing so)

dtlstransport.go
```diff
@@ -362,7 +349,8 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error {
 	if err != nil {
 		t.onStateChange(DTLSTransportStateFailed)
 	}
-	return err
+
+	return t.startSRTP()
```
jeremija, line 353: I'm not that familiar with `dtlstransport.go`, but do we want to call `startSRTP` even when the err above is not nil?

srtp_writer_future.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (s *srtpWriterFuture) Close() error {
+	if value := s.rtcpReadStream.Load(); value != nil {
+		return value.(*srtp.ReadStreamSRTCP).Close()
+	}
+
+	return nil
+}
+
+func (s *srtpWriterFuture) Read(b []byte) (n int, err error) {
```
jeremija, line 61: What if we rewrote this method (and similar ones below) to:

```go
func (s *srtpWriterFuture) Read(b []byte) (n int, err error) {
	value := s.rtcpReadStream.Load()

	if value == nil {
		if err := s.init(); err != nil {
			return 0, err
		}

		value = s.rtcpReadStream.Load()
	}

	return value.(*srtp.ReadStreamSRTCP).Read(b)
}
```

Your code looks neater than my suggestion, but it I'm afraid it might be easy to inadvertently trigger infinite recursions by modifying the implementation of `init()` without thinking about the consequences it might have in these methods (for example if another branch is added that does not set `rtcpReadStream`, but does not return an error).

srtp_writer_future.go
jeremija, line 21: I just checked the implementations of all methods called from `init()` and it seems that it won't be a big deal if `init` is called multiple times (e.g. by calling both `Read()` and `Write` before the `srtpReady` is closed).

Just wanted to double-check if my understanding is correct?

dtlstransport.go
Sean-Der: Oh nice catch! Oof bad mistake :(

srtp_writer_future.go
Sean-Der: Yea wouldn't be a problem at all!

srtp_writer_future.go
Sean-Der: Oh yea, much better to be defensive. I added a `nil` check to guard against that.

rtpsender.go
```diff
@@ -62,8 +61,11 @@ func (api *API) NewRTPSender(track TrackLocal, transport *DTLSTransport) (*RTPSe
 		stopCalled: make(chan interface{}),
```
at-wat, line 61:
```suggestion
		stopCalled: make(chan struct{}),
```

srtp_writer_future.go
at-wat: @Sean-Der what do you think about https://github.com/pion/webrtc/pull/1546/files#r529263738

Sean-Der: @at-wat Oh nice, yes I am all for exposing that! We can add as an additional API. Really nice that it doesn't have to be a breaking change either.

Created https://github.com/pion/webrtc/issues/1553 that could be a nice first contribution for someone :)

rtpsender.go
Sean-Der: Fixed!

srtp_writer_future.go
jeremija: Thanks!

at-wat: #1553 handles only the deadline of the underlying connection, and block due to this `init()` can't be handled.

I created a PR to add `ReadContext` to `packetio.Buffer` and the performance of the context version is better than the current master branch.
https://github.com/pion/transport/pull/104

Sean-Der: Is this PR ok to merge (mind approving?), since we can do the `context` stuff in another PR.

This PR is just moving around existing code, so we should have no behavior differences with `init`. All the function calls inside of it are non-blocking.

at-wat: Ah, right. Previous implementation was also blocking if Read is called without successful DTLS handshake.
https://github.com/pion/webrtc/blob/9ac8e9060faa876e14ca362b867e7596fa513691/rtpsender.go#L221-L226

at-wat (approved):

PR #1548, opened 2020-11-23
Don't run OnNegotiationNeeded if handler not set

Sean-Der:
Reduce the amount of code that is running if we don't need this.

Relates to #1494


Sean-Der:
@tekig @OrlandoCo @tarrencev Just to reduce the blast radius. @tomsun is actively getting bitten by this, so would like to make it easier on him.

tarrencev (approved):

PR #1566, opened 2020-12-02
Write/WriteRTP shouldn't block before SRTP

Sean-Der:
A Track shouldn't block until a PeerConnection is connected. It should
only send to PeerConnections that are fully established. v3 Track API
changed this behavior by mistake.

Related to pion/transport#116

### DIFF
--- a/srtp_writer_future.go
+++ b/srtp_writer_future.go
@@ -19,13 +19,25 @@ type srtpWriterFuture struct {
 	rtpWriteStream atomic.Value // *srtp.WriteStreamSRTP
 }
 
-func (s *srtpWriterFuture) init(ctx context.Context) error {
-	select {
-	case <-s.rtpSender.stopCalled:
-		return io.ErrClosedPipe
-	case <-s.rtpSender.transport.srtpReady:
-	case <-ctx.Done():
-		return ctx.Err()
+func (s *srtpWriterFuture) init(ctx context.Context, returnWhenNoSRTP bool) error {
+	if returnWhenNoSRTP {
+		select {
+		case <-s.rtpSender.stopCalled:
+			return io.ErrClosedPipe
+		case <-s.rtpSender.transport.srtpReady:
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			return nil
+		}
+	} else {
+		select {
+		case <-s.rtpSender.stopCalled:
+			return io.ErrClosedPipe
+		case <-s.rtpSender.transport.srtpReady:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
 
 	srtcpSession, err := s.rtpSender.transport.getSRTCPSession()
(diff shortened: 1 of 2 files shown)


srtp_writer_future.go
```diff
@@ -66,7 +78,7 @@ func (s *srtpWriterFuture) ReadContext(ctx context.Context, b []byte) (n int, er
 		return value.(*srtp.ReadStreamSRTCP).ReadContext(ctx, b)
 	}
 
-	if err := s.init(ctx); err != nil || s.rtcpReadStream.Load() == nil {
+	if err := s.init(ctx, false); err != nil || s.rtcpReadStream.Load() == nil {
```
at-wat, line 81: I guess it should be true for context controlled usage and false for no-context usage. It seems to be true for this case?

srtp_writer_future.go
```diff
@@ -90,7 +102,7 @@ func (s *srtpWriterFuture) Write(ctx context.Context, b []byte) (int, error) {
 		return value.(*srtp.WriteStreamSRTP).WriteContext(ctx, b)
 	}
```
at-wat, line 103: I forgot to change func name from `Write` to `WriteContext` above...

Sean-Der: @at-wat @OrlandoCo I can change that, but I am not sure what the answer is re: context yet?

* Is the performance good enough to do Context only?
* Are we providing two paths (`Write` and `WriteContext`), or removing context?

at-wat: Maybe better to ask @josharian as well

srtp_writer_future.go
at-wat: or all `init()` call in Read/Write/ReadRTP/WriteRTP should set false.

srtp_writer_future.go
Sean-Der: :+1: we can wait for a response!

Personally how do you feel about rolling back the context changes and exposing `SetReadDeadline` and `SetWriteDeadline`? Are you really excited about adding Context, or do you consider it a 'nice to have'. 

@OrlandoCo and @jech are pretty performance sensitive. So I just want to find something that makes sense for everyone.

at-wat: I personally like to control it by context mainly for future maintainability and easiness of managing resources. Exposing SetDeadline can realize equivalent behavior, but bit complicated.
I don't have very urgent need for this.

jech:
> Personally how do you feel about rolling back the context changes and
> exposing SetReadDeadline and SetWriteDeadline?

That would be very, very helpful indeed.  From an API point of view it is
much cleaner (avoids carrying an extra parameter that is generally unused),
and the slight loss of expressivity is not a big deal in practice.

Pleae make sure that you can recover after a missed deadline, i.e. that
the operation can be restarted after it has returned with a missed
deadline error.  (Both bufio and gorilla/websocket have he issue that the
reader or writer needs to be closed after a deadline has been missed.)

Sean-Der:
This is what I think we should do, please correct me if you disagree/am missing anything.

*  Remove `context`, from `pion/webrtc`, `pion/transport` and `pion/srtp`. Make sure we keep these numbers available, and if they become better we can revisit. 
* Add `SetDeadline`/`ReadDeadline` and make sure they are tested/working as expected.

I personally really want to get `v3.0.0` out. Announcing all these improvements is going to bring a lot of new people in that have ignored Pion because it didn't have what they needed before. The longer we wait, the more contributions we miss out on.

@tarrencev @OrlandoCo @jech @at-wat

at-wat:
> Remove context, from `pion/webrtc`, `pion/transport` and `pion/srtp`. Make sure we keep these numbers available, and if they become better we can revisit.

As `pion/transport` supports both and has no performance problem, it can be left as is.

PR #1572, opened 2020-12-04
Allow extensions to be configured with direction

### LINKED ISSUE #1554: PayloadType and Extension Header ID needs to be set at the Transceiver Level
Currently the `MediaEngine` stores these values at the PeerConnection level. We need to store them at the Transceiver level.

One use case where this is important is the fan-out use case. Without this change the SFU would have to choose either Header ID 1 or 2. You would then have to modify the headers as they flow through. This same restriction applies to PayloadType.
```
FireFox (Header ID 1) ->             ------> Track 1 (Header ID 1)
                           SFU  
Chrome (Header ID 2) ->            -------> Track 2 (Header ID 1)
```

By allowing values per media section we remove the need for users to modify/parse inside the SFU. So you could do the following.
```
FireFox (Header ID 1) ->              ------> Track 1 (Header ID 1)
                            SFU  
Chrome (Header ID 2) ->             -------> Track 2 (Header ID 2)
```

Sean-Der:
RegisterHeaderExtension now allows users to enable headers depending on
the type of transceiver that was created.

Also expose GetParameters on RTPSender and RTPReceiver

Co-authored-by: OrlandoCo <luisorlando.co@gmail.com>

Resolves #1554

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -141,6 +141,9 @@ var (
 	// ErrNoPayloaderForCodec indicates that the requested codec does not have a payloader
 	ErrNoPayloaderForCodec = errors.New("the requested codec does not have a payloader")
 
+	// ErrRegisterHeaderExtensionInvalidDirection indicates that a extension was registered with a direction besides `sendonly` or `recvonly`
+	ErrRegisterHeaderExtensionInvalidDirection = errors.New("a header extension must be registered as 'recvonly', 'sendonly' or both")
+
 	errDetachNotEnabled                 = errors.New("enable detaching by calling webrtc.DetachDataChannels()")
 	errDetachBeforeOpened               = errors.New("datachannel not opened yet, try calling Detach from OnOpen")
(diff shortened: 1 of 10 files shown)


Sean-Der:
@OrlandoCo I believe this gives you the features you need, and allows us to avoid some complexity (for a bit)

You can create an Offer (or answer) that has different extensions depending on the transceiver direction. RegisterExtensionHeader allows you to pass directions. By default it assumes all.

That means you can declare
* Incoming media (`recv` transceiver) has TWCC
* Outgoing media (`send` transceiver) doesn't have TWCC

This PR also drops the idea that PayloadType/HeaderExtensionID is declared at the global level. It instead uses `GetParameters`

This PR doesn't support different ExtensionIDs/PayloadTypes I will do that in the next PR and add `SetParameters`

OrlandoCo (approved):
lgtm!

PR #1576, opened 2020-12-05
Export SSRC for RTPSender and RTPReceiver.

### LINKED ISSUE #1575: v3: please export track SSRC and ClockRate
There are a number of track properties that were available in v2 but are only available to interceptors in v3.  In particular, there are no equivalents to `track.SSRC` and `track.Codec().ClockRate`.

Interceptors are a neat idea, but they will take some time to be adopted universally, and some applications might want to use a different approach.  For people to be able to move to v3 incrementally, it is important that anything that is easy to do in v2 is still doable in a similar manner in v3.

Please export SSID and ClockRate in v3.

jech:
This allows retrieving the SSRC for RTPSender and RTPReceiver
in the non-simulcast case.

#### Description

Issues to discuss:
 * what is the right thing to do in the simulcast case?  I'm currently returning 0, which is not a legal SSID if memory serves.
 * what should be done in the WASM case?  Currently, the method is not defined; the alternative would be to return 0.
#### Reference issue
Fixes #1575.

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -69,6 +69,19 @@ func (r *RTPReceiver) GetParameters() RTPParameters {
 	return r.api.mediaEngine.getRTPParametersByKind(r.kind, []RTPTransceiverDirection{RTPTransceiverDirectionRecvonly})
 }
 
+// SSRC returns the SSRC used by the receiver.  If the receiver is doing
+// simulcast, then this returns 0.
+// This is not available in WASM.
+func (r *RTPReceiver) SSRC() SSRC {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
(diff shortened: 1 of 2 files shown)


Sean-Der:
I think we want to extend `GetParameters` to expose the SSRC.

Instead return `RTPSendParameters`[0], and change encodings to a slice (to support Simulcast in the future). The slice will contain one entry for now, and will have the SSRC.

We also need to embed `RTPParameters` in `RTPSendParameters`

[0] https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSendParameters

jech:
Sean has suggested that a better solution would be to fix GetParameters to return the right structure on the sender side, so I'm putting this on hold for the time being.

PR #1578, opened 2020-12-05
Fix RTPSendParameters.

### LINKED ISSUE #1575: v3: please export track SSRC and ClockRate

jech:
RTPSendParameters contains an array of encodings.
RTPSender.GetParameters returns RTPSendParameters.

#### Description

Both these changes bring us closer to the Javascript API.

#### Reference issue
Fixes #1575.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1236,10 +1236,12 @@ func (pc *PeerConnection) startRTPSenders(currentTransceivers []*RTPTransceiver)
 	for _, transceiver := range currentTransceivers {
 		if transceiver.Sender() != nil && transceiver.Sender().isNegotiated() && !transceiver.Sender().hasSent() {
 			err := transceiver.Sender().Send(RTPSendParameters{
-				Encodings: RTPEncodingParameters{
-					RTPCodingParameters{
-						SSRC:        transceiver.Sender().ssrc,
-						PayloadType: transceiver.Sender().payloadType,
+				Encodings: []RTPEncodingParameters{
+					{
+						RTPCodingParameters{
+							SSRC:        transceiver.Sender().ssrc,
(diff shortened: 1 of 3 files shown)


OrlandoCo:
Looks great to me, but not sure of the benefits of making the encodings a slice other than comply with js api, it may confuse some users.

Sean-Der:
LGTM, thanks for doing this @jech 

I am going to add a test and then push now!

Sean-Der:
Merged! Thanks for doing this @jech 

0dea1f91e0f98615e4020158e2e2018aa86ba2fa

jech:
> Merged!

Excellent!  In case somebody's interested, the port to the new API of the
Unnamed SFU is available:

  git clone -b new-api https://www.irif.fr/~jch/software/sfu.git

It's a straight port, I didn't restructure anything to adapt to the new
API.  The only change is that I now keep track of the SSRC rather than
determining it on the fly when I need it.

Expect this branch to be rebased until I decide to merge it.

-- Juliusz

jech:
> Looks great to me, but not sure of the benefits of making the encodings a slice
> other than comply with js api, it may confuse some users.

It's not used in the current code, granted, but it will avoid duplicating
the data structure if we ever implement an equivalent to setParameters, or
if an encoding library (mediadevices?) implements automatic codec
selection constrained by the application.

PR #1582, opened 2020-12-07
Add remote raw rtp reader

OrlandoCo:
#### Description
This method will help getting the raw rtp packet from interceptors

#### Depends on
https://github.com/pion/interceptor/pull/12


Sean-Der:
oh jeez, this is a bigger problem that I understood at first @OrlandoCo! `Read` completely ignores interceptors?

OrlandoCo:
> oh jeez, this is a bigger problem that I understood at first @OrlandoCo! `Read` completely ignores interceptors?

Thats why ReadRTP and others methods call `Read` instead of `read`, so if the chain is using raw rtp and the user is reading from `ReadRTP` it may see the same slice and not a read

Sean-Der:
Closing for #1595

PR #1595, opened 2020-12-13
Update Interceptors to use []byte based API

Sean-Der:
Also update test to assert Attributes get passed all the way through

Resolves pion/interceptor#14


track_remote.go
```diff
@@ -147,34 +122,10 @@ func (t *TrackRemote) Read(b []byte) (n int, err error) {
 	return r.readRTP(b, t)
 }
 
-// peek is like Read, but it doesn't discard the packet read
```
Sean-Der, line 150: moved to bottom of file

OrlandoCo (approved):

PR #1605, opened 2020-12-19
Add buffer factory

OrlandoCo:
#### Description
Allow to configure a buffer factory on SettingEngine to provide a custom buffer for RTP/RTCP

#### References
https://github.com/pion/srtp/pull/126
https://github.com/pion/transport/pull/122


Sean-Der:
@OrlandoCo LGTM! Do we need the `MediaEngine` changes, or were you testing things? You can pull that info from `GetParameters` today.

OrlandoCo:
> @OrlandoCo LGTM! Do we need the `MediaEngine` changes, or were you testing things? You can pull that info from `GetParameters` today.

I was doing a testing I may revert them,  I can use the GetParameters but the buffer is created before the track event is fired, so I wanted a way to get the track info reading the PT from packet, but I opted to cache the first packet until the event is fired and bind the buffer

OrlandoCo:
@Sean-Der  I reverted the mediaengine changes, I think this should be safe to merge including the transport and srtp changes

tarrencev (approved):

PR #1607, opened 2020-12-22
Improve documentation of NewPeerConnection

jech:
It is important to mention that, unlike API.NewPeerConnection, this
registers default interceptors.

#### Description

This is an incompatible change in v3, make sure it is documented.

### DIFF
--- a/interceptor.go
+++ b/interceptor.go
@@ -10,8 +10,8 @@ import (
 	"github.com/pion/rtp"
 )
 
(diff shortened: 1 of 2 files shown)


Sean-Der:
Merged with de913a99a885cf17d14a3ffa3de2f3b99039ad7d 

`golangci-lint` just caught a typo `customised -> customized`

PR #1609, opened 2020-12-23
Fix a race condition in GatheringCompletePromise

lherman-cs:
## Description

It's possible that the ICE gathering is completed after the execution goes to the else statement but before the setGatherCompleteHandler finishes. In other words, it's possible that the handler misses the event and block the caller forever.


Sean-Der (approved):

Sean-Der:
Oof, yea that conditional is worthless! Nice catch :)

Sean-Der:
Also tagged `v3.0.1`

PR #1611, opened 2020-12-24
Added SetReadDeadline to RTPReceiver

Threadnaught:
#### Added SetReadDeadline to RTPReceiver

Previously, Reading RTP or RTCP packets from a peer would block until a packet was received, or the connection was terminated. This change allows you to set a deadline, after which the read function will return a timeout error and you can get on with other things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Suzuki Takeo](https://github.com/BambooTuna)
 * [baiyufei](https://github.com/baiyufei)
 * [pascal-ace](https://github.com/pascal-ace)
+* [Threadnaught](https://github.com/Threadnaught)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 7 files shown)


Sean-Der:
@Threadnaught @mission-liao I pushed some changes, mind giving it a look over!

* Added the ability to do Deadline for RTP or RTCP (not both at once)
* Added a distinct test

Sean-Der (approved):

Sean-Der:
Merging! @Threadnaught and @mission-liao if you see any issues happy to revert :)

mission-liao:
@Sean-Der Sorry missed your changes. The merged version would timeout read operations on all `TrackRemote` instances bound to one `RTPReceiver` (correct me if I'm wrong), and I would wonder if it's reasonable and make sense to provide a way for users to cancel read operations on a single `TrackRemote` instance.

But for my use case, this change works well. Thanks.

Sean-Der:
Oh yea you are right, we need to improve fro Simulcast!

For `SetReadDeadline` on `TrackRemote` it should seek do a lookup for the proper stream. Then we can add a `SetReadDeadlineSimulcast` and the user can choose the rid.

mission-liao:
@Sean-Der Cool, would look into that.

PR #1629, opened 2021-01-10
populate TrackRemote with mid when available

cameronelliott:
#### Description
Mid information is handy when available along with Rid information.

Tested with personal project, and got mid 0,1,2 as per SDP.
OnTrack() call back for simulcast will return zero-string "" for mid, but this can
be changed to the mid of the simulcast section.

#### Reference issue
No issue


cameronelliott:
Turns out this may not be the right solution for the problem I am facing.
If you think this is still valuable/important, please re-open and I can look at setting up unit tests.

PR #1639, opened 2021-01-15
Rtpdump: fix writing 2nd uint16 of packet header

cameronelliott:
After detailed analysis, I think there is a bug in the go code to write the rtpdump packet headers.
The first uint16 of the packet headers includes the header length, the 2nd uint16 of the per-packet headers
does not. This matches hand-examined output from the original rtpdump.
Second, I round tripped original rtpdump input throught the reader, then writer of this lib, and in doing so,
noticed the issue with this library. (binaries were not the same as original after round-tripping)
After making the change in this PR, the binary output from round-tripping was exactly the same.

After lots of time reviewing, checking, I believe this change is warranted/correct.

[struct to for packet header](https://github.com/irtlab/rtptools/blob/3264fecca042253754ddb9c8be1d3f81ef55563c/rtpdump.h#L59)
[code to write packet header](https://github.com/irtlab/rtptools/blob/3264fecca042253754ddb9c8be1d3f81ef55563c/rtpdump.c#L537)
```c
  switch(format) {
    case F_header:
      offset = (dnow - dstart) * 1000;
      packet->p.hdr.offset = htonl(offset);
      packet->p.hdr.plen   = ctrl ? 0 : htons(len);                                 // 2nd uint16
      /* leave only header */
      if (ctrl == 0) len = parse_header(packet->p.data);
      packet->p.hdr.length = htons(len + sizeof(packet->p.hdr));     // 1st uint16 
      if (fwrite((char *)&packet, len + sizeof(packet->p.hdr), 1, out) == 0) {
        perror("fwrite");
        exit(1);
      }
      break;
```


cameronelliott:
@maxhawkins Could you please double check that this makes sense to you?
@Sean-Der maybe we should wait to hear from @maxhawkins to weigh-in/double check this before merging.

maxhawkins:
Could you add a failing test first to verify this patch fixes the problem?

cameronelliott:
> Could you add a failing test first to verify this patch fixes the problem?

Will do.

maxhawkins:
I couldn't remember how this was supposed to work so I took a look at the [rtpdump docs](https://web.archive.org/web/20180809104236/http://www.cs.columbia.edu/IRT/software/rtptools/).

Your change looks good, as long as it makes things compatible with the actual rtpdump utility. I can't understand the purpose of the `+ 8` anymore. It may have been from a mistaken refactor.

So it's easier the next time we need to touch this code, could you copy the struct layout from [the rtptools page](https://github.com/pion/rtp/blob/master/packet.go#L86) as a comment inside of the Marshal method? We [add comments with binary layouts from the rfc](https://github.com/pion/rtp/blob/master/packet.go#L86) inside `Marshal()` methods in other packages like `rtp`. Makes it a lot easier to remember how do the bit packing right.

cameronelliott:
Are you suggestion I add this into the Pion Go code?
```c
typedef struct {
  u_int16 length;    /* length of packet, including this header (may 
                        be smaller than plen if not whole packet recorded) */
  u_int16 plen;      /* actual header+payload length for RTP, 0 for RTCP */
  u_int32 offset;    /* milliseconds since the start of recording */
} RD_packet_t;
```
from here: [http://web3.cs.columbia.edu/irt/software/rtptools/](http://web3.cs.columbia.edu/irt/software/rtptools/)

Is that what you are suggesting?

I am totally on the same page as you with using snippets of the reference from which stuff is derived,
but in this case, I am not sure the original struct+comments help or clear up anything!
(please disagree, if you think otherwise)

IMO, you really need the struct+the original code to understand how the rtptools create these headers,
because I think the comments on the struct are vague and maybe even misleading, and maybe how
we got here in the first place.

Here is how I think they should have written that struct:
```c
typedef struct {
  u_int16 length;    /* value is plen (next field)+sizeof(RD_packet_t) */
  u_int16 plen;       /* actual RTP-header+payload length for RTP, or 0 for RTCP */
  u_int32 offset;    /* milliseconds since the start of recording */
} RD_packet_t;
```
[c-lang struct](https://github.com/irtlab/rtptools/blob/3264fecca042253754ddb9c8be1d3f81ef55563c/rtpdump.h#L59)
[c-lang code](https://github.com/irtlab/rtptools/blob/3264fecca042253754ddb9c8be1d3f81ef55563c/rtpdump.c#L537)
(see lines 537, 540, keep in mind they set the fields in the inverse of the order of the struct)

What do you think about this idea?
If you could carefully review those last two links, and make sure I my proposed changes are correct,
and maybe you could decide whether to use the original comments or my suggested comments?
If you think my comments are better, one of us could also submit a PR to the github repo for rtptools, they should
get that fixed.

What do you think about all that? @maxhawkins

maxhawkins:
You're right, the struct isn't necessary after all since we aren't doing any bit-twiddling like in the other marshalers. Please disregard my comment. Everything looks good.

cameronelliott:
@Sean-Der @maxhawkins
I think this might be ready for baking. :)

Sean-Der:
Yes we should! Thank you so much for fixing this.

On a personal note I am so happy to see you contributing :) we had lunch I think two years ago? It feels great to have you involved and that Pion is working for you. 

Merged with 06f9351bce1d01c3b8f0ade56a65833cb2dd10fa

PR #1649, opened 2021-01-25
Add IPv6 addresses to some tests.

jech:
This only checks they are being parsed with no error, it doesn't
actually do anything useful with them.

#### Reference issue
Fixes #...


Sean-Der:
Thanks @jech !

Sean-Der (approved):

PR #1655, opened 2021-02-01
Add SendBytesText method to Datachannel

OrlandoCo:
This is a convenience method to reduce allocations.

Many marshalers return already a []byte result, so casting to string and bytes again may cause unneccesary allocations.

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -13,11 +13,18 @@ import (
 
 	"github.com/pion/datachannel"
 	"github.com/pion/logging"
+	"github.com/pion/sctp"
 	"github.com/pion/webrtc/v3/pkg/rtcerr"
 )
 
 const dataChannelBufferSize = math.MaxUint16 // message size limit for Chromium
 var errSCTPNotEstablished = errors.New("SCTP not established")
+
+// DataChannelMessageConfig Provides configurations for the datachannels messages.
+// This is helpful when you want to write a string message as bytes
+type DataChannelMessageConfig struct {
+	PPI sctp.PayloadProtocolIdentifier
+}
 
 // DataChannel represents a WebRTC DataChannel
 // The DataChannel interface represents a network channel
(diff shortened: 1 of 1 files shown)


Sean-Der:
What do you think of adding a `PPID` as an optional argument/config? If we add this then an argument could be made for `SendStringBytes`. 

Is it possible to use Byte API and take to string in browser?

OrlandoCo:
> What do you think of adding a `PPID` as an optional argument/config? If we add this then an argument could be made for `SendStringBytes`.
> 
> Is it possible to use Byte API and take to string in browser?

That would be ok for me, but maybe is confusig for others to set the correct `PPID` or understand is meaning

OrlandoCo:
@Sean-Der Should I update the PR with a method that accepts the `PPID` as an argurment? does this method would be an extra method or it will replace others?

Sean-Der:
@OrlandoCo I would add an additional paramater to both `DataChannelSendConfig`(?) that is optional (so no API break)
Then you can add a `PPID` override, and we can add more options in the future if needed! I would be up for accepting right away, feel free to add other people if you disagree/think it needs more discussion. I could totally be wrong.

These were the things I cared about, push back if you disagree!
* Can be zero allocation (have a struct allocated ahead of time that is shared between all)
* Doesn't break the API
* Lets us add new options in the future without creating new methods

ffmiyo:
Didn't  `Send` already do this?
https://github.com/pion/webrtc/blob/285fb782ba71357baec71a927ea855a86bdd4128/datachannel.go#L347-L355

OrlandoCo:
@ffmiyo `Send` sends data as binary format

OrlandoCo:
@Sean-Der I made the requested changes!

Sean-Der:
Looking back on this I don't think this is the right way to go. The receiver can cast if needed.

Our API is pretty extensive already. This adds more things to explain to users and test. If there is a use case that not having this completely backs it may be worth exploring more.

Issue #1656, opened 2021-02-02
Infinite loop in CreateOffer


jech:
Webrtc is at cddf631a7f82092483baecdc1fcfc97c1fd6e7b6.

I've got a reproducible hang in CreateOffer, where `hasLocalDescriptionChanged` persistently returns true although nothing is changing.  Adding a loop to do `println(t.Mid())` yields (0, 1).  Adding `println(desc.SDP)` yields:
```
v=0
o=- 2718194448058297639 1612296369 IN IP4 0.0.0.0
s=-
t=0 0
a=fingerprint:sha-256 49:2B:D2:53:A6:71:31:7D:70:9A:B5:9E:E5:A7:2D:8A:88:35:D2:5F:57:0D:C4:F7:3A:86:53:A0:C8:5E:E0:B3
a=group:BUNDLE 0
m=audio 9 UDP/TLS/RTP/SAVPF 111
c=IN IP4 0.0.0.0
a=setup:actpass
a=mid:0
a=ice-ufrag:KGYfsiPKkRztPQvH
a=ice-pwd:SdJBHmMLpkjuomedoCcusGXqCJNzPqhU
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:111 opus/48000/2
a=fmtp:111 minptime=10;useinbandfec=1
a=ssrc:902986156 cname:zsS95IvrkCqgPy5v3IxScYsNnvesJaXCgOTy
a=ssrc:902986156 msid:zsS95IvrkCqgPy5v3IxScYsNnvesJaXCgOTy e22b328d-aa70-4b24-aa4a-e1f76e6a29fb
a=ssrc:902986156 mslabel:zsS95IvrkCqgPy5v3IxScYsNnvesJaXCgOTy
a=ssrc:902986156 label:e22b328d-aa70-4b24-aa4a-e1f76e6a29fb
a=msid:zsS95IvrkCqgPy5v3IxScYsNnvesJaXCgOTy e22b328d-aa70-4b24-aa4a-e1f76e6a29fb
a=sendrecv
a=candidate:310921987 1 udp 2130706431 192.168.3.100 47255 typ host
a=candidate:310921987 2 udp 2130706431 192.168.3.100 47255 typ host
[... bunch of candidates omitted...]
a=candidate:4031623909 2 udp 2130706431 2a01:[...] 50610 typ host
a=end-of-candidates
m=video 0 UDP/TLS/RTP/SAVPF 0
```
[...]

PR #1658, opened 2021-02-02
Fix check for unchanged transceivers in CreateOffer

jech:
This makes the check in CreateOffer more conservative but also less
fragile.  It works around #1656, but I'm not sure it's the underlying
cause (the SDP quoted in that issue looks broken to my untrained eyes).

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -561,18 +561,16 @@ func (pc *PeerConnection) getStatsID() string {
 	return pc.statsID
 }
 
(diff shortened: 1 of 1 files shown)


Sean-Der (changes requested):
I think we can almost take this, but need to check some stuff.

Right now when generating the offer and we are missing codecs we will generate a non-bundle SDP (which will confuse the remote peer). So I need to investigate this further, but we should leave as a note to make sure this gets done :)

Sean-Der:
I am going to open another PR that gives a better error message for https://github.com/pion/webrtc/issues/1702

PR #1663, opened 2021-02-04
Correct the mimeType

pascal-ace:
#### Description
 change of the mimetype because since their implementations, there are concerns about capitalization.
#### Reference issue
 ivf recorder was not working (never fired )

### DIFF
--- a/examples/save-to-disk/main.go
+++ b/examples/save-to-disk/main.go
@@ -104,10 +104,10 @@ func main() {
 		}()
 
 		codec := track.Codec()
-		if codec.MimeType == webrtc.MimeTypeOpus {
+		if strings.EqualFold(codec.MimeType,webrtc.MimeTypeOpus) {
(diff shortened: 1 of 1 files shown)


examples/save-to-disk/main.go
```diff
@@ -104,10 +104,10 @@ func main() {
 		}()
 
 		codec := track.Codec()
-		if codec.MimeType == "audio/opus" {
+		if codec.MimeType == webrtc.MimeTypeOpus {
```
Sean-Der, line 107: Mind doing `strings.EqualFold(codec.MimeType, webrtc.MimeTypeOpus)`

pascal-ace: Done

Sean-Der:
Merged with 22eea024a224d426b31b2f66ace08a4c0e7fdc13 thanks @pascal-ace :D

Issue #1662, opened 2021-02-04
Guard against MediaEngine being shared between PeerConnections


Sean-Der:
A `MediaEngine` should only be used by one PeerConnection. We need to find a way to assert that it has never been used before.

Sean-Der:
After a little investigation it turns that we can't do this. It would be a massively breaking change for lots of applications. What we can do is copy. The user doesn't need a reference to the MediaEngine. So a PR would look like this.

* Add a `copy` and `equal` function to `MediaEngine` with a test
* In [NewPeerConnection](https://github.com/pion/webrtc/blob/master/peerconnection.go#L135) do a `copy`
* Write a test that if you create a PeerConnection and then modify the `MediaEngine` that was used that DO NOT `equal`

Sean-Der:
We also have the issue of interceptor's being shared. I think we want to allow them to be shared. That is possible sharp edge, but it is unavoidable because we *MAY* want them to be shared.

 Multiple PeerConnections will want to use the same congestion controller etc..

PR #1674, opened 2021-02-10
Improve MediaEngine Copy tests

Sean-Der:
Assert that the MediaEngine used by the PeerConnection is in the state
we expect

Relates to #1662


cameronelliott (approved):
Fantastic!

PR #1678, opened 2021-02-16
Resolve a caching issue when reusing session descriptions.

Mindgamesnl:
#### Description
This resolves a relatively small issue which would return a cached session description despite it being changed. It can be safely cached for internal usage, but should always be updated when requested.

This solution was proposed by Sean DuBois in the #pion slack channel.


Sean-Der:
Merged with cad7b6d34e963e1b765a88df7301473b9641adb3

PR #1682, opened 2021-02-17
Fix updateHeaderExtension for cloned MediaEngine

mission-liao:
#### Description
It addresses the issue that unable to updateHeaderExtension for cloned
MediaEngine, which is resulted from previous commit:

c8b7aa386a92100a0377fd32660c810d6e4e7e28

#### Reference issue
Fixes a possible(?) side-effect from the PR for #1662

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -284,11 +284,15 @@ func (m *MediaEngine) getHeaderExtensionID(extension RTPHeaderExtensionCapabilit
 // copy copies any user modifiable state of the MediaEngine
 // all internal state is reset
 func (m *MediaEngine) copy() *MediaEngine {
-	return &MediaEngine{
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):
Nice catch! I think it’s ok to always initialize, not that much overhead vs complexity.

Tag a new version if you want, thanks so much :)

mission-liao:
OK, would provide another commit to make it always initialize later. It would be less complicated.

mission-liao:
I've looked around codes for `negotiatedHeaderExtensions`, and think it's still better to keep the last commit.

The behavior: `updateHeaderExtension would do nothing if RegisterHeaderExtension is not called` seems reasonable.

Being of two minds.

PR #1683, opened 2021-02-17
Fix ice-lite parsing in SDP when whitespaces are present

f-viktor:
#### Description
Created a test that asserts that if pion receives an ice-lite SDP offer, it will assume ice-controlling role.
pion generates full ice answers, so in this exchange of ice-lite vs full ice, the full ice peer should always assume controlling role.
https://tools.ietf.org/html/rfc5245#section-5.2

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection_go_test.go
+++ b/peerconnection_go_test.go
@@ -1193,3 +1193,38 @@ func TestEmptyCandidate(t *testing.T) {
 		assert.NoError(t, peerConn.Close())
 	}
 }
+
+const liteOffer = `v=0
+o=- 4596489990601351948 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+a=msid-semantic: WMS
+a=ice-lite
+m=application 47299 DTLS/SCTP 5000
+c=IN IP4 192.168.20.129
+a=ice-ufrag:1/MvHwjAyVf27aLu
+a=ice-pwd:3dBU7cFOBl120v33cynDvN1E
+a=fingerprint:sha-256 75:74:5A:A6:A4:E5:52:F4:A7:67:4C:01:C7:EE:91:3F:21:3D:A2:E3:53:7B:6F:30:86:F2:30:AA:65:FB:04:24
+a=mid:data
+`
+
+// this test asserts that if an ice-lite offer is received,
+// pion will take the ICE-CONTROLLING role
+func TestICELite(t *testing.T) {
+	peerConnection, err := NewPeerConnection(Configuration{})
+	assert.NoError(t, err)
+
+	assert.NoError(t, peerConnection.SetRemoteDescription(
+		SessionDescription{SDP: liteOffer, Type: SDPTypeOffer},
+	))
+
+	SDPAnswer, err := peerConnection.CreateAnswer(nil)
+	assert.NoError(t, err)
+
+	assert.NoError(t, peerConnection.SetLocalDescription(SDPAnswer))
+
+	assert.Equal(t, ICERoleControlling, peerConnection.iceTransport.role,
+		"pion did not set state to ICE-CONTROLLED against ice-light offer")
+
+	assert.NoError(t, peerConnection.Close())
+}
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1100,8 +1100,10 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 	}
 
 	remoteIsLite := false
-	if liteValue, haveRemoteIs := desc.parsed.Attribute(sdp.AttrKeyICELite); haveRemoteIs && liteValue == sdp.AttrKeyICELite {
-		remoteIsLite = true
+	for _, a := range desc.parsed.Attributes {
+		if strings.TrimSpace(a.Key) == sdp.AttrKeyICELite {
+			remoteIsLite = true
+		}
 	}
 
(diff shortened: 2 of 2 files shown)


peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
 	return pca, pcb, nil
 }
 
+const liteOffer = `v=0
```
Sean-Der, line 47: I would cut this down to the bare minimum. I think the follow should be enough to confirm behavior.

```
v=0
o=- 1613490992396387 1 IN IP4 123.45.67.89
s=VideoRoom 5538594542389977
t=0 0
a=group:BUNDLE
a=msid-semantic: WMS janus
a=ice-lite
```

peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
...
+
+	peerConnection, err := NewPeerConnection(config)
+	assert.Nil(t, err)
+
+
+	err =  peerConnection.SetRemoteDescription(SDP)
+	assert.Nil(t, err)
+
+	remoteIsLite := false
+	for _, a := range peerConnection.RemoteDescription().parsed.Attributes {
```
Sean-Der, line 129: This change goes [here](https://github.com/pion/webrtc/blob/master/peerconnection.go#L1103). The search today is white-space sensitive, so instead we need to use this version.

peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
...
+		if strings.TrimSpace(a.Key) == sdp.AttrKeyICELite {
+			remoteIsLite = true
+		}
+	}
+	assert.Equal(t,remoteIsLite,true,"SDP was not parsed as ice-lite")
+
+	SDPAnswer, err := peerConnection.CreateAnswer(nil)
+	assert.Nil(t, err)
```
Sean-Der, line 137: `assert.Nil` -> `assert.NoError`

peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
...
+
+	err = peerConnection.SetLocalDescription(SDPAnswer)
+	assert.Nil(t, err)
```
Sean-Der, line 140: `assert.Nil` -> `assert.NoError`

peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
...
+			},
+		},
+		SDPSemantics: SDPSemanticsUnifiedPlanWithFallback,
+	}
+
+	peerConnection, err := NewPeerConnection(config)
+	assert.Nil(t, err)
+
+
+	err =  peerConnection.SetRemoteDescription(SDP)
```
Sean-Der, line 125: `assert.NoError(peerConnection.SetRemoteDescription(SessionDescription{SDP:liteOffer, Type:SDPTypeOffer}))`

This is subjective, but I would rather no allocate a variable that is only used one. It requires the reader to go and find the variable.

peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
...
+		ICEServers: []ICEServer{
+			{
+				URLs: []string{"stun:stun.l.google.com:19302"},
+			},
+		},
+		SDPSemantics: SDPSemanticsUnifiedPlanWithFallback,
+	}
+
+	peerConnection, err := NewPeerConnection(config)
+	assert.Nil(t, err)
```
Sean-Der, line 122: `assert.NoError`

peerconnection_go_test.go
```diff
@@ -43,6 +44,107 @@ func (api *API) newPair(cfg Configuration) (pcOffer *PeerConnection, pcAnswer *P
...
+	config := Configuration{
+		ICEServers: []ICEServer{
+			{
+				URLs: []string{"stun:stun.l.google.com:19302"},
+			},
+		},
+		SDPSemantics: SDPSemanticsUnifiedPlanWithFallback,
+	}
+
+	peerConnection, err := NewPeerConnection(config)
```
Sean-Der, line 121: I think we can just do `Configuration{}` this test isn't a PlanB/Unified-Plan issue.

peerconnection_go_test.go
f-viktor: had to add a few extra lines, I used minimal offer from another test file and trimmed it down.

peerconnection_go_test.go
f-viktor: makes a lot more sense that way :D, moved it over

Sean-Der:
This looks 100% to me! Nice work @f-viktor 

Does this fix your issue locally as well? Hopefully this was the real root cause

f-viktor:
yep, works perfectly in my use-case too!

Sean-Der:
Congrats on your first contribution @f-viktor :D I really hope you stay involved. Your PR was perfect I made zero changes.

8e8567040cd363e62506903bf2ded6f47a1f2c01

PR #1684, opened 2021-02-18
Enable Sender and Receiver Reports by default

Sean-Der:
The play-from-disk examples sees the average bitrate using
Chromium 90.0.4412.3 when enabled on loopback for a 3 minute
session.

Before: 744.443
After: 3927.669

Introduced with pion/interceptor@v0.0.10

---

I uploaded the full video analysis here https://gist.github.com/Sean-Der/b162e879101450f4f27db1c95ca0241d

This is a pretty big change. We are moving more towards interceptors. I want to make sure everyone agrees and this has no negative impacts on people that wish to still maintain their own pipelines.

All of the real work here was done by @aler9, thank you so much for this. This is really exciting and a huge step in taking pion/webrtc forward :)


jeremija:
That's awesome, I'm looking forward to trying it out once I finally switch to v3. I'm more curious about the high level concept of this feature: does this work by terminating the RTCP packets once they reach the SFU?

For example, if we had a room in with three browser peers (left side) and three pion peers in the SFU (right side) and 6 tracks:

(using the same SSRC for tracks between different peers connections for the sake of clarity)

```
+--------+   SSRC 1 audio           +-------------+
| Peer A |------------------------->| SFU peer A' |---------------\
|        |                          |             |               |
|        |   SSRC 2 video           |             |               |
|        |------------------------->|             |-------------- | -\
|        |                          |             |               |  |
|        |   SSRC 3 audio           |             |               |  |
|        |<-------------------------|             |--\            |  |
|        |                          |             |  |            |  |
|        |   SSRC 4 video           |             |  |            |  |
|        |<-------------------------|             |<-| -\         |  |
|        |                          |             |  |  |         |  |
|        |   SSRC 5 audio           |             |  |  |         |  |
|        |<-------------------------|             |<-|--|--\      |  |
|        |                          |             |  |  |  |      |  |
|        |   SSRC 6 video           |             |  |  |  |      |  |
|        |<-------------------------|             |<-|--|--| -\   |  |
+--------+                          +-------------+  |  |  |  |   |  |
                                                     |  |  |  |   |  |
+--------+   SSRC 3 audio           +-------------+  |  |  |  |   |  |
| Peer B |------------------------->| SFU peer B' |--/  |  |  |   |  |
|        |                          |             |     |  |  |   |  |
|        |   SSRC 4 video           |             |     |  |  |   |  |
|        |------------------------->|             |-----+  |  |   |  |
|        |                          |             |     |  |  |   |  |
|        |   SSRC 1 audio           |             |     |  |  |   |  |
|        |<-------------------------|             |<-------|--|---/  |
|        |                          |             |     |  |  |      |
|        |   SSRC 5 video           |             |     |  |  |      |
|        |<-------------------------|             |<-------+  |      |
+--------+                          +-------------+     |  |  |      |
                                                        |  |  |      |
+--------+   SSRC 5 audio           +-------------+     |  |  |      |
| Peer C |------------------------->| SFU peer C' |--------/  |      |
|        |                          |             |     |     |      |
|        |   SSRC 6 video           |             |     |     |      |
|        |------------------------->|             |-----------/      |
|        |                          |             |     |            |
|        |   SSRC 2 video           |             |     |            |
|        |<-------------------------|             |------------------/
|        |                          |             |     |
|        |   SSRC 4 video           |             |     |
|        |<-------------------------|             |<----/
+--------+                          +-------------+
```

Does this change enable sender reports and receiver reports between:

- Peer `A` and `A'` (A' will now send SR for SSRCs 3,4,5,6 and RR for 1, 2)
- Peer `B` and `B'` (B' will now send SR to B for SSRCs 1 and 5 and RR for 3, 4)
- Peer `C` and `C'` (C' will now send SR to C for SSRCs 2 and 4 and RR for 5, 6)

Sean-Der:
@jech as well, sorry I just went through the org and tagged people using media API.

examples/ice-tcp/main.go
```diff
@@ -18,11 +18,6 @@ func doSignaling(w http.ResponseWriter, r *http.Request) {
 	var err error
 
 	if peerConnection == nil {
-		m := webrtc.MediaEngine{}
```
Sean-Der, line 21: This example is only doing DataChannels

jech:
I told you so :-)

On a more serious note, since the report interceptor appears to be generating reasonable loss statistics, it should be safe to deploy in the Internet.  Hence, I support this change.

I'm a little confused about what's going on with the NTP time fields, but I'm not too worried — at the worst, lipsynch will be off.

Sean-Der:
> I told you so :-)

I never pretended to know what I was doing, people just assumed :p Yea I am glad you brought this up, made it way more urgent from me. Glad to hear you are feeling good about this change. I would love to move forward with it. Will give it a few more days and then nag people for approvals.

jech:
@jeremija In an SFU, you want to terminate SCTP at the SFU and run a separate congestion controller for every client.  Let's take a simpler example than yours: A, B, C are in the same group, A is sending a media stream which is begin forwarded to B and C:
```
A --SFU-> B
     |
     +--> C
```
Then you run one congestion controller on the A->SFU link, one congestion controller on the SFU->B link, and one on the SFU->C link, terminating SCTP each time.  In the absence of simulcast, the bitrate you aim for is the minimum of the bitrates indicated by the three congestion controllers.

jeremija:
@jech wow thanks so much for the explanation. This was super useful, I learned some new things and you confirmed some of the things I already knew about but wasn't sure if my understanding is correct. Thank you!

When you say "congestion controller", are you referring to the component that calculates and reports the maximum bitrate (e.g. REMB), or something else?

jech:
The congestion controller is the algorithm that decides the maximum allowable bitrate for a flow.  It usually runs at the sender, with the receiver providing feedback of some sort, but it may run at the receiver, which must then somehow communicate the allowable bitrate to the sender:
 * in TCP-Reno, the congestion controller runs at the sender, with the receiver providing feedback implicitly by sending ACKs (or failing to send them);
 * the loss-based congestion controller in GCC (Google-style WebRTC) runs at the sender, with the receiver providing feedback explicitly by sending loss estimates in receiver reports;
 * the original delay-based congestion controller in GCC runs at the receiver, which then communicates its conclusions to the sender using REMB feedback;
 * the revised delay-based congestion controller in GCC runs at the sender, based on detailed reports provided by the receiver through TWCC feedback.

I'm being told that the revised delay-based controller is better, but I don't understand why — it has the advantage of being similar to traditional congestion controllers (by virtue of running at the sender), but it is way more chatty than the original delay-based controller.

Sean-Der:
@jeremija @jech knows this stuff better then, super lucky to have him in the community :) 

Also check out https://github.com/pion/interceptor/issues/25 it is my plan to get stuff into Pion itself. It will all be opt-in/composable. It will be the second point `Google Style WebRTC` and will use receiver reports.

I want to do TWCC eventually. I don't feel confident I can do it right in the first pass though.

adwpc:
> @jeremija @jech knows this stuff better then, super lucky to have him in the community :)
> 
> Also check out [pion/interceptor#25](https://github.com/pion/interceptor/issues/25) it is my plan to get stuff into Pion itself. It will all be opt-in/composable. It will be the second point `Google Style WebRTC` and will use receiver reports.
> 
> I want to do TWCC/Delay-based eventually. I don't feel confident I can do it right in a good amount of time.

Yep, TWCC is better for SFU.
Im not sure SR/RR shoud be configuable?
There is also a SR/RR in ion-sfu, what do do you think? @OrlandoCo

jech:
> TWCC is better for SFU

Why?

Sean-Der:
@aler9 mind approving this? I haven't heard any push back so I vote to merge and push.

Users can easily opt out of this if they wish. I believe that most users will want this by default.

aler9 (approved):
> The browser will not send higher quality streams unless it has the available bandwidth. You can look at the bandwidth estimation in `chrome://webrtc-internals`. It is under `VideoBwe` when `Read Stats From: Legacy non-Standard` is selected.

This is partially true; in order to obtain the maximum available bitrate, a webrtc implementation must
* implement RR/SR (ok)
* implement bandwidth estimation
* when it's receiving the stream, edit the SDP and add a TIAS bandwidth that contains the maximum bandwidth available

In this way, bandwidths of 20/30 mbps can be achieved.

PR #1687, opened 2021-02-19
fix err in example

figadore:
#### Description
just a quick fix to a reference to the wrong error object in an example


Sean-Der:
Nice catch, thank you @figadore 

Merged with 8cb5bde7c99941895d31997c26bcf0b56441d8b3

PR #1666, opened 2021-02-06
PR1199 redo: Answer with sendonly when remote is recvonly

cameronelliott:
This is a re-application of the commits from PR1199, as the author appears busy with other stuff.
But, it was nice of him to figure this out.

The original PR is the best place for the background:
#1199 

The short introduction from original author longsleep  is:

> When receiving a remote offer with a transceiver which is recvonly, the
> answer must not set the transceiver to a receiving mode to avoid Firefox
> from complaining and failing to set the remote description with the
> following error:  **Answer tried to set recv when offer did not set send**
> While Chrome is more forgiving and happily sets remote transceiver which
> include receive even when the offer did not request ist, this needs to
> be changed to get interop with Firefox.

PR #1689, opened 2021-02-22
Set answer direction as sendonly when offer direction is recvonly

### LINKED ISSUE #717: Answer tried to set recv when offer did not set send

aler9:
#### Description

* When a local peer connection has a single transceiver with a `sendrecv` direction, and the remote has a transceiver with a `recvonly` direction, the local peer connection must change the transceiver direction to `sendonly`
* When a local peer connection has a single transceiver with a `recvonly` direction, and the remote has a transceiver with a `recvonly` direction, the local peer connection must create another transceiver with a `sendonly` direction
* A unit test is added to cover all possible cases

This issue is causing the infamous fatal error "Answer tried to set recv when offer did not set send" with Firefox.

#### Reference issue

This fixes #717 and closes #1199 and #1666

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1049,10 +1049,21 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 				if err != nil {
 					return err
 				}
-				t = pc.newRTPTransceiver(receiver, nil, RTPTransceiverDirectionRecvonly, kind)
+
+				localDirection := RTPTransceiverDirectionRecvonly
+				if direction == RTPTransceiverDirectionRecvonly {
+					localDirection = RTPTransceiverDirectionSendonly
+				}
+
+				t = pc.newRTPTransceiver(receiver, nil, localDirection, kind)
 
 				pc.onNegotiationNeeded()
+			} else if direction == RTPTransceiverDirectionRecvonly {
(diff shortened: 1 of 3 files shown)


Sean-Der:
This is a gorgeous PR @aler9 nice work! Merging now

aler9:
wait a second please :)

aler9:
Now everything seems to be good

Sean-Der:
:fire: really amazing work. This is going to make a lot of peoples lives easier :)

Sean-Der (approved):

PR #1706, opened 2021-03-13
fix writing rtcp empty arrays

OrlandoCo:
#### Description

Fix panic caused when a rtcp array is empty, packet marshal will return an empty byte slice causing and out of range panic

````
panic: runtime error: index out of range [3] with length 0
goroutine 3244 [running]:
encoding/binary.bigEndian.Uint32(...)
c:/go/src/encoding/binary/binary.go:112
github.com/pion/srtp/v2.(*Context).encryptRTCP(0xc00006a8d0, 0x0, 0x0, 0x0, 0xc00166e198, 0x4, 0x8, 0xc0005f7da8, 0x13674b, 0xc00097efc0, ...)
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -147,6 +147,9 @@ var (
 	// ErrSimulcastProbeOverflow indicates that too many Simulcast probe streams are in flight and the requested SSRC was ignored
 	ErrSimulcastProbeOverflow = errors.New("simulcast probe limit has been reached, new SSRC has been discarded")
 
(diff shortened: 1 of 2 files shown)


Sean-Der:
Should we put this in `pion/srtp`. I believe anyone that uses that directly would get this crash also?

OrlandoCo:
Closing to move to srtp.

Issue #1708, opened 2021-03-17
DTLS Fingerprints in SDP Offer/Answer are not verified


Gaukas:
### Your environment.
  - Version: pion/webrtc v3.0.14
  - Browser: N/A
  - Other Information - reproducable with example/data-channels-create & example/data-channels

### What did you do?

- Run both `data-channels-create` and `data-channels` from example. 
- Once the SDP offer has been generated, decode it with base64.
- Randomly edit the DTLS fingerprint value in the SDP offer
- Copy & paste the base64-encoded SDP offer into the waiting `data-channels` 
- Copy & paste the SDP answer generated by `data-channels` into `data-channels-create`
 
### What did you expect?

The built-in fingerprints verification should throw an error and therefore prevent the data-channel from being established. 

### What happened?

The data channel was created as usual.
```
ICE Connection State has changed: checking
ICE Connection State has changed: connected
Data channel 'data'-'824635660400' open. Random messages will now be sent to any connected DataChannels every 5 seconds
Sending 'SBfBWYaFzaFZDiV'
Message from DataChannel 'data': 'EtaKTfGglNgpNNn'
```

Sean-Der:
Hey @Gaukas 

We do have a test for this `TestInvalidFingerprintCausesFailed` that is properly catching this. I will confirm this works and update as appropriate thanks for filing this!

Sean-Der:
Hey @Gaukas what browser are you using to test?

I believe the remote peer isn't the one properly asserting.

thanks

Gaukas:
I am not using any browser. The `offerer` is `data-channels-create` and the `answerer` is `data-channels`. Both are go binaries.

Gaukas:
It may not be that simple because I tried to mess with both fingerprints in offer and answer. The data channel was still created successfully.

Gaukas:
@Sean-Der but thank you for the fast response! I really appreciate it. Please let me know if there's anything else I could help with.

Sean-Der:
Ah I see. So the issue is that we set PeerConnectionState to failed, but we don't actually tear down the connections.

I will fix that and tag a new issue. This probably even warrants a CVE! If you are interested, good for resume and helps people update quicker.

Gaukas:
Thank you @Sean-Der! I'm glad that this issue could be helpful to the project. 
Just for your reference, attached is the log from both instances, in which `2 bytes` from the fingerprint in the offer (generated by `data-channels-create`) have been altered.
```
[...]

PR #1709, opened 2021-03-17
Close DTLS when fingerprint verification fails

Sean-Der:
Before we would set the PeerConnection to failed, but we would leave the
DTLSTransport. This means that a user could still interact with the
other transports.

Relates to #1708

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/pion/dtls/v2"
 	"github.com/pion/dtls/v2/pkg/crypto/fingerprint"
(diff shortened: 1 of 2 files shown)


Sean-Der:
@jech mind giving this a look over!

This fixes the issue. But also wanted to make you aware of the issue.

Sean-Der:
I applied a CVE for this. Lots of companies watch that so gives people more notice.

When we tag + get CVE-ID I will announce on Slack/Twitter. Just want to make sure people update as soon as possible.

OrlandoCo (approved):

PR #1716, opened 2021-03-19
Fix deadlock between RTPSender and TrackLocalStaticRTP

at-wat:
RTPSender.mu.Lock() and TrackLocalStaticRTP.mu.Lock() caused deadlock
when continuously create and close PeerConnection.

### DIFF
--- a/track_local_static.go
+++ b/track_local_static.go
@@ -128,12 +128,14 @@ func (s *TrackLocalStaticRTP) WriteRTP(p *rtp.Packet) error {
 
 // writeRTP is like WriteRTP, except that it may modify the packet p
 func (s *TrackLocalStaticRTP) writeRTP(p *rtp.Packet) error {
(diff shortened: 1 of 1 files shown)


track_local_static.go
```diff
@@ -128,12 +128,14 @@ func (s *TrackLocalStaticRTP) WriteRTP(p *rtp.Packet) error {
 
 // writeRTP is like WriteRTP, except that it may modify the packet p
 func (s *TrackLocalStaticRTP) writeRTP(p *rtp.Packet) error {
+	bindings := make([]trackBinding, len(s.bindings))
```
Sean-Der, line 131:
```
bindings := append([]trackBinding(nil), s.bindings...)
```

It is unfortunate we need to copy :( what is the deadlock? Curious if we could fix another way

at-wat:
The deadlock was occurred in my custom interceptor. Closing.

Issue #1711, opened 2021-03-18
Codec negotiation fails for Firefox


at-wat:
Codec negotiation logic has been changed by https://github.com/pion/webrtc/commit/e5c8c659caa915041dd6dbcaacbb85fe28733820#diff-8b596789ae952f46740e3dabb4df2a5391953429f8f84da7f58820feb5838ae3R111-R114.
It requires that `SDPFmtpLine` should be empty to be partially matched.

For example, Firefox's fmtp line for OPUS codec is `maxplaybackrate=48000;stereo=1;useinbandfec=1`
and pion/webrtc's default fmtp line for OPUS codec is `minptime=10;useinbandfec=1`
https://github.com/pion/webrtc/blob/545613dcdeb5dedb01cce94175f40bcbe045df2e/mediaengine.go#L68
So, OPUS codec won't be negotiated between Firefox and pion/webrtc>=v3.0.13. (unless registering custom codec to `MediaEngine`)

## Possible solutions
1. Make `SDPFmtpLine` of default OPUS codec empty
    - Make them partially matched on the current negotiation logic.
    - User may enable/disable inband FEC, so setting `useinbandfec=1` by default seems wrong.
2. Fix `codecParametersFuzzySearch`
    - For this Firefox case, it is natural to treat `minptime=10;useinbandfec=1` and `maxplaybackrate=48000;stereo=1;useinbandfec=1` are partially matched.
    - Parse contents of `SDPFmtpLine` and treat as partially-matched if two `SDPFmtpLine`s are not incompatible.

Sean-Der:
Thanks for fixing this FireFox issues @at-wat and @jsmouret. I am sorry I haven't been more present I am taking a few weeks off to work on https://webrtcforthecurious.com. If any emergencies come up I am here to help though!

I added @jsmouret to the GitHub org so you can approve each other PRs. Sorry for the breakage :(

at-wat:
Another problem is that when browser requests
- mime=`audio/opus` fmtp=`minptime=10;useinbandfec=1`
- mime=`audio/G722`
and pion requests
- mime=`audio/opus`
- mime=`audio/G722`
, G722 will be always selected.

I think most user expect that `audio/opus` is selected by first priority.
[...]

PR #1717, opened 2021-03-19
Remove OPUS default FMTP

at-wat:
User may enable/disable inband FEC and use other frame.

Also, it works around the codec negotiation failure between pion and Firefox. Firefox default OPUS FMTP is `maxplaybackrate=48000;stereo=1;useinbandfec=1`, so they weren't matched by the current codecParametersFuzzySearch().

### Reference issue
Part of #1711

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


Sean-Der (approved):

at-wat:
replaced by #1724

PR #1693, opened 2021-03-02
Fix mediaengine negotiation when SDPFmtpLine doesn't match

davidzhao:
#### Description

When the remote peer supports a codec with a SDPFmtpLine that local MediaEngine doesn't support, it blindly adds remote offered codec into the negotiated list.

This causes issues Pion peer to accept a codec that it cannot support.. For example, say we supported H264 baseline:

```
		{
			RTPCodecCapability: webrtc.RTPCodecCapability{MimeType: webrtc.MimeTypeH264, ClockRate: 90000, SDPFmtpLine: "level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f", RTCPFeedback: videoRTCPFeedback},
			PayloadType:        102,
		}
```

but remote SDP included the following

```
a=rtpmap:96 H264/90000
a=rtcp-fb:96 goog-remb
a=rtcp-fb:96 transport-cc
a=rtcp-fb:96 ccm fir
a=rtcp-fb:96 nack
a=rtcp-fb:96 nack pli
a=fmtp:96 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=640c1f
a=rtpmap:98 H264/90000
a=rtcp-fb:98 goog-remb
a=rtcp-fb:98 transport-cc
a=rtcp-fb:98 ccm fir
a=rtcp-fb:98 nack
a=rtcp-fb:98 nack pli
a=fmtp:98 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f
```

The current negotiated output would include both offered codecs, even though only the latter is supported

#### Solution

The negotiation process should be using partial matches if exact matches cannot be found. When exact matches are present, they should be used exclusively.

PR #1724, opened 2021-03-20
Parse and compare FmtpLine on codec select

at-wat:
Exact match if fmtp parameters are not inconsistent.

e.g. default OPUS fmtp of the browsers are:
- Chrome:  `minptime=10;useinbandfec=1`
- Firefox: `maxplaybackrate=48000;stereo=1;useinbandfec=1`

They should be treated as matched.

### Reference issue
#1711

### DIFF
--- a/rtpcodec.go
+++ b/rtpcodec.go
@@ -97,19 +97,19 @@ const (
 // Used for lookup up a codec in an existing list to find a match
 // Returns codecMatchExact, codecMatchPartial, or codecMatchNone
 func codecParametersFuzzySearch(needle RTPCodecParameters, haystack []RTPCodecParameters) (RTPCodecParameters, codecMatchType) {
+	needleFmtp := parseFmtp(needle.RTPCodecCapability.SDPFmtpLine)
+
 	// First attempt to match on MimeType + SDPFmtpLine
-	// Exact matches means fmtp line cannot be empty
 	for _, c := range haystack {
 		if strings.EqualFold(c.RTPCodecCapability.MimeType, needle.RTPCodecCapability.MimeType) &&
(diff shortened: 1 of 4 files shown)


rtpcodec.go
```diff
@@ -97,19 +97,19 @@ const (
...
 	for _, c := range haystack {
 		if strings.EqualFold(c.RTPCodecCapability.MimeType, needle.RTPCodecCapability.MimeType) &&
-			c.RTPCodecCapability.SDPFmtpLine == needle.RTPCodecCapability.SDPFmtpLine {
+			fmtpConsist(needleFmtp, parseFmtp(c.RTPCodecCapability.SDPFmtpLine)) {
 			return c, codecMatchExact
 		}
 	}
 
-	// Fallback to just MimeType if either haystack or needle does not have fmtpline set
+	// Fallback to just MimeType
```
at-wat, line 110: I'm not sure this is valid.

Sean-Der: :+1: This is my fault. This didn't come from a spec, I was just trying not to break older clients. Was hoping to support cases where we get an offer with only the MimeType. Would be nice if we could keep it (but understand why it has to go if it breaks active use cases)

https://github.com/pion/webrtc/pull/1693#issuecomment-789396416

at-wat:
@davidzhao what do you think? Test cases you added are satisfied as well in this version.

Sean-Der (approved):

davidzhao:
> @davidzhao what do you think? Test cases you added are satisfied as well in this version.

looks good to me. thanks for fixing it!

davidzhao (approved):

PR #1727, opened 2021-03-21
Handle errors properly in mux readLoop

### LINKED ISSUE #1720: If an incoming authenticated packet exceeds receiveMTU we tear down the demuxer
Just hit this because I was testing locally with ffmpeg. I am not sure where the error handling is failing.

Sean-Der:
Before io.ErrShortBuffer and packetio.ErrTimeout would incorrectly end
the read loop. Now they are only printed.

Resolves #1720

### DIFF
--- a/internal/mux/mux_test.go
+++ b/internal/mux/mux_test.go
@@ -1,13 +1,18 @@
 package mux
 
 import (
+	"io"
 	"net"
 	"testing"
 	"time"
 
 	"github.com/pion/logging"
+	"github.com/pion/transport/packetio"
 	"github.com/pion/transport/test"
+	"github.com/stretchr/testify/assert"
 )
+
+const testPipeBufferSize = 8192
 
 func TestStressDuplex(t *testing.T) {
 	// Limit runtime in case of deadlocks
@@ -34,40 +39,26 @@ func stressDuplex(t *testing.T) {
 		MsgCount: 100,
 	}
 
-	err := test.StressDuplex(ca, cb, opt)
-	if err != nil {
-		t.Fatal(err)
-	}
+	assert.NoError(t, test.StressDuplex(ca, cb, opt))
 }
 
 func pipeMemory() (*Endpoint, net.Conn, func(*testing.T)) {
 	// In memory pipe
 	ca, cb := net.Pipe()
 
-	matchAll := func([]byte) bool {
-		return true
-	}
-
-	config := Config{
+	m := NewMux(Config{
 		Conn:          ca,
-		BufferSize:    8192,
+		BufferSize:    testPipeBufferSize,
 		LoggerFactory: logging.NewDefaultLoggerFactory(),
-	}
+	})
 
-	m := NewMux(config)
-	e := m.NewEndpoint(matchAll)
+	e := m.NewEndpoint(MatchAll)
 	m.RemoveEndpoint(e)
-	e = m.NewEndpoint(matchAll)
+	e = m.NewEndpoint(MatchAll)
 
 	stop := func(t *testing.T) {
-		err := cb.Close()
-		if err != nil {
-			t.Fatal(err)
-		}
-		err = m.Close()
-		if err != nil {
-			t.Fatal(err)
-		}
+		assert.NoError(t, cb.Close())
+		assert.NoError(t, m.Close())
 	}
 
 	return e, cb, stop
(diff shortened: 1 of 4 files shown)


internal/mux/muxfunc.go
```diff
@@ -43,30 +33,12 @@ func MatchRange(lower, upper byte) MatchFunc {
...
-
 // MatchDTLS is a MatchFunc that accepts packets with the first byte in [20..63]
 // as defied in RFC7983
 func MatchDTLS(b []byte) bool {
 	return MatchRange(20, 63)(b)
 }
 
-// MatchTURN is a MatchFunc that accepts packets with the first byte in [64..79]
-// as defied in RFC7983
-func MatchTURN(b []byte) bool {
```
jeremija, line 66: This and above functions seem unrelated to the PR, i assume they just weren't in use?

constants.go
```diff
@@ -7,8 +7,9 @@ const (
 	unknownStr = "unknown"
 	ssrcStr    = "ssrc"
 
-	// Equal to UDP MTU
-	receiveMTU = 1460
+	// Raised to 1500 because this is ffmpeg's max MTU by default
+	// See pion/webrtc#1720 for full discussion
+	receiveMTU = 1500
```
jeremija, line 12: I actually thought that the MTU was supposed to be 1200 for WebRTC?

Firefox sets the max video MTU to 1200[1], and so does Chrome[2]:

1. https://searchfox.org/mozilla-central/source/dom/media/webrtc/libwebrtcglue/VideoConduit.h#49
2. https://source.chromium.org/chromium/chromium/src/+/master:third_party/webrtc/media/base/media_constants.cc;l=19?q=1200%20mtu&ss=chromium

That said, I'm not an expert on the matter. I was just relying on this fact when building a server-to-server transport for Peer Calls to prefix each RTP/RTCP packet with room name over a single port (but maybe there's a better way of doing this).

jeremija: I haven't tried this, but looks like ffmpeg has an option to specify the MTU:

> You can use "pkt_size" option when using udp as output (I don't know why it
is not the default).
> 
>     ffmpeg -i xxxx ....... udp://xx.xx.xx.xx?pkt_size=1316

Source: https://ffmpeg.org/pipermail/ffmpeg-user/2014-September/023233.html

wdouglass: the mtu really should rely on the underlying network hardware... 1200 is a decent default, and should work in most places, but ideally it'd be discovered https://en.wikipedia.org/wiki/Path_MTU_Discovery

internal/mux/mux_test.go
```diff
@@ -1,14 +1,23 @@
...
 	"github.com/pion/transport/test"
+	"github.com/stretchr/testify/assert"
 )
 
+const testPipeBufferSize = 8192
+
+func matchAll([]byte) bool {
+	return true
+}
+
```
wdouglass, line 20: Should matchAll be moved into the `mux` package, and exposed? seems like something people will want to do often...

internal/mux/muxfunc.go
```diff
@@ -8,16 +8,6 @@ import (
...
-// MatchAll always returns true
-func MatchAll(b []byte) bool {
-	return true
-}
-
-// MatchNone always returns false
-func MatchNone(b []byte) bool {
-	return false
-}
-
```
wdouglass, line 20: Why remove this and move it into test? Maybe i'm just not understanding something here...

constants.go
at-wat: Just FYI, 3GPP reported some typical MTU values valid for mobile networks. (usually, mobile networks have smaller MTU than wired internet connections)

> ## Annex C (informative): Link MTU considerations
>  The link MTU value that can prevent fragementation in the backbone network between the MS and GGSN/PGW depends on the actual deployement. Based on the above calculation a link MTU value of 1358 is small enough in most of the network deployements.
>  Note that using a link MTU value smaller than necessary would decrease the efficiency in the network. Moreover a UE may also apply some tunnelling (e.g., DSMIPv6 or VPN). and it is desired to use a link MTU size that assures at least 1280 octets, which is the minimum MTU size in case of IPv6, within the UE tunnel to avoid the fragmentation of the user packets within the tunnel applied in the UE.

ref: 3GPP TS 23.060 https://www.3gpp.org/ftp/Specs/archive/23_series/23.060/23060-g00.zip

internal/mux/muxfunc.go
Sean-Der: Yea! I just deleted anything that was unused. Since it is in `/internal` we should be the only users.

internal/mux/mux_test.go
Sean-Der: Fixed! I deleted this and brought back the other function.

internal/mux/muxfunc.go
Sean-Der: Just an accident on my part. Since this is all in `/internal` just wanted to clean stuff up :)

constants.go
Sean-Der: @jeremija you are right! I was helping a user in Slack, I ended up forcing the `pkt_size` [here](https://github.com/Sean-Der/goexamples/blob/webrtc/rpt-broadcast/publisher/ffmpeg.sh#L3) and it fixed everything.

This is in response to a user who was doing `ffmpeg -> Pion -> Pion`. They wanted to do everything in their LAN so MTU doesn't matter as much to them.

Before the read loop would exit without an error. Now it prints.

```
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/03/22 08:36:11 mux: failed to read from packetio.Buffer short buffer
```

---

@wdouglass this isn't something we can fix with PMTU unfortunately. WebRTC doesn't support it, and in this case someone was just using the `Send` API so they are free to send w/e they want.

Sean-Der:
@at-wat @jeremija @wdouglass I changed the receieve MTU back to 1460. That works today and don't want to possibly raise memory usage/impact others needlessy. We get a nice error message now, so that should be enough!

wdouglass (approved):
lookin good!

jeremija (approved):

PR #1731, opened 2021-03-24
Minor fix for h264 reader and writer

PieerePi:


Sean-Der:
Fantastic work @PieerePi!

Would you mind adding a test. That will make sure someone doesn't break this in the future by accident. I can merge right away after that.

Sean-Der (approved):

PR #1736, opened 2021-03-27
peerconnection: Export NewRTPSender()

zyxar:
#### Description

This adds a public method `NewRTPSender` to ease creating `RTPSender` which could facilitate setting track upon a selected `Transceiver`.

#### Motivation
Currently `AddTrack` doesn't provide any options for selecting a particular `Transceiver` to add media track; instead, it blindly chooses an unused one sequentially, or adds a new one if existing ones don't fit. 
By exporting `NewRTPSender`, application can just traverse pre-added `Transceiver`s, and add the created RTP sender on a chosen one.

#### Alternative
It is also possible to do it if an alternative `AddTrack` method is added, with ability to filter or choose a proper `Transceiver`.

#### Note
`NewRTPReceiver` also exported.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1899,6 +1899,11 @@ func (pc *PeerConnection) Close() error {
 	return util.FlattenErrs(closeErrs)
 }
 
+// NewRTPSender creates a new RTPSender for track
+func (pc *PeerConnection) NewRTPSender(track TrackLocal) (*RTPSender, error) {
+	return pc.api.NewRTPSender(track, pc.dtlsTransport)
+}
+
 func (pc *PeerConnection) newRTPTransceiver(
 	receiver *RTPReceiver,
 	sender *RTPSender,
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1904,6 +1904,11 @@ func (pc *PeerConnection) NewRTPSender(track TrackLocal) (*RTPSender, error) {
 	return pc.api.NewRTPSender(track, pc.dtlsTransport)
 }
 
+// NewRTPReceiver creates a new RTPReceiver for kind
+func (pc *PeerConnection) NewRTPReceiver(kind RTPCodecType) (*RTPReceiver, error) {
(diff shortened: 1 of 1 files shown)


Sean-Der:
Hey @zyxar 

Thanks for opening this PR! Would love to support this functionality, just a few questions first :)

* Is this supported by [webrtc-pc](https://w3c.github.io/webrtc-pc/)? I try and make sure the API matches JS so we can easily compile to WASM. I don't think this matches today (but I could be wrong). If we really get stuck it is ok to diverge IMO.
* What happens in the browser if you set a `recvonly` Transceiver to `sendrecv`? Does it automatically create a `RTPSender`? Then we could do `ReplaceTrack` after that?

zyxar:
Hi @Sean-Der,
- I don't think it's in webrtc-pc either.
- the thing is not on direction or `RTPSender`, but about *selecting* a particular `Transceiver` and exporting `NewRTPSender` can help achieve it.


E.g, there're 5 transceivers pre-added by offerer, and now answerer (in some circumstances) needs to choose the last one (without adding new ones) to send back a certain media stream: `ReplaceTrack` can help in this case, but answerer would have to add some dummy tracks before reaching the desired one. Any other ways to do it?

Sean-Der:
Sorry coming back to this @zyxar. I will make sure to unblock your use case this weekend :( 

* Offerer offers 5 sendrecv transceivers
* Answerer then creates 5 recvonly transceivers
* You want to put a Track on the last Transceiver of the answerer?

zyxar:
For some reason we don't want to dynamically create transceivers and we use pre-added ones by offerer.

Simplified as below:
Offer: (browser or mobile client)
1. audio sendrecv
2. video sendonly
3. video recvonly
4. video recvonly
5. video recvonly

Our unnamed SFU is the other side.

zyxar:
And we need a proper way to distinguish video stream, from camera or from screen

zyxar:
After some internal discussion, I think technically we can make our SFU work without exporting `NewRTPSender()`. 

I'm closing it now.

PR #1737, opened 2021-03-28
PeerConnection: make AddTrack thread-safe

### LINKED ISSUE #1574: Race when replacing track
### Your environment.
  - Version: master (7b1a63243a47e08f713360a481b86861e2d4a650)
  - Browser: no browser

### What did you do?

When replacing tracks multiple threads, AddTrack can return the same free transceiver and try to establish different tracks in it.

Test
```go
func TestRaceReplaceTrack(t *testing.T) {
	lim := test.TimeOut(time.Second * 30)
	defer lim.Stop()

	pc, err := NewPeerConnection(Configuration{})
	assert.NoError(t, err)

	addTrack := func() {
		track, err := NewTrackLocalStaticSample(RTPCodecCapability{MimeType: "video/vp8"}, "foo", "bar")
		assert.NoError(t, err)
		_, err = pc.AddTrack(track) // line 710 invalid state change in RTPTransceiver.setSending
		assert.NoError(t, err)
	}

	for i := 0; i < 100; i++ {
		addTrack()
	}
	for _, tr := range pc.GetTransceivers() {
		assert.NoError(t, pc.RemoveTrack(tr.Sender()))
	}

	var wg sync.WaitGroup
	for i := 0; i < 100; i++ {
		wg.Add(1)
		go func() {
			addTrack() // line 724
			wg.Done()
		}()
	}

	wg.Wait()
	assert.Equal(t, 100, len(pc.GetTransceivers()))
}
```

### What did you expect?
```log
Running tool: C:\Go\bin\go.exe test -timeout 120s -run ^TestRaceReplaceTrack$ github.com/pion/webrtc/v3

--- FAIL: TestRaceReplaceTrack (0.08s)
    d:\projects\github.com\pion\webrtc\peerconnection_media_test.go:710: 
        	Error Trace:	peerconnection_media_test.go:710
        	            				peerconnection_media_test.go:724
        	            				asm_amd64.s:1373
        	Error:      	Received unexpected error:
        	            	invalid state change in RTPTransceiver.setSending
        	Test:       	TestRaceReplaceTrack
    d:\projects\github.com\pion\webrtc\peerconnection_media_test.go:710: 
        	Error Trace:	peerconnection_media_test.go:710
        	            				peerconnection_media_test.go:724
        	            				asm_amd64.s:1373
        	Error:      	Received unexpected error:
        	            	invalid state change in RTPTransceiver.setSending
        	Test:       	TestRaceReplaceTrack
    d:\projects\github.com\pion\webrtc\peerconnection_media_test.go:710: 
        	Error Trace:	peerconnection_media_test.go:710
        	            				peerconnection_media_test.go:724
        	            				asm_amd64.s:1373
        	Error:      	Received unexpected error:
        	            	invalid state change in RTPTransceiver.setSending
        	Test:       	TestRaceReplaceTrack
    d:\projects\github.com\pion\webrtc\peerconnection_media_test.go:710: 
        	Error Trace:	peerconnection_media_test.go:710
        	            				peerconnection_media_test.go:724
        	            				asm_amd64.s:1373
        	Error:      	Received unexpected error:
        	            	invalid state change in RTPTransceiver.setSending
        	Test:       	TestRaceReplaceTrack

FAIL
FAIL	github.com/pion/webrtc/v3	0.374s
FAIL
```

### What happened?
Replacing all tracks without errors

zyxar:
#### Description
Use fine granularity lock to protect `rtpTransceivers` in  `PeerConnection` which shall make `AddTrack()` and `RemoveTrack()` thread-safe.

#### Reference issue
Fixes #1574

Note: 
- the added test `TestPeerConnection_RaceReplaceTrack` is from #1574
- not quite sure about the remaining `GetTransceivers()` callings in `peerconnection.go`; seems they are used as snapshots or cache, may still have data race in this PR.
- previously the coarse lock `mu` is used almost everywhere for exclusive access in `PeerConnection`; changing to a separated lock for `rtpTransceivers` might have implicit conflicts, which I'm not 100% sure by far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Reese](https://github.com/figadore)
 * [David Zhao](https://github.com/davidzhao)
 * [Nam V. Do](https://github.com/namvdo)
+* [Markus Tzoe](https://github.com/zyxar)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -571,7 +571,10 @@ func (pc *PeerConnection) getStatsID() string {
 }
 
 func (pc *PeerConnection) hasLocalDescriptionChanged(desc *SessionDescription) bool {
-	for _, t := range pc.GetTransceivers() {
+	pc.mu.Lock()
+	defer pc.mu.Unlock()
+
+	for _, t := range pc.rtpTransceivers {
 		m := getByMid(t.Mid(), desc)
 		if m == nil {
 			return true
@@ -1532,11 +1535,10 @@ func (pc *PeerConnection) ICEConnectionState() ICEConnectionState {
 }
 
 // GetSenders returns the RTPSender that are currently attached to this PeerConnection
-func (pc *PeerConnection) GetSenders() []*RTPSender {
+func (pc *PeerConnection) GetSenders() (result []*RTPSender) {
 	pc.mu.Lock()
 	defer pc.mu.Unlock()
 
-	result := []*RTPSender{}
 	for _, transceiver := range pc.rtpTransceivers {
 		if transceiver.Sender() != nil {
 			result = append(result, transceiver.Sender())
@@ -1573,7 +1575,8 @@ func (pc *PeerConnection) AddTrack(track TrackLocal) (*RTPSender, error) {
 	}
 
 	var transceiver *RTPTransceiver
-	for _, t := range pc.GetTransceivers() {
+	pc.mu.Lock()
+	for _, t := range pc.rtpTransceivers {
 		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil {
 			transceiver = t
 			break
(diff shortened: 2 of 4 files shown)


Sean-Der:
Fantastic! Thanks for taking this on :) I added you to the organization and I hope we can together a lot in the future. It is hard finding people to contribute as this code base gets more complicated.

What is stopping us from using one lock? Do things work if you take the global PeerConnection lock and then access `pc.rtpTransceivers` directly? I worry when have two mutexs they can get locked on each other. We have had issues with that in the past.

Also if you see anything that is worthy of refactoring feel free to change. Even if it has no behavior changes. I am totally in for the risk of bugs for an increase in code quality!

zyxar:
Thanks.

> What is stopping us from using one lock? Do things work if you take the global PeerConnection lock and then access `pc.rtpTransceivers` directly? I worry when have two mutexs they can get locked on each other. We have had issues with that in the past.

Using global PC lock can also fulfil this but requires some other refactoring. I can try this way. Then `RTPTransceiver` may need refactoring too. Some issues with `RTPTransceiver`:
- `stopped` seems not used;
- using atomic values doesn't make it's methods ( e.g. `SetSender`) thread-safe.

So what do you think of using lock to protect `RTPTransceiver` instead of atomic values?

Sean-Der:
I think using one lock is the way we should go. Unless we have issues with lock contention we can re-visit. We just are going to get into having tag values as `GUARDED_BY` etc.. if we start having multiple locks. We also will need to worry about interlocking. These aren't intractable, but there is a long term cost.

I believe `stopped` is used a lot in `peerconnection.go`. I think it would help readibility if you made it an accessor!

You are right about `SetSender` not being thread safe. I think there is more nuance here though. 
* When we do things that are transactional (more then one operation) we need to take a lock.
* When we do things that are atomic get/set a value. We should use atomics.

zyxar:
> I believe `stopped` is used a lot in `peerconnection.go`. I think it would help readibility if you made it an accessor!
> 

ya, a lot of checking `stopped` but I cannot find nowhere setting an explicit value, so it's always the default `false`.
The `Stop` method doesn't set it to `true`, but can be easily fixed by an `atomicBool`.

> You are right about `SetSender` not being thread safe. I think there is more nuance here though.
> 
> * When we do things that are transactional (more then one operation) we need to take a lock.
> * When we do things that are atomic get/set a value. We should use atomics.

So I'd make another PR for `RTPTransceiver` later.

Sean-Der (approved):

Sean-Der:
Fantastic work @zyxar 

I merged. This made me realize a few other things.

* AddTrack doesn't hold the lock for the entire function. We may still have issues lurking. Would love to see that become a `defer unlock`
* AddTransceiver* functions aren't thread safe.

Want me to make a ticket? Interested in improving this a little bit more, this is super important stuff :)

zyxar:
sure :)

PR #1738, opened 2021-03-30
FAIL: Add certificate save & load

daonb:
This PR adds `GetOrCreateCertificat('cert.pem')` letting programs keep a consistent certificate.

#### Description

Sorry for sending a failed PR as my first, but i'm stuck.

This PR includes the function and it creates a pem file, writes the certificate and private key as blocks, reads it back and parses it. The code seems to be working fine and the test verifies the private key is properly stored and loaded. I just can't get the certificate to be equal. I suspect it has to do with some dark x509 magic and I need a pro.

### DIFF
--- a/certificate.go
+++ b/certificate.go
@@ -5,14 +5,19 @@ package webrtc
 import (
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"math/big"
+	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/pion/dtls/v2/pkg/crypto/fingerprint"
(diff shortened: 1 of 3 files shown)


Sean-Der:
Hey @daonb 

Thanks for the PR! `webrtc.Configuration` does allow you to pass a certificate, does something like [this](https://github.com/pion/offline-browser-communication/blob/master/server.go#L48) give you what you need?

I don't know if this PR will be portable to WASM (but the other one is)

daonb:
> `webrtc.Configuration` does allow you to pass a certificate, does something like [this](https://github.com/pion/offline-browser-communication/blob/master/server.go#L48) give you what you need?

Thanks! You helped me find the bug - I used `NewCertificate` instead of `CertificateFromX509` and it works!

I use the new function to get/create the certificate and pass it to `webrtc.Configuration` so loading the certificate doesn't have to be part of pion. Saving is more complicated as `webrtc.Certificate` fields are private. 


> I don't know if this PR will be portable to WASM (but the current pattern is)

I believe we can store the pem file in local storage but I'm not sure you'd want to do that. Whatsmore, I guess you don't like having pion writing and reading to the file system. 

Maybe a better approach will be to add:

- `webrtc.CertificateFromPEM` accepting a string with the contents of a pem file 
- `webrtc.Certificate.ToPEM` serializing the certificate into a string in pem format. The string will have 2 blocks - one for the the x509 and one for the private key.

PR #1742, opened 2021-04-01
Add CertificateFromPEM and PEM() to Certificate

Sean-Der:
Allow for easier usage of a pre-existing certificate

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [David Zhao](https://github.com/davidzhao)
 * [Nam V. Do](https://github.com/namvdo)
 * [Markus Tzoe](https://github.com/zyxar)
+* [Benny Daon](https://github.com/daonb)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/certificate.go
+++ b/certificate.go
@@ -11,8 +11,10 @@ import (
 	"crypto/x509/pkix"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/pem"
 	"fmt"
 	"math/big"
+	"strings"
 	"time"
 
 	"github.com/pion/dtls/v2/pkg/crypto/fingerprint"
@@ -145,7 +147,7 @@ func GenerateCertificate(secretKey crypto.PrivateKey) (*Certificate, error) {
 		BasicConstraintsValid: true,
 		NotBefore:             time.Now(),
 		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
-		NotAfter:              time.Now().AddDate(0, 1, 0),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
 		SerialNumber:          serialNumber,
 		Version:               2,
 		Subject:               pkix.Name{CommonName: hex.EncodeToString(origin)},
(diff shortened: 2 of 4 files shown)


Sean-Der:
Are you depending on the ability to extract the Certificate from the PeerConnection with `(c Certificate) PEM()` ?

We can't duplicate/port this to WASM unfortunately. I am 100% for `CertificateFromPEM`, but `PEM()` will cause some portability issues. Happy to add if it is a show stopper for you!

daonb:
I'm almost finished with the signaling server and one of the things it's doing is keeping the peers' fingerprints in the db and pass them together with the offer/answer. If the server's certificate is not persistent there is no way to verify the server's identity so it is a show stopper for me. 

That is unless you have a way of keeping a persistent certificate.

BTW - are you sure it doesn't work on WASM? I've tried running the test using `GOOS=js GOARCH=wasm go test -exec="./test-wasm/go_js_wasm_exec" -run PEM .` and it passes.

Sean-Der:
Right now certificate_test.go isn't executed because it starts with `// +build !js`. 

The issue is that the browser doesn't give us the private key for a `RTCCertificate` so we can't provide the `PEM` API. 

So you goal is to generate one certificate and share between all PeerConnections on the server? Would this work?

* Generate certificate yourself (basically copy `GenerateCertificate`) and save to PEM yourself
* Call `GetFingerprints` on that certificate and save to DB
* For each connection call `CertificateFromPEM`
* Pass that into the `Configuration`

daonb:
You're right, I can do all of this in my program. Problem is, the next user that needs certificate consistency will have to figure it out himself and it's hard work. 

Ceritifcates are one of those dark corners of code where there are too many standards, documentation and conflicting terms. I started this PR hoping to help future users save time and frustration.

I understand the browser's limitation - in terminal7 I use indexDB to get the fingerprint - and as pion uses the builtin RTCCertificate there's no way to access the private key. However, if certificate.go is not compiled under wasm I don't see the harm in providing a couple of helper functions. It's already a superset of w3c's interface and I don't see any harm if making it more super.

Sean-Der:
@daonb that's a good point :) here is what I am thinking. If you disagree/think I am worried about things that don't matter I say we merge!

* I don't want to embrace/extend/extinguish the WebRTC API. I want users to be able to switch between Pion and other implementations. If we add a non-standard API it will make it harder to switch and/or force others to implement non-standard things.
* I don't want to make a security mistake. They made a concious decision to not make the PrivateKey exportable in the browser. If we do that and something bad happens will look bad for Pion.

If you think my fears are unfounded merge away!

daonb:
Thanks for letting me decide. I am a firm believer in the python proverb: "We're all consenting adults" and that's why I want you to merge. 

To recap, this PR is about adding certificate serialization. webexec uses it for persistent certificate but it doesn't matter. How the program uses it is up to the programmer. He's an adult and we should treat him as such.

Regarding WASM, it's a different world. In the browser we're not treated as adults. In the browser I had to go through hoops and use low-level IndexDB to get the public key. The private one is well hidden and javascript can't access it.

In the server world, we have a file system and a program can read the private key. When the user runs a program he makes it an "adult" and gives it access to a system wider and deeper than the browser. These capabilities can lead programmers to need more interfaces from pion. WASM compatbility shouldn't stop us from supporting them.

Sean-Der:
@daonb I am 100% in support then! 

Mind fixing the merge conflict, then you can approve+merge. I added you to the Pion organization. I can't merge this PR since I am the one that opened it.

daonb:
Sure. While I'm at it, I want to add mysekf to the authors. Mind if I copy
akl te names to a AUTHERS.txt

בתאריך יום ב׳, 5 באפר׳ 2021, 18:04, מאת Sean DuBois ‏<
***@***.***>:

Sean-Der:
@daonb yes please! You will need to also

* Remove https://github.com/pion/.goassets/blob/master/ci/.github/workflows/lint.yaml#L25-L26
* Modify https://github.com/pion/.goassets/blob/master/ci/.github/assert-contributors.sh to generate `AUTHORS.txt` (not used by CI though)
* Push a new tag

Then you can see all the CI assets update PRs https://github.com/search?q=org%3Apion+type%3Apr+author%3Apionbot+is%3Aopen&type=Issues

We can split the PRs of moving from README.md -> AUTHORS.md

daonb (approved):

daonb:
I'll be happy to do some house cleaning - my projects needs better linting and such and it looks like you got it covered.

 just fixed the conflict and tried to push to pion/webrtc and got permission denied. Maybe the branch is protected?

Sean-Der:
@daonb sorry you have to hit the `Rebase and merge` button on this page. Just to make sure all the tests pass. If you force push to `pem-certificate` branch then we should be good to go!

daonb:
Only problem is, I never accepted the invitation to join pion and now it's expired. Please invite me again.

Sean-Der:
Done!

Sean-Der:
@daonb I am not sure if the timing change is ok. 1 months comes from libwebrtc [here](https://source.chromium.org/chromium/chromium/src/+/master:third_party/webrtc/rtc_base/ssl_identity.h;l=39-43;drc=8ca5d3a5f1d3e18b363549c0edd4c2494cfb70ea)

I haven't found anything in the IETF that dictates certificate length. I don't believe this put users at risk, but I am not sure.

Sean-Der:
Yea this is specified in the W3C [here](https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-generatecertificate) would you mind changing the timing back and we can add a `expires` argument that respects the timing information?

Sean-Der:
It would be best to make it a struct. Frustrating that Javascript does type dispatch.

Sean-Der:
Or we could make it an optional parameter.  WebRTC is capped at 365 days, and the default is 30 days. I am going to revert the date change just to be safe.

daonb:
Sorry about wasting your time. I failed to document the change in the PR description, which was careless. My excuse is that I've been working alone on my app for too long and I got used to change constants too lightly.

Regarding the `expire`, I thought about it some more and IMO pion doesn't need it. If the server wants to generate a permanent certificate he can copy  `GenerateCertificate` to his project and change it there. I can add a wiki page about permanent certificates with code for generating a long-lasting one.

Sean-Der:
You haven't wasted a single second of my time. I am SUPER grateful for your involvement, I hope you stay around long term. Sorry if my communication style comes off as annoyed :(

PR #1747, opened 2021-04-06
Fixes DataChannel panic on sctp transport race

zyxar:
#### Description
`DataChannel.open` would panic during which if `PeerConnection` is closed,
stopping underlying `SCTPTransport` which sets association to `nil`;

`DataChannel.ensureSCTP` method doesn't guarantee `SCTPTransport`'s availability out of it's own scope.

Also fix similar issue in `SCTPTransport.Start`.

#### Reference issue
Fixes #1730

PR #1748, opened 2021-04-06
SCTPTransport: add getAssociation

zyxar:
#### Description
Use assocation() instead of accessing private struct member

member variable `association` in SCTPTransport renamed to `sctpAssociation`

This is a follow-up commit to #1747

### DIFF
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -338,11 +338,7 @@ func TestPeerConnection_Media_Disconnected(t *testing.T) {
 			// Assert that DTLS is done by pull remote certificate, don't tear down the PC early
 			for {
 				if len(vp8Sender.Transport().GetRemoteCertificate()) != 0 {
-					pcAnswer.sctpTransport.lock.RLock()
-					haveAssociation := pcAnswer.sctpTransport.association != nil
-					pcAnswer.sctpTransport.lock.RUnlock()
-
-					if haveAssociation {
+					if pcAnswer.sctpTransport.association() != nil {
 						break
 					}
 				}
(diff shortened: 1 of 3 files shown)


zyxar:
or is it better to rename `getAssocation()` as `Assocation()`?

Sean-Der:
@zyxar that works for me! I would leave it private, I believe Go will have a conflict then though method+member named `assocation`

peerconnection_media_test.go
```diff
@@ -339,7 +339,7 @@ func TestPeerConnection_Media_Disconnected(t *testing.T) {
 			for {
 				if len(vp8Sender.Transport().GetRemoteCertificate()) != 0 {
 					pcAnswer.sctpTransport.lock.RLock()
```
Sean-Der, line 341: Do we still need this lock?

sctptransport.go
```diff
@@ -369,3 +366,13 @@ func (r *SCTPTransport) generateAndSetDataChannelID(dtlsRole DTLSRole, idOut **u
 
 	return &rtcerr.OperationError{Err: ErrMaxDataChannelID}
 }
+
+func (r *SCTPTransport) Association() *sctp.Association {
```
Sean-Der, line 370: I don't believe we should export this. It isn't part of the ORTC ond webrtc-pc API

peerconnection_media_test.go
zyxar: ah, no longer needed.

sctptransport.go
zyxar: 👌
renamed to unexported `association()`

Sean-Der (approved):

Sean-Der:
:fire: 

Love these improvements. Feel free to file any improvements you think would be good. Even if they are big projects we can take them on together.

PR #1759, opened 2021-04-13
Introduce UDPMux in SettingsEngine, single-port

davidzhao:
When set, Pion will handle ICE connections over UDPMux


Sean-Der:
Hey @davidzhao this PR looks really good to me!

If you have the time adding an example would be really great. That would make it easy for users to find/use the feature.

davidzhao:
@Sean-Der I've closed this for now since discovering a few other things that I might need to do in Ice. will re-open this once I figured it out.

PR #1760, opened 2021-04-13
Fixes SDP session id for subsequent offers/answers

### LINKED ISSUE #1751: SDP session id should stay unchanged for subsequent offers/answers
### Your environment.
  - Version: latest/master
  - Browser: N/A
  - Other Information - N/A

### What did you do?
```go
	offer, err := pcOffer.CreateOffer(nil)
	if err != nil {
		return err
	}
	println(offer.parsed.Origin.SessionID)
	offerGatheringComplete := GatheringCompletePromise(pcOffer)
	if err = pcOffer.SetLocalDescription(offer); err != nil {
		return err
	}
	<-offerGatheringComplete
	if err = pcAnswer.SetRemoteDescription(*pcOffer.LocalDescription()); err != nil {
		return err
	}

	answer, err := pcAnswer.CreateAnswer(nil)
	if err != nil {
		return err
	}
	answerGatheringComplete := GatheringCompletePromise(pcAnswer)
	if err = pcAnswer.SetLocalDescription(answer); err != nil {
		return err
	}
	<-answerGatheringComplete
	return pcOffer.SetRemoteDescription(*pcAnswer.LocalDescription())
```

When using the above the negotiation and re-negotiation, the session id (in the `o=` line) is changed every time:
```
6134566055764263477
3690620680687464608
7746218356774445024
955905295255239035
5184175515672053010
6786050404964233948
4553936517277751569
8511417983963443586
2762366857331776204
7954399364685800967
```

### What did you expect?
session id should stay unchanged.

For subsequent offers:
```
o  The fields of the "o=" line MUST stay the same except for the
      <session-version> field, which MUST increment by one on each call
      to createOffer if the offer might differ from the output of the
      previous call to createOffer; implementations MAY opt to increment
      <session-version> on every call.  The value of the generated
      <session-version> is independent of the <session-version> of the
      current local description; in particular, in the case where the
      current version is N, an offer is created and applied with version
      N+1, and then that offer is rolled back so that the current
      version is again N, the next generated offer will still have
      version N+2.
```

and subsequent answers:
```
o  The fields of the "o=" line MUST stay the same except for the
      <session-version> field, which MUST increment if the session
      description changes in any way from the previously generated
      answer.
```

zyxar:
#### Description
This PR makes session id unchanged for subsequent offers/answers and session-version incremented by one on each call

#### Reference issue
Fixes #1751


Sean-Der:
@zyxar Nice work! Mind removing unsafe usage then it is an approve from me.

peerconnection.go
```diff
@@ -55,6 +56,8 @@ type PeerConnection struct {
 	lastOffer  string
 	lastAnswer string
 
+	sdpOrigin *sdp.Origin
```
Sean-Der, line 59: Could we pull this from the `currentLocalDescription` or `pendingLocalDescription`? Would love to find a way to not add more state.

zyxar: using values from `currentLocalDescription` or `pendingLocalDescription` could somehow replace `PeerConnection.updateSDPOrigin()` method;

however, this way doesn't fully follow the spec for _session version_:
>The value of the generated \<session-version\> 
is _independent_ of the \<session-version\> of the
      _current local description_; in particular, in the case where the
      current version is N, an offer is created and applied with version
      N+1, and then that offer is rolled back so that the current
      version is again N, the next generated offer will still have
      version N+2.

before calling `SetLocalDescription()` (or `setDescription()` precisely) `currentLocalDescription` or `pendingLocalDescription` are still `nil`; and session version should be incremented by one regardless _current_ local description's version.

And currently these methods in `PeerConnection` are not well protected by lock:
- `CreateOffer`
- `CreateAnswer`
- `AddTransceiverFromKind`
- `AddTransceiverFromTrack`

But of course the lock issues could be solved later.

Sean-Der: :+1: ok yea I think one extra field is not a big deal at all considering what else we need to untangle.

I do feel strongly about the use of `unsafe` though.

Sean-Der:
Can we just take the PeerConnection lock in CreateOffer/CreateAnswer/SetLocalDescription.

Then this should just be a getter for Createoffer/CreateAnswer, and is only incremented when calling SetLocalDescription?

zyxar:
> Can we just take the PeerConnection lock in CreateOffer/CreateAnswer/SetLocalDescription.
> 
> Then this should just be a getter for Createoffer/CreateAnswer, and is only incremented when calling SetLocalDescription?

The spec is about `CreateOffer`/`CreateAnswer` not `SetLocalDescription`?
Say, multiple calls of `CreateOffer` would have same session version before `SetLocalDescription` is called (though very rare case in real world)?

Another approach maybe is we don't call `sdp.NewJSEPSessionDescription` every time to generate SDP but try to reuse local one. In that case, shall we add a public method for `sdp.SessionDescription` for this?

peerconnection.go
zyxar: I removed the `unsafe` usage :)

Sean-Der (approved):
LGTM! Nice work @zyxar

PR #1761, opened 2021-04-14
New wasm js

cgojin:
1. Use wasm_exec.js in GOROOT to ensure that the wasm file generated by go build is loaded successfully. There is no need to use a specific version of go to build wasm.
2. Delete vendor-wasm.
3. Fix typo in examples.


Sean-Der:
This is awesome! So glad we can get rid of this, never even looked into why we needed it in the first place.

If you can fix the commit message I am happy to merge right away.

Do we still need [test-wasm](https://github.com/pion/webrtc/tree/master/test-wasm) ?

cgojin:
> This is awesome! So glad we can get rid of this, never even looked into why we needed it in the first place.
> If you can fix the commit message I am happy to merge right away.
> Do we still need [test-wasm](https://github.com/pion/webrtc/tree/master/test-wasm) ? Could we have a `testmain` for WASM tests that does this instead? It adds some overhead having this custom testing infra.


Your reply is so fast, thank you.
I haven't tested [test-wasm](https://github.com/pion/webrtc/tree/master/test-wasm), just found that demo.wasm could not be loaded today.

Sean-Der:
This is an amazing PR, great work @cgojin 

I really appreciate that you took the time to clean this up! So great to have a PR that is just deleting code (with zero features lost)

Sean-Der (approved):

PR #1763, opened 2021-04-15
Introduce UDPMux in SettingsEngine, single-port

davidzhao:
When set, Pion will handle ICE connections over UDPMux


davidzhao:
Sean, I know you mentioned adding examples for this.. Let me know what the lowest friction way to do so is. I was thinking docs around this might be a better way to go, since the configuration part would be only ~5 lines of code.

Sean-Der (approved):

PR #1765, opened 2021-04-16
Fixes peerconnection: potential deadlock

zyxar:
#### Description
Potential deadlock:
In `CurrentLocalDescription()` or `PendingLocalDescription()`:
when gathering local candidates in `populateLocalCandidates()` (while holding PeerConnection lock),
agent would try submitting a task via `agent.run()`:
```go
	a.chanTask <- task
```
the `chanTask` is consumed by `taskLoop` routine:
```go
	for {
		select {
		case <-a.done:
			return
		case t := <-a.chanTask:
			t.fn(a.context(), a)
			close(t.done)
			after()
		}
	}
```
while there're other routines using the agent, e.g. calling
`agent.updateConnectionState()`
```go
	a.afterRun(func(ctx context.Context) {
		a.chanState <- newState
	})
```
and the submitted task would be run by the `after()` func in
`taskLoop()`, where the `a.chanState` is consumed by
`agent.startOnConnectionStateChangeRoutine()`

this `after()` task may block the `taskLoop` routine when there's ongoing
event handler `agent.onConnectionStateChange()`, which would
eventually call `PeerConnection`'s `onICEConnectionStateChange()`:
```go
	pc.mu.Lock()
	pc.iceConnectionState = cs
	handler := pc.onICEConnectionStateChangeHandler
	pc.mu.Unlock()
```
trying to hold the PeerConnection lock, causing _deadlock_

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1923,8 +1923,11 @@ func (pc *PeerConnection) newRTPTransceiver(
 // by the ICEAgent since the offer or answer was created.
 func (pc *PeerConnection) CurrentLocalDescription() *SessionDescription {
 	pc.mu.Lock()
-	defer pc.mu.Unlock()
-	return populateLocalCandidates(pc.currentLocalDescription, pc.iceGatherer, pc.ICEGatheringState())
+	localDescription := pc.currentLocalDescription
+	iceGather := pc.iceGatherer
+	iceGatheringState := pc.ICEGatheringState()
+	pc.mu.Unlock()
+	return populateLocalCandidates(localDescription, iceGather, iceGatheringState)
 }
 
 // PendingLocalDescription represents a local description that is in the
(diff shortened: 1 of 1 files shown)


zyxar:
accidentally encountered this deadlock in `TestAddDataChannelDuringRenegotation()`

zyxar:
btw why 
func (pc *PeerConnection) CurrentRemoteDescription() *SessionDescription {
	return pc.currentRemoteDescription
}

zyxar:
BTW, why `CurrentRemoteDescription` and `PendingRemoteDescription` are not protected by lock while local ones are?
Is it intentional or bug?

```go
func (pc *PeerConnection) CurrentRemoteDescription() *SessionDescription {
	return pc.currentRemoteDescription
}

func (pc *PeerConnection) PendingRemoteDescription() *SessionDescription {
	return pc.pendingRemoteDescription
}
```

Sean-Der (approved):
Fantastic catch! That is really scary how easy this is to mess up, nothing should block on a task run when holding the PeerConnection lock.

I wonder if there is something we could write (maybe just tooling that examines the stack?) that could catch this.

zyxar:
By intuition I don't think the after() tasks should be run by the taskLoop, but rather another dedicated routine. But I'm not 100% sure about this idea.

PR #1770, opened 2021-04-20
Fix docker file

cgojin:
1. Removes "host" network_mode 

Steps to reproduce:
$ cd examples/pion-to-pion
$ docker-compose up
  ERROR: for answer  "host" network_mode is incompatible with port_bindings

2. Fix undefined os.ErrDeadlineExceeded

Steps to reproduce:
$ cd examples/pion-to-pion
$ docker-compose up
  #4 13.95 pkg/mod/github.com/pion/ice/v2@v2.1.2/udp_mux.go:242:22: undefined: os.ErrDeadlineExceeded

### DIFF
--- a/examples/pion-to-pion/answer/Dockerfile
+++ b/examples/pion-to-pion/answer/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.14
+FROM golang:1.16
 
 ENV GO111MODULE=on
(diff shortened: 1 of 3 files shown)


Sean-Der:
Thank you so much for fixing this @cgojin !

We should find a way to add this to CI

PR #1782, opened 2021-04-22
samplebuilder keep outdated

ovoshlook:
#### Description
For some cases, when restoring media from the PCAP files, for example, there is no need to drop outdated frames as it already filtered when handled when was stored at the PCAP file by the receiver.

this pull request adds an additional property/flag to sample builder "keepOutdatedPackets" to keep those packets

#### Reference issue
This is not an issue. It is a feature.


ovoshlook:
not needed

PR #1821, opened 2021-05-22
Fix RtpSender: may block infinite in Read

cnderrauber:
In RtpSender, if Read has been called and wait for srtpReady,
then if we call Close() immediately after
rtpSender.trasport.srtpReady, that means srtpWriterFuture's Close
may be called when it's init method is executing.

In Read -> init, goroutine may run at:

	rtpWriteStream, err := srtpSession.OpenWriteStream()
	if err != nil {
		return err
	}

	s.rtcpReadStream.Store(rtcpReadStream)
	s.rtpWriteStream.Store(rtpWriteStream)

In Close, goroutine may run at

	if value := s.rtcpReadStream.Load(); value != nil {
		return value.(*srtp.ReadStreamSRTCP).Close()
	}

	return nil

s.rtcpReadStream could be nil in the check at Close, so it
will not be closed.

And in Read method, the init method initiates the stream and
return. Then read at the stream's buffer, and blocked infinite
because the buffer is not closed. And it will not be closed
if you call RtpSender's close method because it is already
closed.

#### Reference issue
Fixes #...

### DIFF
--- a/srtp_writer_future.go
+++ b/srtp_writer_future.go
@@ -4,6 +4,7 @@ package webrtc
 
 import (
 	"io"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Sean-Der (approved):

Sean-Der:
@cnderrauber sorry this took so long!

PR #1823, opened 2021-05-25
Fix: infinite loop possible in sample builder

### LINKED ISSUE #1810: Infinite loop -> OOM in SampleBuilder edge case
If the s.buildSample call inside of purgeBuffers happens to find a complete packet, s.active is updated to s.consume.tail https://github.com/pion/webrtc/blob/d625f6f6373b6e7140baee0957777035f67b02da/pkg/media/samplebuilder/samplebuilder.go#L227, but then the loop condition is triggered again in purgeBuffers and s.active is reset back to s.filled, causing the same packet to be unmarhsaled over and over in a tight loop. I changed the `continue` to `return` which seems to fix this behavior for me, but there might be a more correct solution

### Your environment.
  - Version: webrtc master https://github.com/pion/webrtc/commit/d625f6f6373b6e7140baee0957777035f67b02da
  - Browser: Chrome 
  - Using samplebuilder in ion-avp

### What did you do?
Saw this issue in SampleBuilder used in ion-avp connected to ion-sfu, caused by a vp8 stream sent from chrome specifically when some minor packet reordering occurred.

### What did you expect?
Reconstruct packets even with reordering
### What happened?
Infinite loop + OOM

robin-raymond:
Active becomes "empty" after consuming the packets, but instead of the
"filled" getting purged by "purgeConsumedBuffers", it doesn't because
active is now empty and empty buffers cannot cause any purging. So the
solution is to purge the "consumed" area first, then "active".

#### Reference issue
Fixes #1810

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -112,13 +112,23 @@ func (s *SampleBuilder) releasePacket(i uint16) {
 // purgeConsumedBuffers clears all buffers that have already been consumed by
 // popping.
 func (s *SampleBuilder) purgeConsumedBuffers() {
-	s.purgeConsumedLocation(s.active)
+	s.purgeConsumedLocation(s.active, false)
 }
 
 // purgeConsumedLocation clears all buffers that have already been consumed
 // during a sample building method.
-func (s *SampleBuilder) purgeConsumedLocation(consume sampleSequenceLocation) {
-	for consume.compare(s.filled.head) == slCompareBefore && s.filled.hasData() {
+func (s *SampleBuilder) purgeConsumedLocation(consume sampleSequenceLocation, forceConsume bool) {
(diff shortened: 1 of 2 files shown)


Sean-Der:
:fire: 

Your reasoning sounds perfect to me! Would it be possible to add a test that would have failed/OOM before? Even if it touches private APIs better then nothing :)

robin-raymond:
Yes, I think I can/will. It's a bit more complicated because it only happens in a perfect storm of things but obviously it's frequent enough if two people hit it.

robin-raymond:
@Sean-Der I added a test which would fail without this fix as you requested.

Sean-Der:
@robin-raymond fantastic! Thank you so much :)

Sean-Der:
Closed by f7f760f8c6e2f1aba1bf7c544fdd84d70f6e05b6 (squashed into one)

PR #1829, opened 2021-06-04
Adds SetCodecPreferences to RTPTransceiver

digitalix:
#### Description

Adds option to set chosen codecs on individual Transceiver.

### DIFF
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -43,18 +43,27 @@ func newRTPTransceiver(
 // SetCodecPreferences sets preferred list of supported codecs
 // if codecs is empty or nil we reset to default from MediaEngine
 func (t *RTPTransceiver) SetCodecPreferences(codecs []RTPCodecParameters) error {
+	mediaEngineCodecs := t.me.getCodecsByKind(t.kind)
+
 	if codecs == nil || len(codecs) < 1 {
 		t.mu.Lock()
 		defer t.mu.Unlock()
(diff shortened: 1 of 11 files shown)


rtptransceiver.go
```diff
@@ -16,6 +17,10 @@ type RTPTransceiver struct {
 	receiver  atomic.Value // *RTPReceiver
 	direction atomic.Value // RTPTransceiverDirection
 
+	me     *MediaEngine
```
Sean-Der, line 20: Can you embed `api *API` instead? That is just the common pattern across `RTC*` objects we have right now.

rtptransceiver.go
```diff
@@ -106,6 +185,11 @@ func (t *RTPTransceiver) Stop() error {
 }
 
 func (t *RTPTransceiver) setReceiver(r *RTPReceiver) {
+	if r != nil {
+		t.mu.RLock()
+		r.codecs = t.codecs
```
Sean-Der, line 190: Is it possible that these could get out of date? Maybe we should store the `RTPTransceiver` on the `RTPReceiver` and  `RTPSender` as well.

digitalix: You're right, it's better if we store the `RTPTransceiver ` on the `RTPReceiver` and `RTPSender`. Done.

Sean-Der:
Moved to PR to branch I have push access on https://github.com/pion/webrtc/pull/1848

PR #1831, opened 2021-06-13
Implemented RTPReceiver SetRTPParameters for ORTC

Antonito:
See discussion with @OrlandoCo https://gophers.slack.com/archives/CAK2124AG/p1623504625397700

- RID is now set on remote tracks, if provided.
- Adds SetRTPParameters which can be used in ORTC to set codec information about the remote track.

I'm not sure if this is the best way to implement this, as the track will still miss the `id` and `streamID` + I'm not a big fan of adding public method for edge cases. 
Maybe it'd be better to  refacto `RTPReceiver.Receive()` to look more like `RTPSender.Send()`?

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -114,7 +114,7 @@ func (r *RTPReceiver) Receive(parameters RTPReceiveParameters) error {
 			track: newTrackRemote(
 				r.kind,
 				parameters.Encodings[0].SSRC,
-				"",
+				parameters.Encodings[0].RID,
 				r,
 			),
 		}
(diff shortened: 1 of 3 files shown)


OrlandoCo:
> I'm not sure if this is the best way to implement this, as the track will still miss the `id` and `streamID` + I'm not a big fan of adding public method for edge cases.

Looks good to me, but since this is not a standard W3C method then I guess we can add the track and stream ID, we can extend `RTPParameters` to include them. Any thoughts @Sean-Der ??

OrlandoCo (approved):

OrlandoCo:
@Sean-Der Any comments before merge this?

Sean-Der:
Sorry for the delay @Antonito And @OrlandoCo 

Code looks good! Could you add a test before merging? Thanks

Issue #1809, opened 2021-05-13
Potential deadlock : createoffer when ice connection state changing


cnderrauber:
### Your environment.
  - Version: fork from v3.0.26
  - Browser:  client use pion
  - Other Information - *stacktraces, related issues, suggestions how to fix, links for us to have context*

```
goroutine 1
5574     2  0x00000000009382e5 in github.com/pion/ice/v2.(*Agent).run
5575         at /home/luog/go/pkg/mod/github.com/pion/ice/v2@v2.1.6/agent.go:177
5576     3  0x000000000093c0f9 in github.com/pion/ice/v2.(*Agent).GetLocalUserCredentials
5577         at /home/luog/go/pkg/mod/github.com/pion/ice/v2@v2.1.6/agent.go:859
5578     4  0x000000000096dc8a in github.com/pion/webrtc/v3.(*ICEGatherer).GetLocalParameters
5579         at /home/luog/go/pkg/mod/github.com/cnderrauber/webrtc/v3@v3.0.24/icegatherer.go:200
5580     5  0x0000000000985d3d in github.com/pion/webrtc/v3.(*PeerConnection).generateMatchedSDP
5581         at /home/luog/go/pkg/mod/github.com/cnderrauber/webrtc/v3@v3.0.24/peerconnection.go:2216
5582     6  0x000000000097a21f in github.com/pion/webrtc/v3.(*PeerConnection).CreateOffer

goroutine 2
5517     3  0x0000000000951405 in github.com/pion/ice/v2.(*Agent).updateConnectionState.func1
5518         at /home/luog/go/pkg/mod/github.com/pion/ice/v2@v2.1.6/agent.go:559
5519     4  0x0000000000950719 in github.com/pion/ice/v2.(*Agent).taskLoop.func1
5520         at /home/luog/go/pkg/mod/github.com/pion/ice/v2@v2.1.6/agent.go:196
5521     5  0x0000000000938485 in github.com/pion/ice/v2.(*Agent).taskLoop
5522         at /home/luog/go/pkg/mod/github.com/pion/ice/v2@v2.1.6/agent.go:225


goroutine 3
5551      7  0x000000000097919a in github.com/pion/webrtc/v3.(*PeerConnection).onICEConnectionStateChange
5552          at /home/luog/go/pkg/mod/github.com/cnderrauber/webrtc/v3@v3.0.24/peerconnection.go:467
[...]

PR #1834, opened 2021-06-15
Fixed deadlock in peerconnection.go

digitalix:
#### Description

In some rare cases during the ice connection stage change may result in deadlock. This fix makes iceConnectionState and connectionState atomic which should prevent deadlock.

Before this fix, some of the tests would randomly timeout, after this, I'm unable to reproduce the issue.
Also, I believe this is the same problem as described in https://github.com/pion/webrtc/issues/1809.

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/1809

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -45,8 +45,8 @@ type PeerConnection struct {
 	currentRemoteDescription *SessionDescription
 	pendingRemoteDescription *SessionDescription
 	signalingState           SignalingState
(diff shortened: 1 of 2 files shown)


Sean-Der:
Woah this is a fantastic PR @digitalix 

Sorry to slow you down with my slow review. I really hope you stay involved :) will try my best to catch up

PR #1846, opened 2021-06-26
Implement DTLS restart

### LINKED ISSUE #1636: DTLS should be restarted when the remote DTLS fingerprint has been changed or the tls-id has changed during applying an answer
### Your environment.
  - Version: *v3.0.3*

### What did you do?
  - Make a pair of `PeerConnection`s, one is at server side, and another one is at App side.
  - The App is crashed or is swipe out from foreground.
  - When the App is back, I would need to create a fresh `PeerConnection` and pair it with the one at server side.
  - The pairing is not working, ICE is connected but not the DTLS part
    - The connection-state of `PeerConnection` at server side shows "connected".
    - The connection-state of `PeerConnection` at App side turns to "failed".

### What did you expect?
  - the connection between them should be able to establish

### What happened?
  - The `PeerConnection` at server side didn't restart its DTLS handshaking. Per JSEP spec 5.11, `PeerConnection` should restart DTLS when corresponding parameters changed in remote session-description as offer.
```
      If the remote DTLS fingerprint has been changed or the tls-id has
      changed, tear down the DTLS connection.  This includes the case
      when the PeerConnection state is "have-remote-pranswer".  If a
      DTLS connection needs to be torn down but the answer does not
      indicate an ICE restart or, in the case of "have-remote-pranswer",
      new ICE credentials, an error MUST be generated.  If an ICE
      restart is performed without a change in tls-id or fingerprint,
      then the same DTLS connection is continued over the new ICE
      channel.  Note that although JSEP requires that answerers change
      the tls-id value if and only if the offerer does, non-JSEP
      answerers are permitted to change the tls-id as long as the offer
      contained an ICE restart.  Thus, JSEP implementations which
      process DTLS data prior to receiving an answer MUST be prepared to
      receive either a ClientHello or data from the previous DTLS
      connection.
```

We can consider to implement this part in our `PeerConnection.SetRemoteDescription`.

Antonito:
https://github.com/pion/srtp/pull/148 is needed before being able to merge

~~The unit test needs https://github.com/pion/transport/pull/141~~

Fixes #1636

~~I'm not sure about `if t.state != DTLSTransportStateNew && t.state != DTLSTransportStateClosed`, one other option would be to set the state to `New` before restarting, but I don't want to trigger a StateChanged event – which could be wrong.~~
Turns out that's what libwebrtc does

Also the condition on which DTLS is restarted might be too permissive – but it doesn't seem to break any test so far.


Sean-Der:
@Antonito This is absolutely amazing work, your PRs are the best I have seen in months. Gives me the energy to get back into it :) 

I think we could make the test a lot simpler. 

* Create two PeerConnections and connect them.
* Send one RTP packet
* Create a new PeerConnection and generate an offer, and send it to one of the PeerConnections in the first flow.
* Send one RTP packet

I think that gets us the coverage we need and keeps the LoC low still? Push back if I am wrong!

Antonito:
> @Antonito This is absolutely amazing work, your PRs are the best I have seen in months. Gives me the energy to get back into it :)
> 
> I think we could make the test a lot simpler.
> 
> * Create two PeerConnections and connect them.
> * Send one RTP packet
> * Create a new PeerConnection and generate an offer, and send it to one of the PeerConnections in the first flow.
> * Send one RTP packet
> 
> I think that gets us the coverage we need and keeps the LoC low still? Push back if I am wrong!

So, I tried to rewrite the unit test in a simpler way, and it uncovered several problems. I'm not sure if there are caused by the DTLS restart or were already there (I'm thinking the later, for most of them).

- Sending a single RTP packet doesn't trigger an `OnTrack` event on the other peer. Sending two in a row doesn't either, but sending 1 / waiting a few ms / sendign another one works.
- When ran multiple times in a row, the test sometimes logs `pc ERROR: 2021/06/28 13:30:33 Incoming unhandled RTP ssrc(1595441956), OnTrack will not be fired. single media section has an explicit SSRC`. I initially thought this was a difference in behavior with the test I already pushed (since with your method, the first peer connection doesn't go to `disconnected` or `failed` before the other one connects), but turns out it's also happening with this one – I just had lucky runs. It doesn't seem to be a race condition, tsan didn't detect anything.
- I found out weird behaviors with DataChannels
  - I added a call to `CreateDataChannel`, behavior unchanged (as expected)
  - I then set an empty `OnDataChannel` callback -> this causes the test to sometimes hang (again I'm not doing anything more with DataChannels here)
  - I then tried to send data from `ClientB` and read with `ClientA1` and `ClientA2`, sometime it hangs, but when it doesn't: 
    - As expected, ClientA1 can read
    - After ClientA2 connects, ClientB tries to send data – which causes a `io: read/write on closed pipe` error log
    - ClientA2 can't read, since ClientB can't send anything (this is most likely an issue with DTLS restart)

Antonito:
I rewrote the unit test without pion/transport, and split it in two very similar tests, for debugging purpose:
- TestPeerConnection_DTLS_Restart_MediaOnly
- TestPeerConnection_DTLS_Restart_MediaAndDataChannel

Before merging, we should only keep `TestPeerConnection_DTLS_Restart_MediaAndDataChannel`.

SCTP restart is now implemented and DataChannels are re-opened, if needed.

There seems to be some kind of race condition, but I really don't think it's related to the stuff this PR modifies – I think the unit tests just show an already existing problem.
It's what I was describing above, and now have a bit more validation of what's really happening.

The `TestPeerConnection_DTLS_Restart_MediaOnly` behaves correctly, in `go test`. To make it work with `tsan`, I had to add an extra: https://github.com/pion/webrtc/blob/0d1e50c3d6cc904f52b596f9f8cd6906eb6e10c9/peerconnection_media_test.go#L1329
I think this is a consequence of the first point I was developing in my previous message (about RTP packets).

`TestPeerConnection_DTLS_Restart_MediaAndDataChannel` behaves correctly with `go test`, but hangs with `tsan`. I also tried to add a few extra RTP packets, but it doesn't change any thing. Somehow, the hang seems to be caused by the simple fact of calling `CreateDataChannel` and `OnDataChannel`.

Antonito:
Following up discussion with @davidzhao, I updated both this PR and https://github.com/pion/srtp/pull/148.

In order to get some progress on this, I think we should first focus on merging https://github.com/pion/srtp/pull/148 – we need to figure out @Sean-Der 's concerns about `rolloverCounter` and `lastSequenceNumber`. I was careful and really don't think this one introduces any security issue, but wouldn't mind a second opinion :)

As for this PR, it requires a bit more work, there are still issues with `TestPeerConnection_DTLS_Restart_MediaAndDataChannel`, and I'm far less confident about security issues here. Unfortunately I don't have much time to dedicate to this for now, I'd gladly welcome any help 😁

davidzhao:
Thanks for picking this back up @Antonito! We'd love to collaborate on this. Will have a look in a few weeks.

edaniels:
hey @Antonito, do you still want to work together to get this in?

JoTurk:
surpassed by https://github.com/pion/webrtc/pull/3555 and the new internal DTLS API.

PR #1848, opened 2021-06-27
Implements SetCodecPreferences in RTPTransceiver

### LINKED ISSUE #1847: Implement RTCRtpTransceiver.setCodecPreferences
This will allow users to set on a per Transceiver basis their preferences around codec choice. The `MediaEngine` is a global list supported codecs still.

[WC3 spec](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiver-setcodecpreferences)

```
setCodecPreferences
The setCodecPreferences method overrides the default codec preferences used by the user agent. When generating a session description using either createOffer or createAnswer, the user agent MUST use the indicated codecs, in the order specified in the codecs argument, for the media section corresponding to this RTCRtpTransceiver.

This method allows applications to disable the negotiation of specific codecs (including RTX/RED/FEC). It also allows an application to cause a remote peer to prefer the codec that appears first in the list for sending.

Codec preferences remain in effect for all calls to createOffer and createAnswer that include this RTCRtpTransceiver until this method is called again. Setting codecs to an empty sequence resets codec preferences to any default value.

NOTE
Codecs have their payload types listed under each m= section in the SDP, defining the mapping between payload types and codecs. These payload types are referenced by the m=video or m=audio lines in the order of preference, and codecs that are not negotiated do not appear in this list as defined in section 5.2.1 of [RFC8829]. A previously negotiated codec that is subsequently removed disappears from the m=video or m=audio line, and while its codec payload type is not to be reused in future offers or answers, its payload type may also be removed from the mapping of payload types in the SDP.
The codecs sequence passed into setCodecPreferences can only contain codecs that are returned by RTCRtpSender.getCapabilities(kind) or RTCRtpReceiver.getCapabilities(kind), where kind is the kind of the RTCRtpTransceiver on which the method is called. Additionally, the RTCRtpCodecCapability dictionary members cannot be modified. If codecs does not fulfill these requirements, the user agent MUST throw an InvalidModificationError.

NOTE
Due to a recommendation in [SDP], calls to createAnswer SHOULD use only the common subset of the codec preferences and the codecs that appear in the offer. For example, if codec preferences are "C, B, A", but only codecs "A, B" were offered, the answer should only contain codecs "B, A". However, [RFC8829] (section 5.3.1.) allows adding codecs that were not in the offer, so implementations can behave differently.
When setCodecPreferences() in invoked, the user agent MUST run the following steps:

Let transceiver be the RTCRtpTransceiver object this method was invoked on.

Let codecs be the first argument.

If codecs is an empty list, set transceiver.[[PreferredCodecs]] to codecs and abort these steps.

Remove any duplicate values in codecs. Start at the back of the list such that the priority of the codecs is maintained; the index of the first occurrence of a codec within the list is the same before and after this step.

Let kind be the transceiver's transceiver kind.

If the intersection between codecs and RTCRtpSender.getCapabilities(kind).codecs or the intersection between codecs and RTCRtpReceiver.getCapabilities(kind).codecs only contains RTX, RED or FEC codecs or is an empty set, throw InvalidModificationError. This ensures that we always have something to offer, regardless of transceiver.direction.

Let codecCapabilities be the union of RTCRtpSender.getCapabilities(kind).codecs and RTCRtpReceiver.getCapabilities(kind).codecs.

For each codec in codecs,

If codec is not in codecCapabilities, throw InvalidModificationError.
Set transceiver.[[PreferredCodecs]] to codecs.

NOTE
If set, the offerer's codec preferences will decide the order of the codecs in the offer. If the answerer does not have any codec preferences then the same order will be used in the answer. However, if the answerer also has codec preferences, these preferences override the order in the answer. In this case, the offerer's preferences would affect which codecs were on offer but not the final order.
```

Sean-Der:
This allows to set supported codecs per transceiver.

Resolves #1847

----

@digitalix Instead of slowing down the review process I am just gonna take a few mins and fix up myself! Feel free to push to this branch also. Will leave notes on what I changed. If you agree when I am done go ahead and merge :)

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1052,7 +1052,7 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 					localDirection = RTPTransceiverDirectionSendonly
 				}
 
-				t = newRTPTransceiver(receiver, nil, localDirection, kind)
+				t = newRTPTransceiver(receiver, nil, localDirection, kind, pc.api)
 				pc.mu.Lock()
 				pc.addRTPTransceiver(t)
 				pc.mu.Unlock()
@@ -1634,7 +1634,7 @@ func (pc *PeerConnection) newTransceiverFromTrack(direction RTPTransceiverDirect
 	if err != nil {
 		return
 	}
-	return newRTPTransceiver(r, s, direction, track.Kind()), nil
+	return newRTPTransceiver(r, s, direction, track.Kind(), pc.api), nil
 }
 
 // AddTransceiverFromKind Create a new RtpTransceiver and adds it to the set of transceivers.
@@ -1668,7 +1668,7 @@ func (pc *PeerConnection) AddTransceiverFromKind(kind RTPCodecType, init ...RTPT
 		if err != nil {
 			return nil, err
 		}
-		t = newRTPTransceiver(receiver, nil, RTPTransceiverDirectionRecvonly, kind)
+		t = newRTPTransceiver(receiver, nil, RTPTransceiverDirectionRecvonly, kind, pc.api)
 	default:
 		return nil, errPeerConnAddTransceiverFromKindSupport
 	}
@@ -2208,7 +2208,7 @@ func (pc *PeerConnection) generateMatchedSDP(transceivers []*RTPTransceiver, use
 				t, localTransceivers = satisfyTypeAndDirection(kind, direction, localTransceivers)
 				if t == nil {
 					if len(mediaTransceivers) == 0 {
-						t = &RTPTransceiver{kind: kind}
+						t = &RTPTransceiver{kind: kind, api: pc.api, codecs: pc.api.mediaEngine.getCodecsByKind(kind)}
 						t.setDirection(RTPTransceiverDirectionInactive)
 						mediaTransceivers = append(mediaTransceivers, t)
 					}
--- a/errors.go
+++ b/errors.go
@@ -202,6 +202,7 @@ var (
 
 	errRTPTransceiverCannotChangeMid        = errors.New("errRTPSenderTrackNil")
 	errRTPTransceiverSetSendingInvalidState = errors.New("invalid state change in RTPTransceiver.setSending")
+	errRTPTransceiverCodecUnsupported       = errors.New("unsupported codec type by this transceiver")
 
 	errSCTPTransportDTLS = errors.New("DTLS not established")
 
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pion/rtp"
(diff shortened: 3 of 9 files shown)


Sean-Der:
* Renamed RTPTransceiver Codec -> getCodecs
  - W3C doesn't have this method. Want to avoid divergence where possible

* Made rtpPayloadTypeBitmask a constant

* Remove determinePayloadType and use checkAndUpdateTrack instead

* Added error handling to checkAndUpdateTrack, before we discarded

peerconnection.go
```diff
@@ -1014,6 +1014,12 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 	detectedPlanB := descriptionIsPlanB(pc.RemoteDescription())
 	weOffer := desc.Type == SDPTypeAnswer
 
+	for _, tr := range localTransceivers {
```
Sean-Der, line 1017: Do we need this? I thought `SetCodecPreferences` was for a user to explicitly set codecs only.

Sean-Der: @digitalix

digitalix: @Sean-Der I think we do. This was needed in case if a peer uses different payload ids for codecs. For example, if the payload id for VP8 in Safari is 100 but in Chrome it's 96. So it will fail if a user uses SetCodecPreferences before SetRemoteDescription.

digitalix: Alternatively, we can set new codecs as "pending" and apply them during re-negotiation. I think this is how it works in libwebrtc.

Sean-Der: @digitalix ah I seel

So I think the right fix is to update `getCodecs` to pull the proper PayloadTypes from the `MediaEngine`?

Similiar to what `SetCodecPreferences` does right now. Maybe we can even drop the input sanitization in `SetCodecPreferences` and filter invalid codecs in getCodecs?

If that is right I can implement, add a test and merge right now!

digitalix: @Sean-Der I think I should have placed a comment there to clarify why it's there :). 

```So I think the right fix is to update getCodecs to pull the proper PayloadTypes from the MediaEngine?```
This is how I had it at the beginning, but I didn't like it that it would happen every time we call `getCodecs`. So I came up with something more elegant IMO. If that's fine then go ahead! :) 

```Maybe we can even drop the input sanitization in SetCodecPreferences```
We can, but this would be different to SetCodecPreferences in specs. 
https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences

Additionally, now the user wouldn't know if it actually worked until it's negotiated, but `SetCodecPreferences` in browsers would fail immediately.

Sean-Der: Oh great point about the error handling!

Ok I am going to make the change to `getCodecs` and write a test that confirms filtering + PayloadTypes are correct.

Really fantastic work on this PR, this is a fantastic change.

digitalix: Thanks!

PR #1855, opened 2021-06-28
Fix H264Reader buffer overflow

### LINKED ISSUE #1734: H264Reader.NextNAL crashes on some inputs
### Your environment.
  - Version: v3.0.19

### What did you do?

```
package main

import (
	"bytes"
	"github.com/pion/webrtc/v3/pkg/media/h264reader"
	"log"
)

func main() {
	data := []byte("\x00\x00\x010\x00\x00\x01\x00\x00\x01") // input 1
// or
// 	data := []byte("\x00\x00\x00\x01\x00\x00\x01") // input 2

	r, err := h264reader.NewReader(bytes.NewReader(data))
	if err != nil {
		log.Fatal(err)
	}
	for {
		nal, err := r.NextNAL()
		if err != nil {
			log.Println(err)
			break
		}
		if nal == nil {
			break
		}
	}
}
```

### What did you expect?

No panics.

### What happened?

`panic: runtime error: slice bounds out of range [:-1]` for input 1 and

`panic: runtime error: index out of range [0] with length 0` for input 2

Antonito:
There was a buffer overflow in H264Reader, it is now fixed.

Fixes #1734

### DIFF
--- a/pkg/media/h264reader/h264reader.go
+++ b/pkg/media/h264reader/h264reader.go
@@ -167,8 +167,10 @@ func (reader *H264Reader) processByte(readByte byte) (nalFound bool) {
 				countOfConsecutiveZeroBytesInPrefix = 3
 			}
 			nalUnitLength := len(reader.nalBuffer) - countOfConsecutiveZeroBytesInPrefix
-			reader.nalBuffer = reader.nalBuffer[0:nalUnitLength]
-			nalFound = true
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

Sean-Der:
@Antonito nice, thanks for grabbing this :)

PR #1860, opened 2021-06-30
Fix RTPSender's streamInfo miss headerExtensions

aggresss:
#### Description

This PR fix `transceiver.Sender().Send()` not contain `HeaderExtensions` when `createStreamInfo`, the result is none of HeaderExtensions at interceptor.

### DIFF
--- a/rtpsender.go
+++ b/rtpsender.go
@@ -104,9 +104,7 @@ func (r *RTPSender) Transport() *DTLSTransport {
 	return r.transport
 }
 
-// GetParameters describes the current configuration for the encoding and
-// transmission of media on the sender's track.
-func (r *RTPSender) GetParameters() RTPSendParameters {
+func (r *RTPSender) getParameters() RTPSendParameters {
 	sendParameters := RTPSendParameters{
 		RTPParameters: r.api.mediaEngine.getRTPParametersByKind(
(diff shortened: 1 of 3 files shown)


rtpsender.go
```diff
@@ -175,14 +181,15 @@ func (r *RTPSender) ReplaceTrack(track TrackLocal) error {
 }
 
 // Send Attempts to set the parameters controlling the sending of media.
-func (r *RTPSender) Send(parameters RTPSendParameters) error {
+func (r *RTPSender) Send() error {
```
Sean-Der, line 184: We can't change this API unfortunately. `RTPSender.Send` comes from the [ORTC](https://draft.ortc.org/#dom-rtcrtpsender-send) spec.

In peerconnection.go can we pull the headerExtension values and call `Send` with them?

Sean-Der (changes requested):
If possible pass Header Extension IDs to RTPSender.Send instead of populating in function.

rtpsender.go
aggresss: Copy that, I resubmitted it.

Sean-Der (approved):

Sean-Der:
Thanks for fixing this!

PR #1861, opened 2021-06-30
Fix AddTrack transceiver reuse per W3C specs

### LINKED ISSUE #1843: pion/webrtc v2.2.26 wrong transceivers reuse, cannot re-add removed tracks on receiver side.
### Your environment.
  - Version: Ubuntu 20.04.2 LTS 64-bit - GNOME Version 3.36.8
  - Browser: Google Chrome Version 91.0.4472.77 (Official Build) (64-bit)
  - Used pion versions:
```
github.com/pion/dtls/v2 v2.0.4 // indirect
github.com/pion/quic v0.1.4 // indirect
github.com/pion/rtcp v1.2.6
github.com/pion/sctp v1.7.11 // indirect
github.com/pion/srtp v1.5.2 // indirect
github.com/pion/stun v0.3.5
github.com/pion/transport v0.12.0 // indirect
github.com/pion/turn/v2 v2.0.5
github.com/pion/webrtc/v2 v2.2.26
```

### What did you do?

First of all, thank you for your product and support!

We are using pion to implement our own SFU.

We are trying to figure out why, although Chrome's webrtc-internal stats display packet reception, negotiation, and transceiver status change are correct, the audio/video stream doesn't show when tracks are removed and then re-added.

#### Steps 

The panelist initiates the call with only the webcam video track (to simplify debugging) and publishes it, attendees joins the call with only the microphone audio track.

The panelist's audio/video stream is then added to the attendees subscriptions and correctly displayed.

The broadcast of the panelist's audio/video stream is stopped and the attendees subscriptions are stopped.

Then the panelist's audio/video stream is added again and the subscriptions restored.

I am attaching the log of the negotiations on both the panelist and the attendant side and the stats.

We don't think this is your issue, but it would be great if we could get support to try and understand why the stream isn't being received again.

##### EDIT 2021-06-24

Note that:
- both negotiations logs are copy/pasted from chrome://webrtc-internals and reflect the javascript side of the client.
- you can also see the SDP generated by your go implementation of webrtc on calls to setRemoteDescription of logs
- the add and removal of screen-sharing works correctly even when done multiple times
- the attendee never remove the initial stream with one track coming from the mic
- you can also see that the id of the tracks remains the same at the next addition which to my knowledge is not normal behavior
- we have verified that the `track.stop()` method is not called for any remote track.

Obviously any information and suggestions will be welcome.

We regret not being able to share more information, we are still investigating, and will keep you updated in case we manage to isolate or solve the problem on our side or on your side.

### What did you expect?

The audio/video stream should be visible again.

### What happened?

The audio/video stream is not restored.

[panelist-negotiations.txt](https://github.com/pion/webrtc/files/6704294/panelist-negotiations.txt)
[attendee-negotiations.txt](https://github.com/pion/webrtc/files/6704325/attendee-negotiations.txt)

panelist-outbound-rtpvideo-stats
attendee-inbound-rtpvideo-stats

EdoaLive:
According to W3C specifications, the transceiver can be reused if "The sender has never been used to send. More precisely, the [[CurrentDirection]] slot of the RTCRtpTransceiver associated with the sender has never had a value of "sendrecv" or "sendonly"".
Current implementation does not have CurrentDirection slot, so the flag usedToSend is set on setDirection change.
This fixes #1843 and maybe other issues (for us it fixes other two issues when reusing transceiver, eg clients going and coming back to a conference).
This has been done on the v2 branch because we currently don't have the resources to switch our software to v3 and test it there. Looking at the source (master) seems this kind of issue is still there.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1197,7 +1197,7 @@ func (pc *PeerConnection) AddTrack(track *Track) (*RTPSender, error) {
 
 	var transceiver *RTPTransceiver
 	for _, t := range pc.GetTransceivers() {
-		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil {
+		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil && t.usedToSend.Load() != true {
 			transceiver = t
 			break
 		}
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -14,6 +14,8 @@ type RTPTransceiver struct {
 	receiver  atomic.Value // *RTPReceiver
 	direction atomic.Value // RTPTransceiverDirection
 
+	usedToSend atomic.Value // track if direction has never had a value of "sendrecv" or "sendonly"
+
 	stopped bool
 	kind    RTPCodecType
 }
@@ -89,6 +91,9 @@ func (t *RTPTransceiver) setReceiver(r *RTPReceiver) {
 }
 
 func (t *RTPTransceiver) setDirection(d RTPTransceiverDirection) {
+	if d == RTPTransceiverDirectionSendrecv || d == RTPTransceiverDirectionSendonly {
+		t.usedToSend.Store(true)
+	}
 	t.direction.Store(d)
 }
 


Sean-Der:
Thank you so much for the fix @EdoaLive !

I will write a test for it and merge to `master` and `v2` in the next couple of days.

edaniels:
hey @EdoaLive , do you still want to work together to get this in?

edaniels:
other than merging the PR, this needs some tests to validate the fix since we're dealing with a state machine. If you get those in, I think we'd be good! No rush on that.

EdoaLive:
Do you mean automatic (unit?) testing or real-world testing? In the second case, we used the patched version for years with many users, so there's that. 
In the first case, we should recreate the exact conditions (those explained in issue #1843 I guess), but if the tests use pion for both sides of the negotiation, it might works anyway because both sides don't comply with W3C specifications, the sides "understand" each other in both cases, but with a different protocol or machine state than the one used in browsers.

Anyway I'll get back tho this when (and if) this issue happens with our new version and with pion v3 implementation.

edaniels:
I do mean unit testing to validate the new code doesn't break any old code. I get what you're saying with needing something other than just two pion peers. It would be nice if we had higher level tests for that, but that's definitely too much for this PR.

Issue #1870, opened 2021-07-05
Samplebuilder packet loss on RTP sequence number wrap


kevindejong:
### Your environment.
  - Version: v3.0.30
  - Browser: N/A
  - Other Information -

### What did you do?

I have pion-to-pion connection which is streaming VP8 video and using the `samplebuilder` to reconstruct the samples.  The sender uses an unaltered `VP8Payloader`, `rtp.Packetizer`, and `rtp.RandomSequencer` though it does set the RTP timestamp in the RTP header using a custom heuristic.

### What did you expect?

During testing over a connection on a local network, I expected the `samplebuilder` not to experience loss when reconstructing samples.  For the most part this is working correctly, however I have noticed that the `samplebuilder` consistently reports packet loss when the RTP sequence number wraps.  

### What happened?

After the sequence number hits 65535, I see it wrap to 0 however I don't see any more samples emitted until it reaches a sequence number of 100 (the configured `maxLate`) after which it emits a sample which reports `PrevDroppedPackets` > 0.  (As an aside, it seems that when the sequence number wraps it can take more than `maxLate` packets for the sample to be emitted as it will be emitted on `maxLate` packets after 0, but this does not include the number of packets received before the wrap.)

As an example, here are some (filtered) logs from my application, `packet` is the RTP header timestamp and `seq` is the RTP sequence number 

```
TRAC[0048] received rtp packet                           codec=video/VP8 len=794 packet=2597305381 seq=65529
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=65530
TRAC[0048] assembled sample                               codec=video/vp8 dropped=0 duration=66.666666ms length=10262 packet=2597305381 timestamp="2021-07-05 19:49:19.599995363 +0000 UTC"      
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=65531
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=65532
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=65533
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=65534
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=65535
TRAC[0048] received rtp packet                           codec=video/VP8 len=1188 packet=2597311381 seq=0
[...]

Issue #1839, opened 2021-06-20
Samplebuilder is broken in v3.0.29


jech:
I've just tried upgrading Galene from v3.0.27 to v3.0.29.  After fixing the issue in #1838, I'm getting corrupted video when attempting to save to disk, at least for VP8.
CC @robin-raymond

robin-raymond:
are you seeing `PrevDroppedPackets` > 0 ever?

jech:
I've only seen it once.

robin-raymond:
is it always corrupted or sometimes?

jech:
It's systematically corrupted.  It takes about 20 seconds before it finds a keyframe (we send regular PLIs until we get a keyframe, so that shouldn't happen), and then there are green blocky artifacts that cover between half and the whole frame.  This is on a lossless connection (streaming from localhost).

robin-raymond:
I think we need to capture RTP packet data for replay as a unit test. I'm unsure why it's breaking down but we need something to demonstrate it failing because it does work in other scenarios. You should see every single time stamp that enters gets popped out. Also the total bytes going to the payloads input should equal the total bytes coming out (up to the same timestamp point).

jech:
I've put a log at https://galene.org/samplebuilder-log.text .  It's generated by the following:
```
diff --git a/diskwriter/diskwriter.go b/diskwriter/diskwriter.go
index 9068a20..97aca0c 100644
--- a/diskwriter/diskwriter.go
+++ b/diskwriter/diskwriter.go
@@ -453,6 +453,10 @@ func (t *diskTrack) Write(buf []byte) (int, error) {
 		return 0, nil
 	}
 
+	if strings.EqualFold(codec.MimeType, "video/vp8") {
+		log.Printf("Packet: %v %v (%v)", p.Timestamp, p.SequenceNumber, len(p.Payload))
+	}
+
 	if strings.EqualFold(codec.MimeType, "video/vp9") {
 		var vp9 codecs.VP9Packet
 		_, err := vp9.Unmarshal(p.Payload)
@@ -478,6 +482,10 @@ func (t *diskTrack) Write(buf []byte) (int, error) {
 			return len(buf), nil
 		}
 
+		if strings.EqualFold(codec.MimeType, "video/vp8") {
+			log.Printf("Sample: %v (%v)", ts, len(sample.Data))
+		}
+
 		keyframe := true
 
 		if strings.EqualFold(codec.MimeType, "video/vp8") ||
```

[...]

PR #1873, opened 2021-07-07
Add test to check whether the samplebuilder drops data

jech:
This test is successful with v3.0.27.  In v3.0.31, it reports large numbers of dropped packets.

Related to #1839 and #1870.

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder_test.go
+++ b/pkg/media/samplebuilder/samplebuilder_test.go
@@ -388,3 +388,38 @@ func TestPopWithTimestamp(t *testing.T) {
 		assert.Equal(t, uint32(0), timestamp)
 	})
 }
+
+type truePartitionHeadChecker struct{}
+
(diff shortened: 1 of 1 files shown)


robin-raymond:
There's an issue with this test. The buffers will hold back because it cannot tell if the first packet in the sequence is truly the first packet. Only once the packet count holdback is exceeded then the packets will release (and there will always be a hold back until the end because there's no 'flush' mechanism to say EOF on the stream).

jech:
That's what the partitionHeadChecker is for.

robin-raymond:
Just because it's the first packet that arrived is a partition head doesn't mean there isn't a packet that might arrive later / in front of the current head that is also a head. Every audio sample is a "head" for example.

robin-raymond:
I added your test and made it pass. The issue is during the wrap around which someone else reported not consistently retaining a backlog of packets. You can cancel this PR.

Sean-Der:
That's great news! Thanks @robin-raymond :)

PR #1874, opened 2021-07-07
revert samplebuilder

### LINKED ISSUE #1839: Samplebuilder is broken in v3.0.29

jech:
This revers the samplebuilder changes, which are uselessly complex
and break existing software.

While it is true that the samplewriter could use a rewrite, a new
version should be a complete rewrite that uses data structures with
better cache locality.  The new APIs introduced in this version would
impede writing such a better version.

Fixes #1839.  Probably also #1870.

- Add test to check whether the samplebuilder drops data
- Revert "Fix infinitate loop possible in sample builder"
- Revert "Refactored samplebuilder logic"


jech:
CC @robin-raymond

robin-raymond:
See https://github.com/pion/webrtc/pull/1873 -- There's an issue with the assumptions in the test and how the sample builder works.

jech:
The sample builder used to have a certain behaviour.  Your patch breaks this behaviour, and in doing so has broken working software.

Perhaps you're right, and the semantics you propose is better.  In that case, you should distribute a separate library, or submit a different package to Pion, and perhaps propose that it replace the samplebuilder in a future v4.  What you have done amounts to sneaking in a change in behaviour into a stable version of Pion which caused people's working software to break.

robin-raymond:
If a behaviour was expected then it needs to be codified into a unit test to demonstrate that behaviour as part of the contract. There was no such test to pass. The previous code was buggy. The solution at this point I think is to codify specific start of buffer behaviours for start/end of buffer ranges rather than a revert.

robin-raymond:
Your test passes. You do not need to revert. The issue was rather easy to fix. You can kill this PR.

PR #1875, opened 2021-07-08
Enable VP8 PictureID by default

at-wat:
For using in SLI.

### Reference issue
Requires https://github.com/pion/rtp/pull/142


jech:
I agree, that's the right thing to do, that's what I thought when I originally saw https://github.com/pion/rtp/commit/52ce162689f06702f26a3615254d7c6b8619524f

Suggestion: instead of doing this here, why not simply change the polarity of the option?  Instead of EnablePictureID, have an option DisablePictureID which has opposite polarity, so that the zero value does the right thing.

jech:
See https://github.com/pion/rtp/pull/144

Sean-Der (approved):

PR #1882, opened 2021-07-13
Add WithTailChecker option to depacketizer

jech:
While Galene is now using its own samplebuilder, it would be good
to avoid splitting the community by having the public interfaces
diverge too much.  This means preventing internal details of the
samplebuilder from leaking into other parts of Pion.

An earlier commit sneaked in the IsDetectedFinalPacketInSequence
method to the rtp.Depacketizer interface.  Not only does this
break compatibility with earlier versions, it requires every single
implementation of rtp.Depacketizer to implement essentially the
same code.  It is also conceptually wrong, as the job of finding
the end of the packet belongs in the samplebuilder, not in the
depacketizer.

In this commit, we treat partition ends similarly to partition
starts: we add a new callback to the samplebuilder whose role
is to determine whether a packet ends a partition.  A new option
WithPartitionTailChecker, analogous to WithPartitionHeadChecker,
can be used to set that field.  Tests are adapted.

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -22,8 +22,11 @@ type SampleBuilder struct {
 	// sampleRate allows us to compute duration of media.SamplecA
 	sampleRate uint32
 
-	// Interface that checks whether the packet is the first fragment of the frame or not
+	// determines whether the packet is the first fragment of the frame
 	partitionHeadChecker rtp.PartitionHeadChecker
+
+	// determines whether the packet is the last fragment of a frame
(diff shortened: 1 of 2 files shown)


Sean-Der:
Since we have `IsDetectedFinalPacketInSequence` I am going to close this as wontfix for v3.*

The API disagreement is unfortunate, but I think we need to stick with it until v4. @jech If you feel strongly I am in support of completely reworking the public API for `v4` we have a couple other things for it as well.

jech:
@Sean-Der in that case, I suppose we move the PartitionHeadChecker into the codec too, in order to avoid having an inconsistent API.  I'll see if I have time to submit a PR in time for 3.1.0.

PR #1883, opened 2021-07-13
Simplify sampleSequenceLocation

jech:
Before I decided that doing a full rewrite is the simpler option
I had started doing some cleanup on the code.  Although Galene has
now switched to a less broken samplebuilder, I guess I might as well
submit the bits I had already done.


Sean-Der:
Really great PR! Thanks for cleaning up this code without breaking the API @jech makes my life a lot easier

Sean-Der (approved):

PR #1885, opened 2021-07-13
Fix RTPReceiver getParameters

Antonito:
Modifications included in https://github.com/pion/webrtc/commit/f524fea32aadec9e1b6b06b886181ae67d32ea17 made https://github.com/pion/webrtc/pull/1831 useless:  when used in ORTC context, `tr` might stay `nil`, and a call to `getParameters` would result in a crash in that case.


Sean-Der (approved):
Sorry about that! Thanks for fixing :)

Antonito:
> Sorry about that! Thanks for fixing :)

Thanks for the quick response time ! 😁

PR #1893, opened 2021-07-20
Fixes ReplaceTrack

digitalix:
When ReplaceTrack was set previously to nil it would be impossible to ReplaceTrack again with a non-nil value as r.track was set to nil.


Sean-Der (approved):
Fantastic PR!

PR #1898, opened 2021-07-27
Check for nil transceivers on get parameters

OrlandoCo:
Add public methods to allow set stream and track id, this is useful
when using ORTC API

### DIFF
--- a/ortc_datachannel_test.go
+++ b/ortc_datachannel_test.go
@@ -1,0 +1,64 @@
+// +build !js
+
+package webrtc
+
+import (
+	"io"
+	"testing"
+	"time"
+
+	"github.com/pion/transport/test"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDataChannel_ORTCE2E(t *testing.T) {
+	lim := test.TimeOut(time.Second * 20)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	stackA, stackB, err := newORTCPair()
+	assert.NoError(t, err)
+
+	awaitSetup := make(chan struct{})
+	awaitString := make(chan struct{})
+	awaitBinary := make(chan struct{})
+	stackB.sctp.OnDataChannel(func(d *DataChannel) {
+		close(awaitSetup)
+
+		d.OnMessage(func(msg DataChannelMessage) {
+			if msg.IsString {
+				close(awaitString)
+			} else {
+				close(awaitBinary)
+			}
+		})
+	})
+
+	assert.NoError(t, signalORTCPair(stackA, stackB))
+
+	var id uint16 = 1
+	dcParams := &DataChannelParameters{
+		Label: "Foo",
+		ID:    &id,
+	}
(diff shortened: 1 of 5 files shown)


Antonito:
I agree this is useful – I've been wanting to do this too.

My only concern (which also applies to https://github.com/pion/webrtc/pull/1831) is we keep adding ORTC specific stuff. Couldn't this be solved by refactoring RTPReceiver stuff?
My point is by using ORTC, ideally we should be able to do the exact same stuff as the PeerConnection does – without introducing extra code.

Sean-Der:
I agree with @Antonito 

I worry if we add non-standard APIs users will expect to use them with WebRTC, and could cause bugs. It also makes the GoDoc bigger.

I haven't used the ORTC Media API myself personally. What happens because we don't do this today? Might be good to add examples/tests first.

OrlandoCo:
@Antonito @Sean-Der the issue is that streamID and trackID are set internally there is no standard method to modify them. Even if we refractor RTPReceiver we would need an outside method to set stream/track ID. The easy way is to modify them on track remote rather receivers where can be more confusing with simulcast.

The other way is making TrackRemote an interface and allow setting custom TrackRemotes for Receivers, which would be a little bit more complicated and confusing.

Sean-Der:
@OrlandoCo I spend some time looking through the [ORTC Draft](https://draft.ortc.org/#rtcrtpreceiver*) and I couldn't find how the id/label would get set for a browser implementation. For WebRTC this information is communicated via SDP, but I don't set an equivalent API for ORTC?

So it looks like for ORTC it is just going to be empty? I still would love to design this in a way that we are 'ORTC Compatible'. I think the right thing here is for you to embed tracks inside another struct that contains any metadata you need.

Sean-Der:
@OrlandoCo that other bug fix looks great though! Would you mind adding some minor ORTC Media tests? It doesn't have to be E2E even, just something to start us off with.

OrlandoCo:
@Sean-Der With ORTC is not necessary the streamID or trackID since you have more controls over the way you signal the information you can add whatever identifier you need. But let me explain you why do we need ORTC, for ion-sfu we are building an API for relay tracks from server to server, to make easy this process is done via ORTC so no need to do negotiation signaling by just forwarding a single message.

But ORTC is only for server to server, clients will still be using webrtc. Thats why we need to set correct Track and stream ID.

Fort tests, I can add them for sure !

Sean-Der (approved):

OrlandoCo:
@Sean-Der What we can do to enable setting the Track/Stream ID?

Sean-Der:
I think adding it would hurt users. For people doing WebRTC 

* It would make it seem like it is a value they can control (would they expect it be signaled to the other side?)
* We can't match the behavior in WASM

Then just for ORTC
* We are shipping non-standard APIs. It makes it harder for users to move between implementations

Why not put the `TrackRemote` in another struct and you can put arbitrary labels/identifiers on it?

Sean-Der:
If you feel really strongly/can't accomplish your use case we can do it though.

PR #1903, opened 2021-07-29
Improved h264 fmtp line parsing

rlshumat:
Implements h264 fmtp parsing based on RFC 6184 Section 8.2.2


Sean-Der:
This PR looks fantastic, nice work @rlshumat!

fmtph264.go
```diff
@@ -0,0 +1,109 @@
+package webrtc
```
Sean-Der, line 1: `fmtp_h264.go` a little easier to read maybe?

Sean-Der: actually would it be possible to put this in `internal`?

rlshumat:
@Sean-Der @jech I've added h264 fmtp parsing based on my interpretation of the RFC, which I believe is correct, but would love some feedback on if you think I may have interpreted something incorrectly. I tried to maintain the existing private api as much as possible.

fmtph264.go
rlshumat: I'll look at putting it in internal. I'm not crazy about this solution in general, but I think a better solution would come with more codec support. For example, replacing the MediaEngine.RegisterCodec() arguments with an Interface where the Interface implements the various parsers and other Codec specific details. That way codec support can be external to pion/webrtc and codec implementations can live on their own. That is probably a v4 change though :)

jech:
Cool, thanks.

The method `Compare` should be called `Equal` (`Compare` is usually a function that returns -1, 0, or 1 and represents an ordering).

The method `Parse` does not need to be part of the interface, it's only ever used on concrete types.

The method `Compare` of genericFmtp is the only one to use a value receiver.  Is that intended?  Please either use a pointer receiver everywhere, or use value receivers everywhere except in Parse.

rlshumat:
> The method `Compare` should be called `Equal` (`Compare` is usually a function that returns -1, 0, or 1 and represents an ordering).

I agree, but thinking more about it, comparing the FMTP line isn't necessarily about Equality here, it's more about Compatibility. Do you think `Compatible` or `IsCompatible` would be a better name?

jech:
Perhaps `Match`?

jech:
In generic match, the test is skipped when `ok` is false.  Is this expected?  Here's a test case:
```
diff --git a/internal/fmtp/fmtp_test.go b/internal/fmtp/fmtp_test.go
index 0ec8b56..52e5691 100644
--- a/internal/fmtp/fmtp_test.go Ok
+++ b/internal/fmtp/fmtp_test.go
@@ -98,7 +98,7 @@ func TestGenericFmtpCompare(t *testing.T) {
                },
                "Inconsistent_OneHasExtraParam": {
                        a:       "key1=value1;key2=value2;key3=value3;key4=value4",
-                       b:       "key1=value1;key2=different_value;key3=value3",
+                       b:       "key1=value1;key2=value2;key3=value3",
                        consist: false,
                },
        }
```

rlshumat:
@jech I'm not sure I follow in regards to the test. You've modified the test case to match the `OneHasExtraParam` test case which based on the previous implementation should be consistent (true). I intended to keep the "generic" implementation with the same logic as the previous implementation, the only difference should be `video/h264` MimeType.

jech:
Then it looks like a bug that's unrelated to your changes.  @Sean-Der, do you know what is the right behaviour in that case?

Sean-Der:
I am not sure what the right behavior is, but I am ok with not changing it in this PR. We should cut an issue and investigate further. 

This PR really looks great to me. Love that it is extensible and putting as little code as possible in `/`. Adds a good feature with no tech-debt that I can see :)

Approve from me, will merge when you agree @jech

jech:
> I am not sure what the right behavior is, but I am ok with not changing it in
> this PR. We should cut an issue and investigate further.

> This PR really looks great to me.

Agreed on both counts.

Sean-Der (approved):

PR #1918, opened 2021-08-11
Fix per-track codec selection

billylindeman:
This fixes per-track codec selection (ie H264 camera + VP8 screenshare)

This may not be the right way to do it though @Sean-Der @OrlandoCo @davidzhao 
#### Reference issue
Fixes #...


Sean-Der:
Fixed by 635bfd9cad37eff2d2b7004e303ca4d509b8de17

PR #1919, opened 2021-08-13
Use time.Ticker in play-from-disk

Sean-Der:
time.Sleep is not precise, instead use a ticker.

Relates to golang/go#44343

### DIFF
--- a/examples/play-from-disk-renegotation/main.go
+++ b/examples/play-from-disk-renegotation/main.go
@@ -167,15 +167,18 @@ func writeVideoToTrack(t *webrtc.TrackLocalStaticSample) {
 
 	// Send our video file frame at a time. Pace our sending so we send it at the same speed it should be played back as.
 	// This isn't required since the video is timestamped, but we will such much higher loss if we send all at once.
-	sleepTime := time.Millisecond * time.Duration((float32(header.TimebaseNumerator)/float32(header.TimebaseDenominator))*1000)
-	for {
+	//
+	// It is important to use a time.Ticker instead of time.Sleep because
+	// * avoids accumulating skew, just calling time.Sleep didn't compensate for the time spent parsing the data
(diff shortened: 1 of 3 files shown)


Sean-Der:
@twometer mind trying this PR out? I have been trying a bunch of different things, time.Ticker is giving me the best results.

If it works for you/better then other timing patterns I am going to merge this with you as Co-author and then make an example-webrtc-applications PR

Twometer:
I've tried this PR, and the audio transmission works flawlessly, no lags whatsoever.
I timed the intervals and it seems the `time.Ticker` also uses some sort of "averaging out" system, these are the intervals I get:

image

Three or four ticks always average out to the expected 20ms/tick and it does so with much less jitter than the other timing pattern I've implemented in the test repository, so I think this might be the best solution!

jech:
Since this is an example, I think you should add a comment explaining the two reasons for the use of a ticker:

 * avoids accumulating skew (the Sleep didn't compensate for the time spent parsing the data);
 * works around latency issues with Sleep in older Go versions.

Other than that, LGTM.

Twometer:
I made a comment explaining the issue in [this example repo](https://github.com/Sean-Der/play-from-disk-h264). Maybe we can update the other example and comment to use the ticker, and then we'd have a consistent explanation and implementation across all examples.

Also, should we also update the `play-from-disk-renegotation` example?

Sean-Der:
Fixed all the suggestions, thank you so much!

* Updated `play-from-disk-renegotation`
* Added comment around Ticker usage to explain why we don't use Sleep anymore
* Added @Twometer as co-author on commits

As an aside @Twometer would you be interested in being added to the Pion organization? Would love your input/guidance on how we can make things better.

Twometer:
Sure, I'd love to help with the project if I can :)!

PR #1920, opened 2021-08-14
Throw error earlier for RTPSender with no codecs

### LINKED ISSUE #1702: Improve cryptic "excessive retries in CreateOffer" error message
If a user doesn't attach a `MediaEngine` with at least one codec in the parameters of `NewAPI`, subsequent calls to `PeerConnection.CreateOffer` will always result in an error with the message `excessive retries in CreateOffer`. This message is a little cryptic and doesn't point to a direct way to resolve it; instead, it could suggest adding a MediaEngine with at least one codec.

# Example code that causes the error

```go
settingsEngine := webrtc.SettingEngine{}

// ... //

rtc = webrtc.NewAPI(webrtc.WithSettingEngine(settingsEngine))

peerConnection, err := rtc.NewPeerConnection(webrtc.Configuration{})

// ... //

peerConnection.CreateOffer(nil)

```

# Example code that resolves the error

```go
settingsEngine := webrtc.SettingEngine{}

// ... //

mediaEngine := webrtc.MediaEngine{}
mediaEngine.RegisterDefaultCodecs()

rtc = webrtc.NewAPI(webrtc.WithSettingEngine(settingsEngine), webrtc.WithMediaEngine(&mediaEngine))

peerConnection, err := rtc.NewPeerConnection(webrtc.Configuration{})

// ... //

peerConnection.CreateOffer(nil)

```

Sean-Der:
It isn't possible to send media if a MediaEngine has no codecs. This
catches a common misconfiguration issues that users find themselves in.

Resolves #1702

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -135,6 +135,10 @@ var (
 	// ErrUnsupportedCodec indicates the remote peer doesn't support the requested codec
 	ErrUnsupportedCodec = errors.New("unable to start track, codec is not supported by remote")
 
+	// ErrSenderWithNoCodecs indicates that a RTPSender was created without any codecs. To send media the MediaEngine needs at
+	// least one configured codec.
+	ErrSenderWithNoCodecs = errors.New("unable to populate media section, RTPSender created with no codecs")
+
 	// ErrRTPSenderNewTrackHasIncorrectKind indicates that the new track is of a different kind than the previous/original
 	ErrRTPSenderNewTrackHasIncorrectKind = errors.New("new track must be of the same kind as previous")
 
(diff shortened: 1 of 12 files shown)


Sean-Der:
@jech This look ok? The original check still exists, I just added an early check for RTPSender + no codecs.

I can't think of a case where this would be useful. Want to make sure I am not breaking any valid use cases.

jech:
I'm confused.  You've removed the definition of errExcessiveRetries, but you're still returning it on line 691.  How does this compile?

Like you, I don't see a usage case for a sender with no codecs.  I'll see if I find the time to check the relevant specs, for now, I suggest you commit this and see if anyone complains.

Sean-Der:
I moved it into `errors.go`. Great will commit thanks for looking it over :)

Issue #1925, opened 2021-08-20
Capability of raising the receive MTU (via SettingEngine option)


brucekim:
## Summary

I am building a media server using pion/webrtc. It receives a large number of streams from tracks in a single peer connection and then performs post processing( composition).
<img width="600" alt="test_comp" src="https://user-images.githubusercontent.com/5209780/130191558-e2c867f5-a0f6-424f-b465-f9546885c540.png">

When the server receives 40 or 50 more streams, then `short buffer` error intermittently has been occurred in readLoop() at mux.go.
https://github.com/pion/webrtc/blob/0735efd344a71601a4a51d45686667029358ce22/internal/mux/mux.go#L113
```bash
mux ERROR: 2021/08/03 21:14:10 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:10 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:11 mux: failed to read from packetio.Buffer short buffer
mux ERROR: 2021/08/03 21:14:11 mux: failed to read from packetio.Buffer short buffer
```

With increasing `receiveMTU`, such error was gone. However, It would be good to have an option to configure the receive MTU with SettingEngine option too. I discussed this with @Sean-Der in slack https://gophers.slack.com/archives/CAK2124AG/p1628486724197500

I have tested with the increased MTU about a week, and could not found side effect.

I would like to hear from others whether the configuration of MTU size in a SettingEngine option in the PR, https://github.com/pion/webrtc/pull/1926, is opinionated or not.

## Motivation

I checked what size of packets are coming when such short buffer error occurs with print message.
https://github.com/pion/transport/blob/55c6994404b2500c83316e83d360f89cb78f05a5/packetio/buffer.go#L251
```go
if copied < count {
				fmt.Printf("packetio err! err=%s copied %d count %d\n", io.ErrShortBuffer.Error(), copied, count)
				return copied, io.ErrShortBuffer
			}
			return copied, nil
```
[...]

PR #1926, opened 2021-08-20
option to configure the receive MTU in SettingEngine option

brucekim:
#### Description

This gives an option to raise the receive MTU as SettingEngine option.
If SettingEngine has not been set the MTU, then default value is used instead, `1460`

#### Reference issue
https://github.com/pion/webrtc/issues/1925

Fixes #...


Sean-Der (approved):

Sean-Der:
Fantastic PR, thank you @brucekim 

Are you interested in staying involved with Pion? I would love to add you to the organization and get your feedback on what can be better/directions we should go :)

thanks!

brucekim:
Thank you! Yes. Pion is very interesting. I would like to follow where pion goes.

PR #1936, opened 2021-08-27
Implement SSRC Based Simulcast

### LINKED ISSUE #1345: Simulcast Improvements
Simulcast has landed! This a list of improvements I think would be great to see before `/v3` is tagged.

* [ ] Support SSRC Based Simulcast
FireFox uses SSRCes instead of `rid` and `mid`. We should support this as well. 

* [x] Refactor/Cleanup RTPReceiver
Rename `trackStreams` -> `readerStreams`.  A lot more stuff will fall into place when multiple `onTrack` paths are removed.
* [ ] ~~Update `examples/simulcast` to respond to remote Session Description
Currently the example is tightly coupled with the frontend code. It would be nice if the backend dynamically constructed tracks depending on how many Simulcast layers the offerer is sending.~~
* [x] Implementing ReadRTCP for Simulcast Streams
Update `Read` to allow users to pass in the SSRC as well? Maybe we don't break the API and assume `readerStreams[0]` if no SSRC is passed? 65400221ece8d2044ed0b5b67bcf08bc1e2055b3
* [x] Remove multiple OnTrack paths
Remove the goroutine from `startReceiver`. Everything should instead just react it incoming SSRCes. We could rename `handleUndeclaredSSRC` to be something like `onNewIncomingSSRC`.
* [x] E2E Tests
Currently we just have unit testing, but no examples of E2E testing. Would be fine to send packets manually now. If easy might be worth adding proper Simulcast client support.
* [x] Fix Simulcast Probing deadlock
Currently if a sender doesn't send enough packets `handleUndeclaredSSRC` can hang forever. This prevents probing taking place on other streams. We should fire a goroutine per stream we want to probe. 70fb90c7fdbca5c728003ee87e5f46d231a058ce

Sean-Der:
Resolves #1345

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1151,12 +1151,19 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 }
 
 func (pc *PeerConnection) startReceiver(incoming trackDetails, receiver *RTPReceiver) {
-	encodings := []RTPDecodingParameters{}
-	if incoming.ssrc != 0 {
-		encodings = append(encodings, RTPDecodingParameters{RTPCodingParameters{SSRC: incoming.ssrc}})
+	encodingSize := len(incoming.ssrcs)
+	if len(incoming.rids) >= encodingSize {
+		encodingSize = len(incoming.rids)
 	}
-	for _, rid := range incoming.rids {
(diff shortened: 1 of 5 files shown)


Sean-Der:
It works! :fire:

2021-08-28-105856_1296x897_scrot

PR #1943, opened 2021-09-04
Fixes RTPTransceiver: Stopped

Sean-Der:
Make RTPTransceiver Stopped an atomic

    Add an accessor to make getting value easy. Also add
    TestPeerConnection_SkipStoppedTransceiver. This commit also cleans
    up RTPTransceiver creation. We used a helper function, when we should
    have just used the provide constructor

### DIFF
--- a/atomicbool.go
+++ b/atomicbool.go
@@ -12,9 +12,20 @@ func (b *atomicBool) set(value bool) { // nolint: unparam
 		i = 1
 	}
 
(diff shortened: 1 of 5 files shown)


Sean-Der:
@zyxar w/e you are happy with this mind dropping an approve? Then we can merge! Sorry this took so long

zyxar (approved):

PR #1946, opened 2021-09-07
Fix double loop in data channel ID generation

### LINKED ISSUE #1945: `(*SCTPTransport).generateAndSetDataChannelID` has exponential loop when finding new channel ID
### Your environment.
  - Version: `github.com/pion/webrtc/v3 v3.0.32`
  - Browser: pion-to-pion
  - Other Information - n/a

### What did you do?

Reproduction repo, the steps are in the README:
https://github.com/deansheather/pion-stuff

If you run the code for a while you'll notice that it takes hundreds of milliseconds to create a data channel. This wait is caused by the `(*SCTPTransport).generateAndSetDataChannelID` call (linked below).

In our test we ran the reproduction script for 20k iterations and it takes > 200ms to open a data channel now.

### What did you expect?

Data channel creation should be have constant timing and not be exponential. Data channels should be removed from the `(*SCTPTransport).dataChannels` slice when they are closed using `dc.Close()`.

### What happened?

We opened a lot of data channels using this package and our CPU gets pinned every time we try to open a new data channel.

It looks like the code in [sctptransport.go](https://github.com/pion/webrtc/blob/c1fa55d5efe20db713953e65f5804f130926d6c5/sctptransport.go#L340-L368) loops exponentially up to `MaxDataChannels` (65535) with an inner loop of all data channels.

My recommendation is to remove closed data channels from `(*SCTPTransport).dataChannels` and to turn it into a `map[uint16]*webrtc.DataChannel` with the data channel ID as the key.

cc @coadler @kylecarbs

deansheather:
#### Description

`(*SCTPTransport).generateAndSetDataChannelID` performed a double loop to find the next available data channel ID. This changes that behavior to generate a lookup map with the taken IDs first, so generating a data channel ID takes much less time.

Before, it would take more than 1000ms to generate the next data channel ID once you had roughly 50k of them. Now it only takes 4ms at that same point.

#### Reference issue
Fixes #1945

cc @coadler @kylecarbs

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -42,6 +42,7 @@ cnderrauber <zengjie9004@gmail.com>
 cyannuk <cyannuk@issart.com>
 David Hamilton <davidhamiltron@gmail.com>
 David Zhao <david@davidzhao.com>
+Dean Sheather <dean@coder.com>
 decanus <7621705+decanus@users.noreply.github.com>
 Denis <Hixon10@yandex.ru>
 digitalix <digitalix4@gmail.com>
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -963,7 +963,6 @@ func TestPeerConnection_Start_Right_Receiver(t *testing.T) {
 // Assert that failed Simulcast probing doesn't cause
 // the handleUndeclaredSSRC to be leaked
 func TestPeerConnection_Simulcast_Probe(t *testing.T) {
-	return
 	lim := test.TimeOut(time.Second * 30) //nolint
 	defer lim.Stop()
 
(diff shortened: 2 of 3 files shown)


Sean-Der:
Fantastic work @deansheather! Thanks for improving this

Sean-Der:
I hope you stay involved with Pion. It is hard to find people that can go and clean up existing code like this. If you have any suggestions on things we should do I would love to hear your suggestions.

Sean-Der (approved):

Sean-Der:
Tagged `v3.1.0-beta.7` so you can move back onto `pion/webrtc` and off of fork!

Also would you be interested in being added to the Pion org @deansheather? Then you can just push branches instead, makes things easier.

deansheather:
@Sean-Der Yeah I'm down. We need to make a couple more changes so avoiding forks would be appreciated 😄

PR #1949, opened 2021-09-08
Placeholder support for ssltcp network type, used by Google

adriancable:
libwebrtc, in addition to udp and tcp ICE candidates, also supports ssltcp candidates. They are not used much in the wild outside of Google but they are used by Google Hangouts and cameras. Per RFC 5245 section 5.1 additional transport protocols can be made available and should be gracefully ignored if unsupported, but right now pion panics if it sees an ssltcp candidate - unnecessarily if there is a totally usable udp candidate. This PR adds placeholder support to ignore ssltcp candidates instead of panicking.


Sean-Der:
Hey @adriancable would you be able to write a small test for this? Just assert that if you call `addicecandidate` or `SetRemoteDescription` that you don't get an error (unlike before?)

If you don't have the time totally understand, I can grab it :)

adriancable:
@Sean-Der - no problem! I'll add that to peerconnection_test.go.

adriancable:
@Sean-Der - test is added. Note that this PR also requires the corresponding PR in pion/ice to be merged. (The supplied test is sufficient for both.)

PR #1958, opened 2021-09-15
Handle Simulcast RepairStream

### LINKED ISSUE #1957: Enabling Transport Wide Congestion Control breaks Simulcast
If I enable TWCC `examples/simulcast` now breaks with 

```
pc ERROR: 2021/09/14 14:44:34 Incoming unhandled RTP ssrc(1780109122), OnTrack will not be fired. incoming SSRC failed Simulcast probing
```

Sean-Der:
Read + Discard packets from the Simulcast repair stream. When a
Simulcast stream is enabled the remote will send packets via the repair
stream for probing. We can't ignore these packets anymore because it
will cause gaps in the feedback reports

Resolves #1957

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -31,6 +31,8 @@ const (
 	incomingUnhandledRTPSsrc = "Incoming unhandled RTP ssrc(%d), OnTrack will not be fired. %v"
 
 	generatedCertificateOrigin = "WebRTC"
+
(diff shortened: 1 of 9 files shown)


Sean-Der:
Not super happy about how this is being solved, but is best solution for now. 

I need to come up with a long term plan for repair streams. Is already ticketed, just no work has been done.

mengelbart (approved):

PR #1964, opened 2021-09-16
Add InterceptorFactory

Sean-Der:
Interceptors are being accidentally misused by users.
The issue is that an Interceptor can be re-used between
multiple PeerConnections. Interceptors were designed to only be
single PeerConnection aware, so state is being corrupted.

Instead we are now going to provide InterceptorFactories. The default
API of pion/webrtc will now be safe to use between PeerConnections.

Resolves webrtc#1956


Sean-Der:
I am really happy with this!

We have API breakage in pion/interceptor, but none in pion/webrtc and solves our re-usage problems.

davidzhao (approved):
lgtm

PR #1967, opened 2021-09-22
Adds WithPacketHeadHandler to SampleBuilder

digitalix:
#### Description
This option allows inspecting head packet for each media.Sample and then lets the user return their custom metadata. This might be useful in case when you need to check whether the given sample is a keyframe.

### DIFF
--- a/pkg/media/media.go
+++ b/pkg/media/media.go
@@ -17,6 +17,7 @@ type Sample struct {
 	Duration           time.Duration
 	PacketTimestamp    uint32
 	PrevDroppedPackets uint16
+	Metadata           interface{}
 }
 
 // Writer defines an interface to handle
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -40,6 +40,9 @@ type SampleBuilder struct {
 
 	// number of packets forced to be dropped
 	droppedPackets uint16
+
+	// allows inspecting head packets of each sample and then returns a custom metadata
+	packetHeadHandler func(headPacket interface{}) interface{}
 }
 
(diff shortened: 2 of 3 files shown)


Sean-Der:
Hey @digitalix  Sorry this took so long. 

I see how this is useful! You are interested in the RTP Payload Header, but are just getting samples.

`interface{}` in+out feels wrong to me. What if we could find a way to expose Unmarshaled packet headers (but not payloads?) Like [VP9Packet](https://github.com/pion/rtp/blob/master/codecs/vp9_packet.go#L115) 

@jech you have opinions on the SampleBuilder. What do you think of the problem?

digitalix:
> Hey @digitalix Sorry this took so long.
> 
> I see how this is useful! You are interested in the RTP Payload Header, but are just getting samples.
> 
> `interface{}` in+out feels wrong to me. What if we could find a way to expose Unmarshaled packet headers (but not payloads?) Like [VP9Packet](https://github.com/pion/rtp/blob/master/codecs/vp9_packet.go#L115)
> 
> @jech you have opinions on the SampleBuilder. What do you think of the problem?

You're right @Sean-Der with `interface{}` as in+out. Hmm... As in we could use `rtp.Depacketizer` but I'd prefer to create an alias for it e.g.`MediaPacket`.

For out we could split `VP9Packet` (and others) into `VP9Packet` and `VP9PacketHeader` that would be embedded within `VP9Packet`.

jech:
I think it's pretty smart.

Sean-Der:
Merging!

Sorry this took so long @digitalix. The usage of `interface{}` I think totally makes sense here.

Users are supplying this data anyway. I think it would be worse for us to force a shape.

Sean-Der (approved):

PR #1968, opened 2021-09-23
Local rtcp-fb values are used when answering

digitalix:
#### Description
Currently, pion uses remote rtcp-fb when answering.
According to https://tools.ietf.org/id/draft-ietf-rtcweb-sdp-08.html a=rtcp-fb lines should list supported feedback mechanisms. This change makes pion use rtcp-fb values provided by the user instead of using whatever was given by the remote pc.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -1,3 +1,4 @@
+//go:build !js
 // +build !js
 
 package webrtc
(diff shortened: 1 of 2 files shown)


edaniels:
hey @digitalix , do you still want to work together to get this in?

Sean-Der:
This is an old one :(

Sorry this didn't get addressed sooner @digitalix! I implemented it to match the spec. It now takes the intersection of local+remote. So the only RTCP Feedback that is enabled is what both sides support.

PR #1972, opened 2021-10-02
Handle non-Simulcast Repair Streams

Sean-Der:
Same issue with TWCC enabled as 11b887. We need to process the
RTX packets so that we can emit proper reports.


Sean-Der:
I created https://github.com/Sean-Der/billylindeman-simulcast-bug to test + confirm

PR #1979, opened 2021-10-09
Use transceiver's codec when getting codecs

boks1971:
Issue:
------
A transceiver's codecs can be modified using `SetCodecPreferences`.
When retrieving codecs from the transceiver after changing codec
preferences of the transceiver (for example changing the SDPFmtpLine),
the filter function was using codec from the media engine. Thus,
the change from `SetCodecPreferences` is lost.

Fix:
----
When a match is found (either exact or partial), use the codec from
the transceiver instead of media engine.

Testing:
--------
- Modify SDPFmtpLine of a codec of a transceiver using `SetCodecPreferences` method of `RTPTransceiver`.
- Invoke `GetParamters` of `RTPReceiver` and ensure that `Codecs` has the SDPFmtpLine modification. Before this change the change was not reflected in the returned `Codecs`.

#### Reference issue
Fixes #...

### DIFF
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -69,7 +69,10 @@ func (t *RTPTransceiver) getCodecs() []RTPCodecParameters {
 	filteredCodecs := []RTPCodecParameters{}
 	for _, codec := range t.codecs {
 		if c, matchType := codecParametersFuzzySearch(codec, mediaEngineCodecs); matchType != codecMatchNone {
-			filteredCodecs = append(filteredCodecs, c)
+			if codec.PayloadType == 0 {
+				codec.PayloadType = c.PayloadType
+			}
+			filteredCodecs = append(filteredCodecs, codec)
 		}
 	}
 
--- a/rtptransceiver_test.go
+++ b/rtptransceiver_test.go
@@ -124,7 +124,7 @@ func Test_RTPTransceiver_SetCodecPreferences_PayloadType(t *testing.T) {
 	assert.NoError(t, err)
 
 	// VP8 with proper PayloadType
-	assert.NotEqual(t, -1, strings.Index(answer.SDP, "a=rtpmap:96 VP8/90000"))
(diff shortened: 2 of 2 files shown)


digitalix:
I'm not sure we should allow this. `SetCodecPreferences` should only allow to use codecs that were registered in MediaEngine. While this still won't allow me to add `VP6` if it wasn't in the MediaEngine it'd allow me to add a variation of VP9 like `profile=3` which is not something that browser version of `SetCodecPreferences` would let me do.

In addition, if I'm not wrong this will break backwards compatibility - if we now use codecs from transceiver we also uses their payloadIDs. Currently if I set VP9 (payloadId=98) on the pc coming from Safari it will work (because MediaEngine will use ids used by Safari and they are different). If this lands sdp negotiation will fail due to payloadIds being different.

boks1971:
@digitalix Thank you very much for the review and comment. Really appreciate it.

Totally understand the negotiation quirks. But, doesn't negotiation try to resolve what the two sides want and settle on the common denominator?

In the case of `SetCodecPreferences` usage, my thought is that the user of the API knows the quirks and uses it in a non-breaking way. An example sequence would be to get codecs from transceiver (receiver/sender), modify it and set it on the transceiver. That would ensure that the remote side's codecs are taken into account before modifying rather than modifying the static entities in the media engine.

If the transceiver's codecs are not picked up at all, I am trying to understand when and where should `SetCodecPreferences` be used so that they take effect?

This is my first PR in Pion and I am still trying to understand how things work. So, really appreciate all the discussion and feedback so that I can understand better. Thank you :bow:

davidzhao:
I'm in agreement with @boks1971 on the approach here.

MediaEngine is static and doesn't allow us to react to fmtp line variations given by the remote peer. `SetCodecPreferences` could give pion users control over that codec selection, particularly the nuance of fmtpline used in negotiation. (i.e. declarative properties).

@Sean-Der @jech @billylindeman do you have thoughts on this?

PR #1987, opened 2021-10-19
WIP: Add bandwidth estimation unit tests

mengelbart:
Adds unit test implementations for the test cases from RFC 8867.

No assertions yet, since we don't have any bandwidth estimation
implemented. After adding first bandwidth estimation implementations
like Google Congestion Control, we can add more metrics and define some
criteria to evaluate the implementations.


mengelbart:
This work continues in a separate repository.

PR #2007, opened 2021-11-01
Add test for negotiated DataChannel.OnOpen

edaniels:
5dc7245 was missing a test for this

### DIFF
--- a/datachannel_test.go
+++ b/datachannel_test.go
@@ -111,6 +111,8 @@ func benchmarkDataChannelSend(b *testing.B, numChannels int) {
 }
 
 func TestDataChannel_Open(t *testing.T) {
+	const openOnceChannelCapacity = 2
+
 	t.Run("handler should be called once", func(t *testing.T) {
 		report := test.CheckRoutines(t)
 		defer report()
@@ -121,7 +123,7 @@ func TestDataChannel_Open(t *testing.T) {
 		}
 
 		done := make(chan bool)
-		openCalls := make(chan bool, 2)
+		openCalls := make(chan bool, openOnceChannelCapacity)
 
 		answerPC.OnDataChannel(func(d *DataChannel) {
 			if d.Label() != expectedLabel {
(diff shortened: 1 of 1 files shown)


datachannel_test.go
```diff
@@ -155,6 +155,63 @@ func TestDataChannel_Open(t *testing.T) {
...
+			Negotiated: &negotiated,
+			Ordered:    &ordered,
+		})
+		assert.NoError(t, err)
+
+		answerDC.OnMessage(func(msg DataChannelMessage) {
+			go func() {
+				// Wait a little bit to ensure all messages are processed.
+				time.Sleep(100 * time.Millisecond)
+				done <- true
```
ashellunts, line 193: Looks like this is not used, right?

datachannel_test.go
```diff
@@ -155,6 +155,63 @@ func TestDataChannel_Open(t *testing.T) {
...
+		report := test.CheckRoutines(t)
+		defer report()
+
+		offerPC, answerPC, err := newPair()
+		if err != nil {
+			t.Fatalf("Failed to create a PC pair for testing")
+		}
+
+		done := make(chan bool)
+		answerOpenCalls := make(chan bool, 2)
```
ashellunts, line 169: Why 2? Please use named constant.

datachannel_test.go
```diff
@@ -155,6 +155,63 @@ func TestDataChannel_Open(t *testing.T) {
...
+				done <- true
+			}()
+		})
+		answerDC.OnOpen(func() {
+			answerOpenCalls <- true
+		})
+
+		offerDC.OnOpen(func() {
+			offerOpenCalls <- true
+			e := offerDC.SendText("Ping")
```
ashellunts, line 202: Is this used in a test?

datachannel_test.go
edaniels: the existing test does the same thing for the sake of ensuring that an open happens only once but gives it the capacity for two in order to check for that failure.

datachannel_test.go
edaniels: done is used in `closePair` in both tests.

datachannel_test.go
edaniels: Yes, `answerDC` uses this message in order to send to `done`

datachannel_test.go
edaniels: added constants

datachannel_test.go
```diff
@@ -155,6 +157,63 @@ func TestDataChannel_Open(t *testing.T) {
...
+			ID:         &dataChannelID,
+			Negotiated: &negotiated,
+			Ordered:    &ordered,
+		})
+		assert.NoError(t, err)
+
+		answerDC.OnMessage(func(msg DataChannelMessage) {
+			go func() {
+				// Wait a little bit to ensure all messages are processed.
+				time.Sleep(100 * time.Millisecond)
```
ashellunts, line 194: Sorry, I don't know exactly this code, but curious what exactly do you wait here?
Is it possible to write the test without sleep?

edaniels: I’m copying this from the test above so I’m not sure. I’d rather keep this the same and then later change both test later.

ashellunts (approved):

Sean-Der:
Thanks for reviewing @ashellunts and sorry I got pulled away @edaniels :(

PR #2008, opened 2021-11-03
Fixes an assignment to entry in nil map panic

mengelbart:
#### Description

Attributes map was not initialized when it is used by an interceptor.


ashellunts (approved):

Sean-Der:
Me and @mengelbart discussed this via PM

If a user isn't using any interceptors this would cause any extra un-needed allocation for every packet sent. For now we should push this check into interceptors. We can revisit this in the future if it becomes a problem.

I am going to close this for now. please re-open if you disagree!

PR #2009, opened 2021-11-03
Remove sleep in TestDataChannel_Open

ashellunts:
The test passes also without sleep.


ashellunts:
Sleep was added in https://github.com/pion/webrtc/commit/9fca58f51048afa5bdfef65fe853990b057774e9
Closing in PR for now.

PR #2012, opened 2021-11-05
Open all DataChannels before accepting remote

Sean-Der:
A race existed with Negotiated DataChannels before this. Before the remote
could sends us a DataChannel message before the datachannel was
negotiated locally. We would then discard the message and print an
error.

This PR moves the processing of remote datachannel messages after we have
registered all local ones.


Sean-Der:
@edaniels is working on this instead.

PR #2015, opened 2021-11-08
Open all DataChannels before accepting remote

edaniels:
A race existed with Negotiated DataChannels before this. Before
the remote could sends us a DataChannel message before the
datachannel was negotiated locally. We would then discard the
message and print an error.

This PR moves the processing of remote datachannel messages after
we have registered all local ones. It also informs
datachannel.Accept of existing datachannels in order to prevent
any initialization logic from proceeding incorrectly.


edaniels:
Depends on https://github.com/pion/datachannel/pull/124

Sean-Der (approved):

PR #2018, opened 2021-11-10
Match remote codec ordering in transceivers created by remote SDP

cnderrauber:
if a transceiver is created by remote sdp, then set
prefer codec same as offer peer.

#### Description
If a client set preference codec, then create offer and send to pion sfu,
then sfu create a recvonly transceiver from the sdp, but the codec seq
is the same as mediaengine's register, I think it should be same as the 
offer peer's prefer codec, so the client can use codec  prefer.

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1064,6 +1064,20 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 				pc.mu.Lock()
 				pc.addRTPTransceiver(t)
 				pc.mu.Unlock()
+
+				// if transceiver is create by remote sdp, set prefer codec same as remote peer
+				if codecs, err := codecsFromMediaDescription(media); err == nil {
+					filteredCodecs := []RTPCodecParameters{}
+					for _, codec := range codecs {
+						if c, matchType := codecParametersFuzzySearch(codec, pc.api.mediaEngine.getCodecsByKind(kind)); matchType == codecMatchExact {
+							// if codec match exact, use payloadtype register to mediaengine
+							codec.PayloadType = c.PayloadType
+							filteredCodecs = append(filteredCodecs, codec)
+						}
+					}
+					_ = t.SetCodecPreferences(filteredCodecs)
+				}
+
 			case direction == RTPTransceiverDirectionRecvonly:
 				if t.Direction() == RTPTransceiverDirectionSendrecv {
 					t.setDirection(RTPTransceiverDirectionSendonly)
--- a/peerconnection_go_test.go
+++ b/peerconnection_go_test.go
@@ -1433,3 +1433,124 @@ func TestPeerConnectionNilCallback(t *testing.T) {
 
 	assert.NoError(t, pc.Close())
 }
+
+func TestTransceiverCreatedByRemoteSdpHasSameCodecOrderAsRemote(t *testing.T) {
+	t.Run("Codec MatchExact", func(t *testing.T) { //nolint:dupl
+		const remoteSdp = `v=0
+o=- 4596489990601351948 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+m=video 60323 UDP/TLS/RTP/SAVPF 98 94 106
+a=ice-ufrag:1/MvHwjAyVf27aLu
+a=ice-pwd:3dBU7cFOBl120v33cynDvN1E
+a=ice-options:google-ice
+a=fingerprint:sha-256 75:74:5A:A6:A4:E5:52:F4:A7:67:4C:01:C7:EE:91:3F:21:3D:A2:E3:53:7B:6F:30:86:F2:30:AA:65:FB:04:24
+a=mid:0
+a=rtpmap:98 H264/90000
+a=fmtp:98 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f
+a=rtpmap:94 VP8/90000
+a=rtpmap:106 H264/90000
+a=fmtp:106 level-asymmetry-allowed=1;packetization-mode=0;profile-level-id=42e01f
+a=sendonly
+m=video 60323 UDP/TLS/RTP/SAVPF 108 98 125
+a=ice-ufrag:1/MvHwjAyVf27aLu
+a=ice-pwd:3dBU7cFOBl120v33cynDvN1E
+a=ice-options:google-ice
+a=fingerprint:sha-256 75:74:5A:A6:A4:E5:52:F4:A7:67:4C:01:C7:EE:91:3F:21:3D:A2:E3:53:7B:6F:30:86:F2:30:AA:65:FB:04:24
+a=mid:1
+a=rtpmap:98 H264/90000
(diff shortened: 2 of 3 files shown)


peerconnection.go
```diff
@@ -1064,6 +1064,22 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 				pc.mu.Lock()
 				pc.addRTPTransceiver(t)
 				pc.mu.Unlock()
+
+				// if transceiber is create by remote sdp, set prefer codec same as remote peer
```
boks1971, line 1068: :nit: `transceiver`

peerconnection_test.go
```diff
@@ -673,6 +673,46 @@ func TestSetRemoteDescriptionInvalid(t *testing.T) {
...
+	api := NewAPI(WithMediaEngine(&m))
+	pc, err := api.NewPeerConnection(Configuration{})
+	assert.NoError(t, err)
+	assert.NoError(t, pc.SetRemoteDescription(SessionDescription{
+		Type: SDPTypeOffer,
+		SDP:  remoteSdp,
+	}))
+	ans, err := pc.CreateAnswer(nil)
+	assert.NoError(t, pc.SetLocalDescription(ans))
+	assert.Equal(t, true, pc.GetTransceivers()[0].getCodecs()[0].MimeType == MimeTypeH264)
```
boks1971, line 713: maybe add a check to ensure that the payload type is correct?

peerconnection.go
cnderrauber: changed

davidzhao:
@cnderrauber The behavior you are going for makes sense. I do wonder if this logic should take place in `mediaengine.go`

We have a method `updateFromRemoteDescription`, that's supposed to update the MediaEngine with what's negotiated with the remote peer. It appears to be doing the same thing in there (with the only difference being it doesn't handle renegotiations)

cnderrauber:
> @cnderrauber The behavior you are going for makes sense. I do wonder if this logic should take place in `mediaengine.go`
> 
> We have a method `updateFromRemoteDescription`, that's supposed to update the MediaEngine with what's negotiated with the remote peer. It appears to be doing the same thing in there (with the only difference being it doesn't handle renegotiations)

@davidzhao I think mediaengine is not proper place to do this. Because pushlisher can set different prefer codec in different media streams, but mediaengine is shared by all rtptransceivers in sfu's peer, that is why we can't send different codec streams for camera and screeshare before this pr.

Sean-Der:
@cnderrauber Do you know what Chrome/FireFox do for this?

If the Offerer does H264 + VP8 and answerer does VP8 + H264 what side does each codec choose? I am in support of w/e the browser does! I will look through the IETF/W3C to see if the spec says anything.

cnderrauber:
> @cnderrauber Do you know what Chrome/FireFox do for this?
> 
> If the Offerer does H264 + VP8 and answerer does VP8 + H264 what side does each codec choose? I am in support of w/e the browser does! I will look through the IETF/W3C to see if the spec says anything.
> 
> The good news is you can accomplish this via the public API right?

@Sean-Der In this situation, the browser will use the VP8, as the answer's prefer (VP8 in front of h264). 
And in https://www.w3.org/TR/webrtc/#bib-rfc8829 , it said: 

> If set, the offerer's codec preferences will decide the order of the codecs in the offer. If the answerer does not have any codec preferences then the same order will be used in the answer. However, if the answerer also has codec preferences, these preferences override the order in the answer. In this case, the offerer's preferences would affect which codecs were on offer but not the final order.

So , if a transceiver is not added manually and created by remote offer sdp, it should use the same order as the offer.

Also, @davidzhao discussed the partial match with me, we should test more partial match case, for now it only test for exact match. 

Another question, the pr's test if fail, even if I revert to master or tag/v3.1.4 and run on my laptop, it still failed. Do you know how to handle this?

davidzhao:
> @davidzhao I think mediaengine is not proper place to do this. Because pushlisher can set different prefer codec in different media streams, but mediaengine is shared by all rtptransceivers in sfu's peer, that is why we can't send different codec streams for camera and screeshare before this pr.

You are right. this is a per track preference, where as media engine is for the connection. You are describing the right behavior here.

cnderrauber:
Updated, now it will use codec match exact as prefer codec if there are some exsit. Consistence as the pion's codec match logic.

davidzhao (approved):
looks good to me. it'd be great to understand the test failures before merging. @Sean-Der any ideas?

peerconnection.go
```diff
@@ -1064,6 +1064,20 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+					filteredCodecs := []RTPCodecParameters{}
+					for _, codec := range codecs {
+						if c, matchType := codecParametersFuzzySearch(codec, pc.api.mediaEngine.getCodecsByKind(kind)); matchType == codecMatchExact {
+							// if codec match exact, use payloadtype register to mediaengine
+							codec.PayloadType = c.PayloadType
+							filteredCodecs = append(filteredCodecs, codec)
+						}
+					}
+					_ = t.SetCodecPreferences(filteredCodecs)
+				}
```
ashellunts, line 1079: The method `SetRemoteDescription` is already too big. What do you think about extracting added code to a function?
You can also remove comment and use method name `SetPreferCodecSameAsFromRemotePeerIfTransceiverIsCreateByRemoteSDP`.

peerconnection_test.go
```diff
@@ -673,6 +673,125 @@ func TestSetRemoteDescriptionInvalid(t *testing.T) {
 	})
 }
 
+func TestTransceiverByRemoteAsSamePreferCodec(t *testing.T) {
```
ashellunts, line 676: IMHO from test name it is not clear what exactly the test is about.
If I understand correctly test check that `WhenTransceiverCreatedByRemoteSDP_PreferredCodecShouldBeSetSameAsOnRemotePeer`.

peerconnection_test.go
```diff
@@ -673,6 +673,125 @@ func TestSetRemoteDescriptionInvalid(t *testing.T) {
...
+	t.Run("MatchExact", func(t *testing.T) { //nolint:dupl
```
ashellunts, line 677: What is matched?

cnderrauber: match codec exactly

peerconnection.go
cnderrauber: A function only for here and can't be reused in other place, I think it may not a better way..

peerconnection_test.go
cnderrauber: sounds good. let me use a clear name but not that long

peerconnection.go
ashellunts: Correct. Why do you want it to be reusable? One can use functions to divide bigger functions to smaller parts.

peerconnection_test.go
ashellunts: Can we include that in test name?

PR #2019, opened 2021-11-11
Fixed locking order in SRTP and DataChannel

davidzhao:
#### Description

Two separate potential deadlocks

1. typically the locking order is SRTP first, then DataChannel. However, when data channel is opened, it's locking srtp when generating an ID. It also would return without unlocking when an error is encountered
2. recursive RLock could be potentially problematic. MediaEngine contained a section doing so.

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -153,10 +153,15 @@ func (d *DataChannel) open(sctpTransport *SCTPTransport) error {
 	}
 
 	if d.id == nil {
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

Sean-Der:
@davidzhao nice! Did you use any tooling that could be automated to get these?

davidzhao:
> @davidzhao nice! Did you use any tooling that could be automated to get these?

Unfortunately it wasn't automated, mass replaced sync.Mutex/RWMutex with https://github.com/sasha-s/go-deadlock, then launched bot army against the SFU

PR #2032, opened 2021-11-20
fix get set isClosed not in atomic

peiying16:
Thanks for this awesome project. 
When I was learning the code, I realized the isClosed on Close function is not in atomic.
if this function is called at the same time on different go routines, the peerConnect may close multiple times.
Here is the fix to make the get and set in atomic.
Hope this help.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -30,6 +30,7 @@ bkim <bruce.kim.it@gmail.com>
 Bo Shi <boshi@mural.co>
 boks1971 <raja.gobi@tutanota.com>
 Brendan Rius <brendan.rius@gmail.com>
+brian <brian@cyalive.com>
 Cameron Elliott <cameron-elliott@users.noreply.github.com>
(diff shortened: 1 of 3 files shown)


Sean-Der:
Thank you so much @peiying16! This LGTM, will merge now :)

If there is anything else that can be improved just tell me. Always looking to make pion better.

Sean-Der (approved):

peiying16:
@Sean-Der Thank you so much to the quick merge. Could you update my name to `Brian (Peiying) Yang<mryang91@hotmail.com>` Thanks.

Sean-Der:
Hey @peiying16 

That is generated from the `git log` so I can't edit it unfortunately. We can block names/emails from appearing in the log though!

Would you like me to add `brian@cyalive.com` to [EXCLUDED_CONTRIBUTORS](https://github.com/pion/.goassets/blob/master/ci/.github/generate-authors.sh#L25) list?

peiying16:
If it is not able to edit, just leave it. Thanks.

PR #2064, opened 2021-12-26
Remove examples that aren't unique

Sean-Der:
data-channels-close demonstrates that you are able to close a
DataChannel. Closing a DataChannel isn't exercising any unique
behaviors. I don't believe it will help users discover anything new or
prevent them from making mistakes.

data-channels-(detach)-create demonstrates creating a DataChannel in Go.
We have a 1:1 mapping with the browser so I think this is expected. We
also have other examples demonstrate this behavior.


Sean-Der:
Since you are the creator of these would love your opinion! We are growing lots of examples so looking to cleanup a bit.

Sean-Der:
@backkem ^

backkem (approved):
Yeah, makes sense.

PR #2071, opened 2022-01-04
Add H265 and AV1 MimeTypes

kevmo314:
This change adds constants for H265 and AV1. Neither are fully
supported in pion, however these constants are still useful and we will
likely send a change to add H265 and AV1 packetizer/depacketizers in
the future.


davidzhao (approved):
looks good!

PR #2074, opened 2022-01-06
Make setMid of Transceiver public

cnderrauber:
in some case like session migration from one sfu node
to another, we need manual set mid of transceiver instead of
auto generate, to make mid consistent between to sfu node.

#### Reference issue
Fixes #...

### DIFF
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -115,8 +115,8 @@ func (t *RTPTransceiver) Receiver() *RTPReceiver {
 	return nil
 }
 
-// setMid sets the RTPTransceiver's mid. If it was already set, will return an error.
-func (t *RTPTransceiver) setMid(mid string) error {
+// SetMid sets the RTPTransceiver's mid. If it was already set, will return an error.
+func (t *RTPTransceiver) SetMid(mid string) error {
 	if currentMid := t.Mid(); currentMid != "" {
 		return fmt.Errorf("%w: %s to %s", errRTPTransceiverCannotChangeMid, currentMid, mid)
 	}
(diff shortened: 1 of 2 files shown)


rtptransceiver.go
```diff
@@ -1,3 +1,4 @@
+//go:build !js
```
boks1971: Is this due to Go version? Does this affect older supported version?

boks1971 (approved):
looks good to me, one question.

rtptransceiver.go
cnderrauber: looks added by IDE, removed.

Sean-Der:
nice change :) 

What do you think of moving this to `RTPParameters` as `muxId`. That would match the spec [here](https://draft.ortc.org/#rtcrtpparameters*]

@cnderrauber That is really exciting that you are moving between SFU nodes! Are you also transferring ICE/SRTP/DTLS/SCTP state?

cnderrauber:
Thanks @Sean-Der . That Sounds good. I will check this

I’m not transfer these state because it is complex. Just use ICE restart to moving to new node. Difficult things is DTLS, because pion will not restart dtls with different fingerprint(but chrome and webrtc dose), so I manipulated fingerprint during negotiate with new sfu node.

Restart DTLS in pion when fingerprint changed during icerestart is useful and consistent with rfc, I will look at this when have time

Sean-Der:
That would be awesome, I think we just need to land https://github.com/pion/webrtc/pull/1846 !

I wasn't able to find the time to give it the quality of review it needs. But if you and @boks1971 are interested that could be a huge improvement. 

We just have to be really careful to keep up code quality/not introduce any security bugs :) that was the part that scared me.

cnderrauber:
Excited! Agree that security is more important than feature, we should be careful about this

Issue #2065, opened 2021-12-27
Get RTPSender's parameter without checking track.


Rollooo:
### Your environment.
  - Version: v3.1.11
  - Browser: All Browser
  - Other Information - Crash point( https://github.com/pion/webrtc/blob/83eb2acc6e3f9749c5767ccab89f25b75388e018/rtpsender.go#L110), (track is nil, `invalid memory address or nil pointer dereference`)

### What did you do?
Stop RTPSender() during SetRemoteDescription()

### What did you expect?
Work well

### What happened?
PeerConnection try to start rtp senders at this line as below 
https://github.com/pion/webrtc/blob/83eb2acc6e3f9749c5767ccab89f25b75388e018/peerconnection.go#L1126

But rtp sender is stopped by any reason during at that time, sender's track be `nil` by ReplaceTrack(nil).

And PeerConnection try to get parameter of sender at this line as below
https://github.com/pion/webrtc/blob/83eb2acc6e3f9749c5767ccab89f25b75388e018/peerconnection.go#L1308
But sender try to get parameter without checking whether track is nil or not. 

So sometimes application is crashed by accessing nil track of sender.

Sean-Der:
Thank you so much for the report @Rollooo. I tagged `v3.1.12`

This turned out not to even be a race condition! You could easily hit this in normal code.

PR #2077, opened 2022-01-07
Fix RTPSender.Send crash

lenaky:
#### Description
If PeerConnection removes track while RTPSender is created and being negotiated,
RTPSender.Send would access nil pointer.
Check If track is nil.

I think it's related with issue #2065 

#### Reference issue
Fixes #...


Sean-Der:
Oh this is fantastic! Thanks for adding more error handling, will merge after tests pass :)

Sean-Der (approved):

Issue #2080, opened 2022-01-11
When negotiating Opus with PT=96, track.Codec is set to VP8


juberti:
### Your environment.
  - Version: 3.1.7
  - Browser: n/a

### What did you do?
- Sent offer with Opus as PT 96
- Looked at resulting track dispatched in onTrack handler

### What did you expect?
- track.Codec is set to Opus

### What happened?
- track.Codec was set to VP8 (even though track.Kind is Audio)

PR #2081, opened 2022-01-11
Use audio negotiated PTs before video builtin PTs

juberti:
This avoids a bug where negotiating Opus with PT 96 in an audio-only
session results in the VP8 codec being picked for a track (because
96 is the built-in type for VP8).

#### Description
Rework `getCodecByPayload` to check negotiated audio types before
built-in video types, which avoids this issue. The code previously
would always check video types before audio types, even if audio
had already been negotiated.

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/2080

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -313,27 +313,39 @@ func (m *MediaEngine) copy() *MediaEngine {
 	return cloned
 }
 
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
This makes sense. Thank you for the fix.

Sean-Der:
:fire: thank you so much! tagged `v3.1.14` with this.

PR #2086, opened 2022-01-14
Add ability to set RTP stream ID on TrackLocal

kevmo314:
This change makes it possible to set the RTP stream ID to
allow forwarding and production of simulcast streams.


davidzhao (approved):
looks good!

Issue #2054, opened 2021-12-16
Send offer twice in very short time, second offer doesn't reflect well


Rollooo:
### Your environment.
  - Version: v3.1.7
  - Browser: Chrome
  - Other Information - [peerconnection.go](https://github.com/pion/webrtc/blob/517a32639a98612814a4a230f1043530112908c4/peerconnection.go#L942)

### What did you do?
Send offer twice in very short time(within 100ms).
In the first offer, I just add datachannel in SDP
In the second offer, I add two tracks(audio, video)

But the second offer doesn't reflects well and see error log in PeerConnection as below
`Incoming unhandled RTP ssrc(123412341234), OnTrack will not be fired. incoming SSRC failed Simulcast probing`

This issue is raised because REMOTE DESCRIPTION is changed during processing the first offer's startRTP().
(startSCTP() takes long time for init)
So I think before completing startRTP() of the first offer, it should prevent to change REMOTE DESCRIPTION or filter the simulcastStream by undeclaredMediaProcessor(). 

And there is no way to check it startRTP() is in progress. Adding some function for checking it in PeerConnection might be helpful for user.

* I call PeerConnection's functions sequentially as below.
```
SetRemoteDescription(offer) -> answer, err := CreateAnswer(nil) -> SetLocalDescription(answer) 
```

### What did you expect?
The second offer should reflect well.

### What happened?
The  second offer doesn't reflect well

Sean-Der:
Hi @Rollooo 

Thanks for filing this issue! Is the issue is that if Pion is called with two different `SetRemoteDescription`s quickly the second one is ignored?

Rollooo:
Hi @Sean-Der 

The second one is not ignored, but offer/answer doesn't have SSRCes those filtered by `undeclaredMediaProcesser()`

If you add sleep at `startRTP()`(ex. <-time.After(time.Second * 2)) and SetRemoteDescription() twice time quickly. You may reproduce this issue easily.

Sean-Der:
Ok I think I understand more!

The issue is that Simulcast times out because the `SetRemoteDescription` hasn't happened yet?

So this would be fixed by setting [simulcastProbeCount](https://github.com/pion/webrtc/blob/master/constants.go#L17) to a very large number (like math.MaxInt)

Maybe we should re-attempt probing after every SetRemoteDescription?

Rollooo:
Okay, I will check this issue after increasing simulcastProbeCount.

Rollooo:
[...]

PR #2087, opened 2022-01-15
Set up RTP Receivers synchronously

boks1971:
References - https://github.com/pion/webrtc/issues/2054

Do the set up of RTP receivers synchronously so that they
are ready to receive media as soon as signalling to remote
side is sent. Once signalling to remote side is sent, the
remote can send media at any time and receiver has to be ready.

Like the bug mentions, a negotiation quickly followed by
a renegotiation left the RTP receivers of the tracks in the
second offer not set up. If the tracks in the renegotiation
happen to be simulcast tracks, they are missed as browsers
send RID only in the first few packets.

The problem can be reproduced by introducing a 1 second
delay in Downstream direction in Network Link Conditioner
and using a modified version of LiveKit JS SDK sample app to
force a double negotiation spaced closely.

With this change, onTrack fires, but the unhandled warning
from RTCP for the highest layer still happens. But, the
track fires almost immediately following that warning
(less than 5 ms later). So, all the simulcast layers
are available.

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -976,6 +976,7 @@ func (pc *PeerConnection) SetLocalDescription(desc SessionDescription) error {
 		if err := pc.startRTPSenders(currentTransceivers); err != nil {
 			return err
 		}
+		pc.configureRTPReceivers(haveLocalDescription, remoteDesc, currentTransceivers)
 		pc.ops.Enqueue(func() {
 			pc.startRTP(haveLocalDescription, remoteDesc, currentTransceivers)
 		})
@@ -1130,6 +1131,7 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 			if err = pc.startRTPSenders(currentTransceivers); err != nil {
 				return err
 			}
+			pc.configureRTPReceivers(true, &desc, currentTransceivers)
 			pc.ops.Enqueue(func() {
 				pc.startRTP(true, &desc, currentTransceivers)
 			})
@@ -1163,6 +1165,8 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 		if err := pc.startRTPSenders(currentTransceivers); err != nil {
 			return err
 		}
+
+		pc.configureRTPReceivers(false, &desc, currentTransceivers)
 	}
 
 	pc.ops.Enqueue(func() {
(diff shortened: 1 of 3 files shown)


Sean-Der:
Fantastic fix, thank you @boks1971! Really appreciate you grabbing this.

Receive is ORTC/would be a breaking change to modify the behavior.  Do you think there is a way we can keep the old behavior/add setupReceive as a private?

I am afk for a bit, can’t look deeper until tonight.

boks1971:
Thank you @Sean-Der . I don't fully understand all the bits yet :-) There is also some Plan B test failing.  I am still working on it. Will check on oRTC behaviour.

boks1971:
@Sean-Der I do not know ORTC spec. So, what I tried to do here is probably not the right thing - https://github.com/pion/webrtc/pull/2087/commits/31f6bc23458518ca48db3121a1eee3a3dfc36252. I just created two private functions `setupReceive` and `startReceive` which are used by peerconnection.go. And a public `Receive` method which just wraps the two methods. I had same arguments to both `setupReceive` and `startReceive`, so wrapping them in a public method made easy. Please let me know if that satisfies the ORTC case. If not, can you please point me to the spec so that I can understand the requirements better and think about better ways to solve it.

I was thinking of a further change too, i. e. handle tracks with SSRCs also just like the RID case, i. e. do a `startReceive` on those also when we see an SSRC in the `unhandledMediaProcessor` (currently that handles the undeclared SSRC and RID based ones). I was thinking of using it for known SSRC too (although the function name `undeclared` would not be correct anymore though :-) )

Sean-Der:
@boks1971 That sounds like a great code improvement! Simulcast is bolted on (at best).

I wasn't even aware Simulcast was a thing when we designed all of these internals. I would love to have it all re-written with that being something considered day one.

The only thing I would suggest is making a list of all the edge cases/behaviors that we support. Most things are covered by tests and I can help with that. The end of `peerconnection_media_test.go` and `peerconnection_go_test.go` have most of them.

peerconnection.go
```diff
@@ -1288,17 +1359,67 @@ func (pc *PeerConnection) startRTPReceivers(incomingTracks []trackDetails, curre
 			unhandledTracks = append(unhandledTracks, filteredTracks[i])
```
Sean-Der, line 1359: @boks1971 Is this a problem?

Sean-Der:
The end of `setupRTPReceivers` and `startRTPReceivers` is the same. Could we make a helper like `getNewReceiver` ?

Sean-Der:
What do you think of `configureReceiver` and `configureRTPReceivers` or maybe `setReceiverParamaters`. 

The more I read the code I understand the flow a bit more, but the naming throws me off. `setup` is a new term we are using, would be nice if we could have something in the existing WebRTC nomenclature.

peerconnection.go
boks1971: Forgot to clean this up. This is used if `remoteIsPlanB`. I could not get `PlanB` test to work properly by splitting the setup and start of receivers. I will try it again today and post exactly what the issue is.

boks1971:
@Sean-Der Will change name from `setup` -> `configure`.

Will also look at consolidating common code between `setup/configure` and `start`.

Sean-Der (approved):

Issue #2092, opened 2022-01-19
Port `copySessionDescription` to all examples


Sean-Der:
In 979aefd702bfe50552cd830a084f9bc339e7d7df @ashellunts added a `copy session description` feature. Instead of requiring that users copy the entire value they can now hit a button. It helps reduce the amount of copy/paste errors.

All jsfiddles need to be updated to include this now.

PR #2102, opened 2022-01-27
refs #2092; port copySessionDescription to all relevant examples; tin…

r-novel:
…y fixes codestyle in api.go file;

#### Description
I've updated all relevant jsfiddles in the example directory.

Tiny fixes codestyle in api.go. Move default initialize before executing option callbacks.

refs #2092 ticket;

### DIFF
--- a/examples/broadcast/README.md
+++ b/examples/broadcast/README.md
@@ -11,7 +11,7 @@ go get github.com/pion/webrtc/v3/examples/broadcast
 ```
 
 ### Open broadcast example page
-[jsfiddle.net](https://jsfiddle.net/1jc4go7v/) You should see two buttons 'Publish a Broadcast' and 'Join a Broadcast'
+[jsfiddle.net](https://jsfiddle.net/ypcsbnu3/) You should see two buttons `Publish a Broadcast` and `Join a Broadcast`
 
 ### Run Broadcast
 #### Linux/macOS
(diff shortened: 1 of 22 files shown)


api.go
```diff
@@ -70,8 +63,12 @@ func WithSettingEngine(s SettingEngine) func(a *API) {
 
 // WithInterceptorRegistry allows providing Interceptors to the API.
 // Settings should not be changed after passing the registry to an API.
-func WithInterceptorRegistry(interceptorRegistry *interceptor.Registry) func(a *API) {
+func WithInterceptorRegistry(ir *interceptor.Registry) func(a *API) {
```
Sean-Der, line 66: Can you add a test to confirm the new behavior? `WithSettingRegistry` that is nil shouldn't segfault

api.go
```diff
@@ -23,28 +23,21 @@ type API struct {
...
+		settingEngine:       &SettingEngine{},
+		mediaEngine:         &MediaEngine{},
+		interceptorRegistry: &interceptor.Registry{},
+	}
 
 	for _, o := range options {
 		o(a)
 	}
 
-	if a.settingEngine == nil {
```
Sean-Der, line 32: Can you also add a `ni` check to `WithSettingEngine` to match WithMediaEngine + WithInterceptorRegistry.

Sean-Der:
Great work @r-novel, awesome job!

Mind sending over the `With..` as a separate change. Will approve and merge right away :) glad to make the code more consistent and cleaner.

Would you mind also updating the README.md of the examples you updated?

api.go
r-novel: Done! But I think this nil check is not necessary because in WithSettingEngine function we pass SettingEngine parameter by value - not by pointer.

r-novel:
Hello, @Sean-Der ! Thank you so much for your code review. I've added unit tests and updated README files.

Sean-Der:
Great job @r-novel merging!

Some of the examples didn't have `log` defined so I switched to `console.log`. I also updated all the jsfiddle links with your new code, thanks for doing this :)

Sean-Der (approved):

PR #2107, opened 2022-01-31
Firefox case when screen and camera offer together

boks1971:
#### Description
Firefox camera tracks do not fire onTrack when screen share
and camera is negotiated in the same offer.

There were two issues

1. When updateRemoteDescription runs, only the first
media description of a specific kind was considered.
So, when screen was before camera, screen did not
save RID in extmap. It is in the camera section as
video is using simulcast. All the extra ones in the
camera section were ignored.

Parse all descriptions and collect extensions to fix it.

2. This one sounds like a Firefox issue. When sending
answer back, if the screen share section has the
RID, Firefox seems to get stuck. It does not send
screen or camera media and after 15 seconds, the
peer connection fails.

To address that, store extensions from SDP in
transceiver and while constructing answer SDP,
cross check against transceiver's negotiated
extensions before adding it to the answer SDP.

The reason I think this is a Firefox issue is
because how it behaves before this change when
camera is negotiated first and screen later in
a separate offer. When screen is negotiated
later, answer does send back RID and it works.
Looks like Firefox has an issue seeing RID
before it expects it.

#### Testing
Used LiveKit JS SDK sample to negotiate screen and
camera together.

This was not an issue with Chrome as I could not
get Chrome to negotiate screen and camera in the
same offer with screen ahead of camera.


mediaengine.go
```diff
@@ -476,10 +476,10 @@ func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) e
 	for _, media := range desc.MediaDescriptions {
 		var typ RTPCodecType
 		switch {
-		case !m.negotiatedAudio && strings.EqualFold(media.MediaName.Media, "audio"):
+		case strings.EqualFold(media.MediaName.Media, "audio"):
 			m.negotiatedAudio = true
 			typ = RTPCodecTypeAudio
-		case !m.negotiatedVideo && strings.EqualFold(media.MediaName.Media, "video"):
+		case strings.EqualFold(media.MediaName.Media, "video"):
```
boks1971, line 482: This is for point #1 in the PR commit notes. More than one media description of the same kind was getting ignored.

mediaengine.go
```diff
@@ -561,18 +561,24 @@ func (m *MediaEngine) getRTPParametersByKind(typ RTPCodecType, directions []RTPT
...
+		}
+	}
+
+	for id, e := range mediaHeaderExtensions {
+		if (typ == RTPCodecTypeVideo && !e.isVideo) || (typ == RTPCodecTypeAudio && !e.isAudio) {
+			continue
+		}
+
+		if haveRTPTransceiverDirectionIntersection(e.allowedDirections, directions) {
+			headerExtensions = append(headerExtensions, RTPHeaderExtensionParameter{ID: id, URI: e.uri})
```
boks1971, line 581: I found this section hard to read to reason. So, I refactored it. Hope it is okay

peerconnection.go
```diff
@@ -1083,6 +1083,14 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
 				}
 
+				if extensions, err := rtpExtensionsFromMediaDescription(media); err == nil {
+					headerExtensions := []RTPHeaderExtensionParameter{}
+					for uri, id := range extensions {
+						headerExtensions = append(headerExtensions, RTPHeaderExtensionParameter{URI: uri, ID: id})
+					}
+					_ = t.SetHeaderExtensions(headerExtensions)
+				}
+
```
boks1971, line 1093: Storing the extension from SDP media description in the transceiver.

peerconnection.go
```diff
@@ -1200,7 +1208,7 @@ func (pc *PeerConnection) startReceiver(incoming trackDetails, receiver *RTPRece
 
 	for _, t := range receiver.Tracks() {
 		if t.SSRC() == 0 || t.RID() != "" {
-			return
+			continue
```
boks1971, line 1211: This is a change from my previous PR to do RTP receiver changes. Realised that `continue` is better code here. Functionally probably makes no difference as all codecs are SSRCs or RIDs.

sdp.go
```diff
@@ -389,11 +389,18 @@ func addTransceiverSDP(d *sdp.SessionDescription, isPlanB, shouldAddCandidates b
 
 	parameters := mediaEngine.getRTPParametersByKind(t.kind, directions)
 	for _, rtpExtension := range parameters.HeaderExtensions {
-		extURL, err := url.Parse(rtpExtension.URI)
-		if err != nil {
-			return false, err
+		for _, te := range t.getHeaderExtensions() {
```
boks1971, line 392: This is for point 2. Cross check with transceiver's extmap and ad only the intersection in the answer.

rtptransceiver.go
```diff
@@ -80,6 +81,33 @@ func (t *RTPTransceiver) getCodecs() []RTPCodecParameters {
...
+}
+
+func (t *RTPTransceiver) hasHeaderExtension(uri string) bool {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
+	// if nothing specific set, accept all
+	if len(t.headerExtensions) == 0 {
+		return true
+	}
```
boks1971, line 100: Had to add this accept all case to fix a test. But, I think this is fine to keep the old behaviour.

rtptransceiver.go
```diff
@@ -80,6 +81,33 @@ func (t *RTPTransceiver) getCodecs() []RTPCodecParameters {
 	return filteredCodecs
 }
 
+// SetHeaderExtensions stores header extension from SDP
+func (t *RTPTransceiver) SetHeaderExtensions(headerExtensions []RTPHeaderExtensionParameter) error {
```
Sean-Der, line 85: Would it be ok if we did [RTPSender.SetParameters](https://www.w3.org/TR/webrtc/#dom-rtcrtpsender-setparameters)

I don't see this exposed on the receiver side. Still looking

Sean-Der:
@boks1971 This LGTM!

What happens if you have multiple MediaDescriptions with different extension IDs?

rtptransceiver.go
boks1971: I went with parallel to `SetCodecPreferences` which is on the transceiver now. I think it could be in the transceiver as sender and receiver on a transceiver should use the same codec/header extensions.

boks1971:
@Sean-Der @davidzhao Any concerns with this PR?

Sean-Der:
@boks1971 I am happy with everything except `SetCodecPreferences`.

I would like to avoid adding SetHeaderExtensions and instead have it match the browser and do RTPSender.SetParameters . This is a long term vision, but would love to have one code base that people could run native + web + mobile clients. 

Making this all work is a long way away, but if we don’t match web makes it harder to do. If it is intractable I understand, but would be nice to avoid!

mediaengine.go
Sean-Der: please refactor away :) I would love whole PRs that just move code around/make it simpler. The more involved you are in the code base the better.

boks1971:
Thank you @Sean-Der . I will try to spend more time on this next week.

Sean-Der:
Hey @boks1971 I am going to resolve this, but please re-open if you would like to see this merged.

I haven't seen/heard this impacting anyone else. Did Mozilla update anything on their side to include header extensions in all media sections?

boks1971:
Thank you @Sean-Der . I have not heard of a fix, but have not heard of more instances of either. Will keep an eye out and re-open if necessary 🙇

davidzhao:
I just verified and looks like this is still an issue. If a FF client offers screen share first, the camera track would not be recognized afterwards.

boks1971:
Thank you @davidzhao . Will re-visit this some time soon.

Sean-Der:
I can take it from here @boks1971 ! I have some time

boks1971:
Thank you so much @Sean-Der .

boks1971:
Closing this as a different PR address the underlying issue - https://github.com/pion/webrtc/pull/2444

PR #2130, opened 2022-02-22
Support for send-side simulcast

boks1971:
@Sean-Der @davidzhao Not a real PR, but setting up one to get feedback on the direction. There are more bits to do here to make this simulcast capable, but wanted to make sure that the general approach is okay before I go further.

This makes sure that the API is not affected.

I was then planning to take encodings supplied in `RTPTransceiverInit` and apply it on the sender. Something like `setParamaters` on `RTPSender`. `setParameters` is much more strict. For v1 of this, I was thinking of not being that strict and apply the encodings. Then use that in `sdp.go` to generate the `simulcast:send` SDP. That will have both SSRC and RID, but I think that is fine.

And then I will have to figure out the `Read` and friends for the simulcast aware sender.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -142,6 +142,9 @@ var (
 	// ErrRTPSenderNewTrackHasIncorrectKind indicates that the new track is of a different kind than the previous/original
 	ErrRTPSenderNewTrackHasIncorrectKind = errors.New("new track must be of the same kind as previous")
 
+	// ErrRTPSenderNewTrackHasIncorrectEnvelope indicates that the new track has a different envelope than the previous/original
+	ErrRTPSenderNewTrackHasIncorrectEnvelope = errors.New("new track must have the same envelope as previous")
+
 	// ErrUnbindFailed indicates that a TrackLocal was not able to be unbind
 	ErrUnbindFailed = errors.New("failed to unbind TrackLocal from PeerConnection")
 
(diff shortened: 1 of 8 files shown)


davidzhao (commented):
Looks good! The API seems to fit browsers' well; and it's better than my idea of providing multiple TrackLocals. 

I think there are a couple of differences to resolve
* currently we are exposing RID on TrackLocal, which makes little sense if encodings/sender is the way to capture that
* today TrackLocal.Bind is called once, we'll need simulcast-aware TrackLocal where bound is called multiple times, and pass them the RID for each

boks1971:
Thank you @davidzhao . 

There will be multiple `TrackLocal`s inside `RTPSender` (something similar to multiple `TrackRemote`s in `RTPReceiver`), one per encoding with its own RID/SSRC etc. I did change the code in `RTPSender::Send()` (sorry about the C++ style of showing a struct method there :-) ) to call `Bind` on all the encodings. Yeah, still need to figure out the Read per TrackLocal and those bits.

Also, `ReplaceTrack` will not be supported. Currently, it passes in a `TrackLocal`. There is no concept of encodings there. So, I have added a check if the existing track has multiple encodings, just fail `ReplaceTrack` with a new error `ErrRTPSenderNewTrackHasIncorrectEnvelope`.

There are other bits like `srtp_writer_future` which is provided the `RTPSender` itself and it reaches into `RTPSender` structure to grab fields. I am not sure how that should all work in the context on multiple encodings. Need to figure out.

boks1971:
One `TrackLocal` per encoding is not according to spec, but I am thinking keeping one `TrackLocal` with multiple encodings is more disruptive.

davidzhao:
> One `TrackLocal` per encoding is not according to spec, but I am thinking keeping one `TrackLocal` with multiple encodings is more disruptive.

got it! if it's one TrackLocal per layer, then writing to it should be the same.

would you add multiple simulcast tracks by calling `PeerConnection.AddTrack` for each simulcast layer? then it would append to the list of encodings in RTPSender?

boks1971:
@davidzhao I went with adding a method `AddEncoding` on `RTPSender` to minimise blast radius on existing stuff. Please look at this commit - https://github.com/pion/webrtc/pull/2130/commits/aaa528da913d4863f20861048b9834f445bd238e.

Signalling works now. Need to have code to send in proper data to test out further.

Sean-Der:
@kevmo314 Does this fit with how you are using Simulcast/RID today?

@boks1971 If you can replace my hacky Simulcast sender in https://github.com/pion/webrtc/blob/master/peerconnection_media_test.go#L1213 that would be amazing! Would be the perfect way to show the API in action :)

boks1971:
> @boks1971 If you can replace my hacky Simulcast sender in https://github.com/pion/webrtc/blob/master/peerconnection_media_test.go#L1213 that would be amazing! Would be the perfect way to show the API in action :)

Will take a look at it @Sean-Der .

boks1971:
@Sean-Der Can you please review this commit when you get a chance and let me know if it looks good to test out the new API? - https://github.com/pion/webrtc/pull/2130/commits/9c9518c12ab974479fdbb0c2110ae596840f27e5

kevmo314:
> @kevmo314 Does this fit with how you are using Simulcast/RID today?
> 
> @boks1971 If you can replace my hacky Simulcast sender in [https://github.com/pion/webrtc/blob/master/peerconnection_media_test.go#L1213](https://github.com/pion/webrtc/blob/master/peerconnection_media_test.go?rgh-link-date=2022-02-23T16%3A50%3A18Z#L1213) that would be amazing! Would be the perfect way to show the API in action :)

Yeah this looks good to me, no concerns here 👍

davidzhao:
One request, is it possible to surface the `interceptor.RTCPReader` in `TrackLocalContext` ? this will make it easier for the TrackLocal to grab its own RTCPReader during Bind

boks1971:
It is simple enough to do @davidzhao . Did it in this commit - https://github.com/pion/webrtc/pull/2130/commits/34e4bec0ffaab0b46b15301a9404bb862288f423. Please give it a try when you get a chance. Not sure if there are potentially bad things that can happen like interceptor leaks and such.

davidzhao:
> It is simple enough to do @davidzhao . Did it in this commit - [34e4bec](https://github.com/pion/webrtc/commit/34e4bec0ffaab0b46b15301a9404bb862288f423). Please give it a try when you get a chance. Not sure if there are potentially bad things that can happen like interceptor leaks and such.

looks good! would need a getter too in order to export it. maybe `RTCPReader` ?

boks1971:
@davidzhao @Sean-Der I have squashed the commits and this is ready to go. David tried an impl in LiveKit server SDK with this and it is working. I may be offline for the next three days. I can merge when I get back or please feel free to merge if it looks good.

davidzhao (approved):

PR #2142, opened 2022-03-09
optimize h.264 reader to reuse pre-allocated buffer during read

leslie-wang:
#### Description
optimize h.264 reader to reuse pre-allocated buffer during read buffer, and may reduce the chance of GC.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/media/h264reader/h264reader.go
+++ b/pkg/media/h264reader/h264reader.go
@@ -14,6 +14,7 @@ type H264Reader struct {
 	countOfConsecutiveZeroBytes int
 	nalPrefixParsed             bool
 	readBuffer                  []byte
+	tmpReadBuf                  []byte
 }
 
 var (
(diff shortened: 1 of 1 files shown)


Sean-Der:
This makes sense to me @leslie-wang!

Can you make this a member of the `H264Reader`? Right now if a user has two thread readers it would cause a race.

dwebfan:
> This makes sense to me @leslie-wang! nice work
> 
> Can you make this a member of the `H264Reader`? Right now if a user has two thread readers it would cause a race.

Yes, good point. Pushed new change.

Sean-Der (approved):

Sean-Der:
thank you so much @leslie-wang !

Are you interested in staying involved with Pion? If you are I would love to add you to the Pion organization and hear your opinion on how we can make the project better :)

leslie-wang:
@Sean-Der Pion is a great project. Really appreciate all your effort. I like to contribute more for sure. I'm going to have another PR on h264reader shortly. Currently the reader assumes input always starts with start code 000001 or 00000001, but sometimes, reader could be one streaming media and may not start with exact prefix. Thus, reader probably need handle alignment case. I'll send a PR shortly.

BTW, Pion will be core part of reference implementation of this fun project: https://opencameras.github.io. It is still at early stage, so keep tune for now. :-)

PR #2144, opened 2022-03-11
return actual error during h264 stream read

leslie-wang:
#### Description
current h264reader will not return actual error, thus, as a user, I don't know error is caused by EOF, or socket close, or other failures. The fix will return the error to caller for debugging and troubleshooting

#### Reference issue
Fixes #...


Sean-Der:
Tagged `v3.1.25`

Issue #2152, opened 2022-03-18
DataChannel breaks when machine is under load (race condition?)


iamcalledrob:
### Your environment.
  - Version: v3.1.25
  - Browser: not using the browser implementation
  - Other Information:
    - go version go1.17.6 darwin/arm64
    - macOS 12.2.1, M1 Pro, 32gb memory

### What did you do?

I've been running into this when using dataChannels in a high-throughput scenario. As a simple repro, the problem is demonstrable using the built-in "data-channels-flow-control" example.

Easy repro steps:
1. Run the data-channels-flow-control example. (`cd examples/data-channels-flow-control; go run main.go`)
2. Make the CPU work hard, maxing it out. (`yes > /dev/null &` x number of cores to keep them all busy)

### What did you expect?
1. The throughput to remain reliable, potentially with lower throughput.

### What happened?
1. The example begins working and showing throughput, e.g. `Throughput: 632.810 Mbps`.
2. `top` shows meaningful CPU usage (~250%) and the process with the `STATE` of `running` most of the time.
3. Soon after the machine is under load, a mux error is printed: ```mux ERROR: 2022/03/18 14:18:20 mux: ending readLoop dispatch error packetio.Buffer is full, discarding write```
4. After the error is printed, no further data is transferred. The write buffer is no longer automatically flushed. The process goes to sleep.
5. In the data-channels-flow-control example, the effect is to show the average throughput dropping endlessly.

---

It seems like there may be a race condition regarding this buffer that's more easily triggered under load?

Notably, in internal/mux/mux.go, there's a comment when setting the packetio buffer suggesting that the buffer should never be able to be full.
https://github.com/pion/webrtc/blob/655daa9689576421c194059dd4f3a05cae544a07/internal/mux/mux.go#L59-L62

Video of repro attached.

https://user-images.githubusercontent.com/87964/159092362-3dae173a-5c4e-4117-99ea-e993b2c47fd8.mp4

iamcalledrob:
Relatedly, the same issue can be seen in the `TestStressDuplex` test by increasing the MsgCount **(mux_test.go, Line 39)**

https://github.com/pion/webrtc/blob/655daa9689576421c194059dd4f3a05cae544a07/internal/mux/mux_test.go#L30-L43

On my machine:
- `MsgSize: 2048, MsgCount: 200` never fails. This makes sense because 2048*200 < 1MB, (the fixed buffer size)
[...]

PR #2153, opened 2022-03-19
Mux: Wait and re-try writes (infinitely) when endpoint buffer is full

iamcalledrob:
#### Description
Mux will now wait and re-try writes (infinitely) when an endpoint's buffer is full.
Other code comments suggest that the buffer should never become full (see line 60 below), but that seems to not always be the case—and this is causing a race condition.
https://github.com/pion/webrtc/blob/655daa9689576421c194059dd4f3a05cae544a07/internal/mux/mux.go#L59-L62

This PR's approach may not be an ideal solution to the underlying problem, but as a band-aid it does appear to fix the issue.

This PR also updates the associated stress test to add CPU load while running, in order to trigger the bug outlined in https://github.com/pion/webrtc/issues/2152.
It's possible to verify this fix by merging only the updated test first—you'll see that quite often the test will stall with `mux ERROR: 2022/03/18 14:18:20 mux: ending readLoop dispatch error packetio.Buffer is full, discarding write` and eventually fail due to timeout.

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/2152

### DIFF
--- a/internal/mux/mux.go
+++ b/internal/mux/mux.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/pion/ice/v2"
 	"github.com/pion/logging"
@@ -57,8 +58,6 @@ func (m *Mux) NewEndpoint(f MatchFunc) *Endpoint {
 	}
 
 	// Set a maximum size of the buffer in bytes.
-	// NOTE: We actually won't get anywhere close to this limit.
-	// SRTP will constantly read from the endpoint and drop packets if it's full.
 	e.buffer.SetLimitSize(maxBufferSize)
 
 	m.lock.Lock()
@@ -146,7 +145,16 @@ func (m *Mux) dispatch(buf []byte) error {
 		return nil
 	}
 
-	_, err := endpoint.buffer.Write(buf)
+	// Attempt again later when encountering a temporarily full buffer (fix for #2152)
+	var err error
+	for {
+		_, err = endpoint.buffer.Write(buf)
+		if !errors.Is(err, packetio.ErrFull) {
+			break
+		}
+		time.Sleep(20 * time.Millisecond)
+	}
+
 	if err != nil {
 		return err
 	}
--- a/internal/mux/mux_test.go
+++ b/internal/mux/mux_test.go
@@ -3,6 +3,7 @@ package mux
 import (
 	"io"
 	"net"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


Sean-Der:
@iamcalledrob Thank you so much for diving deep on this! This is a great change. Might be worth doing a debug log?

So the issue is that `pion/dtls` and then `pion/sctp` aren't pulling buffers fast enough. This change instead will cause us to throw away incoming packets (which I think is better behavior)

Also worth benchmarking and figuring out where things are slowing down for you! If we see any obvious changes I would love to work on it together :)

@at-wat what do you think? Do you see any possible negative impacts?

Sean-Der:
@iamcalledrob oh would you also mind deleting the note around SRTP? It isn't very helpful/misleading.

at-wat:
If system can't handle packets fast enough and retry to process all packets and the condition continues, it will infinitely increase memory usage I guess.
I think it should be disabled by default to avoid decreasing DoS attack tolerance.

iamcalledrob:
@at-wat I'm not sure that this would lead to infinite memory usage, but please correct me if I'm wrong here.

My understanding is that:
1. The kernel's socket receive buffer feeds into pion
2. Pion is currently able to read packets from the kernel faster than it can process them, leading to bug #2152. When this occurs it currently leads to undesirable behaviour—i.e. breaking out of the read loop.
3. This PR would have the effect of slowing down the speed that Pion reads from the socket receive buffer to match the speed that it's able to process the packets—essentially providing some back-pressure.
4. The kernel socket receive buffer is not unbounded, and if it becomes full then incoming UDP packets would get dropped by the kernel automatically.

Assuming this understanding is correct (big assumption!), this wouldn't lead to an increase in memory usage, since excess packets would be dropped rather than accumulating in a buffer somewhere.

@Sean-Der I will update the PR soon to remove the existing misleading comments :)

Sean-Der:
@iamcalledrob Yes you are right! The incoming packets from `pion/ice` will just be dropped. If you turn on trace logging you should see this line being fired.

```
    if _, err := c.agent().buffer.Write(buffer); err != nil {
        log.Warnf("failed to write packet")
    }
```

* Would you also mind adjusting the tests so WASM can pass? Maybe just move that stress test into a file that is marked as non-WASM?
* Can you update the commit message. You can test that it passed the linter with `bash .github/lint-commit-message.sh` I would do something like

```
Wait and re-try when Mux destination ErrFull

Resolves #2152
```

* Fix linter error. You can get it by running `golangci-lint run`

```
 internal/mux/mux_test.go#L52

SA5004: should not have an empty default case in a for+select loop; the loop will spin (staticcheck)
```

iamcalledrob:
@Sean-Der I've addressed the issues you outlined :) Linter and tests should be good-to-go.

Longer-term, a better solution might be for packetio.Buffer to offer a way to block writes until it has space for the write. Or return ErrFull but provide a mechanism to wait until the next Read occurs, so we can retry the write as soon as possible rather than polling every 20ms (a very arbitrary interval!).

Sean-Der (approved):

Sean-Der:
Great work! I approved @iamcalledrob 

Totally agree with future improvements. Blocking sounds like a good idea to me, the only thing we need to watch out for is performance impact. packetio.Buffer is really easy to tweak and then negatively impact things :(

I added you to the Pion org, so hit that `Rebase and merge` button w/e you are ready :) 

You should also push a tag! `git tag v3.1.26 && git push --tags` after you have merged. Make sure you tag the commit in master. Hope you stay involved with Pion. Would love your help in pushing it to the next level re: performance of DataChannels.

PR #2156, opened 2022-03-21
Unlock on error in internal/mux

### LINKED ISSUE #2154: Don't unlock when closing endpoint is failed.
https://github.com/pion/webrtc/blob/93b0f2a75eae5139362faf5346f0b6d53cfbf20d/internal/mux/mux.go#L82

```
func (m *Mux) Close() error {
	m.lock.Lock()
	for e := range m.endpoints {
		err := e.close()
		if err != nil {
			return err // SHOULD ADD m.lock.Unlock() beforeit
		}

		delete(m.endpoints, e)
	}
	m.lock.Unlock()

	err := m.nextConn.Close()
	if err != nil {
		return err
	}

	// Wait for readLoop to end
	<-m.closedCh

	return nil
}
```

Sean-Der:
Resolves #2154

### DIFF
--- a/internal/mux/mux.go
+++ b/internal/mux/mux.go
@@ -78,8 +78,8 @@ func (m *Mux) RemoveEndpoint(e *Endpoint) {
 func (m *Mux) Close() error {
 	m.lock.Lock()
 	for e := range m.endpoints {
-		err := e.close()
-		if err != nil {
+		if err := e.close(); err != nil {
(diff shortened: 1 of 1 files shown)


davidzhao:
thanks for fixing this! I was just digging into a user-reported deadlock in LiveKit that happens when a client's reconnecting *and* has a poor connection. Sounds like this could have been it.

PR #2160, opened 2022-03-28
wrap demo.js within window.onload for swap-tracks example

### LINKED ISSUE #2159: swap-tracks example has a race condition on startup
### Your environment.
  - Version: 3.1.27 (latest)
  - Browser:  Edge 99.0.1150.52
  - Other Information - Windows on Edge, but presumably other browsers as well

### What did you do?

I built the examples app: `cd examples` then `go build`
I ran the examples app: `examples.exe -address :8000`
Opened up Edge to `http://localhost:8000`
Clicked the "swap-tracks" link in order to initiate the swap-tracks demo.

### What did you expect?

I expected the swap-tracks demo web page to have the base64 encoded SDP available in the top text box and to see 3 pulsating circles in the canvas control at the bottom of the page.

### What happened?

The javascript in demo.js hits an error upon startup and aborts. There's no SDP in the text box nor pulsating circles.  The error is this:

       Uncaught TypeError: Cannot read properties of null (reading 'captureStream') at demo.js:31:39

### Root cause

Given the way the `examples` webserver injects each app's `demo.js` into the HTML, this creates the issue. The `demo.js` script attempts to run before the rest of the page has finished loading.  When demo.js invokes `document.getElementById`, the document hasn't finished loading. Hence, the canvas element it attempts to reference is null. The subsequent call to captureStream on null causes the script to abort.

There's a simple fix: Wait for window.onload event to occur before enabling demo.js to run. I will submit a pull request shortly.

jselbie:
#### Description

The fix for [Issue 2159](https://github.com/pion/webrtc/issues/2159). "swap-tracks example has a race condition on startup"

The simple fix is to wrap demo.js within the window.onload event.

The majority of the change is just encapsulating the original code within a `window.onload` event handler. Hence, most of this diff is a result of indenting the original code.

#### Reference issue
Fixes #2159


Sean-Der:
Great job catching this @jselbie!

Can you try the following? I am not against doing the onload, we just have to do it in all the files.

```
diff --git a/examples/example.html b/examples/example.html
index 668d123..eef2d93 100644
--- a/examples/example.html
+++ b/examples/example.html
@@ -11,17 +11,6 @@
                        }
                </style>
                <link rel="stylesheet" type="text/css" href="demo.css">
-               {{ if .JS }}
-               <script src="demo.js"></script>
-               {{ else }}
-               <script src="/wasm_exec.js"></script>
-               <script>
-                       const go = new Go();
-                       WebAssembly.instantiateStreaming(fetch("demo.wasm"), go.importObject).then((result) => {
-                               go.run(result.instance);
-                       });
-               </script>
-               {{ end }}
        </head>
        <body>
                <h1>{{ .Title }}</h1>
@@ -31,4 +20,15 @@
                {{ template "demo.html" }}
                </div>
        </body>
+       {{ if .JS }}
+       <script src="demo.js"></script>
+       {{ else }}
+       <script src="/wasm_exec.js"></script>
+       <script>
+               const go = new Go();
+               WebAssembly.instantiateStreaming(fetch("demo.wasm"), go.importObject).then((result) => {
+                       go.run(result.instance);
+               });
+       </script>
+       {{ end }}
 </html>
 ```

Sean-Der (approved):

PR #2170, opened 2022-04-05
Make TrackLocalContext an interface

kevmo314:
This allows external users to provide their own TrackLocalContext to be
bound to a track.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -60,6 +60,7 @@ Dean Sheather <dean@coder.com>
 decanus <7621705+decanus@users.noreply.github.com>
 Denis <Hixon10@yandex.ru>
 digitalix <digitalix4@gmail.com>
+do-hyung-kim <do_hyung.kim@navercorp.com>
 donotanswer <viktor.ferter@doclerholding.com>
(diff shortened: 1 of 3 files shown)


Sean-Der (approved):

Sean-Der:
That's pretty interesting @kevmo314 !

Are you calling `Bind` yourself? I had only really thought is something that `webrtc` calls for the users.

kevmo314:
> Switching it from a struct to a pointer technically is API breakage
> 
> Let me check `ion-sfu` and `mediadevices` first to make sure we don't cause any problems (I hope not though)

I think because all the properties are private it shouldn't cause any breakages but I'll hold off, let me know when you're comfortable merging :)

Sean-Der:
I tested `pion/mediadevices`, `glimesh/broadcast-box` and `livekit/livekit` (all users of TrackLocalContext)

This could have negative effects on users using reflection. I hope that this doesn't effect anyone. If it does reach out and happy to undo!

thanks

at-wat:
I second this change, but I believe we should migrate versioning rule to semver someday (bump MAJOR version when making incompatible API changes)

PR #2179, opened 2022-04-17
Add testable example for GatheringCompletePromise

jech:
It is not immediately obvious how to do traditional (non-trickle)
ICE in Pion, and the current documentation does not make it
sufficiently clear.  Add a testable example for non-Trickle ICE.

### DIFF
--- a/gathering_complete_promise_example_test.go
+++ b/gathering_complete_promise_example_test.go
@@ -1,0 +1,64 @@
+package webrtc
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ExampleGatheringCompletePromise demonstrates how to implement
+// non-trickle ICE in Pion, an older form of ICE that does not require an
+// asynchronous side channel between peers: negotiation is just a single
+// offer-answer exchange.  It works by explicitly waiting for all local
+// ICE candidates to have been gathered before sending an offer to the peer.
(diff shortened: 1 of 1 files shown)


Sean-Der:
Yes that would be nice! I don't know if users will look in the root of the repo though.

What do you think of putting it in `examples/non-trickle-ice` ?

jech:
The point of a testable example is that it appears in the package documentation.  Moving it somewhere else would put it in a different package, which would defeat the purpose.

Sean-Der:
@jech I didn't know that it would be in the package documentation! Makes sense to me, will merge now

Sean-Der (approved):

jech:
It looks like it worked: https://pkg.go.dev/github.com/pion/webrtc/v3@v3.1.29-0.20220418030300-45061ef3e78e#GatheringCompletePromise

Sean-Der:
That's a great improvement. I will need to add more in-tree examples, I had no idea this feature existed!

jech:
It's not always suitable, but when it is, it's pretty handy.  I particularly like the thing that examples get compiled and optionally tested, so they don't become obsolete.  It's documented here: https://pkg.go.dev/testing#hdr-Examples

PR #2184, opened 2022-04-19
mux: drop packets when buffer is full

### LINKED ISSUE #2180: Infinite loop in Mux.NewEndpoint
Commit 93b0f2a75eae5139362faf5346f0b6d53cfbf20d introduced an infinite loop in a library function.  The discussion in bug #2152 seems to indicate this papers over a bug somewhere else — SCTP should be able to recover from a lost packet, so there's no reason to introduce the busy wait when the mux happens to be full.

At the very least, we should make sure that there is a hard bound on the amount of time we spend in the busy wait.  Ideally, we should find what is the root cause of the issue noted by @iamcalledrob, and fix that properly.

iamcalledrob:
Removes wait-and-retry behavior in favor of dropping packets when the endpoint read buffer is full.

For SCTP, the dropped packets don't matter because they will be re-sent if necessary. I've confirmed that this is happening and that SCTP continues to recover properly.

Removing retries reverts the fix in 93b0f2a75eae5139362faf5346f0b6d53cfbf20d where tests that rely on test.StressDuplex would hang (and eventually timeout panic) if the test machine was under load, leading to a full buffer. The hang occurs because test.StressDuplex blocks until all packets are received. I've added a note about this to the relevant test. It's unlikely to be encountered.

Resolves #2180

### DIFF
--- a/internal/mux/mux.go
+++ b/internal/mux/mux.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"net"
 	"sync"
-	"time"
 
 	"github.com/pion/ice/v2"
 	"github.com/pion/logging"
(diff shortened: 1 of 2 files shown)


iamcalledrob:
> It's unlikely to be encountered.
It was just encountered in the Test / Go i386 tests above 🥲

jech:
Thanks.  I approve this change.  I belive it's the tests that are incorrect.

This change is consistent with the rest of Pion, which doesn't return errors for innocuous failures such as this.  For the record, I believe that a better approach would be to have a notion of innocuous failure, that can be ignored, and to reliably propagate such failures to the client application.  See for example https://datatracker.ietf.org/doc/html/rfc816 (from 1982!) for a much better argument than I could possibly write.

Sean-Der:
Agree, I am removing the test entirely.

We have lots of 'stress tests' and I don't remember them ever catching a bug. In the future I would love to see us take benchmarks more seriously, but I don't have the availability to do that today :/

Sean-Der:
Added `TestNonFatalDispatch` to make sure this doesn't regress/is explained. Merging! Thanks again for taking this on @iamcalledrob :)

Sean-Der (approved):

PR #2198, opened 2022-04-30
Resolve issue with rejecting tracks from Firefox

boushley:
When running the reflect example against Firefox 99.0.1 I was receiving
errors around parsing the SDP from Pion.

The error was:

  SIPCC Failed to parse SDP: SDP Parse Error on line 50:  c= connection
  line not specified for every media level, validation failed.

According to the RFC [4566 5.7](https://datatracker.ietf.org/doc/html/rfc4566#section-5.7)

  A session description MUST contain either at least one "c=" field in
  each media description or a single "c=" field at the session level.
  It MAY contain a single session-level "c=" field and additional "c="
  field(s) per media description, in which case the per-media values
  override the session-level settings for the respective media.

I don't see any errors in the Pion repos for this. Interestingly the
webrtc-rs project that is porting Pion to rust has [an issue](https://github.com/webrtc-rs/webrtc/issues/144) where they're encountering this same error.

### DIFF
--- a/sdp.go
+++ b/sdp.go
@@ -424,12 +424,23 @@ func addTransceiverSDP(
 		}
 
 		// Explicitly reject track if we don't have the codec
+		// We need to include connection information even if we're rejecting a track, otherwise Firefox will fail to
+		// parse the SDP with an error like:
+		// SIPCC Failed to parse SDP: SDP Parse Error on line 50:  c= connection line not specified for every media level, validation failed.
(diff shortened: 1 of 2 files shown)


Sean-Der:
You are on :fire: @boushley 

mind adding a test? If we already have a rejection test just another assert would be perfect.

boushley:
Now with 100% more tests!

I didn't see any existing tests for rejected tracks. Let me know if I missed something!

PR #2202, opened 2022-05-05
Add json serialization support to ICECredentialType

wdouglass:
#### Description
When serializing to/from json, the ICECredentialType should serialize as as a string instead of an enumeration.

### DIFF
--- a/configuration_test.go
+++ b/configuration_test.go
@@ -41,7 +41,7 @@ func TestConfiguration_getICEServers(t *testing.T) {
 
 func TestConfigurationJSON(t *testing.T) {
 	j := `{
-    "iceServers": [{"URLs": ["turn:turn.example.org"],
+    "iceServers": [{"urls": ["turn:turn.example.org"],
                     "username": "jch",
                     "credential": "topsecret"
                   }],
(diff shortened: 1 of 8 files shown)


wdouglass:
marking draft until i figure out the lint/testing stuff

wdouglass:
So, as usually happens, a small change turned into a bigger one. I was not expecting this to have so many commits!

Tests are passing, this is ready for review. Squash if you like!

configuration_test.go
```diff
@@ -41,7 +41,7 @@ func TestConfiguration_getICEServers(t *testing.T) {
 
 func TestConfigurationJSON(t *testing.T) {
 	j := `{
-    "iceServers": [{"URLs": ["turn:turn.example.org"],
+    "iceServers": [{"urls": ["turn:turn.example.org"],
```
Sean-Der, line 44: @wdouglass Is this a breaking change? 

Is it possible we can accept `URLs` and `urls` so we don't break existing users?

wdouglass: I'm not sure how this test  ever worked -- if you look in iceserver.go, that field always serializes/deserializes to "urls" anyway -- this was a faulty test.

also, in the equivalent structure in "upstream" webrtc, the field is "urls"

Sean-Der (approved):

PR #2208, opened 2022-05-08
Fix incorrect conversion between integer types

daenney:
The result of strconv.Atoi uses an architecture dependent bit size. It
is not safe to use when converted to an integer type of a smaller size,
without performing bounds checking.

Instead, use Parse(U)int with the desired base and size. Though this
returns a (u)int64, it can then be safely converted into a (u)int8 etc.


Sean-Der:
nice! Exciting to see you outside of dtls :)

daenney:
> nice! Exciting to see you outside of dtls :)

Ha, thanks. Blame CodeQL, I was checking up on things :smile:.

PR #2211, opened 2022-05-10
RTP header extension IDs should be unique RTP header extension ID mappings should be unique

k0nserv:
The header extension IDs used in a session should be unique accross all
tracks within the session but, as this test case illustrates, they
aren't. The extension ID 1 is re-used between audio and video tracks in
the same media engine.

Ported from [webrtc-rs][0] to confirm that it wasn't a bug in webrtc-rs
itself.

[0]: https://github.com/webrtc-rs/webrtc/pull/154

### DIFF
--- a/mediaengine_test.go
+++ b/mediaengine_test.go
@@ -460,3 +460,62 @@ func TestUpdateHeaderExtenstionToClonedMediaEngine(t *testing.T) {
 	validate(&src)
 	validate(src.copy())
 }
+
+func TestExtensionIdCollision(t *testing.T) {
+	mustParse := func(raw string) sdp.SessionDescription {
+		s := sdp.SessionDescription{}
+		assert.NoError(t, s.Unmarshal([]byte(raw)))
+		return s
+	}
+	sdp_snippet := `v=0
+o=- 4596489990601351948 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+m=audio 9 UDP/TLS/RTP/SAVPF 111
+a=extmap:7 urn:ietf:params:rtp-hdrext:sdes:mid
+a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
+a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
+a=rtpmap:111 opus/48000/2
+`
+
+	m := MediaEngine{}
+	assert.NoError(t, m.RegisterDefaultCodecs())
+
+	assert.NoError(t, m.RegisterHeaderExtension(RTPHeaderExtensionCapability{"urn:3gpp:video-orientation"}, RTPCodecTypeVideo))
+
+	audioExtensions := []string{
+		"urn:ietf:params:rtp-hdrext:ssrc-audio-level",
+		"urn:ietf:params:rtp-hdrext:sdes:mid",
+		"urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id",
+	}
+	for _, extension := range audioExtensions {
+		assert.NoError(t, m.RegisterHeaderExtension(RTPHeaderExtensionCapability{extension}, RTPCodecTypeAudio))
+	}
+
+	assert.NoError(t, m.updateFromRemoteDescription(mustParse(sdp_snippet)))
+
+	assert.True(t, m.negotiatedAudio)
+	assert.False(t, m.negotiatedVideo)
+
+	id, audioNegotiated, videoNegotiated := m.getHeaderExtensionID(RTPHeaderExtensionCapability{sdp.ABSSendTimeURI})
+	assert.Equal(t, id, 0)
+	assert.False(t, audioNegotiated)
+	assert.False(t, videoNegotiated)
+
+	id, audioNegotiated, videoNegotiated = m.getHeaderExtensionID(RTPHeaderExtensionCapability{sdp.SDESMidURI})
+	assert.Equal(t, id, 7)
+	assert.True(t, audioNegotiated)
+	assert.False(t, videoNegotiated)
+
+	id, audioNegotiated, videoNegotiated = m.getHeaderExtensionID(RTPHeaderExtensionCapability{sdp.AudioLevelURI})
+	assert.Equal(t, id, 1)
+	assert.True(t, audioNegotiated)
+	assert.False(t, videoNegotiated)
+
+	params := m.getRTPParametersByKind(RTPCodecTypeVideo, []RTPTransceiverDirection{RTPTransceiverDirectionSendonly})
+	extensions := params.HeaderExtensions
+
+	assert.Equal(t, 1, len(extensions))
+	assert.Equal(t, "urn:3gpp:video-orientation", extensions[0].URI)
+	assert.NotEqual(t, 1, extensions[0].ID)
+}


k0nserv:
I think we are saying the same thing, but I could have been clearer. 

What I meant is that the mapping of id to extension should be unique across the whole session hence if `3` is mapped to `http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01` for an audio stream it must also be mapped to `http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01` in all other streams(audio or video). 

I'm not sure if the inverse applies i.e. is it okay for `http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01` to map to both id `3` and `4`? In any case, that's not desirable because the id space is quite limited.

adriancable:
@k0nserv - are you sure that Pion gets this wrong? If so, can you put together a minimal example to demonstrate it in go? I have just had a look at the code for `RegisterHeaderExtension` in `mediaengine.go`, and there is this:

```
        extensionIndex := -1
        for i := range m.headerExtensions {
                if extension.URI == m.headerExtensions[i].uri {
                        extensionIndex = i
                }
        }

        if extensionIndex == -1 {
                m.headerExtensions = append(m.headerExtensions, mediaEngineHeaderExtension{})
                extensionIndex = len(m.headerExtensions) - 1
        }
```

Unless this code has a subtle bug, it does the right thing. The first part makes sure that any extmap URI that's already registered (audio or video) gets reassigned the same ID. Otherwise, the second part assigns a new ID which is 1 more than the largest ID already used. The logic sounds correct.

k0nserv:
That logic is correct, but it doesn't account of accepting mappings from peers. Look at the test case I added in this PR.

The code you referenced cause `urn:3gpp:video-orientation` to be assigned ID `1`, but it's then possible for Pion to accept a different mapping of `1` to `urn:ietf:params:rtp-hdrext:ssrc-audio-level` at which point using `1` for `urn:3gpp:video-orientation` is no longer legal

adriancable:
@k0nserv - just to clarify re. my request for a minimal example. I did see your test case, but that uses a bunch of internal functions and it isn't clear to me that the sequence of operations in your test matches what pion would actually do.

So when I am asking for a minimal example, I am asking for code that looks something like:

```
...
err = m.RegisterHeaderExtension(webrtc.RTPHeaderExtensionCapability{URI: "urn:ietf:params:rtp-hdrext:sdes:mid"}, webrtc.RTPCodecTypeVideo))
...
err = m.RegisterHeaderExtension(webrtc.RTPHeaderExtensionCapability{URI: "http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time"}, webrtc.RTPCodecTypeAudio))
err = m.RegisterHeaderExtension(webrtc.RTPHeaderExtensionCapability{URI: "urn:ietf:params:rtp-hdrext:sdes:mid"}, webrtc.RTPCodecTypeAudio))
...
offer, err = peerConnection.CreateOffer(nil)
err = peerConnection.SetLocalDescription(offer)
...
err = peerConnection.SetRemoteDescription(webrtc.SessionDescription{ Type: webrtc.NewSDPType("answer"), SDP: 'example' }))
...
```

Then show that the offer and/or the resulting state are incorrect.

k0nserv:
I don't really have time to do that. Suffice to say `updateFromRemoteDescription` is called by `SetRemoteDescription` and `getHeaderExtensionID` is used when creating SDPs. 

If you accept an offer that looks like this with `SetRemoteDescription`

```
v=0
o=- 4596489990601351948 2 IN IP4 127.0.0.1
s=-
t=0 0
m=audio 9 UDP/TLS/RTP/SAVPF 111
a=extmap:7 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=rtpmap:111 opus/48000/2
```

Then `1` will be mapped to `urn:ietf:params:rtp-hdrext:ssrc-audio-level`. If Pion later generates a video track there's nothing that prevents it from using the id `1`  for another registered extension(such as `urn:3gpp:video-orientation` which is what the test case illustrates).

adriancable:
@k0nserv - ah, I see. I think I misunderstood the original issue. I'll dig into it.

k0nserv:
> ah, I see. I think I misunderstood the original issue. I'll dig into it.

No problem, there's pseudo code for the solution I created in the webrtc-rs [PR](https://github.com/webrtc-rs/webrtc/pull/154) but I'd be happy to hear any ides for alternative (simpler) approaches if you can think of something

adriancable:
@k0nserv - I have spent some time thinking about this, and I am still not convinced there is an issue.

Correct me if I am wrong, but: your example test case appears to simulate the following scenario:

1. Let's call A the local (pion) side, B the remote
2. A registers some audio and video header extensions
3. B offers, with audio only and a bunch of extmaps
4. A answers, necessary keeping the same extmap IDs as in the offer in step 2
5. Then you are concerned that if A then adds a video track, you will get conflicting extmap IDs (and yes this is apparently demonstrated by your test case)

But #5 can't happen like this: side A can't add a video track at this point because side B's offer in step #3 has told A it can only accept audio.

For #5 to be allowable, in step #3 side B would have had to offer a video section (in which case it would have decided the extmap IDs and side A would respect them), but in your test case it didn't. Or, there would need to be some kind of out-of-band signalling to ask side B to make another offer with video. In this case side B would also decide the extmap IDs, and side A (the Pion side) would respect those IDs.

What am I missing?

adriancable:
@k0nserv - further to the above, is the scenario you are concerned with the specific case of what happens if sides A and B decide to switch the offer/answer roles, and so prior to step #5, there is some out-of-band signalling which results in side A (the Pion) side generating a new offer? (this seems to be the only case not covered in my above post)

k0nserv:
> But #5 can't happen like this: side A can't add a video track at this point because side B's offer in step #3 has told A it can only accept audio.

My understanding, and please correct me if I'm wrong, is that there's nothing that requires B to have add a video track ahead of time. It might support video but it just wasn't enabled at the time of that negotiation. It can still add it later. For example, B might be joining conference call and only negotiated an audio track(because the software defaults to camera off). Then later B might turn on the camera causing an offer that now includes video.

Continuing with this example let me outline an SFU case that triggers the bug I believe to exist. Again consider A=Pion, B=Web browser and an additional peer, C, which is also a web browser. 

Here's the flow of events:

1. A registers audio and video headers it supports
2. B joins and offers an audio track. 
3. A answers matching the ext mappings for the extensions it supports(This causes id `1` to map to `urn:ietf:params:rtp-hdrext:ssrc-audio-level`) in `negotiatedHeaderExtensions`.
4. C joins and offers a video track which A negotiates with it(Note: this is a separate session with separate mappings).
5. A offers B a video track(corresponding to the one it just got from C)
6. Because `urn:3gpp:video-orientation` was mapped to `1`(via `header_extensions`) that offer includes `a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level`for the `m=` line corresponding to the audio track and `a=extmap:1 urn:3gpp:video-orientation` for the `m=` line corresponding to the video track.

This is exactly what I see happening in my webrtc-rs example and what ends up being fixed by my PR.

**What actually happens:** Pion violates the rule that a single id should always map to a single extension by mapping `1` to both `urn:ietf:params:rtp-hdrext:ssrc-audio-level` and `urn:3gpp:video-orientation`.

**What I believe should happen:** In step six Pion notices that the id is already assigned to a different extension and should pick a new unused id for `urn:3gpp:video-orientation`.

adriancable:
@k0nserv - thanks for the clear example. I think that to make this work only `getRTPParametersByKind` needs to be fixed. I'll put together a PR shortly and you can take a look at that. If you agree my fix doesn't miss anything out I will then talk with @Sean-Der for a second set of eyes and suggest it be merged.

k0nserv:
FWIW I vaguely recall seeing Firefox get mad about the stability of ids you offer i.e. if you offer `http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01` as `1` for an audio track(which Firefox doesn't support that) it'll not be happy if you later offer `http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01` as some other id than `1` for a video track.

adriancable:
@k0nserv - I've created a new PR #2220 which should address this.

I've split the issue into two distinct failure modes (misassignment, and conflict), and also added a test which covers both of these. (Your original test covers the conflict case, but not the misassignment)

Sean-Der:
Thank you so much for taking this on @k0nserv and @adriancable feels great to just show up and have a deep bug fixed :)

Merged with 63b4529ca181d5753c2739dee5c0784d86c0af90

PR #2212, opened 2022-05-10
Upgrade DTLS dependency to 2.1.4

daenney:
The 2.1.4 release contains two security fixes.


daenney:
We should probably tag a new webrtc release after merging this too.

Sean-Der (approved):

PR #2217, opened 2022-05-11
Remove occurrences of Unknonwn ICECredentialType

jech:
In 3.1.37, we removed the possibility for ICECredentialType
to be Unknown.  Remove remaining tests for this case from
the WASM code.


Sean-Der:
Thanks for cleaning this up @jech !

Sean-Der (approved):

PR #2248, opened 2022-06-02
Assign DataChannel ID directly to fix conflicts

kylecarbs:
When opening DataChannels in parallel it was possible
for multiple to be allocated the same ID since the
`generateAndSetDataChannelID` func references the
DataChannel ID, which was being set after the unlock.

I tested opening 1000 DataChannels at once, and it
seems to work perfectly now!


Sean-Der:
Is it safe to modify `d.id` without holding the lock? Maybe we could make the id a `atomic.Value`? Then we drop the `unlock` and `lock`

PR #2262, opened 2022-06-21
Getting `ErrUnsupportedCodec` when registering codec after negotiation 

### LINKED ISSUE #2263: Unable to register multiple codecs after 1st successful negotiation
### Your environment.
  - Version: `v3.1.24` (can be reproduced on latest `master`)
  - Browser: NA
  - Other Information - Created test case #2262 to reproduce the below steps

### What did you do?

I saw that i'm not able to subscribe multiple tracks of different codecs if the codecs are registered on the fly (just before the transceiver has to be created). E.g. Initialise `MediaEngine` with 0 codecs. Steps, 
- say for `PeerConnection `offerer` initialised with `MediaEngine` having no codecs registered initially and other `PeerConnection` `answerer` on client side with H264 and VP8 codecs registered  
- Some chrome based client (say 'chrome') publishes track-A with codec `H264`
- register codec to `MediaEngine` of `offerer` pc and add this track to `offerer`'s transceiver
- do the offer/answer sdp negotiation b/w `offerer` and `answerer` pc --  ✅ Success with H264 registered in `MediaEngine` of `offerer` 
- 'chrome' publishes another track track-B with codec VP8 ->
- register codec to `MediaEngine` of `offerer` pc and add this track to `offerer`'s transceiver 
- redo the offer/answer sdp negotiation b/w `offerer` and `answerer` pc --  ❌ `offerer.SetRemoteDescription(answer)` Fails because VP8 was not pushed to `MediaEngine.negotiatedCodecs` 

### What did you expect?
Expected `SetRemoteDescription` to finish without any errors

### What happened?
`SetRemoteDescription` raised `ErrUnsupportedCodec` even when `me.RegisterCodec(VP8)` was successful. 

What happens internally I saw is that the second time we call `SetRemoteDescription` -> `updateFromRemoteDescription` does not register the new codec because `negotiatedVideo` was set to true when the first negotiation was successful.

In [**`ion-sfu/mediaengine.go`**](https://github.com/pion/ion-sfu/blob/master/pkg/sfu/mediaengine.go) as the subscriber pc does not have any codecs and each codec is registered when a new `DownTrack` is added.  This creates problem when a new track with different codec is added.

adityaa30:
#### Reference issue
Fixes #2263


Sean-Der:
Thanks for the PR @adityaa30! I don't think merging is the right thing to do. 

I responded with the reasoning/design of the API in https://github.com/pion/webrtc/issues/2263#issuecomment-1666377485. Will also copy here to make easier

-----

In earlier versions of Pion it was possible to share a copy of the MediaEngine between multiple PeerConnections. This would fail when shared between multiple types of remote agents (FireFox/Chrome use different PayloadTypes etc...)

For the application you have described here I think the 'right' thing to do here is create a MediaEngine with VP8 and H264. Then if you want to force the remote to use a specific codec use [SetCodecPreference](https://pkg.go.dev/github.com/pion/webrtc/v3#RTPTransceiver.SetCodecPreferences) on each individual transceiver.

PR #2274, opened 2022-07-07
Add failing test for stopped transceivers

k0nserv:
#### Description

A small failing test were upon stopping a transceiver it is not rejected in the following offer.


JoTurk:
@k0nserv sorry that your PR got ghosted, I'll try to get this behavior and your commit for pion@4.3.0, thank you.

k0nserv:
I'm not fussed personally (I don't use Pion). I recall @Sean-Der and I discussed this on Slack when I raised this PR, but really it's pretty big change in behaviour that needs to be considered carefully

PR #2287, opened 2022-07-23
Fix pion-to-pion example in docker-compose

### LINKED ISSUE #2286: pion-to-pion example build fail
```
webrtc-master/examples/pion-to-pion $ docker-compose up
Creating network "pion-to-pion_default" with the default driver
Building answer
Step 1/5 : FROM golang:1.18
1.18: Pulling from library/golang
d836772a1c1f: Pull complete
66a9e63c657a: Pull complete
d1989b6e74cf: Pull complete
c28818711e1e: Pull complete
9d6246ba248c: Pull complete
59e23c1b5fe8: Pull complete
4adeeae69874: Pull complete
Digest: sha256:9349ed889adb906efa5ebc06485fe1b6a12fb265a01c9266a137bb1352565560
Status: Downloaded newer image for golang:1.18
 ---> e3c0472b1b62
Step 2/5 : ENV GO111MODULE=on
 ---> Running in 57c18ed4d07a
Removing intermediate container 57c18ed4d07a
 ---> cfefcae629ba
Step 3/5 : RUN go get -u github.com/pion/webrtc/v3/examples/pion-to-pion/answer
 ---> Running in 738ae817365b
go: go.mod file not found in current directory or any parent directory.
        'go get' is no longer supported outside a module.
        To build and install a command, use 'go install' with a version,
        like 'go install example.com/cmd@latest'
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
ERROR: Service 'answer' failed to build : The command '/bin/sh -c go get -u github.com/pion/webrtc/v3/examples/pion-to-pion/answer' returned a non-zero code: 1
```

mengelbart:
Fixes the command line arguments to use correct addresses and uses go
install instead of now unsupported go get.

Fixes #2286


examples/pion-to-pion/docker-compose.yml
```diff
@@ -13,5 +14,6 @@ services:
       - answer
     container_name: offer
     build: ./offer
+    command: offer -answer-address answer:60000
```
ashellunts, line 17: answer port is 50000 (line 10), isn't it?

mengelbart: I am actually not sure if line 10 is needed at all since we only want the two containers to talk to each other.

The flag value `answer:60000` is needed to let the offerer know where to find the answer. By default, this would be `:60000` on localhost (see [`answer/main.go:35`](https://github.com/pion/webrtc/blob/fix/pion-to-pion-docker-compose/examples/pion-to-pion/answer/main.go#L35)). But when both run in different containers, they need the name of the other container instead.

ashellunts (approved):

examples/pion-to-pion/docker-compose.yml
mengelbart: I don't see how that is related to the issue this PR is supposed to fix. It may be useful to expose ports if you want to connect from the host machine instead.

ashellunts: As you said answer container works on port 60000, so exposing port 50000 does not make sense.

PR #2301, opened 2022-09-14
Add currentDirection to RTPTransceiver

cnderrauber:
add currentDirection to RTPTransceiver, don't reuse transceiver if its currentDirection is sendrecv or sendonly

#### Description
If remove track from transceiver then reuse it to add track without re-negotiation to transit its direction of sdp, then safari will not fire `removetrack` and `ontrack` events for that transceiver, so the mediatrack is lost.
Step:
1. sfu offer with 2 tracks, negotiation with safari, all things work, safari fire 2 `onTrack` event
2. sfu call RemoveTrack then AddTrack, will reuse rtptransceiver, after that, re-negotiation with safari. No new mid/transceiver in safari, the reused trasceiver changed its stream attributes to the new track.
3. safari can't fire `onTrack` for the new track.

According to https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addtrack, if the transceiver can be reused only if it's currentDirection never be sendrecv or sendonly. But that will cause sdp inflate in sfu. So we only check currentDirection's current value, that's worked for all browsers.

> If any [RTCRtpSender](https://www.w3.org/TR/webrtc/#dom-rtcrtpsender) object in senders matches all the following criteria, let sender be that object, or null otherwise:
> 
> The sender's track is null.
> 
> The [transceiver kind](https://www.w3.org/TR/webrtc/#dfn-transceiver-kind) of the [RTCRtpTransceiver](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiver), associated with the sender, matches kind.
> 
> The [[[Stopping]]](https://www.w3.org/TR/webrtc/#dfn-stopping) slot of the [RTCRtpTransceiver](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiver) associated with the sender is false.
> 
> The sender has never been used to send. More precisely, the [[[CurrentDirection]]](https://www.w3.org/TR/webrtc/#dfn-currentdirection) slot of the [RTCRtpTransceiver](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiver) associated with the sender has never had a value of "[sendrecv](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiverdirection-sendrecv)" or "[sendonly](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiverdirection-sendonly)".

#### Reference issue
I think #1843 is related to this fix, but chrome works well on my side, only safari have issue, with this PR, safari works well

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -984,6 +984,7 @@ func (pc *PeerConnection) SetLocalDescription(desc SessionDescription) error {
 	weAnswer := desc.Type == SDPTypeAnswer
 	remoteDesc := pc.RemoteDescription()
 	if weAnswer && remoteDesc != nil {
+		_ = setRTPTransceiverCurrentDirection(&desc, currentTransceivers, false)
 		if err := pc.startRTPSenders(currentTransceivers); err != nil {
 			return err
 		}
@@ -1143,6 +1144,7 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 
 	if isRenegotation {
 		if weOffer {
+			_ = setRTPTransceiverCurrentDirection(&desc, currentTransceivers, true)
 			if err = pc.startRTPSenders(currentTransceivers); err != nil {
 				return err
 			}
@@ -1172,6 +1174,7 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error { 
 	// Start the networking in a new routine since it will block until
 	// the connection is actually established.
 	if weOffer {
+		_ = setRTPTransceiverCurrentDirection(&desc, currentTransceivers, true)
 		if err := pc.startRTPSenders(currentTransceivers); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


peerconnection.go
```diff
@@ -1706,7 +1754,13 @@ func (pc *PeerConnection) AddTrack(track TrackLocal) (*RTPSender, error) {
 	pc.mu.Lock()
 	defer pc.mu.Unlock()
 	for _, t := range pc.rtpTransceivers {
-		if !t.stopped && t.kind == track.Kind() && t.Sender() == nil {
+		currentDirection := t.getCurrentDirection()
+		// According to https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addtrack, if the
+		// transceiver can be reused only if it's currentDirection never be sendrecv or sendonly.
+		// But that will cause sdp inflate in sfu. So we only check currentDirection's current value,
```
boks1971, line 1760: Maybe remove reference to the work `sfu`? Just mention that without re-use of transceivers, SDP will inflate for every track add/removal.

boks1971 (approved):

peerconnection.go
cnderrauber: removed. just thinking only sfu will have much track remove/add

PR #2303, opened 2022-09-21
Fix ORTC accept datachannel sctpTransport is nil

jerry-tao:
#### Description

When using ORTC API, the new accepted channel don't set the sctpTransport, see https://github.com/pion/webrtc/blob/master/sctptransport.go#L219.

it will panic if someone try to use it:

```
pair, err := channel.Transport().Transport().ICETransport().GetSelectedCandidatePair()
```


Sean-Der:
Sorry I missed this @jerry-tao ! I will merge right now, thank you so much for the contribution :)

Sean-Der (approved):

PR #2316, opened 2022-09-29
Add IPFilter to SettingEngine

cnderrauber:
When machine's network interface have more than one ip address and user don't want expose one of these ips to remote peer, interface filter can't work in this case, so add a ip filter for that

### DIFF
--- a/settingengine.go
+++ b/settingengine.go
@@ -5,6 +5,7 @@ package webrtc
 
 import (
 	"io"
+	"net"
 	"time"
 
 	"github.com/pion/dtls/v2"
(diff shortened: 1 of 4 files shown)


settingengine.go
```diff
@@ -157,6 +159,14 @@ func (e *SettingEngine) SetInterfaceFilter(filter func(string) bool) {
 	e.candidates.InterfaceFilter = filter
 }
 
+// SetIPFilter sets the filtering functions when gathering ICE candidates
+// This can be used to exclude certain ip  from ICE. Which may be
```
davidzhao, line 163:
```suggestion
// This can be used to exclude certain ip from ICE. Which may be
```

davidzhao (approved):

WofWca:
Corresponding ICE MR: https://github.com/pion/ice/pull/478

PR #2319, opened 2022-09-30
Fix generated Mid conflict with user set Mid

cnderrauber:
Inside one cycle renegotiation, if user both has manually set Mid and auto generated Mid for transceivers, CreateOffer will failed for Mid conflict.


boks1971 (approved):

Sean-Der:
Nice! Thanks for adding the tests :)

I think you just need to close the PeerConnections and should pass!

Issue #2323, opened 2022-10-06
Backport SampleBuilder bugfixes from webrtc-rs (if needed)


tmatth:
Quoting @k0nserv (see https://github.com/webrtc-rs/webrtc/pull/309/)

> Improve Sample Builder handling of padding packets
> 
> When sending video streams that are fairly static, such as a screen
> share of an inactive screen, libWebRTC will insert padding RTP packets.
> These were not correctly handled by sample builder, because of this
> SampleBuilder would throw away packets carrying actual media.
> 
> There was also a bug where dropping packets because the sample head was
> not at a partition start would double count the dropped packets(once in
> build_sample and once in purge_buffers.

Apparently Pion's samplebuilder may have the same issues.

The steps I guess would be to:
- [ ] confirm with e.g. a testcase that these bugs are present in pion
- [ ] backport the changes as needed

### Your environment.
  - Version: 5c13a6ecd6ff37c7958a059d1850baa96356af17
  - Browser: N/A

k0nserv:
https://github.com/webrtc-rs/webrtc/pull/303 might also be of interest.

The way to reproduce the issues I've observed is to send a screen share video track of relatively static content(I a tab with chrome://webrtc-internals) and then run this through `SampleBuilder`. Then attempt to play back this samples, either by saving them to disk or by sending them out again as RTP.

The test cases in https://github.com/webrtc-rs/webrtc/pull/309 could be ported directly and should showcase the issue if it exists.

jech:
Interesting.  Could you please explain why you only drop empty packets when the timestamp is constant, rather than dropping all empty packets?

k0nserv:
There's no specific code that drops empty packets, what happens is: SampleBuilder tries to build samples by looking for continuous(in terms of sequence number) stretches of RTP packets where the first packet is identified as the start of a partition(this is codec specific) and the last packet is identified as the end of a partition(either via the RTP marker bit being set or by detecting a differing timestamp). 

What we've done in the webrtc-rs is add code that can identify these empty padding packets to make it possible to distinguish between dropped padding packets(which are inconsequential) and dropped packets that carry media. Further, we discovered that in some instances the code to drop packets would get confused by these padding packets and, incorrectly, drop media packets. This happens with sequences like

```
p1(t=1, start=true)
p2(t=1)
p3(t=1)
p4(t=1, marker=true)
p5(t=1, padding=true)
p6(t=1, padding=true)
p7(t=1, padding=true)
[...]

PR #2328, opened 2022-10-10
Samplebuilder: add tests for padding packets

tmatth:
Refs #2323

Co-authored-by: Juliusz Chroboczek <jch@irif.fr>

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder_test.go
+++ b/pkg/media/samplebuilder/samplebuilder_test.go
@@ -38,6 +38,12 @@ func (f *fakeDepacketizer) IsPartitionHead(payload []byte) bool {
 		// the tests should be fixed to not assume the bug
 		return true
 	}
+
+	// skip padding
+	if len(payload) < 1 {
+		return false
+	}
+
 	for _, b := range f.headBytes {
 		if payload[0] == b {
 			return true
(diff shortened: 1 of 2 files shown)


pkg/media/samplebuilder/samplebuilder_test.go
```diff
@@ -223,6 +223,27 @@ func TestSampleBuilder(t *testing.T) {
...
+		{
+			// shamelessly stolen from webrtc-rs
+			message: "Padding",
+			packets: []*rtp.Packet{
+				{Header: rtp.Header{SequenceNumber: 5000, Timestamp: 1}, Payload: []byte{1}},
+				{Header: rtp.Header{SequenceNumber: 5001, Timestamp: 1}, Payload: []byte{2}},
+				{Header: rtp.Header{SequenceNumber: 5002, Timestamp: 1, Marker: true}, Payload: []byte{3}},
+				{Header: rtp.Header{SequenceNumber: 5003, Timestamp: 1}, Payload: []byte{}},
+				{Header: rtp.Header{SequenceNumber: 5004, Timestamp: 1}, Payload: []byte{}},
+				{Header: rtp.Header{SequenceNumber: 5005, Timestamp: 3}, Payload: []byte{1}},
```
k0nserv, line 235: It's worth having a test that also has `Marker: true` here, i.e. a packet that follows a run of padding packets and is both a start and end of a partition i.e. a very small video partition that fits inside 1200 bytes. This is something I observed with Chrome and it's covered in the webrtc-rs code.

tmatth: @k0nserv would that be this one ?
https://github.com/webrtc-rs/webrtc/pull/309/files#diff-0e05f283d453fbc18b4b99c7bf67c1ccd8e630a44e57047ee7a2a5fcc70dfbccR1055

k0nserv: That's testing that it can built a sample from such a packet in isolation. This one test such a packet following a padding run https://github.com/webrtc-rs/webrtc/pull/309/files#diff-0e05f283d453fbc18b4b99c7bf67c1ccd8e630a44e57047ee7a2a5fcc70dfbccR1099 where "Valid packet 3"  is both start and end of a partition

tmatth: I've added a test for this with some caveats:
- for some reason setting `withHeadChecker: true` crashes on trying to index 0 of an empty array, I'm guessing it's due to missing padding handling?
- likewise there is no `PrevPaddingPackets` field yet
- and the expected results are not the same as in webrtc-rs, presumably again because padding isn't yet handled here

Sean-Der:
@tmatth Mind looking at this one more time before I merge?

I rebased and removed the new public API members. No one has requested them, so don't want to expand the public API unless needed!

Sorry this took so long :( just working on getting PRs down to zero now.

tmatth:
> @tmatth Mind looking at this one more time before I merge?
> 
> I rebased and removed the new public API members. No one has requested them, so don't want to expand the public API unless needed!
> 
> Sorry this took so long :( just working on getting PRs down to zero now.

Thanks! LGTM, I still never really got to the bottom of the main issue I wanted to address whereby A/V sync would diverge if there was a long interruption in video packets (not lost, just nothing being sent by the publisher for a few seconds). Any thoughts?

Sean-Der:
@tmatth i think you will need to delay using the sender reports? I can help implement that

Sean-Der (approved):

tmatth:
> @tmatth i think you will need to delay using the sender reports? I can help implement that

Can this be done internally in `SampleBuilder` or is it up to users of `SampleBuilder` to handle that?

Issue #2226, opened 2022-05-18
Inactive tracks should not be stopped


k0nserv:
### Your environment.
  - Version: 3.1.0
  - Browser: Any
  - Other Information - N/A
  - 
### What did you do?

1. Create a track on the browser side
2. Negotiate the track with Pion
3. Remove the track on the browser side with `removeTrack`
4. Negotiate with Pion

### What did you expect?

For Pion to change the direction of the transceiver in question to `inactive` and not stop it.

### What happened?

Pion changes the direction of the transceiver in question to `inactive` **and** stops it:

https://github.com/pion/webrtc/blob/ea7964c12c1a1e51fe9d86c8cc9cae8cbca22e58/peerconnection.go#L1058-L1061

This relates to the issue https://github.com/webrtc-rs/webrtc/issues/192 which in turns is related to https://github.com/webrtc-rs/webrtc/issues/191(I'm not sure if this happens in Pion too)

adriancable:
@k0nserv - if it's correct that changing the direction to inactive on the Pion side shouldn't stop the track (and this sounds quite plausible, but I haven't checked the RFC) then I think removing L1058-1061 in `webrtc/peerconnection.go` is necessary but not sufficient.

We would also need to introduce the concept, I suppose, of a 'paused' track, whereby received RTP (but not RTCP?) data is discarded, and sending RTP (but not RTCP?) fails. We'd also need some way of aborting an in-progress ReadRTP and WriteRTP in that track if it enters the pause state while in the middle of those calls. I don't quite have the right mental model yet of the cleanest way to do that.

Thoughts from some experts on this? @davidzhao maybe?

davidzhao:
Far from being an expert on this, but I think the current behavior is actually correct.

[RTPTransceiver.Stop()](https://github.com/pion/webrtc/blob/ea7964c12c1a1e51fe9d86c8cc9cae8cbca22e58/rtptransceiver.go#L151) in Pion doesn't work exactly the same as the browser. The current implementation of Stop is more like a paused state. According to [the specs](https://datatracker.ietf.org/doc/html/rfc8829#section-4.2.1), once a transceiver is paused, it cannot be restarted. In Pion, it'd happily re-use the same transceiver in order to send other tracks.

k0nserv:
[...]

PR #2336, opened 2022-10-18
Populate RTPTransceiver stopped flag

boks1971:
There is use of `t.stopped` in `peerconnection.go`, but that flag is not set/used in `RTPTransceiver` at all. This PR sets that when a transceiver is stopped and provides an API to check if a transceiver is stopped.

@Sean-Der @davidzhao @cnderrauber The fact that code does not seem to get tripped makes me think I am missing something, but I noticed the ☝️ (i. e. `stopped` flag in `RTPTransceiver` was not set/used) and made this PR to address this. Have tested this change with LiveKit and it seems fine.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -371,56 +371,57 @@ func (pc *PeerConnection) checkNegotiationNeeded() bool { //nolint:gocognit
 		// 	return true
 		// }
 		m := getByMid(t.Mid(), localDesc)
+
 		// Step 5.2
-		if !t.stopped && m == nil {
+		if m == nil {
 			return true
 		}
-		if !t.stopped && m != nil {
-			// Step 5.3.1
-			if t.Direction() == RTPTransceiverDirectionSendrecv || t.Direction() == RTPTransceiverDirectionSendonly {
-				descMsid, okMsid := m.Attribute(sdp.AttrKeyMsid)
-				sender := t.Sender()
-				if sender == nil {
-					return true
-				}
-				track := sender.Track()
-				if track == nil {
-					// Situation when sender's track is nil could happen when
-					// a) replaceTrack(nil) is called
-					// b) removeTrack() is called, changing the transceiver's direction to inactive
-					// As t.Direction() in this branch is either sendrecv or sendonly, we believe (a) option is the case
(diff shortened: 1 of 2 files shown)


cnderrauber:
This PR might change transceiver behavior, pion's transceiver don't have a real `stopped` state now, it call `stop` method when state go to inactive, so in pion, a transceiver after `stop` (actually it not set stopped flag, just transit to `inactive`) can be reused. #2226 also mentioned this. Actually,  https://datatracker.ietf.org/doc/html/rfc8829#section-4.2.1 said that a `stopped` can't be reused, so this PR would comply with the rfc, but changed transceiver reuse behavior, might cause problem if application relied on it. 
There are #1943 and #1969 do similar things and revert it, don't know why, @Sean-Der might have suggestion about this.

boks1971:
Thank you @cnderrauber . Did not realise this path had already been done. Maybe, we should just remove the `stopped` flag to avoid confusion.

edaniels:
hey @boks1971, do you still want to work together to get this in?

boks1971:
> #1969

@edaniels Would be good to clean up that unused flag, but the discussions surrounding it which @cnderrauber pointed out makes me reluctant to change things. There was also a similar commit by @Sean-Der and a revert. Have not checked exactly why it was reverted.

I am still thinking that getting rid of just that field is good as it is has only one value, i. e. `false`.  The long discussion in #2226 is more around the expected behaviour of stop vs inactive and not around that field value per se.

Getting rid of that field looks straightforward except this one place which is conditioned on `if t.stopped` which is never `true` now. Unclear whether that if block should be completely removed if `stopped` field is removed OR the condition needs to be different - https://github.com/pion/webrtc/blob/fc3521e85c997a6f04f59e49470b52eb7afc06d4/peerconnection.go#L409

edaniels:
I think it's safe and good to remove the always false bool then. Separately, I think we need to get the behavior of stopped, ended, and friends in line with the W3 spec. I may try to tackle that soon.

Sean-Der:
I removed the `stopped` flag! In the future if someone wants to come back and take that on again I am in support.

Sean-Der (approved):

Sean-Der (approved):

PR #2351, opened 2022-11-13
Export media engine default audio and video codec parameters

asticode:
Hey guys,

Would you be open to exporting media engine default audio and video codec parameters?

I want to use default codec parameters provided by Pion but need to filter out some codecs from the list.

Cheers

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -70,11 +70,9 @@ type MediaEngine struct {
 	mu sync.RWMutex
 }
 
-// RegisterDefaultCodecs registers the default codecs supported by Pion WebRTC.
-// RegisterDefaultCodecs is not safe for concurrent use.
-func (m *MediaEngine) RegisterDefaultCodecs() error {
-	// Default Pion Audio Codecs
-	for _, codec := range []RTPCodecParameters{
+// DefaultAudioCodecParameters returns default audio codec parameters
+func (m *MediaEngine) DefaultAudioCodecParameters() []RTPCodecParameters {
+	return []RTPCodecParameters{
 		{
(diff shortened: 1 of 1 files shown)


kevmo314:
Would it make more sense to export individual constants for each codec? That seems like it would be easier to work with as far as adopting Pion's default codecs. Sort of similar to what pion/mediadevices does https://github.com/pion/mediadevices/blob/master/pkg/codec/codec.go#L24-L89.

asticode:
It would mean exporting codecs based on their payload type, something along the lines of:

```go
type DefaultRTPCodec struct {
  Parameters RTPCodecParameters
  Type RTPCodecType
}

var DefaultRTPCodecWithPayloadType95 = DefaultRTPCodec{
  Parameters: ...
  Type: ...
}
```

If you're ok with this I can make the change

kevmo314:
How about something simpler by extracting the current parameters into an exported var?

```
var DefaultOpusRTPCodecParameters = RTPCodecParameters{...}
var DefaultH264ProfileID42001fRTPCodecParameters = RTPCodecParameters{...}
...

func (m *MediaEngine) RegisterDefaultCodecs() error {
	for _, codec := range []RTPCodecParameters{DefaultOpusRTPCodecParameters, ...} {
		if err := m.RegisterCodec(codec, RTPCodecTypeAudio); err != nil {
			return err
		}
    }
	for _, codec := range []RTPCodecParameters{DefaultH264ProfileID42001fRTPCodecParameters, ...} {
		if err := m.RegisterCodec(codec, RTPCodecTypeVideo); err != nil {
			return err
		}
    }
}
```

asticode:
This wouldn't really help my use case and I don't mind copying actual default codecs into my application. I'll close this PR 👍 

Cheers

PR #2352, opened 2022-11-15
Close unhandled stream when peerconnection closed

cnderrauber:
When PeerConnection is closed, need to call the unhandled stream's close method to release the buffer created by BufferFactory, otherwise might cause resource leak if the user provides BufferFactory need close method to release the resource.

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1608,68 +1608,71 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream io.Reader, ssrc SSRC) err
 
 // undeclaredMediaProcessor handles RTP/RTCP packets that don't match any a:ssrc lines
 func (pc *PeerConnection) undeclaredMediaProcessor() {
-	go func() {
-		var simulcastRoutineCount uint64
-		for {
-			srtpSession, err := pc.dtlsTransport.getSRTPSession()
-			if err != nil {
-				pc.log.Warnf("undeclaredMediaProcessor failed to open SrtpSession: %v", err)
-				return
-			}
+	go pc.undeclaredRTPMediaProcessor()
+	go pc.undeclaredRTCPMediaProcessor()
+}
 
-			stream, ssrc, err := srtpSession.AcceptStream()
-			if err != nil {
-				pc.log.Warnf("Failed to accept RTP %v", err)
-				return
-			}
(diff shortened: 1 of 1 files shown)


peerconnection.go
```diff
@@ -1632,34 +1633,41 @@ func (pc *PeerConnection) undeclaredMediaProcessor() {
...
+				pc.dtlsTransport.storeSimulcastStream(stream)
 				continue
 			}
 
 			go func(rtpStream io.Reader, ssrc SSRC) {
-				pc.dtlsTransport.storeSimulcastStream(stream)
-
 				if err := pc.handleIncomingSSRC(rtpStream, ssrc); err != nil {
 					pc.log.Errorf(incomingUnhandledRTPSsrc, ssrc, err)
+					pc.dtlsTransport.storeSimulcastStream(stream)
```
cnderrauber, line 1643: Only store stream if it is not handled, this change can shorten the stream's lifecycle and can be recycled earlier if the stream is closed by renegotiation, since stream has a reference to the buffer, so this change can decrease memory usage in a long run session.

peerconnection.go
```diff
@@ -1607,61 +1608,71 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream io.Reader, ssrc SSRC) err
...
-	go func() {
-		for {
-			srtcpSession, err := pc.dtlsTransport.getSRTCPSession()
-			if err != nil {
-				pc.log.Warnf("undeclaredMediaProcessor failed to open SrtcpSession: %v", err)
-				return
+		go func(rtpStream io.Reader, ssrc SSRC) {
+			if err := pc.handleIncomingSSRC(rtpStream, ssrc); err != nil {
+				pc.log.Errorf(incomingUnhandledRTPSsrc, ssrc, err)
+				pc.dtlsTransport.storeSimulcastStream(stream)
```
cnderrauber, line 1647: Only store the stream if it is not handled, this change can shorten the stream's lifecycle and can be recycled earlier if the stream is closed by renegotiation, since the stream has a reference to the buffer, this change can decrease memory usage in a long run session.

boks1971 (approved):
Good find @cnderrauber !!!

PR #2376, opened 2022-12-22
Fix negotiation needed panic on nil Sendonly track

edaniels:
In step 5.3.1 (https://www.w3.org/TR/webrtc/#dfn-update-the-negotiation-needed-flag), since we do not have a sender we can only choose to signal that we need negotiation. This typically happens in a case where the offerer asks for a recv only, but the answering side does not provide it. This only triggers if an on negotiation needed handler is set.

I believe the fix I applied is mostly correct by interpreting 5.3.1 and how in the Google WebRTC code, a non-nil sender is created during SetRemoteDescription. My only concern with the fix is that this would basically rely on the handler of negotiation needed to determine they need to go add that track. If they don't you can imagine an endless loop between offerer and answer.

@Sean-Der, while working on this, I realized that `SetRemoteDescription` may be non-compliant due to https://www.rfc-editor.org/rfc/rfc8829#section-5.10-6.6.2.2.2.2. However, if I make us compliant by removing https://github.com/pion/webrtc/blob/1c20cd4d7b48d2ab8f166486730eb2475f3f8333/peerconnection.go#L1084-L1088, the newly added test fails. Not sure how to reconcile that one as my knowledge & reasoning stops there :).


edaniels:
Also not sure why this fails with WASM but maybe that points to something.

yurifedoseev:
@edaniels thank you for the PR! I was looking for solving the panic problem too. Could you update your branch to make it actual with the base branch?

yurifedoseev:
Hi @Sean-Der, could you help to review this MR - to solve panic in pion?

Sean-Der (approved):

PR #2395, opened 2023-01-31
Avoid including SSRC in single media offers

streamer45:
#### Description

Sending simulcast tracks from a pion sender to a pion receiver is failing unless there are other medias added:

```
Incoming unhandled RTP ssrc(xxxx), OnTrack will not be fired. single media section has an explicit SSRC
```

I admit I am not well versed in SDP semantics so please let me know if there's more to this that we should consider. The proposed changes are incredibly dumb as we are simply checking if there are multiple media sections and if not we avoid adding the ssrc attribute.

#### Reference issue

This is strictly related to https://github.com/pion/webrtc/pull/2299 which I guess is also a working solution although I am thinking that if we are being strict on the receiving side we should probably do the same on the sending side.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1486,6 +1486,8 @@ func (pc *PeerConnection) handleUndeclaredSSRC(ssrc SSRC, remoteDescription *Ses
 	onlyMediaSection := remoteDescription.parsed.MediaDescriptions[0]
 	streamID := ""
 	id := ""
+	hasRidAttribute := false
+	hasSSRCAttribute := false
 
 	for _, a := range onlyMediaSection.Attributes {
 		switch a.Key {
@@ -1495,10 +1497,16 @@ func (pc *PeerConnection) handleUndeclaredSSRC(ssrc SSRC, remoteDescription *Ses
 				id = split[1]
 			}
 		case sdp.AttrKeySSRC:
-			return false, errPeerConnSingleMediaSectionHasExplicitSSRC
+			hasSSRCAttribute = true
 		case sdpAttributeRid:
-			return false, nil
+			hasRidAttribute = true
 		}
+	}
+
+	if hasRidAttribute {
+		return false, nil
+	} else if hasSSRCAttribute {
+		return false, errPeerConnSingleMediaSectionHasExplicitSSRC
 	}
 
 	incoming := trackDetails{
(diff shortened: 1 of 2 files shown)


Sean-Der:
Nice catch! Thanks for the PR @streamer45 I believe something else is broken here. For Simulcast we can't put a SSRC in the MediaSection since we will have multiple.

Are you able to write a test of `Pion <-> Pion` Simulcast? We might even have one already let me look for it.

streamer45:
@Sean-Der I believe we have:

https://github.com/pion/webrtc/blob/9dd2c42999700f89906096411e626342d818897a/peerconnection_media_test.go#L1213

Maybe I could build on that to cover the changes in here. Let me know.

Sean-Der:
@streamer45 that would be fantastic! I don’t fully understand what is broken, but if you have a failing test would help a lot.

streamer45:
@Sean-Der Added a pion <-> pion test as well which would have failed before these changes, specifically in:

https://github.com/pion/webrtc/blob/132ebfa817e12b1f5f2fe3b1627faa169f70e4f9/peerconnection.go#L1485

Let me know if that helps.

streamer45:
@Sean-Der Let me know if there's anything else needed to move this forward. Thanks :)

Sean-Der:
Shit, sorry I missed this @streamer45!

I have some time off work and will be able to work on this now

Sean-Der:
@streamer45 Nope nothing is needed. I just got caught up in day job. With this repro I have everything I need. 

I don't know exactly what is wrong/the fix. I should have a full understanding tonight and understand how to move forward.

Sean-Der:
Thank you so much for this PR @streamer45! That was a gnarly issue, lots of niche parts interacting :) This is the commit I am landing.

````
Author: streamer45 <cstcld91@gmail.com>
Date:   Tue Jan 31 17:25:12 2023 -0600

    Handle Simulcast Offer with one Media Section

    Everytime we receieve a new SSRC we probe it and try to determine the
    proper way to handle it. In most cases a Track explicitly declares a
    SSRC and a OnTrack is fired. In two cases we don't know the SSRC
    ahead of time
    * Undeclared SSRC in a single media section (pion/webrtc#880)
    * Simulcast

    The Undeclared SSRC processing code would run before Simulcast.
    If a Simulcast Offer/Answer only contained one Media Sec
```

I am sorry I have been so slow on Pion lately. I am going to be working on it more. I would love to have you involved more also. If you have any feedback on the project (and how I can make it better I would love to hear)

I am going to drive down our PRs/Issues and cut a new release.

Sean-Der (approved):

streamer45:
Thanks Sean, appreciate the explanation there and the extra work on this 🙌

PR #2408, opened 2023-02-07
Use new pion/transport Net interface

stv0g:
This change adapts pion/ice to use a new interface for most network related operations. The interface was formerly a simple struct vnet.Net which was originally intended to facilicate testing. By replacing it with an interface we have greater flexibility and allow users to hook into the networking stack by providing their own implementation of the interface.


stv0g:
Currently pending on: https://github.com/pion/ice/pull/471

enobufs:
@stv0g I did a quick review and it looks good to me. Hopefully, we see all builds succeed once we have the updated pion/ice integrated. 🤞

edaniels (approved):
LGTM, nothing stands out

at-wat:
@stv0g Mind releasing it if it's ready?

PR #2426, opened 2023-02-25
Transceivers state  change from inactive to active

zjzhang-cn:
#### Description
Reuse  transceivers in remote peer。

In the existing  RTC Transceivers, when the state of the remote transceiver changes from inactive to active, the state of the pion also needs to change from inactive to active。

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1897,6 +1897,44 @@ func (pc *PeerConnection) AddTransceiverFromKind(kind RTPCodecType, init ...RTPT
 	return t, nil
 }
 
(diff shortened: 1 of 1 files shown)


cnderrauber:
Thank you for refining this, can you add a test case for this?

zjzhang-cn:
Client cooperation is required.

Sean-Der:
Hey @zjzhang-cn thank you so much for the contribution!

Since this isn't a `webrtc-pc` API I don't believe merging it is the right thing to do.  If a user wants to re-use a Transceiver they should find an inactive Transceiver and call ReplaceTrack I believe?

If I am wrong please re-open! I would love to see Pion be as good as possible.

thank you

PR #2434, opened 2023-03-01
fix: return correct bytes written on WriteRTCP on dtlsTransport

itzmanish:
#### Description

Currently WriteRTCP function in `dtlstransport.go` doesn't return bytes written info. This PR fixes this and handles error and success response correctly. 
#### Reference issue
Fixes #...


Sean-Der:
Thank you so much @itzmanish !

Sean-Der (approved):

PR #2435, opened 2023-03-03
Fix peerconnetcion connecting state

cnderrauber:
The peerconnection state should transit to connecting if icestate is checking or dtls connecting

#### Reference issue
Fixes #...


Sean-Der (approved):

Sean-Der:
A unit test for `updateConnectionState` might be nice! Seems like it would be easy to do (might even have one already) 

but approval from me! If you don't think a test is needed that is totally fine

PR #2444, opened 2023-03-31
Match header extensions to all media sections in offer

ellenfkh:
#### Description

We noticed that once pion has negotiated headers and codecs for a given Kind, when a subsequent offer SDP adds a media section for that same Kind with different header extensions, the generated answer uses the original headers for all media sections. 

This doesn't happen in Chrome, which never seems to send mixed headers like this. Firefox does, though.

We ran into this when sending an offer in which the first video media section had simulcast disabled, and thus didn't set this attribute
```
a=extmap:8/sendonly urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
```
but the second video media section had simulcast enabled, so included that header.

The resulting answer didn't include that header in either media section, but the second media section did correctly have these attributes 
```
a=rid:f recv
a=rid:h recv
a=rid:q recv
a=simulcast:recv f;h;q
```
Resulting in this error 
```
Incoming unhandled RTP ssrc(3870936274), OnTrack will not be fired. stream id RTP Extensions required for Simulcast
```

This patch fixes our manual test case, but I suspect there are more places in the code where this assumption is being made. I'm also not sure whether Plan B vs Unified expects different behavior in this case.

#### Reference issue
Fixes #...

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -455,6 +455,30 @@ func (m *MediaEngine) matchRemoteCodec(remoteCodec RTPCodecParameters, typ RTPCo
 	return matchType, nil
 }
 
+// Update header extensions from a remote media section
+func (m *MediaEngine) updateHeaderExtensionFromMediaSection(media *sdp.MediaDescription) error {
+	var typ RTPCodecType
+	switch {
+	case strings.EqualFold(media.MediaName.Media, "audio"):
+		typ = RTPCodecTypeAudio
+	case strings.EqualFold(media.MediaName.Media, "video"):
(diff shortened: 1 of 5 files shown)


adriancable (approved):
This looks good to me. If one other person can quickly eyeball this (@Sean-Der / @davidzhao ?) then I am happy to merge it.

davidzhao (approved):
Thank you for the PR. It looks good!

Sorry for the delays getting to this one.

cnderrauber:
cc @boks1971  as you have a related pr #2107 for same issue

PR #2455, opened 2023-04-14
Fix simulcast track set to incorrect stream id

cnderrauber:
When there are more than one simulcast media sections in SDP, we need to check mid when parsing detail for the simulcast track.


boks1971 (approved):
Nice find @cnderrauber

cnderrauber:
@Sean-Der The go 1.18 seems to be removed from test workflow, but is still required by the merge check, make the PR can't be merged in. Can you verify this? Thanks!

PR #2456, opened 2023-04-18
Update CI configs to v0.10.7

pionbot:
Update lint scripts and CI configs.


Sean-Der:
@stv0g What should we do about the new licensing lint? Do you know of any tooling that could auto append the license to every file?

I can write a bash script that can do that, and we can split the repos!

stv0g:
Hi @Sean-Der,

> What should we do about the new licensing lint?

I've started adding the license headers to most of the ICE/STUN/TURN related repos with the following command:

```bash
find . -name "*.go" -exec sed -i '1i // SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>\n// SPDX-License-Identifier: MIT\n' {} \;
```

But in the end, I think we manually need to check each file anyway.

There is also a local CLI command to test the REUSE compliance which I use to download licenses and check the headers: https://github.com/fsfe/reuse-tool

Sean-Der (approved):

PR #2469, opened 2023-05-05
Controlling the generation of SSRC from outside

FizikRoot:
#### Description

Control ssrc from outside of webrtc module

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -1256,7 +1256,7 @@ func TestPeerConnection_Simulcast(t *testing.T) {
 		pcOffer, pcAnswer, err := NewAPI(WithMediaEngine(m)).newPair(Configuration{})
 		assert.NoError(t, err)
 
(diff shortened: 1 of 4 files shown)


Sean-Der:
Hi @FizikRoot 

thanks for the contribution! What problem are you trying to solve?

We won’t be able to store SSRC on the TrackLocal unfortunately. They will need to be per RTPSender. Not against providing a hook for this, I just don’t want to break other stuff.

Maybe using ORTC might be better here?

Sean-Der:
Both super cool use cases! Yea we need this.

ok let me look at API tonight. This needs to be on RTPSender. If that doesn’t work we will add SettingEngine.SSRCGenerator

hailt-genki:
Hi @Sean-Der, could you please provide information on this issue? My use case is as follows, and I would like to modify the SSRC number

Sean-Der:
This is now available! Sorry for the delay @FizikRoot @lamhai1401 

See [here](https://github.com/pion/webrtc/blob/master/rtptransceiverinit_go_test.go#L54) for a test that sets the SSRC to 5000 explicitly.

Issue #2516, opened 2023-07-24
webrtc.NewAPI is inherently confusing and frustrating


ForestJohnson:
I get it that Pion WebRTC is a highly technical thing and we are supposed to be "smart" and "Able to Read"  in order to use it...  But please hear me out. 

## Summary

I think `webrtc.NewAPI` docstring needs needs an **ALL CAPS BOLD WARNING TEXT**  and a link to an example of how to set it up to so that the resulting `webrtc.NewPeerConnection()`s will behave exactly the same as the ones you get from the default `webrtc.NewPeerConnection()`.

## Motivation

https://gophers.slack.com/archives/CAK2124AG/p1690162030755169

When you call `webrtc.NewAPI()` and then create a new `PeerConnection` with it, it behaves radically differently from the default  `PeerConnection` that you get from `webrtc.NewPeerConnection()`, and the developer has no way of being informed why this is the case or how to "fix" it. Specifically, they have no way of easily getting their `webrtc.NewAPI()` to reach "feature parity" with the default from the `webrtc` package, they have no easy way of being assured that their `PeerConnection` should work roughly the same as the "normal" one.

## Describe alternatives you've considered

I considered asking for `webrtc.NewAPI()` to come fully pre-configured by default.  But I thought there's gotta be a reason why it wasn't done that way, it probably would be even more confusing in a lot of cases.

## Additional context

Just look at the slack thread to see my despair live. 

EDIT:   OMFG :facepalm:   I can't believe I missed this;  the exact example I was looking for is... right where you would expect it, in the body of the `webrtc.NewPeerConnection()` function.

edaniels:
Closing with merging of https://github.com/pion/webrtc/pull/2517

PR #2517, opened 2023-07-24
Document the "empty defaults" behaviour of NewAPI fn

ForestJohnson:
#### Description

I think `webrtc.NewAPI` docstring needs needs an **ALL CAPS BOLD WARNING TEXT**  and a link to an example of how to set it up to so that the resulting `webrtc.NewPeerConnection()`s will behave exactly the same as the ones you get from the default `webrtc.NewPeerConnection()`.

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/2516

### DIFF
--- a/api.go
+++ b/api.go
@@ -25,6 +25,10 @@ type API struct {
 }
 
 // NewAPI Creates a new API object for keeping semi-global settings to WebRTC objects
+//
+// WARNING: the resulting API won't work the same as the default webrtc.NewPeerConnection()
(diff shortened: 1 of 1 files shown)


Sean-Der:
Sorry this took so long @ForestJohnson I really appreciate your feedback. I am ignorant to how it feels as a new user to use Pion. Your feedback/notes are really important. Anything else that frustrated you?

Merged with 6b22634c1d1ca4ff24e36c80f662a46217d2128c

Issue #618, opened 2019-04-12
Add PCAP logger


maxhawkins:
## Summary

Implement a pcap logger that can be enabled in debug mode to generate a packet dump of a PeerConnection.

## Motivation

It's difficult to debug issues in the DTLS or RTP stacks because an external packet logger like tcpdump does not have access to the DTLS encryption keys.

To make it easier to track down issues, it should be possible to save a packet dump for analysis.

## Design

To aid debugging, the pcap should include all the types of packets Pion sends and receives over its ICE connections:

* STUN requests
* DTLS handshakes
* Decrypted DTLS packets
* RTP packets
* RTCP packets

These should be mapped to IP addresses and port numbers where possible so that issues with STUN and firewall hole punching can be diagnosed.

It should be possible to:
* Enable or disable the writer on a per-PeerConnection level
* Attach a pcap writer before or after a PeerConnection has been created
* Write packets from multiple PeerConnections to a single pcap

Here's a sketch of a design:

```go
// Logger can be attached via SettingEngine
f, _ := os.Open("out.pcap")
s := &SettingEngine{
  PacketLogger: NewPCAPLogger(f),
}
api := NewAPI(WithSettingEngine(s))
pc, _ := api.NewPeerConnection(Configuration{})

(...)

// Later, the logger can be changed or removed
s.PacketLogger = NewPacketLogger(otherfile)
s.PacketLogger = nil
```

```go
type PacketLogger interface {
  WriteUDP(b []byte, dst, src *net.UDPAddr)
  WriteTCP(b []byte, dst, src *net.TCPAddr)
}
```

### Notes

Related to pion/website#29
Partially implemented PCAP format: #521

Sean-Der:
@maxhawkins This is an interesting one! Going back through old tickets and I want to build this! This would be amazing for people to learn and debug.

I don't think we should add toggles to enable/disable it during logging. If users want that they can implement that behavior on their own. I also don't think we can support adding them during runtime. The complexity of plumbing that in I think could be costly/error prone.

What do you think of making it configurable? Via the SettingEngine you can do the following
* STUN Traffic
* DTLS handshakes
* DTLS Packets (Encrypted or Decrypted might be nice?)
* RT(C)P Packets (Encrypted or Decrypted might be nice?)

This could then be broken up into smaller tasks pretty easily.

* [ ] Add a `PacketInterceptor` field that is a PacketConn to `ice.AgentConfig`. If set it would wrap interfaces and write.
[...]

PR #2522, opened 2023-07-31
Add DTLS KeyLog configuration option in WebRTC API

npmenard:
#### Description
Add an option in the setting engine to log TLS key material when a DTLS connection is established with a peer. The option exists in pion/dtls but is not easily accessible. A user would use it very similarly to the crypto/tls package : 
```go
...
if keyLog := os.Getenv("SSLKEYLOGFILE"); len(keyLog) != 0 {
		w, _ := os.OpenFile(keyLog, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
		settingEngine.SetDTLSKeyLogWriter(w)
}
...
```
This is somewhat related to #618 as it lets a user use Wireshark or a similar tools to capture and decrypt a particular session.

### DIFF
--- a/settingengine.go
+++ b/settingengine.go
@@ -71,6 +71,7 @@ type SettingEngine struct {
 		clientAuth                *dtls.ClientAuthType
 		clientCAs                 *x509.CertPool
 		rootCAs                   *x509.CertPool
+		keyLogWriter              io.Writer
 	}
 	sctp struct {
 		maxReceiveBufferSize uint32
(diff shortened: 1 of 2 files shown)


settingengine.go
```diff
@@ -422,6 +423,12 @@ func (e *SettingEngine) SetDTLSRootCAs(rootCAs *x509.CertPool) {
 	e.dtls.rootCAs = rootCAs
 }
 
+// SetDTLSKeyLogWriter sets the destination of the TLS key material for debugging.
+// Logging key material compromise security and should only be use for debugging.
```
edaniels, line 427:
```suggestion
// Logging key material compromises security and should only be use for debugging.
```

edaniels (approved):
Nice! Just one change to make.

Sean-Der:
This is really cool @npmenard 

If you are interested I would like to make a post on using this! Debugging WebRTC is hard, this is a lot easier then the alternative.

PR #2527, opened 2023-08-02
Add RTPTransceiver func

do-hyung-kim:
Hi there! I want to use RTPTransceiver in on track func. RTP TRANSCEIVER is provided in the ON TRACK WEB API, but when I looked at it, it was difficult to access the RTP TRANSCEIVER in PION's ON_TRACK. I'm raising the PR because I would like to add a function to get the TR from the receiver rather than changing the supplied arguments of the ON_TRACK FUNC.

#### Description:
-   Add Transceiver func in rtp receiver
-   on_track api [https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/track_event](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/track_event)

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -126,6 +126,15 @@ func (r *RTPReceiver) Tracks() []*TrackRemote {
 	return tracks
 }
 
+// RTPTransceiver returns the RTPTransceiver this
+// RTPReceiver belongs too, or nil if none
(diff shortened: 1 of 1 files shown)


Sean-Der (approved):

Sean-Der:
@do-hyung-kim tagged `v3.2.14`

Issue #2472, opened 2023-05-15
High audio bitrate results in invalid opus file


onthegit:
Hello,
when using audio bitrate higher than 96000 and the following audio preset:
  echoCancellation: true,
  noiseSuppression: true,
  sampleRate: 48000,
  channelCount: 2,
  autoGainControl: false,

https://github.com/livekit/server-sdk-go/tree/main/examples/filesaver

Results in invalid opus file (CRC is invalid)

Pion @v3.2.1

Chrome version:
Version 110.0.5481.177

I posted it here since the oggwriter is in this repo.

onthegit:
It appears that if the bitrate is above 48000  any ogg file generated is with invalid CRC.

Up until 96000 it is playable. Above 96000 the ogg file is not playable.

PR #2531, opened 2023-08-11
OggWriter: add packet segmentation

streamer45:
#### Description

PR adds packet segmentation capability to `OggWriter`. This makes it possible to accept payloads bigger than 255 bytes which are actually more common that I thought (e.g. Firefox seems to be generating quite a few). Without this change we'd be defaulting to a single segment causing data loss when writing and corrupted files as consequence.

There's still a theoretical limit of 255 segments per page that we are not handling (packets spanning pages) but for encoded audio data I am hoping that shouldn't ever become an issue.

#### Reference issue

Likely related to https://github.com/pion/webrtc/issues/2472

### DIFF
--- a/pkg/media/oggwriter/oggwriter.go
+++ b/pkg/media/oggwriter/oggwriter.go
@@ -149,7 +149,9 @@ const (
 
 func (i *OggWriter) createPage(payload []uint8, headerType uint8, granulePos uint64, pageIndex uint32) []byte {
 	i.lastPayloadSize = len(payload)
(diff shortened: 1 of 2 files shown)


Sean-Der:
Really nice PR @streamer45! Nice debugging on this, that must have been fun debugging :)

Sean-Der (approved):

Sean-Der:
Tagged `v3.2.17` (seems like a useful thing to have right away!)

PR #2535, opened 2023-08-18
Revert breaking change in ICECandidateStats

### LINKED ISSUE #2533: webrtc v3.2.15 has a breaking change
### Your environment.
  - Version: v3.2.15
  - Browser: N/A
  - Other Information: See below for my suggested fix.

### What did you do?

I updated to the latest version of webrtc (v3.2.15) in our program, which uses the struct field webrtc.ICECandidateStats.NetworkType, and then tried to build it. There were build errors like this:

```
r.local.NetworkType.String undefined (type string has no field or method String)
```

### What did you expect?

No breaking changes in a patch release.

### What happened?

Version 3.2.15 of webrtc has a breaking change where the type of the struct field webrtc.ICECandidateStats.NetworkType has been changed from NetworkType to string.

The breaking change was made in this commit: https://github.com/pion/webrtc/commit/2926b99173b44266550855079501ae1cd7d89a92

The line that was changed in a breaking way can be viewed here: https://github.com/pion/webrtc/commit/2926b99173b44266550855079501ae1cd7d89a92#diff-9f6cddeabc26d57e3837fe66df90a6288f56e8a0e898cb6004d098da521583aeL1393

My understanding of how Go modules is supposed to work is that if the major version is 1 or higher and you want to do breaking changes, you need to increase the major version. It looks to me that webrtc v3.2.15 isn't following that contract.

### Discussion ###

If you want to change the type of the NetworkType field and follow the contract of Go modules, you need to do that in webrtc v4.0.0.

Since  RTCIceCandidateStats.networkType is removed from the latest version of the Identifiers for WebRTC's Statistics API specification, I guess it makes more sense to just remove the NetworkType field than change the type (unless you want to continue to support old specs). If you want to change the type or remove the struct field is up to you, but both choices are breaking changes.

### Suggested changes ###

My suggested fix is that you revert the breaking change in webrtc.ICECandidateStats and then retract version 3.2.15 in the go.mod file. https://go.dev/ref/mod#go-mod-file-retract

### References ###

Links to the RTCIceCandidateStats dictionary in the Identifiers for WebRTC's Statistics API specification.
- Latest version: https://www.w3.org/TR/webrtc-stats/#icecandidate-dict*
- First version where the networkType member was removed: https://www.w3.org/TR/2020/CR-webrtc-stats-20200114/#icecandidate-dict*
- Last version where the networkType member was present: https://www.w3.org/TR/2018/CR-webrtc-stats-20180703/#icecandidate-dict*

hansgylling:
#### Description
Commit 2926b99173b44266550855079501ae1cd7d89a92 had a breaking change where the type of the struct field webrtc.ICECandidateStats.NetworkType changed from NetworkType to string. This PR reverts that type change and changes it back to NetworkType.

The addition of the omitempty struct tag to the NetworkType field from commit 2926b99173b44266550855079501ae1cd7d89a92 is left as is. I don't think that is a breaking change.

#### Reference issue
Fixes #2533


Sean-Der:
This looks perfect to me. Congrats on your first Open Source PR :) 

Will merge after tests pass and tag new version

Sean-Der (approved):

Sean-Der:
@hansgylling Tagged `v3.2.16`

hansgylling:
@Sean-Der I built our library that uses pion/webrtc and also our application that uses that library using v3.2.16 and there were no problems.

PR #2549, opened 2023-09-03
fix comments

zhngs:
one agent full, one lite: the full agent must take the controlling role, and the lite agent must take the controlled role


Sean-Der:
Nice catch! Thank you @zhngs

PR #2555, opened 2023-09-04
Close PeerConnection on DTLS CloseNotify

### LINKED ISSUE #1767: PeerConnection should support shutdown on DTLS Close
### Your environment.
  - Version: v2.2.26
  - Browser: firefox 88.0b8

### What did you do?
I'm calling the `PeerConnection.Close` function.

### What did you expect?
The connection should be closed, and I should see the connection on the remote peer being disconnected.

### What happened?
The connection on the remote peer does not go to disconnected state instantly. It does after the `SettingsEngine.tiemout.ICEConnection` timeouts (which is 30 seconds by default).

I'm perhaps missing something here since I'm fairly new to the webrtc field.

Sean-Der:
Resolves #1767
Resolves pion/dtls#151

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -44,7 +44,8 @@ type DTLSTransport struct {
 	state                 DTLSTransportState
 	srtpProtectionProfile srtp.ProtectionProfile
 
-	onStateChangeHandler func(DTLSTransportState)
+	onStateChangeHandler   func(DTLSTransportState)
+	internalOnCloseHandler func()
 
 	conn *dtls.Conn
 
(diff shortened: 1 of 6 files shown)


Sean-Der:
Would love your take @boks1971 @cnderrauber 

Everyone should see much faster hang ups and should more closely match browser behavior. I hope I am not making any assumptions though.

cnderrauber (approved):
Looks good!

boks1971 (approved):
@Sean-Der This is great. LGTM!

One curiosity question: I could not quite understand the test you added, i. e. how does it trigger the new path? It is basically the new path is going to be triggered pretty much most of the time? If yes, is there is a need for a new test (guessing existing tests would cover it).

Sean-Der:
@boks1971 The test doesn’t explicitly close both PeerConnections. It then uses the ‘leaked threads checker’ so it can’t complete until both close. 

I bet the new path is happening in most tests! As soon as you close a PeerConnection it sends the Close. It’s a race between that and closing both connections in each test :)

boks1971:
> @boks1971 The test doesn’t explicitly close both PeerConnections. It then uses the ‘leaked threads checker’ so it can’t complete until both close.
> 
> I bet the new path is happening in most tests! As soon as you close a PeerConnection it sends the Close. It’s a race between that and closing both connections in each test :)

aah, got it, thank you Sean.

Issue #2586, opened 2023-09-23
Datachannel has a race in handleOpen


sukunrt:
### Your environment.
  - Version: v3.2.20
  - Other Information - triggered while working on libp2p-webrtc implementation: https://github.com/libp2p/go-libp2p/actions/runs/6283700827/job/17064424760?pr=2576

### What did you do?
The race is straightforward, the two fields onBufferedAmountLow and onBufferedAmountLowThreshold are protected by the mutex and should only be accessed with the lock held. 
https://github.com/pion/webrtc/blob/master/datachannel.go#L315

Sean-Der:
Fixed by @sukunrt in https://github.com/pion/webrtc/pull/2587

thank you!

PR #2587, opened 2023-09-23
fix race in datachannel handleOpen

sukunrt:
#### Description
The two variables used here are protected by the mutex `mu` they should only be accessed when the lock is held. 

#### Reference issue
https://github.com/pion/webrtc/issues/2586


Sean-Der:
Thank you @sukunrt !

Sean-Der:
@sukunrt I tagged `v3.2.21` for this

sukunrt:
Oh wow, this is awesome! Thanks for the release @Sean-Der

PR #2591, opened 2023-09-25
Deletion of existing sample_builder implementation codes

midepeter:
This commit deletes the existing implementation of sample builder and replaces with raw jech implementation. No further modification has been made in this commit.

#### Reference issue
Fixes #...


midepeter:
@Sean-Der  I made this draft push to show the state before i start making necessary additions and other iterations

jech:
@midepeter why is this being committed with you as the author?  Since I am the author of the code, I would expect to be credited.  Please use the `--author` flag to the git command to properly credit me as the author.

midepeter:
@jech oh, I will do that

sorry about that. This literally my first time in the open source space 

I will do that. Apologies

SerialVelocity:
Hey, I gave this a go and found two issues:
1. You've removed some functionality like `WithMaxTimeDelay` but the options silently apply. I'm assuming this is what you meant when you said "to show the state before i start making necessary additions and other iterations"
2. If the first few packets sent are padding packets, then no packets are processed until the padding packets end up being dropped when `maxLate` is hit.

Sean-Der:
Replaced with https://github.com/pion/webrtc/pull/2679

PR #2592, opened 2023-09-28
Transparently handle incoming distinct-SSRC RTX packets

adriancable:
Support incoming RTX packets on separate RTP stream if `video/rtx` track is used. They are handled transparently by `TrackRemote.Read()` with some additional attributes being returned (rtx_ssrc, rtx_payload_type, rtx_sequence_number) if the packet is an RTX packet, so the caller can distinguish RTX if needed.

Replaces PR #2519.


adriancable:
Per the discussion in #2519, this implements the 'transparent' approach to handling RTX packets on a separate RTP stream, as preferred by @jech and @kcaffrey. Please take a look, feedback welcome.

jech:
So this is rewriting the packet transparently before it reaches the application?  Perhaps it would be better to provide a utility function that does the rewriting, and let the application do the rewriting if desired?

adriancable:
@jech - the argument is that the reception of RTX packets should 'by default' be transparent to the caller. (Otherwise we could just implement a separate ReadRTX() function like the original PR.) That way everything 'just works' if there's an RTX track with no code changes.

The rewriting doesn't lose any information, since it's made available to the caller in the attributes if desired.

track_remote.go
```diff
@@ -125,12 +127,23 @@ func (t *TrackRemote) Read(b []byte) (n int, attributes interceptor.Attributes,
 		}
 	}
 
-	n, attributes, err = r.readRTP(b, t)
-	if err != nil {
-		return
+	// If there's a separate RTX track and an RTX packet is available, return that
+	if rtxPacket, rtxAttributes := r.readRTX(t); rtxPacket != nil {
+		n, err = rtxPacket.MarshalTo(b)
```
kcaffrey, line 132: It may not be worth the optimization without benchmarks, but I wonder if it is worth somehow saving the marshaled packet (maybe in attributes?), since `ReadRTP` will have to unmarshal immediately anyway. I checked some profiles from our deployment, and it looks like `Unmarshal()` is only ~2.8% of the CPU time used in `ReadRTP()` for us, and `MarshalTo()` is less than 1% of `WriteRTP()`.

rtpreceiver.go
```diff
@@ -411,13 +418,42 @@ func (r *RTPReceiver) receiveForRtx(ssrc SSRC, rsid string, streamInfo *intercep
...
+				// BWE probe packet, ignore
+				continue
+			}
+
+			// RTX packets have a different payload format. Move the OSN in the payload to the RTP header and rewrite the
+			// payload type and SSRC, so that we can return RTX packets to the caller 'transparently' i.e. in the same format
+			// as non-RTX RTP packets
+			attributes.Set("rtx_payload_type", pkt.Header.PayloadType)
+			attributes.Set("rtx_ssrc", pkt.Header.SSRC)
+			attributes.Set("rtx_sequence_number", pkt.Header.SequenceNumber)
```
kcaffrey, line 446: Perhaps define some exported constants for the fields in the attributes? They are effectively part of the API, and that would be more clear there were exports.

kcaffrey (approved):
It seems good to me, but I only have passing familiarity with this part of the code base, so it's probably better to wait for someone else to review if you wanted more confidence!

track_remote.go
adriancable: I thought about this too, but I don't think this is worth it. If we store the marshalled packet (so we can reuse it in `ReadRTP`) we are just trading off saving a tiny amount of CPU for a tiny amount of extra memory usage. The CPU saving would only be when processing RTX packets, which are typically only 5% of all received packets anyway. So by doing this we would be looking at saving 5% of 2.8% ~= 0.14% in Read + ReadRTP.

rtpreceiver.go
adriancable: Good idea, I'll make the change.

jech:
> the argument is that the reception of RTX packets should 'by default' be transparent to the caller.

I understand the purpose.  I'd like more information on the cost.

1. Are we discarding any data when rewriting the packet?  Is any of the data we discard impossible to recover?
2. How many allocations are being done while rewriting the packet?  If you give me the raw packet, I can copy the data into my data structures without incurring any extra allocations.

As a general rule, I tend to be cautious about library code doing too much magic behind my back.  So if you're arguing in favour of doing magic, you need to justify that the magic is not prohibitively expensive.

adriancable:
@jech - the approach taken by this PR follows your own recommendation from the original PR, where you said e.g.

> What about the following API: the existing Read/ReadRTP functions return either an original or an RTX packet, indifferently;

> we need to retrofit the existing Read functions to return RTX packets when available, as this is the only way to avoid modifying all existing clients.

Achieving this requires some 'background magic' since as you know RTX packets have a different payload structure. The costs of the magic were analysed by @kcaffrey and myself and there are comments in the @kcaffrey's review notes here with the numbers, so you can take a look there. It is almost certainly possible to optimize further e.g. reduce allocations by doing all work within a single buffer. I am very supportive of this, and we don't need to achieve the optimum in the first shot.

To your first question, we're not discarding any data when rewriting the packet. All the data that's moved out is made available via the attributes.

adriancable:
@jech - I've rewritten the logic to handle RTX packets allocationless, i.e. there's no extra unmarshal/marshal and everything is done inline in the returned slice. Hope this addresses your concern.

Sean-Der:
This is amazing @adriancable thank you so much for doing this!

I am in support of merging this into `master` when you get tests/lints to pass. This weekend I will cover it with a bunch of tests.

cnderrauber:
There might have a data race in the allocationless buffer, when the `TrackRemote.Read` reads a rtx packet from the channel and copying the data from the buffer, the buffer can be written by the goroutine of `receiveForRTX` at same time.

adriancable:
@cnderrauber - thanks for your fix!

PR #2595, opened 2023-10-01
Fix data race of RTX packet

cnderrauber:
When the TrackRemote.Read reads a rtx packet from the channel and copying the data from the buffer, the buffer can be written by the goroutine of receiveForRTX at the same time, use a buffer pool to fix it.


adriancable (approved):

adriancable:
@cnderrauber - this is a good spot! Thanks for fixing, this looks good.

PR #2617, opened 2023-11-16
ReplaceTrack resets all track encodings

biglittlebigben:
#### Description
This PR addresses an issue where calling `RTPSender.ReplaceTrack` with a nil parameter on a sender with more than 1 encoding (simulcast) would only cause the 1st encoding to be unbound, breaking common publisher reconnection workflows with simulcast enabled.


cnderrauber (approved):

davidzhao (approved):
lgtm!

PR #2632, opened 2023-12-11
Allow multiple negotiated A/V codecs respectively

pougetat:
#### Description

Pion currently restricts the negotiated audio codec and negotiated video codec to the first exact / partial matches it finds in a remote description.

This causes issues in the following scenario:

- Call `RegisterCodec()`.
- Call `AddTransceiver()` for multiple video codecs / multiple audio codecs.
- Pion instance receives a remote description, call `SetRemoteDescription()` => This updates the list of negotiated codecs.
- Pion instance creates an answer, call `CreateAnswer()`.
- Pion instance sets the local description, call `SetLocalDescription` => There is a conflict here between what transceivers have been configured and the negotiated codecs filtered down from the remote description.

#### Reference issue
Fixes #...

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -490,10 +490,10 @@ func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) e
 	for _, media := range desc.MediaDescriptions {
 		var typ RTPCodecType
 		switch {
-		case !m.negotiatedAudio && strings.EqualFold(media.MediaName.Media, "audio"):
+		case strings.EqualFold(media.MediaName.Media, "audio"):
 			m.negotiatedAudio = true
 			typ = RTPCodecTypeAudio
-		case !m.negotiatedVideo && strings.EqualFold(media.MediaName.Media, "video"):
+		case strings.EqualFold(media.MediaName.Media, "video"):
 			m.negotiatedVideo = true
 			typ = RTPCodecTypeVideo
 		default:
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -72,7 +72,7 @@ type RTPReceiver struct {
 	rtxPool sync.Pool
 }
 
-// NewRTPReceiver constructs a new RTPReceiver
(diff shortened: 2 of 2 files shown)


streamer45:
FWIW I am hitting this same issue when trying to receive multiple differently encoded tracks.

streamer45:
@Sean-Der Any thoughts on this? We may actually need to make similar changes on our end so it would be great to get some feedback on what could be the best way forward to send multi codec tracks.

Sean-Der:
Hey @streamer45 

Is this what you are trying to implement? You want to have one `TrackLocal` that can be used to send many different codecs?

streamer45:
> Hey @streamer45
> 
> Is this what you are trying to implement? You want to have one `TrackLocal` that can be used to send many different codecs? Since you are forwarding media created by someone else, you don't know the codec at PeerConnection creation time?

Hey @Sean-Der, to be honest I may have lost a bit of context on this one as I can't seem to reproduce the original issue (failing to negotiate the track) so maybe it was a simple misimplementation on our part. 

Our use case is rather simple: we have one peer connection (sender) pushing two tracks with different encodings and then want to selectively forward one of the two to our receivers, depending on what they support.

itzmanish:
Hi @Sean-Der, I think this PR makes sense. If I understand the region behind having a check of the `negotiated` Flag on each media section, that is for having a track with a single encoding because the browser may send a lot of codecs to negotiate, and it's not ideal to create a track for all the codecs. 

However, in the scenario where the client/browser indeed wants to have multiple tracks with different encodings, this check prohibits that.

In my opinion, having a configuration around this on the media engine should be a better fix where by default the media engine will only select one codec for subsequent media sections.

itzmanish:
> Our use case is rather simple: we have one peer connection (sender) pushing two tracks with different encodings and then want to selectively forward one of the two to our receivers, depending on what they support.

@streamer45 were you able to achieve that without using this PR?

JoeTurki:
closed in favor of https://github.com/pion/webrtc/pull/3018

PR #2634, opened 2023-12-12
Answer to paused simucalst stream correctly

cnderrauber:
#### Description
If the publisher sends an offer with paused simulcast stream like `a=simucalst:send 1;~2;~3`, the sfu needs to respond with paused stream too, otherwise, it will make the publisher's inactive encoding to become active unintentionally.

#### Reference issue
Fixes #...

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -22,6 +22,8 @@ const (
 
 	sdpAttributeRid = "rid"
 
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
LGTM!

PR #2655, opened 2024-02-02
Skip padding packet for simulcast probe

cnderrauber:
Skip padding packet for simulcast probe

Fix rtx attributes panic for nil map

### DIFF
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -1280,7 +1280,28 @@ func TestPeerConnection_Simulcast(t *testing.T) {
 
 		assert.NoError(t, signalPair(pcOffer, pcAnswer))
 
-		for sequenceNumber := uint16(0); !ridsFullfilled(); sequenceNumber++ {
+		// padding only packets should not affect simulcast probe
+		var sequenceNumber uint16
+		for sequenceNumber = 0; sequenceNumber < simulcastProbeCount+10; sequenceNumber++ {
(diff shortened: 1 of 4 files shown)


boks1971 (approved):

adriancable (approved):

davidzhao (approved):
lgtm!

especially fixing that panic! I guess no one's using this path :)

peerconnection_media_test.go
```diff
@@ -1280,7 +1280,28 @@ func TestPeerConnection_Simulcast(t *testing.T) {
...
+
+			for _, track := range []*TrackLocalStaticRTP{vp8WriterA, vp8WriterB, vp8WriterC} {
+				pkt := &rtp.Packet{
+					Header: rtp.Header{
+						Version:        2,
+						SequenceNumber: sequenceNumber,
+						PayloadType:    96,
+						Padding:        true,
+					},
+					Payload: []byte{0x00, 0x02},
```
boks1971, line 1296: @cnderrauber Sorry, missed this. This has two byte payload, but the check for padding only checks for `Padding == true` && payload length is zero. Here it is not. I am guessing this is not fulfilling rids because the extension is not set.

boks1971: Oops, please ignore me. The payload last byte is padding size. My bad. Sorry about the fake news.

PR #2664, opened 2024-02-05
Refine rtx support

cnderrauber:
Always handle header extensions from packet read
from interceptor, let interceptor has consistent
chance to process headers

Fix rtx is not negotiated when there is multiple
codecs has same mime but different profile (H264)

Fix rtx stream info missed when SSRC group attr
shows after base track's ssrc attr.

### DIFF
--- a/mediaengine_test.go
+++ b/mediaengine_test.go
@@ -308,8 +308,18 @@ a=rtpmap:96 VP8/90000
 o=- 4596489990601351948 2 IN IP4 127.0.0.1
 s=-
 t=0 0
-m=video 60323 UDP/TLS/RTP/SAVPF 94 96 97
+m=video 60323 UDP/TLS/RTP/SAVPF 94 95 106 107 108 109 96 97 
 a=rtpmap:94 VP8/90000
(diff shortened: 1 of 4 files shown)


mediaengine_test.go
```diff
@@ -318,22 +328,62 @@ a=fmtp:97 apt=96
...
 		assert.True(t, m.negotiatedVideo)
 
-		_, _, err := m.getCodecByPayload(97)
+		vp9Codec, _, err := m.getCodecByPayload(96)
+		assert.NoError(t, err)
+		assert.Equal(t, vp9Codec.MimeType, MimeTypeVP9)
+		vp9RTX, _, err := m.getCodecByPayload(97)
+		assert.NoError(t, err)
+		assert.Equal(t, vp9RTX.MimeType, "video/rtx")
+		h264P1Codec, _, err := m.getCodecByPayload(106)
```
boks1971, line 371: :nit: add an empty line between these different codec + rtx test combinations, will make it easier to read.

boks1971 (approved):
LGTM!

PR #2666, opened 2024-02-05
RTX support for v3

cnderrauber:
#### Reference issue
Fixes #...

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -39,10 +39,12 @@ const (
 
 	sdesRepairRTPStreamIDURI = "urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id"
 
-	// Attributes returned when Read() returns an RTX packet from a separate RTX stream (distinct SSRC)
-	attributeRtxPayloadType    = "rtx_payload_type"
-	attributeRtxSsrc           = "rtx_ssrc"
-	attributeRtxSequenceNumber = "rtx_sequence_number"
+	// AttributeRtxPayloadType is the interceptor attribute added when Read() returns an RTX packet containing the RTX stream payload type
+	AttributeRtxPayloadType = "rtx_payload_type"
+	// AttributeRtxSsrc is the interceptor attribute added when Read() returns an RTX packet containing the RTX stream SSRC
+	AttributeRtxSsrc = "rtx_ssrc"
+	// AttributeRtxSequenceNumber is the interceptor attribute added when Read() returns an RTX packet containing the RTX stream sequence number
+	AttributeRtxSequenceNumber = "rtx_sequence_number"
 )
 
 func defaultSrtpProtectionProfiles() []dtls.SRTPProtectionProfile {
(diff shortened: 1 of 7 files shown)


adriancable:
This looks great (and of course I am a fan of RTX support, having done the implementation in v4 :smiley:). But I am wondering, in general what is our approach regarding backporting stuff to v3 vs. focussing on v4?

My own view (not strongly held) is that we should get v4 out of beta so it becomes the 'de facto' base for new users, and basically freeze v3 except maybe for important bug fixes. I think this is what happened when we went from v2 to v3. But we should probably (as maintainers) decide the approach collectively because there is a time/effort cost in backporting things to v3.

adriancable:
@alexpokotilo - RTX receiving *over a distinct SSRC* is supported in v4 and, with this PR, is backported to v3. RTX sending *over a distinct SSRC* is currently not supported as far as I know. RTX sending/receiving *over the same track/SSRC as the main media* i.e. via the NACK mechanism has been supported for a while by Pion's NACK generator/sender interceptor.

alexpokotilo:
@adriancable NACK over the main works fine, you are right! The problem with not supporting RTX in Pion is that we cannot implement Bandwidth Estimation as we need to add bandwidth probing. Bandwidth probing from another hand cannot be added using main SSRC since sending the same RTP packet (with the same payload) would lead browsers(Chrome and Firefox) think we are doing Replay Attack. I tried to send some RTP packets twice changing TWCC header, but browsers reject my duplicates since payload is the same as SSRC is the same and Sequence Number of RTP packet is the same.Thanks to @lminiero I was able to find why browsers reject these RTPs. So we need to send probing packets over distinct SSRC. 
We will not be able to implement BWE for Pion without RTX sending over a distinct SSRC support as padding must be added before we can increase stream bitrate and there is no way to measure network other than send some probing packets over time.
But now we cannot do that. 
What should I theoretically do to add RTX support for distinct SSRC ?
I really appreciate your help !

adriancable:
@alexpokotilo - for video streams with same-track RTX via NACK feedback, BWE probes are implemented as empty-payload (i.e. padding only) RTP packets. You do not need distinct-SSRC RTX for this.

Distinct-SSRC RTX has an advantage for BWE in that empty-payload probes are generally not needed because you can use repeat RTP packets as probes instead. But you do not need distinct-SSRC for BWE probes.

See: https://chromium.googlesource.com/external/webrtc/+/master/pc/g3doc/rtp.md

cnderrauber:
@adriancable Agree to focus on release v4 instead of backporting features to v3. We also have a plan to upgrade to v4 in our product. I add the RTX to v3 because it is important to us now, after evaluating the effort and risk between upgrading to v4 and supporting it in v3, I chose the latter. It also contains bug fixes merged to master first and cherry-pick to v3 so it is focusing on v4 too. Will re-evaluate v4 later and discuss with other developers about the plan to release v4.

PR #2677, opened 2024-02-20
Export unexported functions

potaga:
#### Description
For performance reasons, some projects need to bypass part of the standard pion path and access the SRTP session directly. This PR exports the GetSRTPSession().
Related to: [https://github.com/pion/srtp/pull/259](https://github.com/pion/srtp/pull/259)

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -236,6 +236,11 @@ func (t *DTLSTransport) startSRTP() error {
 	return nil
 }
 
+// GetSRTPSession returns the SRTP session
+func (t *DTLSTransport) GetSRTPSession() (*srtp.SessionSRTP, error) {
+       return t.getSRTPSession()
+}
+
 func (t *DTLSTransport) getSRTPSession() (*srtp.SessionSRTP, error) {
 	if value, ok := t.srtpSession.Load().(*srtp.SessionSRTP); ok {
 		return value, nil


Sean-Der:
I spoke with @potaga via Slack and I don't believe this change is needed! You can access the underlying SRTP session of a TrackLocal by defining your own version like this.

You create a `trackLocalRaw` then call `AddTrack` on the PeerConnection. This replaces TrackLocalStaticRTP (or sample)

You have direct access to `trackLocalRaw.WriteStream`. You need to know the SSRC+PayloadType that needs to be used. 

```
type trackLocalRaw struct {
	WriteStream  webrtc.TrackLocalWriter

	kind        webrtc.RTPCodecType
	id, rid, streamID string
}

func (t *trackLocalRaw) Bind(ctx webrtc.TrackLocalContext) (webrtc.RTPCodecParameters, error) {
	t.writeStream = ctx.WriteStream()
	return ctx.CodecParameters()[0], nil
}

func (t *trackLocalRaw) Unbind(webrtc.TrackLocalContext) error {
	return nil
}

func (t *trackLocalRaw) ID() string       { return t.id }
func (t *trackLocalRaw) RID() string      { return t.rid }
func (t *trackLocalRaw) StreamID() string { return t.streamID }
func (t *trackLocalRaw) Kind() webrtc.RTPCodecType {
	return kind
}
```

PR #2696, opened 2024-03-04
Drop reference to detached datachannels

### LINKED ISSUE #2672: PeerConnection should drop reference to datachannel when it's closed
### Your environment.
  - Version: 3.x

### What did you do?
libp2p is a peer to peer networking library used in IPFS among other things. We are using webrtc in go-libp2p to provide connectivity to nodes running in the browser. For our use, we create a peer connection and open lots of data channels over the life time of the connection. 

Currently the reference to datachannels is not dropped when a datachannel is closed. `PeerConnection.sctpTransport.dataChannels` always has a reference to the existing data channels which has a reference to the sctp stream. Over the life time of the connection this leads to a lot of memory usage. 

### What did you expect?

I expected the memory used by the datachannel to be garbage collected. 

### Proposed fix

`Datachannel` has an OnClose handler. This handler should remove the datachannel from the sctpTransport
To preserve existing behaviour we should:
1. Only keep the reference to the ID that was used on the datachannel
2. Create a `DataChannelStats` object for the stats on the datachannel to preserve collectStats behaviour.

sukunrt:
This allows users of detached datachannels to garbage collect resources associated with the datachannel and the sctp stream. There is no functional change here.

Fixes #2672

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -420,7 +420,6 @@ func (d *DataChannel) ensureOpen() error {
 // resulting DataChannel object.
 func (d *DataChannel) Detach() (datachannel.ReadWriteCloser, error) {
 	d.mu.Lock()
-	defer d.mu.Unlock()
 
 	if !d.api.settingEngine.detach.DataChannels {
 		return nil, errDetachNotEnabled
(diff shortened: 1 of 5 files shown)


datachannel.go
```diff
@@ -432,7 +431,28 @@ func (d *DataChannel) Detach() (datachannel.ReadWriteCloser, error) {
...
-	return d.dataChannel, nil
+	dataChannel := d.dataChannel
+	d.mu.Unlock()
+
+	// Remove the reference from SCTPTransport so that the datachannel
+	// can be garbage collected on close
+	d.sctpTransport.lock.Lock()
+	n := len(d.sctpTransport.dataChannels)
+	j := 0
+	for i := 0; i < n; i++ {
```
MarcoPolo, line 442: Does `d.sctpTransport.dataChannels` need to remain in order? Can you swap remove instead?

MarcoPolo (approved):
Seems reasonable to me. Thanks @sukunrt!

datachannel.go
sukunrt: I think I can. I just didn't want to change that property here.

Sean-Der:
Thank you so much @sukunrt! Really appreciate the tests :)

Rebasing + merging now

Sean-Der (approved):

PR #2697, opened 2024-03-05
handle the case where TrackRemote payload type is 0 (PCMU)

mklingb:
#### Description
In the case where a remote track is sending PCMU with payload type 0, `checkAndUpdateTrack` will fail to update the track codec and params (because `t.PayloadType()` is already 0). Add an extra check to handle this case.


edaniels (commented):
This is great! Would it be much effort to add a test for this?

Sean-Der:
cherry-picked as 6deb47d3bf148db556cf3df905f236770a15b24d

thank you @mklingb !

PR #2702, opened 2024-03-13
Make OnConnectionStateChange wait for handler to finish

sukunrt:
`OnConnectionStateChange` is used to determine whether the peer connection has succeeded and track connection state by users. It should be called in sync. Right now state changes might get reordered which would lead to tracking errors for users.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -510,13 +510,7 @@ func (pc *PeerConnection) onConnectionStateChange(cs PeerConnectionState) {
 	pc.connectionState.Store(cs)
 	pc.log.Infof("peer connection state changed: %s", cs)
 	if handler, ok := pc.onConnectionStateChangeHandler.Load().(func(PeerConnectionState)); ok && handler != nil {
-		var wg sync.WaitGroup
-		wg.Add(1)
-		go func() {
-			handler(cs)
-			wg.Done()
-		}()
-		wg.Wait()
+		handler(cs)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


peerconnection.go
```diff
@@ -510,7 +510,13 @@ func (pc *PeerConnection) onConnectionStateChange(cs PeerConnectionState) {
...
 	pc.log.Infof("peer connection state changed: %s", cs)
 	if handler, ok := pc.onConnectionStateChangeHandler.Load().(func(PeerConnectionState)); ok && handler != nil {
-		go handler(cs)
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			handler(cs)
+			wg.Done()
+		}()
+		wg.Wait()
```
Wondertan, line 519: Why wg instead of just removing `go`?

Sean-Der: @Wondertan this is so the callback doesn’t block the ICE Agent

hopefully we have a test for that :)

Sean-Der: I believe we have this issue with this PR. The code right now would allow a callback to block the PeerConnection internals.

Sean-Der (changes requested):

peerconnection.go
sukunrt: >Why wg instead of just removing go?

Yeah sorry this is not required at all.

sukunrt:
I have moved the updates to a operations queue. That'll run the notifications sequentially. This ensures that users can do things like `pc.Close` from the notification.

peerconnection.go
Wondertan: No worries @sukunrt, all good

Sean-Der:
This was fixed by https://github.com/pion/ice/commit/67cc918a518d3b3dd1887b07434331dabc846702, going to close here!

Will be tagging a new version of pion/ice today! thank you @sukunrt :)

sukunrt:
It's not exactly fixed. The state reported on the callback isn't reliable. Users need to ignore the state provided on the call back and call `pc.ConnectionState()` for the accurate state of the connection.

PR #2718, opened 2024-04-02
Fix deprecated golangci-lint config

mogren:
Executing `golangci-lint run` using old config generates the following warnings:

WARN [config_reader] The configuration option `run.skip-dirs-use-default` is deprecated, please use `issues.exclude-dirs-use-default`. 
WARN [config_reader] The configuration option `linters.govet.check-shadowing` is deprecated. Please enable `shadow` instead, if you are not using `enable-all`.

#### Description
Before change:
Screenshot 2024-04-01 at 20 59 13

After change:
Screenshot 2024-04-01 at 21 26 42

#### Reference issue
No issue filed for this warning.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,7 +3,8 @@
 
 linters-settings:
   govet:
-    check-shadowing: true
+    enable:
+      - shadow
   misspell:
     locale: US
   exhaustive:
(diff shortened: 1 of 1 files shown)


Sean-Der:
Thank you so much @mogren excited to get your help. Can you make this against https://github.com/pion/.goassets/blob/master/ci/.golangci.yml

when you push a tag to that repo it syncs across everything then.

mogren:
Thanks! Just testing so that I have things set up right 😄

mogren:
Moved to https://github.com/pion/.goassets/pull/208

PR #2751, opened 2024-04-18
Fix receiveForRtx usage for simulcast

aalekseevx:
Hello! We have encountered some issues with simulcast after latest upgrade. All layers, except for one, were shutting down very quickly after the connection was established. After some investigation, we found that this issue is related to this specific [commit](https://github.com/pion/webrtc/commit/5da72784c8d9425fe61bbf66de7b658eb94d6717). This change does not seem to work correctly when receiveForRtx method is [called](https://github.com/pion/webrtc/blob/master/peerconnection.go#L1625) to receive a new simulcast layer. Therefore, this PR is intended to restore the previous behavior in order to ensure that simulcast continues to function properly.

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -33,11 +33,6 @@ const (
 	generatedCertificateOrigin = "WebRTC"
 
 	sdesRepairRTPStreamIDURI = "urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id"
-
-	// Attributes returned when Read() returns an RTX packet from a separate RTX stream (distinct SSRC)
-	attributeRtxPayloadType    = "rtx_payload_type"
-	attributeRtxSsrc           = "rtx_ssrc"
-	attributeRtxSequenceNumber = "rtx_sequence_number"
 )
 
 func defaultSrtpProtectionProfiles() []dtls.SRTPProtectionProfile {
--- a/constants.go
+++ b/constants.go
@@ -34,12 +34,10 @@ const (
 
 	sdesRepairRTPStreamIDURI = "urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id"
 
-	// AttributeRtxPayloadType is the interceptor attribute added when Read() returns an RTX packet containing the RTX stream payload type
-	AttributeRtxPayloadType = "rtx_payload_type"
-	// AttributeRtxSsrc is the interceptor attribute added when Read() returns an RTX packet containing the RTX stream SSRC
-	AttributeRtxSsrc = "rtx_ssrc"
-	// AttributeRtxSequenceNumber is the interceptor attribute added when Read() returns an RTX packet containing the RTX stream sequence number
-	AttributeRtxSequenceNumber = "rtx_sequence_number"
+	// Attributes returned when Read() returns an RTX packet from a separate RTX stream (distinct SSRC)
+	attributeRtxPayloadType    = "rtx_payload_type"
+	attributeRtxSsrc           = "rtx_ssrc"
+	attributeRtxSequenceNumber = "rtx_sequence_number"
 )
 
 func defaultSrtpProtectionProfiles() []dtls.SRTPProtectionProfile {
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -414,11 +414,9 @@ func (m *MediaEngine) matchRemoteCodec(remoteCodec RTPCodecParameters, typ RTPCo
 		}
 
 		aptMatch := codecMatchNone
-		var aptCodec RTPCodecParameters
 		for _, codec := range exactMatches {
 			if codec.PayloadType == PayloadType(payloadType) {
 				aptMatch = codecMatchExact
-				aptCodec = codec
 				break
 			}
 		}
@@ -427,7 +425,6 @@ func (m *MediaEngine) matchRemoteCodec(remoteCodec RTPCodecParameters, typ RTPCo
 			for _, codec := range partialMatches {
 				if codec.PayloadType == PayloadType(payloadType) {
 					aptMatch = codecMatchPartial
-					aptCodec = codec
 					break
 				}
 			}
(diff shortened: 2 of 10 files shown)


Sean-Der:
That’s not good, thank you @aalekseevx 

Is it possible to get a failing test? If you can fix the lint would love to merge right now!

aalekseevx:
Sure, will think of the failing test and fix the linter

Sean-Der:
@aalekseevx If you don't have the time I am happy to fix this tonight! After work I have some time to do Pion stuff :)

aalekseevx:
@Sean-Der, thank you! I'll fix it, just need some time ☺️

aalekseevx:
I couldn't write a simple test to reproduce it because we need to send simulcast RTX to receive it, which is not yet supported as far as I understand. It seems that the root of the problem is that SSRCs are not parsed correctly for simulcast, and `HasRTX()` returns false for such streams. Therefore, `readRTX` does not work, but `receiveForRtx` does work, resulting in a deadlock.

Logged RTPDecodingParameters for simulcast:
```
{RTPCodingParameters:{RID:hi SSRC:0 PayloadType:0 RTX:{SSRC:0}}}
{RTPCodingParameters:{RID:med SSRC:0 PayloadType:0 RTX:{SSRC:0}}}
{RTPCodingParameters:{RID:low SSRC:0 PayloadType:0 RTX:{SSRC:0}}}
```

Blocked goroutine:
```
1 @ 0x291732e 0x29283a5 0x322d570 0x294b0c1
#	0x322d56f	github.com/pion/webrtc/v3.(*RTPReceiver).receiveForRtx.func1+0x56f	/-S/vendor/github.com/pion/webrtc/v3/rtpreceiver.go:486
```
I think the best way to fix it is to make simulcast RTX work via the newly created readRTX. However, it may be worth returning to the old behavior for now to fix what's broken.

aalekseevx:
@Sean-Der, updated!

Sean-Der:
@aalekseevx Revert sounds good to me! Want to use this PR for it?

I will work with original author on relanding with tests

Sean-Der (approved):

Sean-Der:
Want to use this PR and just do revert commits?

aalekseevx:
My commit is literally a revert, but only for simulcast. It fixes the issue for me. Do we need anything else?

Sean-Der:
@aalekseevx Could we revert the commit in whole? 

I would do just a normal `git revert` on the RTX changes. We can revisit them considering all use cases.

aalekseevx:
No problem

aalekseevx:
@Sean-Der, do I need to open a separate PR to v3?

Sean-Der:
Nope! I think cherry-pick is fine when this lands.

I didn’t realize this would break public API! I need to be more careful with this

@boks1971 @cnderrauber could I get your help on this? I don’t know the RTX code well enough to fix, so I was going to revert and re-land.

I don’t think that’s a good idea because of API breakage. Do you think we could easily test/fix this issue or is it ambiguous?

cnderrauber:
I remember there was a developer said they enabled rtx in prod and don't get problem, we enabled it for some users too. I will try to fix it before revert the feature if you have the steps to reproduce this issue.

cnderrauber:
I think the code to update rsid is already there and it works well in livekit sfu while it is using buffer instead of track.Read to read rtp packet. I'll test the normal read method.

Sean-Der:
@cnderrauber any progress on testing? I added d9a59a594982f950c56cc49a870054f165aa2560, I will add a test for Simulcast tomorrow

cnderrauber:
> @cnderrauber any progress on testing? I added [d9a59a5](https://github.com/pion/webrtc/commit/d9a59a594982f950c56cc49a870054f165aa2560), I will add a test for Simulcast tomorrow

Uses examples/simulcast, I have no problem to send all 3 simulcast layers back to the browser. @aalekseevx which browser are you using, can you try it with the simulcast example in the repo?

aalekseevx:
The original issue was encountered with chromium,  I'll work on reproduction based on examples/simulcast

aalekseevx:
@cnderrauber, I have reproduced the unwanted behaviour in my [branch](https://github.com/aalekseevx/webrtc/commit/f8a4a765629189cd4749782f50f9e33ee929be8f). It may be not 100% accurate, but I hope it will be helpful.

I run it as follows:
```
 go test -v -run ^TestPeerConnection_Simulcast$/^E2E$ -timeout 3s
=== RUN   TestPeerConnection_Simulcast
=== RUN   TestPeerConnection_Simulcast/E2E
readRTX not reading from t.repairStreamChannel, reader.HasRTX == false
readRTX not reading from t.repairStreamChannel, reader.HasRTX == false
readRTX not reading from t.repairStreamChannel, reader.HasRTX == false
readRTX not reading from t.repairStreamChannel, reader.HasRTX == false
readRTX not reading from t.repairStreamChannel, reader.HasRTX == false
readRTX not reading from t.repairStreamChannel, reader.HasRTX == false
trying to write to track.repairStreamChannel
trying to write to track.repairStreamChannel
trying to write to track.repairStreamChannel
    peerconnection_media_test.go:1382: 
                Error Trace:    /home/alekseev-dev/webrtc/peerconnection_media_test.go:1382
                Error:          "0" is not greater than "0"
                Test:           TestPeerConnection_Simulcast/E2E
                Messages:       no rtx packet read
--- FAIL: TestPeerConnection_Simulcast (2.07s)
    --- FAIL: TestPeerConnection_Simulcast/E2E (1.67s)
FAIL
exit status 1
FAIL    github.com/pion/webrtc/v4       2.072s
```

cnderrauber:
Your test case is useful, although it needs to change to match the real case. The root cause is TrackRemote relies on HasRTX(rtxssrc!=0) to determine if it needs to read the rtx packet but it dosen't have that in the simulcast case because rtx ssrc is not included in sdp. The result is TrackRemote.Read/ReadRTP method can't get any rtx packet in this case but only get media packet so examples/simucalst can pass the test.

PR #2755, opened 2024-04-22
Add recvonly transceiver for simulcast sample

cnderrauber:
Fix one of the layers can't send back video because the transceiver is used for receiving.

#### Reference issue
Fixes #...


boks1971 (approved):
lgtm!

Issue #1516, opened 2020-11-11
Reduce DataChannel.readLoop allocations


bshimc:
## Summary

I'm evaluating webrtc for a use case that stresses DataChannels.  In profiling, I tracked down some pathological alloc behavior to readLoop.  A simple patch that uses `sync.Pool` to minimize allocs (see patch below) resulted in what appear to be significant GC improvements.  That said, the patch in question changes the semantics of the datachannel OnMessage handler; that's to say the contents of the message buffer are only valid inside the lifetime of the handler function.  For my use case this was acceptable but that's unlikely to be the case in general (great if it is).

## Motivation

Performance.

## Describe alternatives you've considered

To guarantee compatibility we'd need to allocate a new buffer sized to `n` and copy the contents to be passed on before releasing the larger `dataChannelBufferSize` buffers.  I haven't tried this.  For our purposes it would be nice to have a configuration that could toggle between no-copy and legacy-safe behaviors.

## Additional context

```diff
diff --git a/vendor/github.com/pion/webrtc/v3/datachannel.go b/vendor/github.com/pion/webrtc/v3/datachannel.go
index ba1269d..7a32c70 100644
--- a/vendor/github.com/pion/webrtc/v3/datachannel.go
+++ b/vendor/github.com/pion/webrtc/v3/datachannel.go
@@ -311,11 +311,16 @@ func (d *DataChannel) onError(err error) {
        }
 }

+var rlBufPool = sync.Pool{New: func() interface{} {
+       return make([]byte, dataChannelBufferSize)
+}}
+
 func (d *DataChannel) readLoop() {
        for {
-               buffer := make([]byte, dataChannelBufferSize)
+               buffer := rlBufPool.Get().([]byte)
                n, isString, err := d.dataChannel.ReadDataChannel(buffer)
                if err != nil {
+                       rlBufPool.Put(buffer)
                        d.setReadyState(DataChannelStateClosed)
                        if err != io.EOF {
                                d.onError(err)
@@ -325,6 +330,7 @@ func (d *DataChannel) readLoop() {
                }

                d.onMessage(DataChannelMessage{Data: buffer[:n], IsString: isString})
+               rlBufPool.Put(buffer)
        }
 }
```

Top (before) bottom (after) memory allocation profiles.  The second set of profiles actually is generating a lot more QPS through the established datachannels.

![image](https://user-images.githubusercontent.com/69528604/98773956-d62cd780-239e-11eb-9ff7-ad20beb2b291.png)

[...]

PR #2758, opened 2024-04-24
Replace pool with bytes in readLoop

cnderrauber:
Replace pool with bytes in readLoop

#### Reference issue
Fixes #...

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -349,18 +349,11 @@ func (d *DataChannel) onError(err error) {
 	}
 }
 
-// See https://github.com/pion/webrtc/issues/1516
-// nolint:gochecknoglobals
-var rlBufPool = sync.Pool{New: func() interface{} {
-	return make([]byte, dataChannelBufferSize)
-}}
-
 func (d *DataChannel) readLoop() {
+	buffer := make([]byte, dataChannelBufferSize)
 	for {
-		buffer := rlBufPool.Get().([]byte) //nolint:forcetypeassert
 		n, isString, err := d.dataChannel.ReadDataChannel(buffer)
 		if err != nil {
-			rlBufPool.Put(buffer) // nolint:staticcheck
 			d.setReadyState(DataChannelStateClosed)
 			if !errors.Is(err, io.EOF) {
 				d.onError(err)
@@ -371,8 +364,6 @@ func (d *DataChannel) readLoop() {
 
 		m := DataChannelMessage{Data: make([]byte, n), IsString: isString}
 		copy(m.Data, buffer[:n])
-		// The 'staticcheck' pragma is a false positive on the part of the CI linter.
-		rlBufPool.Put(buffer) // nolint:staticcheck
 
 		// NB: Why was DataChannelMessage not passed as a pointer value?
 		d.onMessage(m) // nolint:staticcheck


boks1971 (approved):
lgtm!

Sean-Der:
@cnderrauber Would you mind explaining the reasoning for the change?

Does this reduce allocation for a single data channel? What is the impact if the user has many?

The original change was driven with data. Would be nice to match that!

cnderrauber:
@Sean-Der It is not a revert to the change https://github.com/pion/webrtc/commit/159ba5aca3b96b85f7a653329102ea12046a1b1f but a simplify, the oldest behavior was allocating a 65KB buffer for each read call, then @bshi change it to get buffer from pool for every loop, but we don't need a pool here if the readLoop always need a bytes buffer, just allocate it once at the head and release it at end

cnderrauber:
```
goos: darwin
goarch: arm64
pkg: github.com/pion/webrtc/v4
                     │    old.txt    │                  new.txt                  │
                     │    sec/op     │     sec/op      vs base                   │
DataChannelSend2-10     1.860µ ± ∞ ¹   1.837µ ±   ∞ ¹        ~ (p=1.000 n=1+3) ²
DataChannelSend4-10     3.882µ ± ∞ ¹   2.175µ ±   ∞ ¹        ~ (p=0.400 n=3+2) ²
DataChannelSend8-10     2.797µ ± ∞ ¹   3.271µ ±   ∞ ¹        ~ (p=1.000 n=1)   ²
DataChannelSend16-10   13.133µ ± ∞ ¹   5.842µ ±   ∞ ¹        ~ (p=0.333 n=2)   ²
DataChannelSend32-10     1.119 ± ∞ ¹    1.030 ± 19%          ~ (p=0.683 n=4+8)
geomean                 49.49µ         37.94µ          -23.33%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                     │    old.txt    │                  new.txt                  │
                     │     B/op      │      B/op        vs base                  │
DataChannelSend2-10    1.274Ki ± ∞ ¹   1.243Ki ±   ∞ ¹  -2.45% (n=1+3)
DataChannelSend4-10    1.305Ki ± ∞ ¹   1.252Ki ±   ∞ ¹       ~ (p=0.200 n=3+2) ²
DataChannelSend8-10    1.284Ki ± ∞ ¹   1.256Ki ±   ∞ ¹       ~ (p=1.000 n=1)   ²
DataChannelSend16-10   1.303Ki ± ∞ ¹   1.318Ki ±   ∞ ¹       ~ (p=1.000 n=2)   ²
DataChannelSend32-10   6.831Mi ± ∞ ¹   7.465Mi ± 22%         ~ (p=0.683 n=4+8)
geomean                7.208Ki         7.226Ki          +0.25%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                     │   old.txt    │                 new.txt                  │
                     │  allocs/op   │   allocs/op     vs base                  │
DataChannelSend2-10     33.00 ± ∞ ¹    32.00 ±   ∞ ¹       ~ (p=1.500 n=1+3) ²
DataChannelSend4-10     33.00 ± ∞ ¹    32.00 ±   ∞ ¹  -3.03% (n=3+2)
DataChannelSend8-10     33.00 ± ∞ ¹    32.00 ±   ∞ ¹       ~ (p=1.000 n=1)   ²
DataChannelSend16-10    24.50 ± ∞ ¹    32.00 ±   ∞ ¹       ~ (p=0.333 n=2)   ²
DataChannelSend32-10   87.43k ± ∞ ¹   73.69k ± 68%         ~ (p=0.808 n=4+8)
geomean                 150.4          150.5          +0.08%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

bshi:
> but we don't need a pool here if the readLoop always need a bytes buffer, just allocate it once at the head and release it at end

The reason that the pool is required is that, under high concurrency and GC pressure, memory allocation explodes and performance is crippled.

From the https://github.com/pion/webrtc/commit/159ba5aca3b96b85f7a653329102ea12046a1b1f - note how **allocations/op** stays constant as parallelism increases (and goes to 0 with a pooled approach).  Note also how time/op stays relatively constant because GC pressure has been alleviated.


image


Two things to note about the the most recent benchmark numbers in the previous comment are;

* the # of measurements needs to be higher to have more confidence in them (`-count`)
* The more interesting thing is that the old measurements already show signs of GC pressure.  At some point, some other changes in and around this code (not just this one) have made performance regressions.  It's probably worth repeating the profiling exercise in #1516 at this point.

```
                     │    old.txt    │                  new.txt                  │
                     │    sec/op     │     sec/op      vs base                   │
DataChannelSend2-10     1.860µ ± ∞ ¹   1.837µ ±   ∞ ¹        ~ (p=1.000 n=1+3) ²
DataChannelSend4-10     3.882µ ± ∞ ¹   2.175µ ±   ∞ ¹        ~ (p=0.400 n=3+2) ²
DataChannelSend8-10     2.797µ ± ∞ ¹   3.271µ ±   ∞ ¹        ~ (p=1.000 n=1)   ²
DataChannelSend16-10   13.133µ ± ∞ ¹   5.842µ ±   ∞ ¹        ~ (p=0.333 n=2)   ²
DataChannelSend32-10     1.119 ± ∞ ¹    1.030 ± 19%          ~ (p=0.683 n=4+8) <-- 1 second per operation (!!!)
geomean                 49.49µ         37.94µ          -23.33%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

I would recommend rolling back this change and profiling to figure out where this regression came from (or perhaps just bisect and benchmark for a quick search).  Instead of eliminating pooling, I suspect the team needs to *increase* the use of pools.

CC @Sean-Der

cnderrauber:
This is different with code before 1516, the problem of the original code is it allocates 65KB bytes in every read that cause the gc pressure, and this change allocates the buffer only once for a data channel readloop and reuse it for in the entire channel lifecycle, I don't think the data channel creation is a high frequency operation that needs a pool for it.

PR #2759, opened 2024-04-25
Update RtxSSRC for simulcast track remote

cnderrauber:
Fix #2751, updates remote track's rtx ssrc for
simulcast track doesn't contain rtx ssrc in sdp
since readRTX relies on rtx ssrc to determine if
it has a rtx stream.

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection_media_test.go
+++ b/peerconnection_media_test.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"regexp"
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
can't claim I understand the test fully, but looks good as far as I could understand!

aalekseevx:
Thank you very much! Porting this to v3 would be greatly appreciated

PR #2769, opened 2024-05-14
fix TestPeerConnection_Simulcast_RTX on v3

aalekseevx:
Newly added test could not pass as simulcast extension headers are not enabled by default on v3


Sean-Der:
Thank you @aalekseevx !

Issue #1912, opened 2021-08-05
Fix ftmp matching for all codecs


jech:
Pion used to match m lines in SDP by doing a simple comparison of the fmtp lines.  This does not work in general, as some fmtp parameters are symmetric (they must have the same value on both sides for the m lines to match) and others are declarative (they only give additional information about what the announcing peer is able to send or receive).  It might also be the case that some symmetric parameters can be omitted, indicating that the announcer will accept whatever value the other side finds convenient (not sure, the RFCs are unclear on this subject).

Here's the situation, as far as I can tell from a quick browse of the relevant documents:

* VP8 https://datatracker.ietf.org/doc/html/rfc7741
  * both parameters `max-fr` and `max-fs` are declarative.
* VP9 https://datatracker.ietf.org/doc/html/draft-ietf-payload-vp9
  * the parameter `profile-id` is symmetric, but may be omitted;
  * the parameters `max-fr` and `max-fs` are declarative.
* Opus https://datatracker.ietf.org/doc/html/rfc7587
  * all parameters are declarative.
* H.264 https://datatracker.ietf.org/doc/html/rfc6184
  * the parameters `profile-level-id` and `packetization-mode` are symmetric, with weird matching for `profile-level-id` (see https://github.com/pion/webrtc/pull/1903);
  * `level-asymmetry-allowed` changes the matching algorithm;
  * all other parameters are declarative;
  * there are weird rules for `sprop-parameter-sets`.

aler9:
The situation is the following:

Symmetric parameters now play a role in matching codecs between senders and receivers, in particular:

* H264 symmetric parameters were implemented with #1903
* VP9 symmetric parameters were implemented with #2772 and #2773

The only thing that remains is that the library does not prevent matching incompatible codecs, since there's a fallback matching algorithm that takes into consideration the mime type only:

https://github.com/pion/webrtc/blob/2d529be5717a7c8a454abbafad4c6851d30ca793/rtpcodec.go#L112-L117

Sean-Der:
The final issue was fixed with https://github.com/pion/webrtc/pull/3021 (correct me if I'm wrong @aler9!

If you define codecs with explicit values (ClockRate etc...) the matching will be explicit now.  The fallback behavior still happens if the values are empty.

aler9:
[...]

PR #2772, opened 2024-05-27
Fix AV1 and VP9 codec matching

aler9:
#### Description

Currently, AV1 or VP9 formats are matched regardless of the profile parameter. This was not noticeable until browsers started advertising multiple VP9 formats at the same time, each with a different profile ID, in order to allow the counterpart to send different streams on the basis of supported profiles.

This causes two issues: first, the library includes in the SDP all formats passed by the browser, regardless of the fact that the profile ID is registered in the API or not. Then, the library is unable to choose the correct format for streaming, causing an intermittent failure.

This patch fixes the matching algorithm and also covers the case in which the profile ID is missing, by using values dictated by specifications.

Tests were refactored since previous ones covered the same lines multiple times.

Related to #1912


internal/fmtp/av1.go
```diff
@@ -0,0 +1,40 @@
...
+	return "video/av1"
+}
+
+func (h *av1FMTP) Match(b FMTP) bool {
+	c, ok := b.(*av1FMTP)
+	if !ok {
+		return false
+	}
+
+	// AV1 RTP specification:
```
edaniels, line 20: can you link to the spec?

edaniels (approved):
This LGTM. Just one comment. You can merge if you don't want to address it :)

aler9:
@edaniels  thanks for the review, the issue in editing this patch is that is has already been merged inside the v3 branch (#2773), therefore editing it would de-synchronize the master branch from the v3 branch. I'll create a dedicated patch that adds the comment you requested.

PR #2777, opened 2024-05-29
Avoid dropping packets for simulcast probes

cptpcrd:
#### Description

Currently, when Pion is receiving simulcast, `PeerConnection.handleIncomingSSRC` reads (at least) 2 packets: one packet is used to determine the payload type, and one (or more) packets are used to determine the MID, RID, and repair stream ID. This means that the first video frame is always partially lost, which is undesirable.

To fix this:
1. Get the initial payload type from the SRTP read stream to avoid having to consume the first packet.
2. Peek (instead of reading) from the SRTP read stream to avoid having to consume the second packet (assuming that the MID/RID/repair stream ID come in right away).

(This patch relies on corresponding changes to `pion/srtp` which I will PR shortly and link back here.)

#### Reference issue

N/A

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -12,7 +12,6 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
-	"io"
 	"strconv"
 	"strings"
 	"sync"
@@ -1480,7 +1479,7 @@ func (pc *PeerConnection) startSCTP() {
 	}
 }
 
-func (pc *PeerConnection) handleUndeclaredSSRC(ssrc SSRC, remoteDescription *SessionDescription) (handled bool, err error) {
+func (pc *PeerConnection) handleUndeclaredSSRC(ssrc SSRC, payloadType PayloadType, remoteDescription *SessionDescription) (handled bool, err error) {
 	if len(remoteDescription.parsed.MediaDescriptions) != 1 {
 		return false, nil
 	}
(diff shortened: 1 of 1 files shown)


adriancable:
@cptpcrd - does your issue #1 (consuming the first packet) also impact non-simulcast?

cptpcrd:
> @cptpcrd - does your issue #1 (consuming the first packet) also impact non-simulcast?

As far as I'm aware, no.

jech:
Is this still relevant?  It looks like a thing that's worth fixing.

Sean-Der:
Sorry this took so long to get merged. With the holidays I hope to have more time to get through backlog.

Super impressed that you diagnosed and fixed this!

track_remote.go
```diff
@@ -116,43 +122,38 @@ func (t *TrackRemote) Codec() RTPCodecParameters {
...
-		t.mu.Unlock()
-		// someone else may have stolen our packet when we
-		// released the lock.  Deal with it.
-		if data != nil {
-			n = copy(b, data)
-			err = t.checkAndUpdateTrack(b)
-
-			return n, attributes, err
-		}
+	if receiver.haveClosed() {
```
Sean-Der, line 132: Before calling `Read` would return peaked packets (even though the Transceiver) was closed.

Sean-Der:
Good from me! 

@JoeTurki @boks197 @jech 1 just as a 'heads up' I am changing a core code path, want to make sure I am not messing anything up. If you don't mind looking it over

peerconnection_media_test.go
```diff
@@ -2493,3 +2493,105 @@ func Test_PeerConnection_RTX_E2E(t *testing.T) { //nolint:cyclop
 	closePairNow(t, pcOffer, pcAnswer)
 	assert.NoError(t, wan.Stop())
 }
+
+// Simulcast probe would lose packets. This test asserts that we don't drop
+// any packets during the probe.
```
boks1971: maybe remove comment which says `Simulcast probe would lose packets`. This PR fixes it. So, would be good to just say that this test ensures no drops during probe.

JoeTurki (approved):
tested it :)

peerconnection_media_test.go
Sean-Der: Fixed!

PR #2791, opened 2024-06-24
Fix tracking of successfully read bytes in oggreader

Kostassoid:
This is a one-liner fix for tracking `bytesReadSuccesfully` in `oggreader`. Currently, it's never updated so it remains at 0 even after reading pages successfully. As a result, it's impossible to use `oggreader` in a streaming context.

@Sean-Der, could you please help with reviewing/merging this one?


edaniels (approved):
This looks good!

Issue #2774, opened 2024-05-28
DataChannel negotiated peer connection's server.AddTrack doesn't always trigger client.OnTrack 


nicksanford:
### Your environment.
  - 2 webrtc peers using pion/webrtc/[v3.2.40](https://github.com/pion/webrtc/releases/tag/v3.2.40)

### What did you do?
1. Established a PeerConnection negotiated over a data channel between a client (recvonly) and a server (sendonly)
2. Once `pc.OnOpen` is called on both the client & server, the server creates a `NewTrackLocalStaticSample` sending h264, passes it to `pc.AddTrack`, and starts calling `trackLocal.WriteSample`
3. Client waits for `pc.OnTrack` to be invoked so it can start receiving the `RTP` packets written by server

Repro test case: https://github.com/pion/webrtc/pull/2775

### What did you expect?
The Client's `pc.OnTrack` callback function is always invoked

### What happened?
Sometimes the client's `pc.OnTrack` callback function is not invoked

### Additional Notes:
I think the problem might be in peerconnection.go 's `onNegotiationNeeded` state machine which changes states even if the `onNegotiationNeededHandler` is not installed.  If I uncomment [these lines](https://github.com/pion/webrtc/pull/2775/files#diff-c94e5058864f96b58c5e039ac0485b3a9640613f32a9e31bba905dc3f8438e7aR298-R300) the repro test case doesn't fail.

edaniels:
> @nicksanford @Sean-Der These changes (pulled from `v3.2.42`) are causing a regression on our end. A pion-based client is, at times, triggering step 2.2 and ending up in an endless loop.
> 
> https://github.com/pion/webrtc/blob/43a6a6996bff40982aa2d43a354da7325db8a558/peerconnection.go#L330-L334
> 
> I am trying to assemble a simple reproducer but let me know if you have any thoughts.

Taking a look now

edaniels:
So it turns out my patch was partially wrong and this issue is a non-issue. Firing the event handler again after negotiation is already needed is not allowed. This means @nicksanford that if you want to see a negotiation needed event, you must set the handler prior to any background operations happening that could cause the event to fire.

edaniels:
@streamer45 published https://github.com/pion/webrtc/releases/tag/v3.2.43. let me know if you see the same issue.

streamer45:
Thanks @edaniels for the incredibly quick turnaround on this! The new changes are working as expected.

edaniels:
[...]

PR #2792, opened 2024-06-24
Make onNegotiationNeeded conform to spec

edaniels:
- Removes non-canon logic
- Addresses #2774
- Effectively reverts ca05618c7f2fa0777dfbad03a1cb745b666716fd


edaniels:
@Sean-Der merging in since tests are passing and this fixes a probably nasty bug. We can revert if needed.

PR #2798, opened 2024-07-01
Make pc.Close wait on spawned goroutines to close

edaniels:
This ensures no goroutines are running when a PeerConnection is closed. I'd also like to backport this into v3 so that existing users can benefit. Relies on https://github.com/pion/ice/pull/706 (tests will fail before that's in).

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -40,6 +40,7 @@ type DataChannel struct {
 	readyState                 atomic.Value // DataChannelState
 	bufferedAmountLowThreshold uint64
 	detachCalled               bool
+	readLoopActive             chan struct{}
 
 	// The binaryType represents attribute MUST, on getting, return the value to
 	// which it was last set. On setting, if the new value is either the string
@@ -327,6 +328,7 @@ func (d *DataChannel) handleOpen(dc *datachannel.DataChannel, isRemote, isAlread
 	defer d.mu.Unlock()
 
 	if !d.api.settingEngine.detach.DataChannels {
+		d.readLoopActive = make(chan struct{})
 		go d.readLoop()
 	}
 }
@@ -350,6 +352,7 @@ func (d *DataChannel) onError(err error) {
 }
 
 func (d *DataChannel) readLoop() {
+	defer close(d.readLoopActive)
 	buffer := make([]byte, dataChannelBufferSize)
 	for {
 		n, isString, err := d.dataChannel.ReadDataChannel(buffer)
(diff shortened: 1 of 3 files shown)


Sean-Der:
What’s the downside of making this the default behavior? Close takes longer?

I would say having two Close methods is worse then blocking in Go. It’s a extra cognitive load on the user

edaniels:
> What’s the downside of making this the default behavior? Close takes longer?
> 
> I would say having two Close methods is worse then blocking in Go. It’s a extra cognitive load on the user

I'm cool with it being just one 8)

edaniels:
@Sean-Der updated Close instead.

datachannel.go
```diff
@@ -457,6 +460,12 @@ func (d *DataChannel) Close() error {
 		return nil
 	}
 
+	if d.readLoopActive != nil {
```
Sean-Der, line 463: What do you think of dropping the `nil` check or logging a `error` message? I would rather have a crash/error message over inconsistent behavior. 

Is closing a `nil` chan safe?

peerconnection.go
```diff
@@ -2268,8 +2291,11 @@ func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, re
 	}
 
 	pc.dtlsTransport.internalOnCloseHandler = func() {
-		pc.log.Info("Closing PeerConnection from DTLS CloseNotify")
+		if pc.isClosed.get() {
```
Sean-Der, line 2294: Is this getting fired multiple times right now?

peerconnection.go
```diff
@@ -2061,7 +2085,6 @@ func (pc *PeerConnection) Close() error {
 	// 2. A Mux stops this chain. It won't close the underlying
 	//    Conn if one of the endpoints is closed down. To
 	//    continue the chain the Mux has to be closed.
-	closeErrs := make([]error, 4)
```
Sean-Der, line 2064: Why remove?

edaniels: I moved it up

peerconnection.go
edaniels: It doesn't, but it can get fired from a DTLS closenotify while Close is happening, if the timing is right.

peerconnection.go
```diff
@@ -2097,6 +2114,19 @@ func (pc *PeerConnection) Close() error {
 	// https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close (step #11)
 	pc.updateConnectionState(pc.ICEConnectionState(), pc.dtlsTransport.State())
 
+	// non-canon steps
+	pc.sctpTransport.lock.Lock()
```
edaniels, line 2118: @Sean-Der I moved these steps lower. I also made Close not block since it's not clear to me if that's allowed per the spec.

datachannel.go
edaniels: I'm okay doing a check for `!d.api.settingEngine.detach.DataChannels` instead, which is when we have a read loop. What would you prefer?

edaniels:
@Sean-Der cleaned up Close and made a blocking/non-blocking version to behave with when/where it's called (PeerConnection.Close (block) or directly (cannot block due to spec))

Sean-Der (approved):

Sean-Der:
Thanks for doing this @edaniels! Hopefully this means one less annoyed user :)

Were you doing tests that looked for leaked goroutines in a code base that used Pion?

edaniels:
Yep! It came up for Viam and I saw this years ago and we just ignored it. It's become more important as more sophisticated testing has ramped up.

edaniels:
I'll put up a backport later!

PR #2814, opened 2024-07-19
Avoid dropping packets immediately in mux

lactyy:
#### Description
As discussed in [this](https://gophers.slack.com/archives/CAK2124AG/p1703996245888079) thread on the Gophers Slack, internal/mux needs a pending queue for packets that would dropped in some situations. (e.g. ICE transport blocks until it is in a connected state, losing any DTLS packets received in the meantime)

Sorry, I accidentally closed the issue before. I've reopened an issue with fixes.

#### Reference issue
N/A

### DIFF
--- a/internal/mux/mux.go
+++ b/internal/mux/mux.go
@@ -15,8 +15,13 @@ import (
 	"github.com/pion/transport/v3/packetio"
 )
 
-// The maximum amount of data that can be buffered before returning errors.
-const maxBufferSize = 1000 * 1000 // 1MB
+const (
+	// The maximum amount of data that can be buffered before returning errors.
+	maxBufferSize = 1000 * 1000 // 1MB
+
+	// How many total pending packets can be cached
+	maxPendingPackets = 15
+)
 
 // Config collects the arguments to mux.Mux construction into
 // a single structure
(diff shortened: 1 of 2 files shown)


Sean-Der:
Thank you @lactyy! Great to see this finally fixed.

What do you think of spawning goroutine only if something needs to go in the pending queue? When the queue is drained the routine exits

Otherwise each PeerConnection gets another goroutine that will mostly go unused. I have users report doing 10k+ on a machine, so I worry what the impact will be.

Would it be possible to add a unit test also? You could spin up some fake dispatch and confirm that it gets the packet eventually.

lactyy:
Hello, @Sean-Der! Sorry for the late response, was this okay? I changed some of the code and added a unit test.

Sean-Der:
Hey @lactyy mind trying this again now that it is merged!

I instead process the queue w/e a new dispatch is added. I set a limit of 15 packets (but we can raise it!)

lactyy:
Hello @Sean-Der, I can confirm this has been fixed in the latest commit. thank you!

Sean-Der:
Thank you for opening the PR @lactyy! I am sure that bothered lots of people, but they never put the time in to help out :)

PR #2816, opened 2024-07-19
Properly handle non-media probes

Sean-Der:
libwebrtc has started sending media probes on an unannounced SSRC(0). This adds a special case to accept SSRC 0 and Read the RTP packets. This causes the TWCC reports to properly be generated.

### DIFF
--- a/interceptor_test.go
+++ b/interceptor_test.go
@@ -225,12 +225,59 @@ func Test_InterceptorRegistry_Build(t *testing.T) {
 		},
 	})
 
-	peerConnectionA, err := NewAPI(WithInterceptorRegistry(ir)).NewPeerConnection(Configuration{})
-	assert.NoError(t, err)
-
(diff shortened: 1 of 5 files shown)


adriancable:
Yay! For more info on when this can happen, it requires RTX to be enabled and it requires SetAllowProbeWithoutMediaPacket to be enabled (modules/pacing/bitrate_prober.cc in libwebrtc). In this case BWE probes get sent as soon as the ICE connection is established, before tracks start. The probes have SSRC 0.

This change was made recently in the Chromium repo, I am not sure if it is in production yet (or just in canary) but if it isn't happening now, it soon will so this PR is super important. Right now Pion ignores this 'probe track' which means TWCC/RTCP RRs do not get sent for it. This makes the sender think there is 100% packet loss on the probes, which causes it to adjust sender bandwidth to hit the defined bitrate floor - not very nice!

cnderrauber:
looks good!

cnderrauber (approved):
Looks good!

boks1971 (approved):
lgtm!

PR #2819, opened 2024-07-20
Added support for SRTP_NULL_HMAC_SHA1_80 cipher

### LINKED ISSUE #2273: Add interface to disable DTLS SRTP
## Summary
As the title describes, I would like to have an interface to disable DTLS SRTP, just like the `enable_dtls_srtp` defined in C++

## Motivation
To analyze RTP payload.

I'm working on rtmp 2 webrtc, the workflow is as follow：
- rtmp receive origin aac data
- decode aac to pcm
- slice pcm to 10ms pcm list (the opus encoder requires)
- encode pcm to opus
- pack opus as an rtp package
- send rtp package to TrackLocalStaticRTP

Unfortunately, the received sound on the other end is noisy.
I saved the opus data before sending it to the RTPSender, and it sounded fine.

So I wanted to use Wireshark to see what happened, but the rtp payload was encrypted.

By searching at code，I could'n find a way to turn it off.

@Sean-Der

sirzooro:
Added support for SRTP_NULL_HMAC_SHA1_80 protection profile (cipher). It is disabled by default. You need to use SettingEngine and set list of allowed SRTP protection profiles using its SetSRTPProtectionProfiles function called with dtls.SRTP_NULL_HMAC_SHA1_80 as a parameter. You need to do this for both pion peers. For non-pion ones you may need to enable it somewhere too, as NULL cipher is usually disabled for security reasons.

#### Reference issue
Fixes #2273

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -377,6 +377,8 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error {
 		t.srtpProtectionProfile = srtp.ProtectionProfileAeadAes256Gcm
 	case dtls.SRTP_AES128_CM_HMAC_SHA1_80:
 		t.srtpProtectionProfile = srtp.ProtectionProfileAes128CmHmacSha1_80
+	case dtls.SRTP_NULL_HMAC_SHA1_80:
+		t.srtpProtectionProfile = srtp.ProtectionProfileNullHmacSha1_80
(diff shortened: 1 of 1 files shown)


Sean-Der (approved):
Amazing, thank you so much @sirzooro this was a lot of work I appreciate it

PR #2822, opened 2024-07-22
Bump ice to v2.3.30

edaniels:


edaniels:
Will be fixed by https://github.com/pion/transport/pull/304

PR #2825, opened 2024-07-22
Close unhandled ssrc rtp/rtcp simulcast streams

edaniels:
Not doing this can leak packetio.Buffers


edaniels:
There was a test failure around probing but I'm not sure how this would make it ***worse***.

edaniels:
Working on fixing this test

edaniels:
for anyone that sees this. it didn't make sense to have the logic in this PR because we would accidentally close rtp/rtcp streams that were non-multicast

PR #2826, opened 2024-07-22
Close unhandled rtcp simulcast streams

edaniels:
handleIncomingSSRC will call streamsForSSRC which opens rtp/rtcp streams that if unhandled can be leaked resources. Now we will proactively open them before calling handleIncomingSSRC and close then later. In the future it would be better to do this inside handleIncomingSSRC to protect other callers.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1670,25 +1670,41 @@ func (pc *PeerConnection) undeclaredRTPMediaProcessor() {
 			return
 		}
 
-		stream, ssrc, err := srtpSession.AcceptStream()
+		srtcpSession, err := pc.dtlsTransport.getSRTCPSession()
+		if err != nil {
+			pc.log.Warnf("undeclaredMediaProcessor failed to open SrtcpSession: %v", err)
+			return
+		}
+
+		srtpReadStream, ssrc, err := srtpSession.AcceptStream()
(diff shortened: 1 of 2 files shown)


edaniels:
I noticed the srtcp leaking packetio.buffer's when I added a goroutine to them in https://github.com/pion/transport/pull/304

peerconnection.go
```diff
@@ -1670,26 +1683,42 @@ func (pc *PeerConnection) undeclaredRTPMediaProcessor() {
...
+			if err = srtpReadStream.Close(); err != nil {
 				pc.log.Warnf("Failed to close RTP stream %v", err)
 			}
+			if err = srtcpReadStream.Close(); err != nil {
+				pc.log.Warnf("Failed to close RTCP stream %v", err)
+			}
 			continue
 		}
 
+		pc.dtlsTransport.storeSimulcastStream(srtpReadStream, srtcpReadStream)
```
edaniels, line 1715: an observation while working on this: we hang onto these for the lifetime of the PC which could increase memory over time. It would be better to know that these are strictly 1. not valid when expecting simulcast and 2. not valid ssrcs when we are non-simulcast

cnderrauber (approved):
Looks good!

PR #2833, opened 2024-07-26
Fix our-of-order twcc fb cause by rtx blocked

### LINKED ISSUE #2830: Target Bitrate As Reported by Chrome is slow to adjust when ConfigureTWCCSender is Being Used
### Your environment.
  - Go: 1.22.2
  - Version: Pion Webrtc v3.2.49/v4.0.0-beta.24
  - Browser: Chrome  126.0.6478.126
  - Other Information - Ubuntu 22.04

### What did you do?
Used the Broadcast Example, created a new RegisterDefaultInterceptors without webrtc.ConfigureTWCCSender
```go
func RegisterDefaultInterceptors(mediaEngine *webrtc.MediaEngine, interceptorRegistry *interceptor.Registry) error {
	if err := webrtc.ConfigureNack(mediaEngine, interceptorRegistry); err != nil {
		return err
	}
	if err := webrtc.ConfigureRTCPReports(interceptorRegistry); err != nil {
		return err
	}
	if err := webrtc.ConfigureSimulcastExtensionHeaders(mediaEngine); err != nil {
		return err
	}
	return nil
}
```
replaced 
```go
// Use the default set of Interceptors
	if err := webrtc.RegisterDefaultInterceptors(m, i); err != nil {
		panic(err)
	}
```

### What did you expect?
Bandwidth Ramp-up should be fast with webrtc.ConfigureTWCCSender

### What happened?
looks like introduced in v3.2.25
when using webrtc.ConfigureTWCCSender the rampup was slow and in a sea-saw pattern and got stuck at 1 mbps
when not using it it was still slow but went upto 2.5 mbps, please see the Target Bitrate as reported on the graph

With ConfigureTWCCSender , Slow, Seesaw and stuck at 1mbps
With ConfigureTWCCSender , Seesaw and stuck at 1mbps

Without ConfigureTWCCSender, slow but goes upto 2.5mbps
Screenshot from 2024-07-24 10-30-41

With ConfigureTWCCSender in v3.2.24
Screenshot from 2024-07-24 15-31-34

cnderrauber:
Fix #2830. The TrackRemote.Read could block in
readRTP if the buffer is empty then rtx packets
arrival before next media rtp packet will be
readed after the next media rtp packet and cause
out-of-order fb and mess up remote peer's
bandwidth estimation.

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -433,7 +433,7 @@ func (r *RTPReceiver) receiveForRtx(ssrc SSRC, rsid string, streamInfo *intercep
 	track.repairInterceptor = rtpInterceptor
 	track.repairRtcpReadStream = rtcpReadStream
 	track.repairRtcpInterceptor = rtcpInterceptor
-	track.repairStreamChannel = make(chan rtxPacketWithAttributes)
+	track.repairStreamChannel = make(chan rtxPacketWithAttributes, 50)
 
 	go func() {
 		for {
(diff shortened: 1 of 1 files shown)


Sean-Der:
Thank you so much @cnderrauber!

When I hear that this fixes things I will merge and tag right away :)

Sean-Der (approved):

PR #2836, opened 2024-07-27
Change time.Sleep to time.Ticker in insertable streams example

### LINKED ISSUE #2835: Inconsistency in some of the examples
## Information

Hello, 

I was looking at the examples in the WebRTC/examples folder and came across the following comment in the "play-from-disk" example:

(line 196-198)
```
// It is important to use a time.Ticker instead of time.Sleep because
// * avoids accumulating skew, just calling time.Sleep didn't compensate for the time spent parsing the data
// * works around latency issues with Sleep (see https://github.com/golang/go/issues/44343)
```

However, the "insertable streams" example seems to use sleep instead of the ticker (I think at least):

(85-108)
```
// Send our video file frame at a time. Pace our sending so we send it at the same speed it should be played back as.
// This isn't required since the video is timestamped, but we will such much higher loss if we send all at once.
sleepTime := time.Millisecond * 
time.Duration((float32(header.TimebaseNumerator)/float32(header.TimebaseDenominator))*1000)
for {
        frame, _, ivfErr := ivf.ParseNextFrame()
	if errors.Is(ivfErr, io.EOF) {
		fmt.Printf("All frames parsed and sent")
		os.Exit(0)
	}

	if ivfErr != nil {
		panic(ivfErr)
	}

	// Encrypt video using XOR Cipher
	for i := range frame {
		frame[i] ^= cipherKey
	}

	time.Sleep(sleepTime)
	if ivfErr = videoTrack.WriteSample(media.Sample{Data: frame, Duration: time.Second}); ivfErr != nil {
		panic(ivfErr)
	}
}
```

Is it supposed to be a sleep here?

AfonsoVilalonga:
#### Description
Hello, 

While looking at the WebRTC examples, I came across the following comment in the "play-from-disk" example:

(line 196-198)
```
// It is important to use a time.Ticker instead of time.Sleep because
// * avoids accumulating skew, just calling time.Sleep didn't compensate for the time spent parsing the data
// * works around latency issues with Sleep (see https://github.com/golang/go/issues/44343)
```

However, the "insertable streams" example still uses time.Sleep. I changed the code of the example to use time.Ticker like the "play-from-disk" example.

#### Reference issue
Fixes #2835

### DIFF
--- a/examples/insertable-streams/main.go
+++ b/examples/insertable-streams/main.go
@@ -84,8 +84,12 @@ func main() {
 
 		// Send our video file frame at a time. Pace our sending so we send it at the same speed it should be played back as.
 		// This isn't required since the video is timestamped, but we will such much higher loss if we send all at once.
-		sleepTime := time.Millisecond * time.Duration((float32(header.TimebaseNumerator)/float32(header.TimebaseDenominator))*1000)
-		for {
+		//
+		// It is important to use a time.Ticker instead of time.Sleep because
+		// * avoids accumulating skew, just calling time.Sleep didn't compensate for the time spent parsing the data
+		// * works around latency issues with Sleep (see https://github.com/golang/go/issues/44343)
+		ticker := time.NewTicker(time.Millisecond * time.Duration((float32(header.TimebaseNumerator)/float32(header.TimebaseDenominator))*1000))
+		for ; true; <-ticker.C {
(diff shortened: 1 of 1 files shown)


Sean-Der:
Works great @AfonsoVilalonga 

Thank you so much! If you are interested I hope you stay involved with the project, always looking for more contributors :)

PR #2839, opened 2024-08-01
Avoid leaking ticker

jech:
A ticker needs to be explicitly stopped when it's no longer
useful.


Sean-Der:
We have this anti-pattern in a few examples :/

Mind updating the other examples also? If you don't have the time I can grab it.

Hopefully this bad code hasn't ended up in too many other code bases :/

jech:
Superseded by #2843.

PR #2840, opened 2024-08-01
RID order guaranteed by avoiding iteration over a map

### LINKED ISSUE #2838: Undefined iteration order when parsing SDP RID ids
Hi,

I've been observing a strange behavior in my WebRTC application, due to the fact that some times the simulcast layers are provided by Pion in a different order than the one in SDP messages.

After quite a bunch of debugging and printfs, ended up finding this line:

https://github.com/pion/webrtc/blob/2364173cedda65030a8a5f2110416387deb0083e/sdp.go#L197-L207

`getRids` returned a `map[string]*simulcastRid`, and then it gets traversed so the RIDs get appended into `simulcastTrack.rids`.

However, a big problem that I think is what I'm observing, is that Go does not guarantee any ordering on map traversal:

> 3. The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

[Source](https://go.dev/ref/spec#For_range)

**Concrete example**

This is the relevant part of an SDP Offer that is received:

```
a=rid:0 send
a=rid:1 send
a=rid:2 send
a=simulcast:send 0;1;2
a=sendonly
```

If then we add a printf of `incomingTrack.rids` right before this line: https://github.com/pion/webrtc/blob/v3.2.50/peerconnection.go#L1414
What *some* times we get is this: `[1 2 0]`.

So I believe that to respect the same ordering as given in the SDP's `a=simulcast` line, a different method of construction should be used that avoids unspecified orderings.

What do you think?

j1elo:
Map iteration order is not guaranteed by Go, so it's an error to iterate over a map in places where maintaining the same ordering is important.

This change replaces the map of simulcastRid{} with an array of the same type. The simulcastRid{} type is extended to hold the rid-id which previously was used as the key in the map.

Accesses to the map are replaced with range loops to find the desired rid-id for each case.

Default code format applied.
Unit tests passed.

Fixes #2838

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2520,7 +2520,7 @@ func (pc *PeerConnection) generateMatchedSDP(transceivers []*RTPTransceiver, use
 			mediaTransceivers := []*RTPTransceiver{t}
 
 			extensions, _ := rtpExtensionsFromMediaDescription(media)
-			mediaSections = append(mediaSections, mediaSection{id: midValue, transceivers: mediaTransceivers, matchExtensions: extensions, ridMap: getRids(media)})
+			mediaSections = append(mediaSections, mediaSection{id: midValue, transceivers: mediaTransceivers, matchExtensions: extensions, rids: getRids(media)})
 		}
 	}
 
--- a/sdp.go
+++ b/sdp.go
@@ -202,8 +202,8 @@ func trackDetailsFromSDP(log logging.LeveledLogger, s *sdp.SessionDescription) (
 				id:       trackID,
 				rids:     []string{},
 			}
-			for rid := range rids {
-				simulcastTrack.rids = append(simulcastTrack.rids, rid)
+			for _, rid := range rids {
+				simulcastTrack.rids = append(simulcastTrack.rids, rid.id)
 			}
 
 			tracksInMediaSection = []trackDetails{simulcastTrack}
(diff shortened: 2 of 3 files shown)


j1elo:
A comment to the side: I'm thinking that what the original code did (and it's still done by the changed code) is to present the RIDs in the same order as the `a=rid` lines are found, not strictly speaking in the order of the `a=simulcast` attribute.

However, I'm not sure if that'd be an actual problem. I know that theoretically RIDs should be treated or given in the order found in `a=simulcast`... but in practice, does this matter? I've yet to see an instance where the `a=rid` lines come in a different order than the order of `a=simulcast`.

Practical SDP Offer example for readers who might be confused:
```
m=video 9 UDP/TLS/RTP/SAVPF ...
a=rid:layer0 send
a=rid:layer1 send
a=rid:layer2 send
a=simulcast:send layer0;layer1;layer2
```

Sean-Der (approved):

j1elo:
Great! thanks for the lint fixes, and for the work merging this.
A couple questions:
* Is the linter a default Go tool? I'd love to read more about it and see how to run it locally to avoid those naming mistakes in the future.
* Will this change be backported to `v3`?

Sean-Der:
https://github.com/golangci/golangci-lint isn't a default Go tool, but is pretty popular! I don't know how I ended up on it.

I also added you to the org. Feel free to push tags and make branches w/e you need :)

PR #2847, opened 2024-08-02
Add PeerConnection.GracefulClose

edaniels:
Take 2: This is a more thorough attempt at closing gracefully. If anything, every path that graceful touches becomes opt-in, so it shouldn't affect any existing users and it makes it safe for backporting.

GracefulClose/Stop added to:
- PeerConnection
- DataChannel
- ICEGatherer
- ICETransport
- operations


edaniels:
First test run had https://github.com/pion/webrtc/issues/2848 but verified it happens on `master`

peerconnection.go
```diff
@@ -2092,13 +2096,41 @@ func (pc *PeerConnection) writeRTCP(pkts []rtcp.Packet, _ interceptor.Attributes
 	return pc.dtlsTransport.WriteRTCP(pkts)
 }
 
-// Close ends the PeerConnection
+type shouldGracefullyClose bool
+
+const (
+	shouldGracefullyCloseNo  shouldGracefullyClose = false
+	shouldGracefullyCloseYes shouldGracefullyClose = false
```
Sean-Der, line 2103: should this be true?

Sean-Der: It's a matter of style, but this pattern is easier/more popular IMO

```
return pc.close(false /* shouldGracefullyClose */)
```

datachannel.go
```diff
@@ -40,6 +40,8 @@ type DataChannel struct {
 	readyState                 atomic.Value // DataChannelState
 	bufferedAmountLowThreshold uint64
 	detachCalled               bool
+	readLoopActive             chan struct{}
+	isUserClosed               bool
```
Sean-Der, line 44: `isGracefulClose`? I see this set via `Close` or `GracefulClose` only

peerconnection.go
edaniels: yep and sure good idea

datachannel.go
edaniels: good idea, more specific :)

internal/mux/mux.go
```diff
@@ -129,6 +129,10 @@ func (m *Mux) readLoop() {
 		}
 
 		if err = m.dispatch(buf[:n]); err != nil {
+			if errors.Is(err, io.ErrClosedPipe) {
```
edaniels, line 132: I don't see a reason to have this spuriously log when we are the ones closing the packetio.Buffer.

Sean-Der (approved):

PR #2853, opened 2024-08-04
Fix: IVF writer writing invalid timestamp value into frame headers

### LINKED ISSUE #2384: IVF writer writing invalid timestamp value into frame headers?
Hey maintainers, thank you for this awesome work.

While working a little bit with some VP8-based streams, I have noticed something interesting related to IVF writers. Is it possible that the value written for the IVF frame timestamp is incorrect?

- what's being written is a sequence number: https://github.com/pion/webrtc/blob/master/pkg/media/ivfwriter/ivfwriter.go#L112
- the reader correctly expects it to be a timestamp: https://github.com/pion/webrtc/blob/master/pkg/media/ivfreader/ivfreader.go#L97,

And, of course, Duck IVF says that 8 bytes at offset 4 are 64-bit presentation timestamp: https://wiki.multimedia.cx/index.php/Duck_IVF.

Shouldn't the writer write the RTP header timestamp value into the IVF frame header?

I would propose the following:

- https://github.com/radekg/boos/pull/3/files#diff-652bfa3efcbaa4c9f2876f1b72e4c566f6539428d6ac738050aa074370b192f1R112-R123

with the following call-site changes:

- https://github.com/radekg/boos/pull/3/files#diff-652bfa3efcbaa4c9f2876f1b72e4c566f6539428d6ac738050aa074370b192f1R156
- https://github.com/radekg/boos/pull/3/files#diff-652bfa3efcbaa4c9f2876f1b72e4c566f6539428d6ac738050aa074370b192f1R172

With these changes, I can handle variable stream framerate when reading frames.

I can provide a PR, if this change is acceptable.

### Your environment.
  - Version: *Release or SHA*
  - Browser: *include version*
  - Other Information - *stacktraces, related issues, suggestions how to fix, links for us to have context*

radekg:
#### Description

This PR fixes the incorrect header timestamp written into the frame headers.

#### Reference issue

Fixes #2384

### DIFF
--- a/pkg/media/ivfwriter/ivfwriter.go
+++ b/pkg/media/ivfwriter/ivfwriter.go
@@ -109,10 +109,10 @@ func (i *IVFWriter) writeHeader() error {
 	return err
 }
 
-func (i *IVFWriter) writeFrame(frame []byte) error {
+func (i *IVFWriter) writeFrame(frame []byte, timestamp uint64) error {
 	frameHeader := make([]byte, 12)
 	binary.LittleEndian.PutUint32(frameHeader[0:], uint32(len(frame))) // Frame length
-	binary.LittleEndian.PutUint64(frameHeader[4:], i.count)            // PTS
+	binary.LittleEndian.PutUint64(frameHeader[4:], timestamp)          // PTS
 	i.count++
 
 	if _, err := i.ioWriter.Write(frameHeader); err != nil {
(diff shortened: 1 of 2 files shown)


edaniels:
Thanks @radekg

edaniels (approved):
LGTM, ty for the fix and added test

edaniels:
@radekg can you fix up the error from the meta lint (commit message length)?

PR #2858, opened 2024-08-08
Provide SCTP Association OnClose callback

sukunrt:
The only way to detect a remote connection close is to check the state of the SCTP transport.
See:
https://github.com/w3c/webrtc-pc/issues/1799#issuecomment-373394934

>In case it wasn't clear, I'm arguing for an exception for data
channels, saying pc1.close() should fire close on both ends of a data channel, because A) that is useful, and the only way to detect remote closing of a peer connection, and B) it's a workaround for the fact that closing of peer connections isn't advertised to the other end.

Depending on datachannel close to notify us requires out of band protocol agreement as datachannels can be closed by the remote for other reasons. Having access to the SCTP layer we can provide an accurate SCTP Assocation closed event.

### DIFF
--- a/sctptransport.go
+++ b/sctptransport.go
@@ -45,6 +45,7 @@ type SCTPTransport struct {
 	// OnStateChange  func()
 
 	onErrorHandler func(error)
+	onCloseHandler func(error)
 
 	sctpAssociation            *sctp.Association
 	onDataChannelHandler       func(*DataChannel)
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):
LGTM! Merge w/e CI passes

thanks for doing this @sukunrt

sukunrt:
I removed the shadowing.

PR #2885, opened 2024-08-23
add shutdown to close

craymond12:
#### Reference issue
Fixes #...

### DIFF
--- a/sctptransport.go
+++ b/sctptransport.go
@@ -7,6 +7,7 @@
 package webrtc
 
 import (
+	"context"
 	"errors"
 	"io"
 	"math"
@@ -153,6 +154,10 @@ func (r *SCTPTransport) Stop() error {
 	if r.sctpAssociation == nil {
 		return nil
 	}
+
+	var ctx, _ = context.WithTimeout(context.Background(), 10*time.Second)
+	r.sctpAssociation.Shutdown(ctx)
+
 	err := r.sctpAssociation.Close()
 	if err != nil {
 		return err


Sean-Der:
@craymond12 Do we not do a graceful close currently without this?

craymond12 (commented):
all good

craymond12:
> @craymond12 Do we not do a graceful close currently without this?

sorry, i was supposed to push on my fork to test things
Currently when we close the webrtc connection we do not send the shutdown message

This often cause the webrtc stream to hang until timeout on the producer side when the consumer leave
Chrome do send it

Sean-Der:
@craymond12 we can run the tests here! If you open the PR again I can run the tests and if they pass we can merge things.

Would it be possible to write a test? Otherwise someone might undo your fix accidentally.

craymond12:
> @craymond12 we can run the tests here! If you open the PR again I can run the tests and if they pass we can merge things.
> 
> Would it be possible to write a test? Otherwise someone might undo your fix accidentally.

this was kind of hackish, would have to think more about the context sent to the shutdown

craymond12:
> @craymond12 we can run the tests here! If you open the PR again I can run the tests and if they pass we can merge things.
> 
> Would it be possible to write a test? Otherwise someone might undo your fix accidentally.

the root of the issue is a lot of onvif camera support 2 to 4 streams, Having the webrtc hang in the camera for around 30 seconds can be an issue

I intended to explore this behavior in a fork to avoid creating some noise here
But once satisfied I will publish a PR here

craymond12:
> @craymond12 we can run the tests here! If you open the PR again I can run the tests and if they pass we can merge things.
> 
> Would it be possible to write a test? Otherwise someone might undo your fix accidentally.

I managed to do some test with some real Axis device and it work

I am thinking of having a new peerConnection Close method to which we will be able to pass down the context for cancellation
I will be off for 1 week and in my return i will update the pr and look at the tests

edaniels:
> > @craymond12 we can run the tests here! If you open the PR again I can run the tests and if they pass we can merge things.
> > Would it be possible to write a test? Otherwise someone might undo your fix accidentally.
> 
> I managed to do some test with some real Axis device and it work
> 
> I am thinking of having a new peerConnection Close method to which we will be able to pass down the context for cancellation I will be off for 1 week and in my return i will update the pr and look at the tests

Given how new the method is, you can probably put the context into GracefulClose

PR #2897, opened 2024-09-04
prevent ICETransport start/stop deadlock

edaniels:


icetransport.go
```diff
@@ -200,28 +204,31 @@ func (t *ICETransport) GracefulStop() error {
...
 	}
 
+	// mux and gatherer can only be set when ICETransport.State != Closed.
+	mux := t.mux
+	gatherer := t.gatherer
+	t.lock.Unlock()
+
 	if t.mux != nil {
 		var closeErrs []error
-		if shouldGracefullyClose && t.gatherer != nil {
```
edaniels, line 213: I looked at these closes and they should be safe to call concurrently. If they aren't, we'll see some other problems crop up quickly. But I haven't found those to happen in any concurrent PC close test runs.

icetransport.go
```diff
@@ -157,6 +157,10 @@ func (t *ICETransport) Start(gatherer *ICEGatherer, params ICEParameters, role *
 		return err
```
benjirewis, line 157: Can't comment on the line but it looks like you're `Lock`ing the `t.lock` above and never unlocking; is that on purpose?

edaniels: yeah there's a defer unlock above. not sure how i feel about that pattern!

benjirewis: I don't like it haha, but since it was pre-existing and I have basically no context here will approve.

benjirewis (approved):

PR #2901, opened 2024-09-09
Reduce Mux logging on Close

sukunrt:
This removes warning logs like:
```
Warning: mux: no endpoint for packet starting with 21
```
when closing the connection. 

Replace RWMutex with Mutex because there's only 1 goroutine reading.


edaniels (approved):
Looks great

PR #2912, opened 2024-10-02
fix: fix slice init length

rusttech:
#### Description

The intention here should be to initialize a slice with a capacity of len(m.pendingPackets)  rather than initializing the length of this slice.

#### Reference issue
Fixes #...


Sean-Der:
Hi @rusttech 

Unfortunately this change doesn’t compile. Mind checking it again, thank you!

PR #2923, opened 2024-10-07
Add ice transport api to get selected pair stats

boks1971:
In use cases like SFU, it is useful to get just the selected candidate pair stats to have access to current RTT on the peer connection. The standard has a way to do `GetSelectedCandidatePair` on `ICETransport`, but does not have a way to get stats of that pair.

Although not in standard, adding a method to `ICETransport` to get selected candidate pair along similar lines of above method.

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -308,46 +308,12 @@ func (g *ICEGatherer) collectStats(collector *statsReportCollector) {
 		for _, candidatePairStats := range agent.GetCandidatePairsStats() {
 			collector.Collecting()
 
-			state, err := toStatsICECandidatePairState(candidatePairStats.State)
+			stats, err := toICECandidatePairStats(candidatePairStats)
 			if err != nil {
 				g.log.Error(err.Error())
+				continue
 			}
 
-			pairID := newICECandidatePairStatsID(candidatePairStats.LocalCandidateID,
-				candidatePairStats.RemoteCandidateID)
-
-			stats := ICECandidatePairStats{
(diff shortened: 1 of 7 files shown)


icegatherer.go
```diff
@@ -308,46 +308,12 @@ func (g *ICEGatherer) collectStats(collector *statsReportCollector) {
 		for _, candidatePairStats := range agent.GetCandidatePairsStats() {
 			collector.Collecting()
 
-			state, err := toStatsICECandidatePairState(candidatePairStats.State)
+			stats, err := toICECandidatePairStats(candidatePairStats)
 			if err != nil {
 				g.log.Error(err.Error())
+				continue
```
boks1971, line 314: A slight change in behaviour here. Consolidate the conversion from `ice.CandidatePairStats` -> `webrtc.ICECandidatePairStats` in `stats.go` so that it can be use both here and in `getSelectedCandidatePairStats()` 👇 . In fact, linter flagged it for repeated code 🎉 .

Previously, code continued to add stats after just logging the error. Now, it does not. The consolidation code in `stats.go` returns an empty stats if there is an error here.

I think the state error should never happen. There is also a comment in code like
```
    default:                                                                                                                                                    
        // NOTE: this should never happen[tm]                                                                                                                   
        err := fmt.Errorf("%w: %s", errStatsICECandidateStateInvalid, state.String())                                                                           
        return StatsICECandidatePairState("Unknown"), err
```

So, just skipping adding the stat on error.

boks1971:
Thank you @Sean-Der !!!

PR #2931, opened 2024-10-14
feat: allow media engine to register custom payloader on mimetype

### LINKED ISSUE #2928: Enable user to register custom codec implementation
## Summary

Allow user to register custom codec implementation other than provided one.

## Motivation

Right now we have a way to register custom codec via MediaEngine.RegisterCodec, but it'll throw "the requested codec does not have a payloader" error when the codec implementation is not found. I would like to bind a custom codec to a particular mime type. The matching of codec is private and couldn't be modified by user

```
func payloaderForCodec(codec RTPCodecCapability) (rtp.Payloader, error) {
	switch strings.ToLower(codec.MimeType) {
	case strings.ToLower(MimeTypeH264):
		return &codecs.H264Payloader{}, nil
	case strings.ToLower(MimeTypeOpus):
		return &codecs.OpusPayloader{}, nil
	case strings.ToLower(MimeTypeVP8):
		return &codecs.VP8Payloader{
			EnablePictureID: true,
		}, nil
	case strings.ToLower(MimeTypeVP9):
		return &codecs.VP9Payloader{}, nil
	case strings.ToLower(MimeTypeAV1):
		return &codecs.AV1Payloader{}, nil
	case strings.ToLower(MimeTypeG722):
		return &codecs.G722Payloader{}, nil
	case strings.ToLower(MimeTypePCMU), strings.ToLower(MimeTypePCMA):
		return &codecs.G711Payloader{}, nil
	default:
		return nil, ErrNoPayloaderForCodec
	}
}
```
In the above code, pion should try to match a custom user implemented codec before throwing an error

LeeTeng2001:
#### Description

Allow media engine to register custom payloader depending on the mime type

#### Reference issue
Fixes #2928

### DIFF
--- a/track_local_static.go
+++ b/track_local_static.go
@@ -31,6 +31,7 @@ type TrackLocalStaticRTP struct {
 	mu                sync.RWMutex
 	bindings          []trackBinding
 	codec             RTPCodecCapability
+	payloader         func(RTPCodecCapability) (rtp.Payloader, error)
 	id, rid, streamID string
 }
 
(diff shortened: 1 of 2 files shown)


Sean-Der:
Looks great @LeeTeng2001!

* Can you put it on the `MediaEngine`. That way users can configure per PeerConnection (or globally) if they want
* Would it be possible to assert that `Payload` actually gets called?

LeeTeng2001:
@Sean-Der just to clarify, should I put it in both MediaEngine and global scope so that we ca either register it through global scope and media engine scope?

LeeTeng2001:
hi, I've updated a pr with payload assert and  MediaEngine custom payloader scope

Sean-Der (approved):

PR #2933, opened 2024-10-15
Add SetFireOnTrackBeforeFirstRTP

edaniels:
This will allow getting OnTrack events before RTP packets are received (non-simulcast)

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1245,6 +1245,10 @@ func (pc *PeerConnection) startReceiver(incoming trackDetails, receiver *RTPRece
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


edaniels:
> @edaniels Could we put it behind SettingEngine?

Yeah let's do that. Making Codec/PayloadType blocking feels bad as well

edaniels:
@Sean-Der how about this?

Sean-Der (approved):
beautiful :)

Sean-Der:
@edaniels sorry that didn't happen in `v4` I think it's a good change!

`v5` for sure. I should start a list somewhere .

Issue #2403, opened 2023-02-06
SRTP does not include padding bytes


bixycler:
### What did you do?
Establish a peer connection to send a track then call `webrtc.TrackLocalStaticRTP.WriteRTP()` to write RTP packets with **non-empty payload and payload padding** (`PaddingSize > 1`), e.g. with packets like this:
```
rtp.Packet{
	Header: rtp.Header{
		Version: 2,
		Padding: true,
	},
	PaddingSize: 4,
	Payload: []byte{0x01, 0x02, 0x03},
}
```

### What did you expect?
The sent ciphertext (SRTP) should have corresponding plaintext containing trailing padding bytes, e.g. of the plaintext like this:
```
[160 0 0 0 0 0 0 0 0 0 0 0 1 2 3 0 0 0 4]
// padding bytes ----------------^-^-^ ^----- PaddingSize
```

### What happened?
The sent ciphertext (SRTP) has corresponding plaintext with `Header.Padding == true` but no padding bytes, e.g. of the plaintext like this:
```
[160 0 0 0 0 0 0 0 0 0 0 0 1 2 3]
```

### Causes of the problem:
In `webrtc.TrackLocalStaticRTP.writeRTP(p *rtp.Packet)`, the packet's header and payload are _passed as is **without marshaling**_, via `b.writeStream.WriteRTP(&p.Header, p.Payload)`, to `TrackLocalWriter`, then to `interceptor`, then to `srtp.SessionSRTP.writeRTP(header *rtp.Header, payload []byte)`, and finally to `srtp.Context.encryptRTP(dst []byte, header *rtp.Header, payload []byte)`. Currently, there's only one way to generate padding bytes, that is to marshal RTP packet via `rtp.Packet.Marshal()`(See [rtp#155](https://github.com/pion/rtp/pull/155/commits/762664cff63b7431b93e69a1cd2cf76a05c8040c#diff-0f1ddc9367ddb7a72b6dfae6035cba60b8d821e7a82371ab0115ce5ad1eaa012)). 

### Suggested solutions:
Either call `rtp.Packet.Marshal()` or add the padding appending codes to update the payload somewhere before passing it to `srtp`.

alexpokotilo:
I don't use webrtc.TrackLocalStaticRTP, but for webrtc.TrackLocalStaticSample I ended up adding padding-only packets and packets with payload and padding into rtp.Payloader here https://github.com/alexpokotilo/rtp/tree/packetizer-padding-support

[...]

PR #2957, opened 2024-11-24
fixed bug in examples/rtp-forwarder where packet state is not properl…

strangesast:
fixes corrupt RTP packet issue related to memory initialization/reset in examples/rtp-forwarder that causes frames to be distorted

### DIFF
--- a/examples/rtp-forwarder/main.go
+++ b/examples/rtp-forwarder/main.go
@@ -143,7 +143,6 @@ func main() {
 		}
 
 		b := make([]byte, 1500)
-		rtpPacket := &rtp.Packet{}
 		for {
 			// Read
 			n, _, readErr := track.Read(b)
@@ -152,6 +151,7 @@ func main() {
 			}
 
 			// Unmarshal the packet and update the PayloadType
+			rtpPacket := &rtp.Packet{}
 			if err = rtpPacket.Unmarshal(b[:n]); err != nil {
 				panic(err)
 			}


Sean-Der:
@strangesast are you able to reproduce this reliably? 

`Unmarshal` should be safe to use like this! Possibly `Padding` isn't cleared properly? We might need a deeper fix in `pion/rtp`

strangesast:
yes I get frames that are distorted like this with VP8 encoding.  I had tried h.264 and get a different flavor of distortion.
Screenshot 2024-11-26 at 9 33 56 AM

when testing I noticed that I could get the demo to work with TrackRemote.ReadRTP but realized that was just a convenience wrapper around the same logic

it looks like there's an issue with how rtp.Packet.PaddingSize is set by Unmarshal.  or at least that's the only value that differs between these approaches.

it's almost certainly related to [this](https://github.com/pion/webrtc/issues/2403) and [this](https://github.com/pion/rtp/pull/155#issuecomment-1455042108)

Sean-Der:
Ok great, I think we should fix that instead. Users have been copying/using this code so we should fix it deeper.

Would you mind making the change [here](https://github.com/pion/rtp/blob/master/packet.go#L210) to zero padding size? I believe that's the issue

strangesast:
not at all.  I've created [this pr](https://github.com/pion/rtp/pull/284) over there

strangesast:
the PR in pion/rtp should address the issue brought up here.  thanks for the insight and help with the fix

PR #2958, opened 2024-11-24
refactor: clarify filter func return val meaning

WofWca:
#### Description

This should it easier to understand whether you're supposed to return `true` or `false`.
I know it's pretty much the standard that `true` is for `keep`, but this is a simple change that should make developers think less.

The same change probably should be implemented for pion/ice, where these functions are actually invoked.

#### Reference issue
No


WofWca:
CI failures seem unrelated?

Sean-Der:
Great change, thank you @WofWca !

Sean-Der (approved):

PR #2959, opened 2024-11-26
jitterbuffer: Use jitterbuffer in SampleBuilder 

thatsnotright:
#### Reference issue
Fixes #...

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/pion/interceptor/pkg/jitterbuffer"
 	"github.com/pion/rtp"
 	"github.com/pion/webrtc/v4/pkg/media"
 )
@@ -16,7 +17,7 @@ import (
 type SampleBuilder struct {
 	maxLate          uint16 // how many packets to wait until we get a valid Sample
 	maxLateTimestamp uint32 // max timestamp between old and new timestamps before dropping packets
-	buffer           [math.MaxUint16 + 1]*rtp.Packet
+	buffer           *jitterbuffer.JitterBuffer
 	preparedSamples  [math.MaxUint16 + 1]*media.Sample
 
 	// Interface that allows us to take RTP packets to samples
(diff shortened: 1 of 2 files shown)


jech:
The current JitterBuffer induces a fixed delay of 50 packets.  I don't think that it's a good idea to use it in SampleBuilder until that is fixed.

thatsnotright:
> The current JitterBuffer induces a fixed delay of 50 packets. I don't think that it's a good idea to use it in SampleBuilder until that is fixed.

@jech What would you propose the interface look like, or mechanism be for adjusting that delay?

jech:
> What would you propose the interface look like, or mechanism be for adjusting that delay?

Before we discuss the interface, we need to understand what is required.

For an interactive display, the jitter buffer delay should be set to something like smoothed_RTT + 4*variance(RTT).  The RTT and its variance can be recomputed at various times: when we receive a receiver report, when we receive TWCC feedback, or when we receive an ICE consent.  Yeah, it's a mess.

What is more, the delay might be influenced by external factors, such as when a receiver is performing audio-video sync.

For the samplebuilder, the amount of delay will depend on the application: it will be as above for low-latency applications, but it can be much larger than that for semi-interactive applications (such as streaming).  For non-interactive applications, such as saving to disk, there is no need for a fixed delay, the samplebuilder can add as much delay as required to avoid packet loss.

In summary: there are many contradictory requirements, and meeting them all without having dozens of special cases will require good taste and careful thought.  I am not able to propose a fully general interface at this time.

Sean-Der:
@jech This is the behavior we have today (unfortunately) 

I agree we should improve it! We shouldn't block merging this for that though.

Correct me if I'm am wrong @thatsnotright, but this causes 0 behavior change. We are just trying to push as much 'RTP code' into pion/interceptor. That lets people doing non-WebRTC things use it also.

thatsnotright:
> Correct me if I'm am wrong @thatsnotright, but this causes 0 behavior change.

Aside from potentially a little more delay there shouldn't be, it was designed to strictly replace how the current SampleBuilder buffered packets. It should reduce the amount of memory used.

jech:
I see, thanks.

I hold no opinion on whether this patch is desirable. I do have some 
misgivings about including in Pion an API that is not useful in its 
current state without as much as a warning in the comments, but that ship 
has sailed.

PR #2960, opened 2024-11-28
Fix nil rtcpInterceptor on track.Bind

cnderrauber:
Create rtcpInterceptor before bind called.

#### Reference issue
Fixes #...


boks1971 (approved):
lgtm!

PR #2961, opened 2024-11-29
Add option to disable close by dtls

cnderrauber:
Close peerconnection on DTLS.CloseNotify
could break ice restart with dtls restart,
when the dtls finger-print changed, the
browser could teardown the old dtlstransport
and establish new one then pion could close
the peerconnection and restart failed. So
browser don't do this and spec also don't
say peerconnection should close when dtls
is closed.
Remove this feature might break applications 
already relies on this behavior for quick close 
so add a SettingEngine option to disable it. 

#### Reference issue
Fixes #...

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2417,7 +2417,7 @@ func (pc *PeerConnection) startTransports(iceRole ICERole, dtlsRole DTLSRole, re
 	}
 
 	pc.dtlsTransport.internalOnCloseHandler = func() {
-		if pc.isClosed.get() {
+		if pc.isClosed.get() || pc.api.settingEngine.disableCloseByDTLS {
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
lgtm

Sean-Der:
What browser/client supports DTLS Restart? Would love to see it in action! If we detected fingerprint changes should we just ignore the close?

cnderrauber:
> What browser/client supports DTLS Restart? Would love to see it in action! 

@Sean-Der  Chrome already supports it, we are using it to migrate webrtc session into different nodes without interruption.

> If we detected fingerprint changes should we just ignore the close?

There is no problem in normal ice-restart case since pion doesn't support ice-restart, but in migration case it is a problem since the different finger-prints are received by different nodes (old node doesn't receive the new offer/answer sdp). We are updating the pion v4 and find the migration feature is broken (as well as a panic)

Sean-Der:
Cool!

Sorry about the panic @cnderrauber if you need any help debugging I am here!

cnderrauber:
> Cool!
> 
> Sorry about the panic @cnderrauber if you need any help debugging I am here!

@Sean-Der Please don't feel that about this great project, I had already fixed it. I'm glad to update to v4 with the new features!

PR #2968, opened 2024-12-10
Add EnableDataChannelBlockWrite to SettingEngine

cnderrauber:
The settingengine can enabled block write if
datachannel is detached, makes it working like
a normal net.Conn.

#### Reference issue
Fixes #...


cnderrauber:
>   - (*DataChannel).Detach: changed from func() (github.com/pion/datachannel.ReadWriteCloser, error) to func() (github.com/pion/datachannel.ReadWriteCloserDeadliner, error)
>   - (*SCTPTransport).Start: changed from func(SCTPCapabilities) error to func(SCTPCapabilities, ...SCTPOption) error
> 

The API check complains compatibility issues but they should be ok for existed codes, any suggestion on this change? @Sean-Der

boks1971 (approved):
lgtm!

Sean-Der:
@cnderrauber Could we do `DetachWithDeadline`? We can return a `datachannel.ReadWriteCloseDeadliner` from `Detach` the extra method will just be ignored?

API break I believe could be a problem. Users could have code that returns a `datachannel.ReadWriteCloser`

PR #2976, opened 2024-12-23
Fix detached datachannels handling

sukunrt:
https://github.com/pion/webrtc/pull/2696 introduced removing datachannels from the sctptransport for better garbage collection.

That PR introduced a race condition for data channels created before connection establishment. When an out of band negotiated data channel, created before peerconnection establishment is detached, there's a race between the data channel being removed from `r.dataChannels` and it being copied in to the existing data channel slice in the acceptDataChannels goroutine.

This PR fixes this race by copying the slice before any datachannels could be detached.

### DIFF
--- a/sctptransport.go
+++ b/sctptransport.go
@@ -142,7 +142,7 @@ func (r *SCTPTransport) Start(_ SCTPCapabilities) error {
 	r.dataChannelsOpened += openedDCCount
 	r.lock.Unlock()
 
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):
Thank you! Mind fixing lint error and we can land :)

p-shahi:
@sukunrt @Sean-Der can we get a patch release out with this fix?

sukunrt:
I'll cut a release tomorrow.

PR #2978, opened 2024-12-24
fix examples/bandwidth-estimation-from-disk

### LINKED ISSUE #2896: examples/bandwidth-estimation-from-disk currentTimestamp not updated correctly
In the `examples/bandwidth-estimation-from-disk/main.go` file, the `currentTimestamp` variable is intended to track the current video frame's timestamp to ensure correct playback when switching video quality levels. However, `currentTimestamp` is not being updated correctly and remains at its initial value.

Update currentTimestamp after each video frame is read and sent. Here is the code snippet:

```
    // ... 
for ; true; <-ticker.C {
    targetBitrate := estimator.GetTargetBitrate()
    switch {

    // ... 

    default:
        frame, frameHeader, err = ivf.ParseNextFrame() // modify this section
    }

    switch {
    // If we have reached the end of the file, start again
    case errors.Is(err, io.EOF):
        ivf.ResetReader(setReaderFile(qualityLevels[currentQuality].fileName))

    // No error, write the video frame
    case err == nil:
        currentTimestamp = frameHeader.Timestamp // Update currentTimestamp
        if err = videoTrack.WriteSample(media.Sample{Data: frame, Duration: time.Second}); err != nil {
            panic(err)
        }
    // Handle other unknown errors
    default:
        panic(err)
    }
}
```

howjmay:
#### Reference issue
Fixes #2896

### DIFF
--- a/examples/bandwidth-estimation-from-disk/main.go
+++ b/examples/bandwidth-estimation-from-disk/main.go
@@ -225,7 +225,7 @@ func main() {
 
 		// Adjust outbound bandwidth for probing
 		default:
-			frame, _, err = ivf.ParseNextFrame()
+			frame, frameHeader, err = ivf.ParseNextFrame()
 		}
 
 		switch {


Sean-Der:
Fantastic! Thank you so much @howjmay

PR #2985, opened 2025-01-02
Upgrade linters

joeturki:
#### Description

Upgrade the linters, I will ignore those that require refactor (will be addressed in a separate PR).

#### Reference PR
https://github.com/pion/.goassets/pull/220


iceserver_test.go
```diff
@@ -117,10 +117,6 @@ func TestICEServer_validate(t *testing.T) {
 }
 
 func TestICEServerZeroValue(t *testing.T) {
-	server := ICEServer{
```
Sean-Der, line 120: Can you bring these values back?

Sean-Der:
Amazing! This is exactly what we need @joeturki 

If you are overwhelmed by the lints you can just upgrade versions/disable more and we can iterate on it!

iceserver_test.go
joeturki: Done, and I disabled `unusedwrite` again.

.golangci.yml
```diff
@@ -44,68 +47,66 @@ linters:
...
+    - nlreturn         # nlreturn checks for a new line before return and branch statements to increase code clarity
     - noctx            # noctx finds sending http request without context.Context
     - predeclared      # find code that shadows one of Go's predeclared identifiers
     - revive           # golint replacement, finds style mistakes
     - staticcheck      # Staticcheck is a go vet on steroids, applying a ton of static analysis checks
     - stylecheck       # Stylecheck is a replacement for golint
     - tagliatelle      # Checks the struct tags.
     - tenv             # tenv is analyzer that detects using os.Setenv instead of t.Setenv since Go1.17
-    - tparallel        # tparallel detects inappropriate usage of t.Parallel() method in your Go test codes
+    - testpackage      # linter that makes you use a separate _test package
```
Sean-Der, line 87: We can't do `testpackage` quite a few tests touch Pion internals. Someday would be nice to fix!

joeturki: I was making a list of internal tests to rename them to _internal_tests, but will disable it for now :)

.golangci.yml
```diff
@@ -25,17 +25,37 @@ linters-settings:
       - ^os.Exit$
       - ^panic$
       - ^print(ln)?$
+  varnamelen:
+    max-distance: 12
+    min-name-length: 2
+    ignore-type-assert-ok: true
+    ignore-map-index-ok: true
+    ignore-chan-recv-ok: true
+    ignore-decls:
```
seaduboi-openai, line 34: We can't do `ignore-decls` because they are shared between repos :(

I really like this `varnamelen`. `wg` and `pc` are only things you are gonna know if you have wrote a lot of Go/WebRTC. This is really going to help people learn better!

joeturki: > We can't do ignore-decls because they are shared between repos :(

Even if I only apply it to shared decls like `n int`? :( I don't think it's necessary for `pc` and `wg` anymore since I changed `min-name-length` to 2.

Issue #2659, opened 2024-02-02
Data channel does not close if dataChannel.Close is called in the OnDataChannel handler


timmattison:
### Your environment.
  - Version: 3.2.24
  - Browser: N/A

### What did you do?
I set up a handler for OnDataConnection on the PeerConnection struct. This handler checks a few things to see if it's interested in the data channel or not. If it decides that it isn't interested it calls dataChannel.Close(). Otherwise, it sets up the OnOpen and OnMessage handlers.

### What did you expect?
I expected the data channel to close on both sides of the connection immediately.

### What happened?
The data channel's state went from connecting to open and stayed open. The call to dataChannel.Close() returns a nil error.

If I put the logic in the OnOpen handler it works. But I would've expected an error from Close() or for it to work no matter when it is called.

stephanrotolante:
@timmattison would you mind sharing some code?

stephanrotolante:
My initial thoughts is that you could be calling `Close()` on the peer connection's data channel before the channel was opened??

PR #2991, opened 2025-01-06
Gracefully close connecting channels

joeturki:
#### Description

This PR fixes an issue where calling `dataChannel.Close` on 'connecting' state channels didn't work as expected; It adds the ability to reject data channels directly from the `OnDataConnection` handler.

#### Reference issue

Fixes https://github.com/pion/webrtc/issues/2659

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -320,8 +320,13 @@ func (d *DataChannel) onMessage(msg DataChannelMessage) {
 
 func (d *DataChannel) handleOpen(dc *datachannel.DataChannel, isRemote, isAlreadyNegotiated bool) {
 	d.mu.Lock()
-	if d.isGracefulClosed {
+	if d.isGracefulClosed { // The channel was closed during the connecting state
 		d.mu.Unlock()
+		if err := dc.Close(); err != nil {
+			d.log.Errorf("Failed to close DataChannel that was closed during connecting state %v", err.Error())
+		}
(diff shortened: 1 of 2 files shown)


datachannel.go
```diff
@@ -320,8 +320,11 @@ func (d *DataChannel) onMessage(msg DataChannelMessage) {
 
 func (d *DataChannel) handleOpen(dc *datachannel.DataChannel, isRemote, isAlreadyNegotiated bool) {
 	d.mu.Lock()
-	if d.isGracefulClosed {
+	if d.isGracefulClosed { // The channel was closed during the connecting state
 		d.mu.Unlock()
+		dc.Close() // nolint:errcheck,gosec
```
jech, line 325: I think a comment is required here, since it's not immediately obvious why that is needed.

Sean-Der: I am fixing and merging that right now actually! I am going to log the error.

Also going to get involved with RTX, sorry I have been busy.

jech: No worries, I'm going to implement TWCC, and come back to RTX when you guys are ready.

Sean-Der (approved):
Amazing work @joeturki congrats on the first merge 🥳

datachannel.go
joeturki: I used to log the error but I remove it in another commit my bad \:)

PR #2999, opened 2025-01-11
Rewrite VP8 isKeyFrame check

xdrudis:
#### Description
[RFC 6386](https://www.rfc-editor.org/rfc/rfc6386#section-9.1) describes the least significant bit of the first byte of the header as:
```
A 1-bit frame type (0 for key frames, 1 for interframes).
```

The change is functionally a no-op, but the naming implies the wrong logic (you would assume `isKeyFrame == 1` means it is a key frame, but the opposite is true).

#### Reference issue
Fixes #...


Sean-Der (approved):
Nice catch :)

you are really looking into the details!

PR #3016, opened 2025-01-22
Fix addCodec to return error if payload type exists in codecs list

### LINKED ISSUE #3015: RegisterCodec can register two different codecs with same payload type
### Your environment.
  - Version: feeeebf251166faabdb00bd07b1ce2f6bd1064ad
  - Browser: Chrome/132.0.0.0
  - Other Information - https://gophers.slack.com/archives/CAK2124AG/p1737531701059119

### What did you do?

I added two different codecs with the same payload type and upon negotiation with the browser the browser threw `Failed to parse SessionDescription. a=rtpmap:100 VP8/90000 Duplicate payload type with conflicting codec name or clock rate` error.

### What did you expect?
`RegisterCodec` should have returned an error and should not add codec with duplicate payload type.

### What happened?
The addCodec method is called from updateFromRemoteDescription and RegisterCodec as well. The issue is if there are different codecs with the same payload type, for eg- 100 VP8 (Firefox) and 100 VP9 (Chrome), getting registered. https://github.com/pion/webrtc/blob/feeeebf251166faabdb00bd07b1ce2f6bd1064ad/mediaengine.go#L249 This line will treat them as different codecs and add both of them in codecs slice.
When the getCodecsByKind method is called during SDP generate for either answer or offer, both codecs will be included in SDP. Chrome throws an error `Failed to parse SessionDescription. a=rtpmap:100 VP8/90000 Duplicate payload type with conflicting codec name or clock rate` when it gets the same payload type with different codec information.

itzmanish:
#### Description
This PR adds changes to return an error if the codec already exists on the codecs list. This way now the implementing party should take care of duplicate codecs addition if the payload type is the same.
#### Reference issue
Fixes #3015

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -7,6 +7,7 @@
 package webrtc
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -54,6 +55,9 @@ const (
 	// Note: Matching should be case insensitive.
 	MimeTypeFlexFEC = "video/flexfec"
 )
+
+// ErrCodecAlreadyRegistered indicates that a codec has already been registered for the same payload type.
+var ErrCodecAlreadyRegistered = fmt.Errorf("codec already registered for same payload type")
 
 type mediaEngineHeaderExtension struct {
 	uri              string
@@ -244,17 +248,20 @@ func (m *MediaEngine) RegisterDefaultCodecs() error {
 }
 
 // addCodec will append codec if it not exists.
-func (m *MediaEngine) addCodec(codecs []RTPCodecParameters, codec RTPCodecParameters) []RTPCodecParameters {
+func (m *MediaEngine) addCodec(codecs []RTPCodecParameters, codec RTPCodecParameters) ([]RTPCodecParameters, error) {
 	for _, c := range codecs {
-		if c.MimeType == codec.MimeType &&
-			fmtp.ClockRateEqual(c.MimeType, c.ClockRate, codec.ClockRate) &&
-			fmtp.ChannelsEqual(c.MimeType, c.Channels, codec.Channels) &&
-			c.PayloadType == codec.PayloadType {
-			return codecs
+		if c.PayloadType == codec.PayloadType {
+			if c.MimeType == codec.MimeType &&
+				fmtp.ClockRateEqual(c.MimeType, c.ClockRate, codec.ClockRate) &&
+				fmtp.ChannelsEqual(c.MimeType, c.Channels, codec.Channels) {
+				return codecs, nil
+			}
+
+			return codecs, ErrCodecAlreadyRegistered
 		}
 	}
 
-	return append(codecs, codec)
+	return append(codecs, codec), nil
 }
 
 // RegisterCodec adds codec to the MediaEngine
@@ -263,17 +270,18 @@ func (m *MediaEngine) RegisterCodec(codec RTPCodecParameters, typ RTPCodecType) 
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	var err error
 	codec.statsID = fmt.Sprintf("RTPCodec-%d", time.Now().UnixNano())
 	switch typ {
 	case RTPCodecTypeAudio:
-		m.audioCodecs = m.addCodec(m.audioCodecs, codec)
+		m.audioCodecs, err = m.addCodec(m.audioCodecs, codec)
 	case RTPCodecTypeVideo:
-		m.videoCodecs = m.addCodec(m.videoCodecs, codec)
+		m.videoCodecs, err = m.addCodec(m.videoCodecs, codec)
 	default:
 		return ErrUnknownType
 	}
 
-	return nil
+	return err
 }
 
 // RegisterHeaderExtension adds a header extension to the MediaEngine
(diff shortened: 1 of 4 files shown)


itzmanish:
While this PR fixes the issue of duplicate payload type codec entry on RegisterCodec, There is still one issue that this PR doesn't cover. 

What if we have some registered codecs and the client sends the codecs with different payload types and one of the payload types the client is using clashes with our registered codecs, in that case, the client's codec won't be added to the negotiated codecs list. That's an issue because the server will now be missing that codec.

The flow -
```
Pion server codec list - []{..,PayloadType: 96, Mime: "video/VP8"..}
Browser sends codecs on sdp which triggers `updateFromRemoteDescription` - []{..., PayloadType: 96, Mime: "audio/OPUS"...}
```

This PR will make the media engine discard the opus codec. However, the issue remains even without this PR,
```
Pion server codec list - []{..,PayloadType: 96, Mime: "video/VP9", "profile-id=0"..}
Browser sends codecs on sdp which triggers `updateFromRemoteDescription` - []{..., PayloadType: 96, Mime: "video/VP9", "profile-id=1"...}
```
Again the client-given codec won't be added and most probably the browser will use `"profile-id=0"` one only (I am not sure about specs here, so please correct me here).

mediaengine.go
```diff
@@ -573,9 +577,9 @@ func (m *MediaEngine) updateHeaderExtension(id int, extension string, typ RTPCod
 func (m *MediaEngine) pushCodecs(codecs []RTPCodecParameters, typ RTPCodecType) {
 	for _, codec := range codecs {
 		if typ == RTPCodecTypeAudio {
-			m.negotiatedAudioCodecs = m.addCodec(m.negotiatedAudioCodecs, codec)
+			m.negotiatedAudioCodecs, _ = m.addCodec(m.negotiatedAudioCodecs, codec)
```
JoeTurki, line 580: Should we consider handling this error explicitly? If someone mistakenly changes matchRemoteCodec.

I think it might be better to scope the logic of your PR to specific cases, such as only handling it only for videoCodecs and audioCodecs (calls originating from RegisterCodec). This would make sure we’re not inadvertently affecting unrelated handling.

mediaengine.go
```diff
@@ -573,9 +577,9 @@ func (m *MediaEngine) updateHeaderExtension(id int, extension string, typ RTPCod
...
 		} else if typ == RTPCodecTypeVideo {
-			m.negotiatedVideoCodecs = m.addCodec(m.negotiatedVideoCodecs, codec)
+			m.negotiatedVideoCodecs, _ = m.addCodec(m.negotiatedVideoCodecs, codec)
```
JoeTurki, line 582: See above.

itzmanish:
> However, for sendonly and sendrecv streams, the answer might indicate
different payload type numbers for the same codecs, in which case,
the offerer MUST send with the payload type numbers from the answer.

This line. let's say, the server is an offerer and generates SDP with PT 96, mime video/VP8 and PT 97, mime video/VP9. Could it be possible for the client to generate and answer SDP with PT 96, mime video/VP9 and PT 97, mime video/VP8? If yes, according to how we update negotiatedCodecs, it won't get updated.
But the spec says `the offerer MUST send with the payload type numbers from the answer`, With what PT number the server will send the packet?

JoeTurki (changes requested):

JoeTurki:
> Maybe need to re-trigger the commit linter to run. I have updated the PR title and last commit message to pass, but it's not getting triggered for newer commits. @JoeTurki could you please re-trigger the gh action for lint metadata?

PR title doesn't affect it, You'll need to reword your old commits, but since we only allow for one commit per PR it's better if you squash your commits into a single one and fix the lint issues :)

Thanks :)

mediaengine.go
itzmanish: >Should we consider handling this error explicitly?

Honestly, we don't have to handle the error here, because push codec only happens on either partial or full match. If it's a full match, the codec won't be pushed anyway, and if it's partial I don't think it will be possible for the client to send two different codecs with the same payload type on SDP. So error handling doesn't matter here.

itzmanish:
yes. I get it now. The negotiated codec always respects the answer's codec preferences.

mediaengine.go
JoeTurki: This is true, but my comment was more about if someone updates `matchRemoteCodec` in the future, or any side effects. Maybe we only introduce this change only to the public API `RegisterCodec` or what do you think?

>  So error handling doesn't matter here.

I think we should return an error but I think we should do that in another PR.

I'm not entirely sure about this PR, maybe we should wait for someone with better understanding of the codebase, But I agree that `RegisterCodec` should return an error for dynamic PT reuse.

@Sean-Der

JoeTurki: >  It's just currently the least possible change path where without doing another loop to find if the payload type already exists or not, we can achieve the error returning in case the codec is available.

You can just pass a Boolean (unique) to  addCodec and conditionally return an error if it's true or refactor it to two functions one for internals, and one for matchedcodecs. Or any other option, The implementations doesn't matter, we can fix the code later; What matters is the behavior.

> Is there any possible scenario you can think of, where the client sends two different codecs with the same payload type?

Yes, A misconfigured client, You found this "bug" because Chrome returned an error when you re-used the PT for different codecs. And I think we should return an error, to make sure we don't have an unexpected behavior.

itzmanish: > Yes, A misconfigured client, You found this "bug" because Chrome returned an error when you re-used the PT for different codecs. And I think we should return an error, to make sure we don't have an unexpected behavior.

I totally agree. I can make those changes in the same PR or wait for @Sean-Der's input. It makes sense to handle this error on `updateFromRemoteDescription`.

itzmanish:
Hi @JoeTurki, Sorry for the delay on my end. Please re-review the PR, I have updated it to propagate the error on the updateRemoteDescription flow as well.

JoeTurki (approved):
Thank you Manish :)

PR #3018, opened 2025-01-22
Add multi codec negotiation feature

itzmanish:
#### Description
This extends PR #2632 and Adds the capability to have multi-codec negotiation just by setting a flag in the media engine.
#### Reference issue
Fixes #...

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -358,10 +358,9 @@ func (m *MediaEngine) copy() *MediaEngine {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	cloned := &MediaEngine{
-		videoCodecs:          append([]RTPCodecParameters{}, m.videoCodecs...),
-		audioCodecs:          append([]RTPCodecParameters{}, m.audioCodecs...),
-		headerExtensions:     append([]mediaEngineHeaderExtension{}, m.headerExtensions...),
-		negotiateMultiCodecs: m.negotiateMultiCodecs,
+		videoCodecs:      append([]RTPCodecParameters{}, m.videoCodecs...),
+		audioCodecs:      append([]RTPCodecParameters{}, m.audioCodecs...),
+		headerExtensions: append([]mediaEngineHeaderExtension{}, m.headerExtensions...),
 	}
 	if len(m.headerExtensions) > 0 {
 		cloned.negotiatedHeaderExtensions = map[int]mediaEngineHeaderExtension{}
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -44,16 +44,16 @@ type MediaEngine struct {
 	mu sync.RWMutex
 }
 
-// SetMultiCodecNegotiation enables or disables the negotiation of multiple codecs.
-func (m *MediaEngine) SetMultiCodecNegotiation(negotiateMultiCodecs bool) {
+// setMultiCodecNegotiation enables or disables the negotiation of multiple codecs.
+func (m *MediaEngine) setMultiCodecNegotiation(negotiateMultiCodecs bool) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
 	m.negotiateMultiCodecs = negotiateMultiCodecs
 }
 
-// MultiCodecNegotiation returns the current state of the negotiation of multiple codecs.
-func (m *MediaEngine) MultiCodecNegotiation() bool {
+// multiCodecNegotiation returns the current state of the negotiation of multiple codecs.
+func (m *MediaEngine) multiCodecNegotiation() bool {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -33,6 +33,7 @@ type mediaEngineHeaderExtension struct {
 type MediaEngine struct {
 	// If we have attempted to negotiate a codec type yet.
 	negotiatedVideo, negotiatedAudio bool
+	negotiateMultiCodecs             bool
 
 	videoCodecs, audioCodecs                     []RTPCodecParameters
 	negotiatedVideoCodecs, negotiatedAudioCodecs []RTPCodecParameters
@@ -41,6 +42,22 @@ type MediaEngine struct {
 	negotiatedHeaderExtensions map[int]mediaEngineHeaderExtension
 
 	mu sync.RWMutex
+}
+
+// SetMultiCodecNegotiation enables or disables the negotiation of multiple codecs.
+func (m *MediaEngine) SetMultiCodecNegotiation(negotiateMultiCodecs bool) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.negotiateMultiCodecs = negotiateMultiCodecs
+}
+
+// MultiCodecNegotiation returns the current state of the negotiation of multiple codecs.
+func (m *MediaEngine) MultiCodecNegotiation() bool {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return m.negotiateMultiCodecs
 }
 
 // RegisterDefaultCodecs registers the default codecs supported by Pion WebRTC.
@@ -579,9 +596,9 @@ func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) e
 		}
 
 		switch {
-		case !m.negotiatedAudio && typ == RTPCodecTypeAudio:
+		case (!m.negotiatedAudio || m.negotiateMultiCodecs) && typ == RTPCodecTypeAudio:
 			m.negotiatedAudio = true
-		case !m.negotiatedVideo && typ == RTPCodecTypeVideo:
+		case (!m.negotiatedVideo || m.negotiateMultiCodecs) && typ == RTPCodecTypeVideo:
 			m.negotiatedVideo = true
 		default:
 			// update header extesions from remote sdp if codec is negotiated, Firefox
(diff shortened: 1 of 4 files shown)


mediaengine.go
```diff
@@ -615,9 +632,9 @@ func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) e
 		}
 
 		switch {
-		case !m.negotiatedAudio && typ == RTPCodecTypeAudio:
+		case (!m.negotiatedAudio || m.negotiateMultiCodecs) && typ == RTPCodecTypeAudio:
 			m.negotiatedAudio = true
-		case !m.negotiatedVideo && typ == RTPCodecTypeVideo:
+		case (!m.negotiatedVideo || m.negotiateMultiCodecs) && typ == RTPCodecTypeVideo:
 			m.negotiatedVideo = true
 		default:
```
JoeTurki, line 639: I'm thinking instead of modifying the case conditions, I suggest adding a condition before the continue in the next default block:

``` if !m.negotiateMultiCodecs || (typ != RTPCodecTypeAudio && typ != RTPCodecTypeVideo)```

This ensures that m.negotiatedVideo and m.negotiatedAudio are set properly.

**Comment by JoeTurki** [comment]
I'm thinking maybe the original continue specifically intended for data sections, making it safe to consider adding this without a flag? not sure, I'll have to check, but thanks again Manish!

File: mediaengine.go (hunk 1 above)
Comment (nils-ohlmeier) [reply mediaengine.go]: I updated the PR to follow your suggestion since it appears to be the safer version.

**Comment by nils-ohlmeier** [comment]
I updated the PR with a bugfix: since the MediaEngine gets copied for every new PeerConnection this new setting needs explicitly been taken care of in the `copy()` function.

> I'm thinking maybe the original continue specifically intended for data sections, making it safe to consider adding this without a flag? not sure, I'll have to check, but thanks again Manish!

Even though this PR makes Pion more RFC compliant I would suggest to start with an optional setting, like in this PR. Followed by eventually switching the default of the setting. And hopefully if nobody opens any issues explaining why they depend on the old, non-RFC compliant behavior, we can remove the setting.

**Review by JoeTurki — APPROVED** [review]
Thanks You guys, would have been cooler if we could add this setting to `SettingEngine` instead, but i don't think that there is a clean way to do that :(

**Comment by nils-ohlmeier** [comment]
@JoeTurki here is an alternative proposal:

- we add a new multi codec setting to the `SettingEngine`
- here https://github.com/pion/webrtc/blob/master/peerconnection.go#L158 we manually go through the `mediaEngine` related settings in `SettingEngine` and call setters on the new copy of `MediaEngine`.

What I'm not clear on is if we should do this only if `disabledMediaEngineCopy` is false, or if we should do that regardless of the value of `disableMediaEngineCopy`.

**Comment by nils-ohlmeier** [comment]
The latest commit has my alternative proposal. Let me know which version everyone likes better.

**Comment by JoeTurki** [comment]
Thank you so much @nils-ohlmeier  this makes so much sense, @itzmanish what do you think? i think we should make the `mediaEngine` setting private and expose it through Nils proposal via the settings engine?

**Comment by Sean-Der** [comment]
What would it break if we turned this on by default?

if a remote didn’t actually enable a codec in a media section things wouldn’t work?

**Comment by Sean-Der** [comment]
I’m curious what people think is best for users. If it’s not on by default people won’t benefit/have to debug and then will turn it on

**Comment by nils-ohlmeier** [comment]
My only concern here is that this changes how Pion so far handled things. Maybe there are services/implementations out there which for some reason rely on the fact that Pion does answer with the same codec for all media sections. Personally I would find that pretty strange, but you never know.

I guess the more likely scenario is that folks learned to work around the current behavior some way. For example I'm not sure if renegotiating one media section after another, each going through a full SDP offer/answer round,  would allow you to choose different codecs. But if people have developed such workarounds I guess it is pretty unlikely that these would break.

So maybe you are right and we could make this the new default and only leave the new setting in as a way to get back to the old behavior if someone needs it.

**Comment by Sean-Der** [comment]
That's a great analysis :)

My gut says it is better for users to have this on by default. I would even leave the setting out! Until someone complains then we add it back. Otherwise it turns into one of those things that is in the code base forever that never gets cleaned up. 

Maybe I should be more conservative. I feel bad we are shipping something today that is subpar, I just want users to `git pull` and get the better version without jumping through hoops. Especially if no one is disrupted!

**Comment by itzmanish** [comment]
> what do you think? i think we should make the mediaEngine setting private and expose it through Nils proposal via the settings engine?

yeah, because not everyone uses custom MediaEngine struct and pass on the API to create PeerConnection and it would be nice to have this as a config in setting engines so everyone can use it easily.

**Comment by itzmanish** [comment]
A different issue I have seen, though I am not sure how this works with the spec guideline. The `updateHeaderExtensionFromMediaSection` is called from the default block of the 2nd switch case as well as at the end of the method https://github.com/pion/webrtc/pull/3018/files#diff-045bc12c2f054653bc94bdbcbdd27f63b14536b7b5b34f451633f14ea1d8c1b0L677.

Now even if we don't consider multi codec enabled scenario, the default case will be hit for all subsequent codecs in offer even the audio/video is negotiated. In that case if the offer have different `extensionId` for header extensions they will be stored in `negotiatedHeaderExtension` map which is being used by peerconnection when doing `handleIncomingSSRC`. This may pick up wrong extension id for the track on the incoming MID. 

Questions - 
1. Is it allowed to have a different extension ID for the same extension in different media sections?
2. Should Pion handle different extension IDs for the same extension for different media sections and correctly use the extension ID for a MID's incoming track?

**Comment by nils-ohlmeier** [comment]
> That's a great analysis :)
> 
> My gut says it is better for users to have this on by default. I would even leave the setting out! Until someone complains then we add it back. Otherwise it turns into one of those things that is in the code base forever that never gets cleaned up.
> 
> Maybe I should be more conservative. I feel bad we are shipping something today that is subpar, I just want users to `git pull` and get the better version without jumping through hoops. Especially if no one is disrupted!

Alright. I see your point for making this the new default behavior. Makes sense.

I would suggest to have the setting in there for folks to easily disable if needed. I can put a reminder in my calendar to remove it, or maybe we could put it on the todo list for the next release after the next? But if you @Sean-Der prefer to config option I'll remove it. Just let me know.

**Comment by itzmanish** [comment]
yeah, let's discuss that seperately #3087 to avoid getting out of track from the scope of this PR.

File: mediaengine.go [hunk 2]
```diff
@@ -43,6 +44,22 @@ type MediaEngine struct {
 	mu sync.RWMutex
 }
 
+// SetMultiCodecNegotiation enables or disables the negotiation of multiple codecs.
+func (m *MediaEngine) SetMultiCodecNegotiation(negotiateMultiCodecs bool) {
```
JoeTurki, line 48: Do we need to also expose `SetMultiCodecNegotiation` and `MultiCodecNegotiation` from the `MediaEngine` since we expose them from `SettingsEngine` I'm just asking for clarification

settingengine.go
```diff
@@ -396,6 +397,16 @@ func (e *SettingEngine) DisableMediaEngineCopy(isDisabled bool) {
 	e.disableMediaEngineCopy = isDisabled
 }
 
+// DisableMediaEngineMultipleCodecs disables the MediaEngine negotiating different codecs for
+// multiple media sections in the SDP. This is the new default behvior, because it makes
+// Pion more spec compliant. The value of this setting will get copied to every copy of the
+// MediaEngine generated for new PeerConnections (assuming DisableMediaEngineCopy is set to
+// false).
+// Note: this setting is targeted to be removed in release 4.2.0 (or later)
+func (e *SettingEngine) DisableMediaEngineMultipleCodecs(enable bool) {
```
JoeTurki, line 406: Just a random nitpick, you're free to ignore, I think disable(enable) is a bit confusing, maybe setEnable(enable) or setDissable(disable)

peerconnection.go
```diff
@@ -156,6 +156,7 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
 		pc.api.mediaEngine = api.mediaEngine
 	} else {
 		pc.api.mediaEngine = api.mediaEngine.copy()
+		pc.api.mediaEngine.setMultiCodecNegotiation(!api.settingEngine.disableMediaEngineMultipleCodecs)
```
3DRX, line 159: Just a small question, why is `setMultiCodecNegotiation` called only when `disableMediaEngineCopy` is false?

JoeTurki: Good question if you look at nils comment: 
> What I'm not clear on is if we should do this only if disabledMediaEngineCopy is false, or if we should do that regardless of the value of disableMediaEngineCopy.

I think we should set it regardless, We're currently waiting on @itzmanish review for the changes!

nils-ohlmeier: Indeed I'm unsure on how to handle that one.
One argument is that if copying is disable no settings from the initial MediaEngine should be transferred over. 
I think the other view is: this is now a global setting and thus should get applied to all MediaEngine instances. 

I'm okay doing it either way.

itzmanish: Since this is a global setting now, I think we should always respect that when copying the mediaengine. However api have setting engine inside it and it won't get changed once the api is made, I don't think we need to explicitly set the multiCodecNegotiation flag from setting engine because it would be same.

Also currently I don't think it's possible to set mediaengine's multicodec flag directly because they are now private field which means if you want to disable multicodec in the middle (which is weird case where you want multicodec for some instances of peerconnection but not for other), you have change the settingengine directly.

itzmanish:
@JoeTurki The changes are fine from my side and let's merge this.

PR #3021, opened 2025-01-25
Fix matching codecs with different rate or channels

aler9:
#### Description

In pion/webrtc, currently codecs are matched regardless of the clock rate and the channel count, and this makes impossible to fully support codecs that might have a clock rate or channel count different than the default one, in particular LPCM, PCMU, PCMA and multiopus (the last one is a custom Opus variant present in the Chrome source code to support multichannel Opus).

For instance, let's suppose a peer (receiver) wants to receive an audio track encoded with LPCM, 48khz sample rate and 2 channels. This receiver doesn't know the audio codec yet, therefore it advertises all supported sample rates in the SDP:

```
LPCM/44100
LPCM/48000
LPCM/44100/2
LPCM/48000/2
```

The other peer (sender) receives the SDP, but since the clock rate and channel count are not taken into consideration when matching codecs, the sender codec `LPCM/48000/2` is wrongly associated with the receiver codec `LPCM/44100`. The result is that the audio track cannot be decoded correctly from the receiver side.

This patch fixes the issue and has been running smoothly in MediaMTX for almost a year.

Unfortunately, in lots of examples and tests, clock rate and/or channels are not present (and in fact they are producing horrible SDPs that contain `VP8/0` instead of `VP8/90000` and are incompatible with lots of servers) therefore this new check causes troubles in existing code. In order to maintain compatibility, default clock rates and channels are provided for most codecs.

In the future, it might be better to update examples (i can do it in a future patch) and remove the exception.

### DIFF
--- a/internal/fmtp/fmtp.go
+++ b/internal/fmtp/fmtp.go
@@ -8,6 +8,32 @@ import (
 	"strings"
 )
 
+func defaultClockRate(mimeType string) uint32 {
+	defaults := map[string]uint32{
+		"audio/opus": 48000,
+		"audio/pcmu": 8000,
+		"audio/pcma": 8000,
+	}
+
+	if def, ok := defaults[strings.ToLower(mimeType)]; ok {
+		return def
+	}
+
+	return 90000
+}
+
+func defaultChannels(mimeType string) uint16 {
+	defaults := map[string]uint16{
+		"audio/opus": 2,
+	}
+
+	if def, ok := defaults[strings.ToLower(mimeType)]; ok {
+		return def
+	}
+
+	return 0
+}
+
 func parseParameters(line string) map[string]string {
 	parameters := make(map[string]string)
 
@@ -22,6 +48,61 @@ func parseParameters(line string) map[string]string {
 	}
 
 	return parameters
+}
+
+// ClockRateEqual checks whether two clock rates are equal.
+func ClockRateEqual(mimeType string, valA, valB uint32) bool {
+	// Lots of users use formats without setting clock rate or channels.
+	// In this case, use default values.
+	// It would be better to remove this exception in a future major release.
+	if valA == 0 {
+		valA = defaultClockRate(mimeType)
+	}
+	if valB == 0 {
+		valB = defaultClockRate(mimeType)
+	}
+
+	return valA == valB
+}
+
+// ChannelsEqual checks whether two channels are equal.
+func ChannelsEqual(mimeType string, valA, valB uint16) bool {
+	// Lots of users use formats without setting clock rate or channels.
+	// In this case, use default values.
+	// It would be better to remove this exception in a future major release.
+	if valA == 0 {
+		valA = defaultChannels(mimeType)
+	}
+	if valB == 0 {
+		valB = defaultChannels(mimeType)
+	}
+
+	// RFC8866: channel count "is OPTIONAL and may be omitted
+	// if the number of channels is one".
+	if valA == 0 {
+		valA = 1
+	}
+	if valB == 0 {
+		valB = 1
+	}
+
+	return valA == valB
+}
+
+func paramsEqual(valA, valB map[string]string) bool {
+	for k, v := range valA {
+		if vb, ok := valB[k]; ok && !strings.EqualFold(vb, v) {
+			return false
+		}
+	}
+
+	for k, v := range valB {
+		if va, ok := valA[k]; ok && !strings.EqualFold(va, v) {
+			return false
+		}
+	}
+
+	return true
 }
 
 // FMTP interface for implementing custom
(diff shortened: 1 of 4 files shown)


internal/fmtp/fmtp.go
```diff
@@ -24,6 +24,74 @@ func parseParameters(line string) map[string]string {
 	return parameters
 }
 
+// ClockRateEqual checks whether two clock rates are equal.
+func ClockRateEqual(mimeType string, valA, valB uint32) bool {
```
JoeTurki, line 28: Correct me if I'm wrong, but I feel like this function `ClockRateEqual` might not belong in the fmtp library. Could this logic be more appropriate elsewhere? same for `ChannelsEqual` etc

internal/fmtp/fmtp.go
```diff
@@ -24,6 +24,74 @@ func parseParameters(line string) map[string]string {
...
+	// Clock rate and channel checks have been introduced quite recently.
+	// Existing implementations often use VP8, H264 or Opus without setting clock rate or channels.
+	// Keep compatibility with these situations.
+	// It would be better to remove this exception in a future major release.
+	switch {
```
JoeTurki, line 33: Do we need to handle this for other codecs like H.264 and VP9 as well?

Should we also match if either clock rate is 0 regardless of the codec (to maintain compatibility with cases where clock rate is not explicitly set)?

rtpsender_test.go
```diff
@@ -36,7 +36,7 @@ func Test_RTPSender_ReplaceTrack(t *testing.T) { //nolint:cyclop
 	trackA, err := NewTrackLocalStaticSample(RTPCodecCapability{MimeType: MimeTypeVP8}, "video", "pion")
 	assert.NoError(t, err)
 
-	trackB, err := NewTrackLocalStaticSample(RTPCodecCapability{MimeType: MimeTypeH264}, "video", "pion")
+	trackB, err := NewTrackLocalStaticSample(RTPCodecCapability{MimeType: MimeTypeH264, ClockRate: 90000}, "video", "pion")
```
JoeTurki, line 39: Can we keep this test as it was to ensure we fallback to matching without the `clockRate` for backward compatibility? Instead, we could add a new test

track_local_static_test.go
```diff
@@ -400,7 +400,7 @@ func Test_TrackLocalStatic_Payloader(t *testing.T) {
 
 	customPayloader := &customCodecPayloader{}
 	track, err := NewTrackLocalStaticSample(
-		RTPCodecCapability{MimeType: mimeTypeCustomCodec},
+		RTPCodecCapability{MimeType: mimeTypeCustomCodec, ClockRate: 90000},
```
JoeTurki, line 403: Can we keep this test as it was to ensure we fallback to matching without the `clockRate` for backward compatibility? Instead, we could add a new test

JoeTurki (commented):
Thank you so much for this! Just few random thoughts; although I'm not fully confident with the Pion codebase yet to approve this so you'll have to wait for @Sean-Der , this change does make sense to me. 

Thank you again <3

JoeTurki (approved):

internal/fmtp/fmtp.go
aler9: @JoeTurki the fmtp internal library is intended to parse and compare formats. Clock rate and channels are part of a format. Furthermore, in case of misunderstanding, everything that is touched by this PR is in the internal folder, so it is not exposed and can be changed without creating a major release.

internal/fmtp/fmtp.go
aler9: i'll add defaults for H264 and VP9 too. I think i'd better add defaults for every existing codec instead of hand-picking them.

track_local_static_test.go
aler9: i'll try to find out why this test fails without the change.

aler9:
@JoeTurki thank you very much for the detailed review. I'm gonna rewrite this in order to provide defaults for every supported codec and support any existing situation, and at the same time solve the original issue that led to this PR (that was the fact that LPCM/48000/2 was wrongfully associated with LPCM/44100/1, leading to corrupted audio).

track_local_static_test.go
aler9: the test was restored to its previous state

internal/fmtp/fmtp.go
aler9: defaults were added for H264, H265, VP9 too

JoeTurki:
@aler9  Thank you so much <3

Sean-Der:
@aler9 @JoeTurki no need to wait on me! As long as we don't break existing code push/tag whatever you need for forward progress :)

Issue #519, opened 2019-03-15
Address differences between Go API and Wasm bindings


albrow:
## Summary

This issue lists all known API differences between the Go API and the Wasm bindings. Some of these already have their own issues.

## List of Differences

- [ ] If no certificates are provided during initialization, `GetConfiguration` returns the default generated certificate in the Go API but no certificates in the Wasm bindings (See #513).
- [x] (Fixed in https://github.com/pions/webrtc/pull/527). Some properties of `PeerConnection` are implemented as struct fields in the Go API but are methods in the Wasm bindings. Specifically:
  - `CurrentLocalDescription`
  - `PendingLocalDescription`
  - `CurrentRemoteDescription`
  - `PendingRemoteDescription`
  - `SignalingState`
  - `ICEGatheringState`
  - `ConnectionState`
- [x] (Fixed in https://github.com/pions/webrtc/pull/524). Similarly, some properties of `DataChannel` are implemented as struct fields in the Go API but are methods in the Wasm bindings:
  - `Label`
  - `Ordered`
  - `MaxPacketLifeTime`
  - `MaxRetransmits`
  - `Protocol`
  - `Negotiated`
  - `ID`
  - `ReadyState`
  - `BufferedAmount`
  - `BufferedAmountLowThreshold`
- [x] (Fixed in https://github.com/pions/webrtc/pull/546) In the Go API, ICE candidate gathering is synchronous but in the Wasm bindings, ICE candidate gathering happens in the background. The Go API is missing the `OnICECandidate` and `OnNegotiationNeeded` methods.
- [ ] The Wasm bindings do not support certificates or credentials (See #499). This might not be addressed before the v2 release.
- [x] The Wasm bindings do not support tracks, streams, or media (See #500). This might not be addressed before the v2 release.
- [ ] The Wasm bindings do not support the `BufferedAmount` methods (introduced in https://github.com/pions/webrtc/pull/572).
- [ ] The Wasm bindings can't properly subscribe to OnOpen callback when creating a DataChannel. This function might return an already opened DataChannel.

albrow:
Once we have addressed this differences, it should be possible to create a (probably unexported) interface to help enforce homogeneity between the two APIs going forward. This is what @backkem [suggested here](https://github.com/pions/webrtc/pull/479#issuecomment-469630837).

Sean-Der:
All the subtasks have been resolved.

The API differences between WASM/Go are unfortunate, but surprisingly no user has brought it up!

[...]

PR #3030, opened 2025-02-10
feat: implement js,wasm DataChannel OnError & OnClosing

WofWca:
#### Description
- **feat: implement js,wasm DataChannel.OnError**
- **feat: implement js,wasm DataChannel.OnClosing**

FYI `OnClosing` is not implemented in the non-JS version of Pion,
so this is a JS-only method. Please consider if this is a good idea.

This is a follow-up to https://github.com/pion/webrtc/pull/479.

#### Reference issue

Partially addresses https://github.com/pion/webrtc/issues/519#issuecomment-2649343163.

***

If I may, @albrow thank you so much for the initial implementation!
It's a bliss that it "just works"!


Sean-Der:
Thank you @WofWca!

I went to write some tests, but `wrtc` supports neither :(

PR #3051, opened 2025-03-04
Fix missing RTX codec when using SetCodecPreferences

### LINKED ISSUE #2996: Using SetCodecPrefences to constrain the track type disables negotiation of RTX
In Galene, I'm using SetCodecPreferences to ensure that we negotiate the right codec for a track forwarded to a client.  Unfortunately, this has the side effect of disabling RTX.

Here's a full test:

```
package rtxtest

import (
	"strings"
	"testing"

	"github.com/pion/webrtc/v4"
)

func TestRTX(t *testing.T) {
	t.Run("NoSetCodecPreferences", func(t *testing.T) { testRTX(t, false) })
	t.Run("SetCodecPreferences", func(t *testing.T) { testRTX(t, true) })
}

func testRTX(t *testing.T, setPreferences bool) {
	pc, err := webrtc.NewPeerConnection(webrtc.Configuration{})
	if err != nil {
		t.Fatalf("NewPeerConnection: %v", err)
	}
	transceiver, err := pc.AddTransceiverFromKind(
		webrtc.RTPCodecTypeVideo,
		webrtc.RTPTransceiverInit{
			Direction: webrtc.RTPTransceiverDirectionRecvonly,
		},
	)

	if setPreferences {
		codec := webrtc.RTPCodecCapability{
			"video/vp8", 90000, 0, "", nil,
		}
		err := transceiver.SetCodecPreferences(
			[]webrtc.RTPCodecParameters{
				{
					RTPCodecCapability: codec,
				},
			},
		)
		if err != nil {
			t.Errorf("SetCodecPreferences: %v", err)
		}
	}

	if err != nil {
		t.Fatalf("AddTransceiverFromKind: %v", err)
	}

	offer, err := pc.CreateOffer(nil)
	if err != nil {
		t.Fatalf("CreateOffer: %v", err)
	}

	if !strings.Contains(offer.SDP, "apt=96") {
		t.Errorf("Didn't try to negotiate RTX")
	}
}
```

stephanrotolante:
#### Description

When you specify the codecs that you want to use, we need to add the RTX codec related to the specified codecs

#### Reference issue
Fixes #2996

### DIFF
--- a/rtpcodec.go
+++ b/rtpcodec.go
@@ -143,7 +143,7 @@ func codecParametersFuzzySearch(
 	return RTPCodecParameters{}, codecMatchNone
 }
 
-// Given a CodecParameters find the RTX CodecParameters if one exists.
+// Given a CodecParameters find the RTX Payload if one exists.
 func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) PayloadType {
 	aptStr := fmt.Sprintf("apt=%d", needle)
 	for _, c := range haystack {
@@ -153,6 +153,18 @@ func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) Paylo
 	}
 
 	return PayloadType(0)
+}
+
+// Given a CodecParameters find the RTX CodecParameters if one exists.
+func findRTXPCodec(needle PayloadType, haystack []RTPCodecParameters) *RTPCodecParameters {
+	aptStr := fmt.Sprintf("apt=%d", needle)
+	for _, c := range haystack {
+		if aptStr == c.SDPFmtpLine {
+			return &c
+		}
+	}
+
+	return nil
 }
 
 func rtcpFeedbackIntersection(a, b []RTCPFeedback) (out []RTCPFeedback) {
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -86,6 +86,17 @@ func (t *RTPTransceiver) getCodecs() []RTPCodecParameters {
 		}
 	}
 
+	// check if direction is specified
+	if t.direction.Load() != nil {
+		for _, c := range filteredCodecs {
+			if rtxCodec := findRTXPCodec(c.PayloadType, mediaEngineCodecs); rtxCodec != nil {
+				if _, matchType := codecParametersFuzzySearch(*rtxCodec, filteredCodecs); matchType == codecMatchNone {
+					filteredCodecs = append(filteredCodecs, *rtxCodec)
+				}
+			}
+		}
+	}
+
 	return filteredCodecs
 }
 
--- a/rtptransceiver_test.go
+++ b/rtptransceiver_test.go
@@ -135,3 +135,55 @@ func Test_RTPTransceiver_SetCodecPreferences_PayloadType(t *testing.T) {
 
 	closePairNow(t, offerPC, answerPC)
 }
+
+func Test_RTPTransceiver_SDP_Codec(t *testing.T) {
+	tests := []struct {
+		Label          string
+		setPreferences bool
+	}{
+		{
+			Label:          "NoSetCodecPreferences",
+			setPreferences: false,
+		},
+		{
+			Label:          "SetCodecPreferences",
+			setPreferences: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.Label, func(t *testing.T) {
+			pc, err := NewPeerConnection(Configuration{})
+			assert.NoError(t, err)
+
+			transceiver, err := pc.AddTransceiverFromKind(
+				RTPCodecTypeVideo,
+				RTPTransceiverInit{
+					Direction: RTPTransceiverDirectionRecvonly,
+				},
+			)
+			assert.NoError(t, err)
+
+			if test.setPreferences {
+				codec := RTPCodecCapability{
+					"video/vp8", 90000, 0, "", nil,
+				}
+
+				err = transceiver.SetCodecPreferences(
+					[]RTPCodecParameters{
+						{
+							RTPCodecCapability: codec,
+						},
+					},
+				)
+				assert.NoError(t, err)
+			}
+
+			offer, err := pc.CreateOffer(nil)
+			assert.NoError(t, err)
+
+			assert.Equal(t, true, strings.Contains(offer.SDP, "apt=96"))
+			assert.NoError(t, pc.Close())
+		})
+	}
+}


stephanrotolante:
broken test cases.... I am looking into that

rtpcodec.go
```diff
@@ -155,6 +155,29 @@ func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) Paylo
 	return PayloadType(0)
 }
 
+// Given a RTPCodecParameters check if it exists in a list of other RTPCodecParameters.
+func doesCodecExistInList(codec RTPCodecParameters, codecList []RTPCodecParameters) bool {
+	for _, c := range codecList {
+		if codec.SDPFmtpLine == c.SDPFmtpLine && codec.MimeType == c.MimeType {
```
JoeTurki, line 161: We match codecs with different rate or channels now, can we use the fmtp helpers that Alessandro's made in his PR instead, for the sake of behavioral consistency?

https://github.com/pion/webrtc/blob/e4ff415b2bff31382bdb803b62e8b08d6d000911/internal/fmtp/fmtp.go#L169

stephanrotolante: sure, let me see what I can do

stephanrotolante: now that I review the code, I wonder if I should just use `codecParametersFuzzySearch`

JoeTurki: Yeah if you can use `codecParametersFuzzySearch` that would be better!

JoeTurki: Thank you

itzmanish:
Hi, Please correct me if my understanding of the issue is wrong. 

This PR fixes the scenario where someone sets CodecPreferences with one codec, and it disables the RTX for that codec. Right? 

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences#codecs), the parameter for `setCodecPreferences` is a list of codecs. Pion's API is the same as well. So my question is, why should pion try to find the RTX codec from mediaengine instead of the user providing the list of codecs with the RTX codec in `setCodecPreferences`?

stephanrotolante:
@itzmanish that is a great question and I thought about that as well while making this PR

If we want the RTX codec when using `setCodecPreferences`, why not just specify it?

> why should pion try to find the RTX codec from mediaengine instead of the user providing the list of codecs with the RTX codec > in setCodecPreferences?

I guess you could say this is a pion specific feature, that will handle this for you..... 

Maybe I should ask the author of the issue why they cannot just pass in the RTX codec :thinking:

jech:
Sure, no problem.  However, since that would be a divergence from the JavaScript API, it would need to be clearly documented.

If you choose to do that, please add a test to make sure that a codec with an associated RTX track can be specified, since I'm afraid that's code that might rot and since things will still sort of work if the RTX track is incorrect, we might not notice that it's broken.

stephanrotolante:
@JoeTurki @itzmanish 

What do you think? Should we stay aligned with the nodejs api or do our own thing??

I am thinking we stay aligned

jech:
> Should we stay aligned with the javascript api or do our own thing?

The JavaScript API is fairly high level.  Pion is useful for low-level software, such as SFUs, so it might make sense to give more control to the programmer.  But perhaps it's not worth the extra complexity in this particular case.

In other words, I have no clue.

itzmanish:
> Pion is useful for low-level software, such as SFUs, so it might make sense to give more control to the programmer. But perhaps it's not worth the extra complexity in this particular case.

This ^

In my view, the reward for introducing extra complexity and logic is unfulfilling. The more pion/webrtc operates behind the scene logic, the more unpredictable the API will be, and it will be more challenging for users using the library.

JoeTurki:
I too think we should stay aligned!

jech:
It's a little bit more complicated than that.

When using SetCodecPreferences in Pion, we set the ptype, and in particular we can set a ptype that is not defined in the mediaengine.  If we do that, there is no way for Pion to choose the right ptype for the RTX track.

This is not the case in the JavaScript case, where we do not set the ptype.  See here: <https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences>

So the API is already not aligned with the JavaScript case, and not being explicit will cause the API to fail for mysterious reasons.

stephanrotolante:
@Sean-Der Do you have any suggestions?

stephanrotolante:
```
let peerConnection = new RTCPeerConnection();

  const videoCodecs = RTCRtpSender.getCapabilities("video").codecs;

  const transceiver = peerConnection.addTransceiver('video', { direction: "recvonly" });

  const codecs = RTCRtpSender.getCapabilities("video").codecs;

  const preferredCodecs = codecs.filter(c => c.mimeType === "video/VP8");

  transceiver.setCodecPreferences(preferredCodecs);

  peerConnection.createOffer().then(offer => {
    console.log(offer)
    return peerConnection.setLocalDescription(offer);
  });
  ```
  
  ```
v=0
o=- 219046925391310381 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0
a=extmap-allow-mixed
a=msid-semantic: WMS
m=video 9 UDP/TLS/RTP/SAVPF 96
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:uvYX
a=ice-pwd:Jm/Z2PV2vCIuKg+XG7owMHj9
a=ice-options:trickle
a=fingerprint:sha-256 A5:BA:59:DE:81:84:71:C9:D9:CE:1E:EE:5E:7A:B9:8E:9F:A3:D5:CD:E3:85:EF:DA:D8:94:A6:2C:8B:99:ED:C4
a=setup:actpass
a=mid:0
a=extmap:1 urn:ietf:params:rtp-hdrext:toffset
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 urn:3gpp:video-orientation
a=extmap:4 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:5 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:6 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:8 http://www.webrtc.org/experiments/rtp-hdrext/color-space
a=extmap:9 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:10 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:11 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
a=recvonly
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:96 VP8/90000
a=rtcp-fb:96 goog-remb
a=rtcp-fb:96 transport-cc
a=rtcp-fb:96 ccm fir
a=rtcp-fb:96 nack
a=rtcp-fb:96 nack pli
```

PR #3064, opened 2025-03-13
Feat rtp processor #3063

### LINKED ISSUE #3063: Pion repeatedly sends the same RTCP (NACK) packet even after successfully receiving and processing the corresponding RTX packet.
**Environment**

- Version: v4.0.12
- OS: Zorin OS 17.1 Core

 **What I Did**

I simulated packet loss by generating UDP packet drops, triggering RTCP (NACK) requests in Pion WebRTC with RTX enabled.

**Expected Behavior**

The RTCP (NACK) requests should stop once the missing packet is successfully received via RTX, preventing further redundant requests.

**Actual Behavior**

Pion continuously sends RTCP (NACK) requests for the same packet, even after the RTX packet is received and processed.

This happens because RTX packets are only logged in the receiver log of the RTX track, not the corresponding RTP track. 

The tight coupling of the RTPReader within the interceptor prevents RTX packets from passing through the correct interceptor chain for the RTP track.

Image
Image
Image

**Root Cause**

The ReadRTP function reads an RTP packet and passes it through the interceptor chain.
However, when an RTX packet is received, it isn’t processed through the corresponding RTP track’s interceptor chain. 
As a result, it isn’t logged in the RTP track’s receiver log, causing the system to treat the packet as missing and repeatedly issue NACKs.

**Proposed Fix**

To resolve this, I propose separating the reading and processing of RTP packets:

Read RTP: Handle packet reading independently to avoid tight coupling with the interceptor.

Process RTP: Pass the packet (including converted RTX packets) through the interceptor chain explicitly.

Currently, RTX packets are converted to RTP packets but not passed through the interceptor chain. By splitting the process, the converted RTX packets can be processed through the appropriate interceptors and logged correctly in the RTP track’s receiver log.

Image
Image
Image

arjunshajitech:
This is part 2/2 of a cross-repository series.

Part 1 :  https://github.com/pion/interceptor/pull/317

#### Note
Before merging, make sure to remove:

```
replace github.com/pion/interceptor v0.1.37 => github.com/arjunshajitech/interceptor v0.0.0-20250312123050-c9d476a510a1
```

#### Description

Use RTPProcessor with RTPReader
RTPReader simply reads packets from the buffer, while the processor retains them through the interceptors.

#### Reference issue
Fixes [#3063](https://github.com/pion/webrtc/issues/3063)


JoTurk:
@arjunshajitech we started the plan for `v5` do you want to break the API and fix this correctly? we have a `next` branch now, I can suggest a few ways to do it too

https://github.com/pion/webrtc/issues/3561

PR #3075, opened 2025-03-31
FEC mimeType

3DRX:
#### Description

Add const mime type for flexfec03, flexfec20 and ulpfec (flexfec03 and ulpfec is currently supported by most browsers, and pion is currently in the process of getting flexfec03 and flexfec20 implemented).

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -54,6 +54,12 @@ const (
 	// MimeTypeFlexFEC FEC MIME Type
 	// Note: Matching should be case insensitive.
 	MimeTypeFlexFEC = "video/flexfec"
+	// MimeTypeFlexFEC03 FlexFEC03 MIME Type
+	// Note: Matching should be case insensitive.
+	MimeTypeFlexFEC03 = "video/flexfec-03"
+	// MimeTypeUlpFEC UlpFEC MIME Type
+	// Note: Matching should be case insensitive.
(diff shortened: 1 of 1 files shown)


JoeTurki (commented):
My concern is that both `video/flexfec-03` and `video/flexfec-20` are not officially registered with IANA, and I haven't seen any implementation that uses `video/flexfec-20`, where is that used and why not `video/flexfec` for `video/flexfec-20`?

I invited you to the org so you can branches directly, and run the CI!

3DRX:
Thanks for adding me as a member! I'm only adding video/flexfec-20 here because https://github.com/pion/interceptor/tree/master/pkg/flexfec have a wip implementation there. I think you are right about just using `video/flexfec` for flexfec20, but perhaps `video/flexfec-03` is necessary for now, since chromium will only consider decoding using flexfec when the mimeType is set to `video/flexfec-03` in my own test. (If sending flexfec03 using mimeType `video/flexfec-03`, chromium is able to detect that it's fec, but the entry `fecBytesReceived` of stat type inbound-rtp for video is always undefined, which to me sounds like it's ignoring packets with video/flexfec mimeType).

JoeTurki:
Sure i think adding `video/flexfec-03` is fine because chrome uses it!

JoeTurki (approved):
Thank you Jingyang

PR #3078, opened 2025-04-02
Fix pion2pion example readme

amanakin:
#### Description
This PR updates the pion-to-pion example README to include `@latest`, aligning it with other examples.
To install the module without being inside the WebRTC repository, explicitly specifying the version is required.

For reference, see other examples, e.g., [reflect example](https://github.com/amanakin/webrtc/tree/master/examples/reflect).


JoeTurki (approved):
Thank you.

PR #3085, opened 2025-04-07
Fix MIME type comparisons to be case-insensitive

San9H0:
#### Description

The `RegisterCodec` function is public and takes the mimeType as a string. Therefore, it should compare values using strings.EqualFold.
In general, MIME type comparisons should be done in a case-insensitive manner.

#### Reference issue
Fixes #...

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -240,7 +240,7 @@ func (m *MediaEngine) RegisterDefaultCodecs() error {
 func (m *MediaEngine) addCodec(codecs []RTPCodecParameters, codec RTPCodecParameters) ([]RTPCodecParameters, error) {
 	for _, c := range codecs {
 		if c.PayloadType == codec.PayloadType {
-			if c.MimeType == codec.MimeType &&
+			if strings.EqualFold(c.MimeType, codec.MimeType) &&
 				fmtp.ClockRateEqual(c.MimeType, c.ClockRate, codec.ClockRate) &&
 				fmtp.ChannelsEqual(c.MimeType, c.Channels, codec.Channels) {
 				return codecs, nil
@@ -798,7 +798,7 @@ func payloaderForCodec(codec RTPCodecCapability) (rtp.Payloader, error) {
 
 func (m *MediaEngine) isRTXEnabled(typ RTPCodecType, directions []RTPTransceiverDirection) bool {
 	for _, p := range m.getRTPParametersByKind(typ, directions).Codecs {
-		if p.MimeType == MimeTypeRTX {
+		if strings.EqualFold(p.MimeType, MimeTypeRTX) {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


Sean-Der:
This looks great to me @San9H0!

would you mind adding a simple test? Just to make sure no one breaks your fix

San9H0:
Thanks for the review! I've added a test case to verify the case-insensitive MIME type comparison. Let me know if you need any changes.

JoeTurki (approved):

JoeTurki:
Thank you :)

mediaengine.go
```diff
@@ -820,7 +820,7 @@ func (m *MediaEngine) isRTXEnabled(typ RTPCodecType, directions []RTPTransceiver
 
 func (m *MediaEngine) isFECEnabled(typ RTPCodecType, directions []RTPTransceiverDirection) bool {
 	for _, p := range m.getRTPParametersByKind(typ, directions).Codecs {
-		if strings.Contains(p.MimeType, MimeTypeFlexFEC) {
+		if strings.Contains(strings.ToLower(p.MimeType), MimeTypeFlexFEC) {
```
JoeTurki, line 823: Can we also test this?

mediaengine.go
```diff
@@ -810,7 +810,7 @@ func payloaderForCodec(codec RTPCodecCapability) (rtp.Payloader, error) {
 
 func (m *MediaEngine) isRTXEnabled(typ RTPCodecType, directions []RTPTransceiverDirection) bool {
 	for _, p := range m.getRTPParametersByKind(typ, directions).Codecs {
-		if p.MimeType == MimeTypeRTX {
+		if strings.EqualFold(p.MimeType, MimeTypeRTX) {
```
JoeTurki, line 813: Can we also test this?

San9H0: Of course. I've added test cases for both isRTXEnabled and isFECEnabled functions to verify the case-insensitive MIME type comparisons.

San9H0:
OK. It's all done on my side. Let me know if anything else is needed

JoeTurki (approved):

3DRX:
Can we call `strings.ToLower` when any MIME type is passed into media engine? Then for future

3DRX:
Can we call `strings.ToLower` when MIME type is passed into media engine? Then internally it will always be lowercase and we don't have to worry about case insensitiveness. I'm thinking about this because if future changes adds more MIME type comparison, more tests on MIME type comparison needs to be added.

(Mistakenly closed this for a second because I hit the wrong button when I tried to comment, sorry about that)

JoeTurki:
> Can we call strings.ToLower when MIME type is passed into media engine? Then internally it will always be lowercase and we don't have to worry about case insensitiveness. I'm thinking about this because if future changes adds more MIME type comparison, more tests on MIME type comparison needs to be added.

I agree that would make much more sense

PR #3091, opened 2025-04-12
Added missing lock

sirzooro:
My tests executed with race detector enabled found that code tries to read value of `pc.sctpTransport.dataChannelsRequested` without lock.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2880,6 +2880,9 @@ func (pc *PeerConnection) generateMatchedSDP(
 		}
 	}
 
+	pc.sctpTransport.lock.Lock()
+	defer pc.sctpTransport.lock.Unlock()
+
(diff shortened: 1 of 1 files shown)


edaniels (approved):

JoeTurki (approved):
Makes sense, Also matches `generateMatchedSDP` would be cool if we're able to make a test for it.

sirzooro:
> Makes sense, Also matches `generateMatchedSDP` would be cool if we're able to make a test for it.

I am not sure if we need any new test here, this function is called during normal offer/answer exchange. In my case I had problem with incorrect handling of incoming data channels in my code and sometimes initial message which should be received just after accepting it was lost. My test was checking that messages are no longer lost. It was ran in a loop many times to verify fix and at some point race detector found this bug.

Issue #2994, opened 2025-01-07
No way to use the RTX SSID of a TrackLocalStaticRTP


jech:
I have a `TrackLocalStaticRTP`, and I have successfully negotiated an associated RTX SSID.  Unfortunately, there is no way I can find to write a packet on the associated RTX SSID.

(The obvious workaround would be to use an interceptor, but this is difficult in my case, since I'm sharing a single NACK buffer between multiple PeerConnections, and there is no way to pass the packet buffer to the interceptor.  It could probably be hacked around, but it seems simpler to provide a direct handle to the RTX SSID.)

jech:
The obvious solution is to add a method `WriteRTX` to `TrackLocalStaticRTP`, which I'm quite willing to do.

Any better ideas?

joeturki:
I think adding it to `TrackLocalStaticRTP` maybe wouldn't be a good abstraction.

jech:
Joe, we're only just getting to grips with v4.  If you change everything, you're going to lose a lot of people.

joeturki:
@jech no, I'm not going to make any breaking changes

joeturki:
I'm just saying we should add more tooling to handle RTX streams, including exposing an RTXWriter type and an RTPPacketHistory type for example. these additions shouldn't break anything, 

I think just adding `WriteRTX` and `WriteRTPRTX` would be easiest fix for this issue but we should have a better RTX abstraction. 

@jech what do you think?

jech:
The RTXWriter suggestion seems fairly uncontroversial.  For RTPPacketHistory, please see https://github.com/jech/galene/tree/master/packetcache, and please make sure it can be adapted to your interface.

joeturki:
Thank you, I'm going to read this, I think i will make a proposal for the suggestion (no code, just design), before I implement anything :)

jech:
Please see https://github.com/pion/interceptor/pull/297 and https://github.com/pion/webrtc/pull/3002

While that series is aimed at TWCC, it would easily allow me to put my RTX code into an ad hoc interceptor, and therefore solve this problem.

JoTurk:
Sorry for missing track on this on, I'll try to start working on it soon.

PR #3095, opened 2025-04-13
Automatically configure RTX codecs

3DRX:
#### Description

Add a setting engine flag to enable automatically configure RTX codecs in `MediaEngine`.

1. NACK codec only added to video codecs that have Type: "nack" in it's RTCPFeedback, for example:
```go
RTCPFeedback: []RTCPFeedback{
	{Type: "nack", Parameter: ""},
	{Type: "nack", Parameter: "pli"},
},
```
2. Automatically added NACK codec's payload type is original video's payload type + 1 (a common behavior).
3. If original video's payload type + 1 is already taken, an error is returned in `api.NewPeerConnection`.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -273,6 +273,44 @@ func (m *MediaEngine) RegisterCodec(codec RTPCodecParameters, typ RTPCodecType) 
 	return err
 }
 
+func (m *MediaEngine) autoConfigRTXCodecs() error {
+	additionalRTXCodecs := []RTPCodecParameters{}
+	for _, codec := range m.videoCodecs {
+		// ignore FEC & RTX
+		if strings.Contains(codec.MimeType, MimeTypeFlexFEC) || codec.MimeType == MimeTypeRTX {
+			continue
+		}
+		haveNACK := false
+		for _, fb := range codec.RTCPFeedback {
+			if fb.Type == "nack" {
+				haveNACK = true
+
+				break
+			}
+		}
+		if haveNACK {
+			additionalRTXCodecs = append(additionalRTXCodecs, RTPCodecParameters{
+				RTPCodecCapability: RTPCodecCapability{
+					MimeType:     MimeTypeRTX,
+					ClockRate:    90000,
+					Channels:     0,
+					SDPFmtpLine:  fmt.Sprintf("apt=%d", codec.PayloadType),
+					RTCPFeedback: nil,
+				},
+				PayloadType: codec.PayloadType + 1,
+			})
+		}
+	}
+	for i := range additionalRTXCodecs {
+		err := m.RegisterCodec(additionalRTXCodecs[i], RTPCodecTypeVideo)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // RegisterHeaderExtension adds a header extension to the MediaEngine
 // To determine the negotiated value use `GetHeaderExtensionID` after signaling is complete.
 //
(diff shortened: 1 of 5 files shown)


mediaengine.go
```diff
@@ -245,6 +245,44 @@ func (m *MediaEngine) RegisterCodec(codec RTPCodecParameters, typ RTPCodecType)
...
+		if haveNACK {
+			additionalRTXCodecs = append(additionalRTXCodecs, RTPCodecParameters{
+				RTPCodecCapability: RTPCodecCapability{
+					MimeType:     MimeTypeRTX,
+					ClockRate:    90000,
+					Channels:     0,
+					SDPFmtpLine:  fmt.Sprintf("apt=%d", codec.PayloadType),
+					RTCPFeedback: nil,
+				},
+				PayloadType: codec.PayloadType + 1,
```
JoeTurki, line 272: This can cause conflicts, no?

JoeTurki (commented):
My primary concerns about this pull request are:

We should only introduce new settings if they improve behavior by aligning more closely with the specification, and we can't break existing behavior. or a options that cannot be configured otherwise for example, SetNAT1To1IPs.

I think we should try to minimize the number of settings to keep the API as compatible with the JavaScript API as possible, especially considering that i think most users are unaware of the existence of the settingsEngine :)

I'm not convinced where this change would fit, and it appears to make assumptions about how the other side implements RTX and NACK

What do you think @Sean-Der @jech 

Thank you so much @3DRX

mediaengine.go
3DRX: Yes, currently this PR’s behavior when conflict occurs is to return an error in api.NewPeerConnection.

3DRX:
@JoeTurki I understands your concerns, if it’s not that straightforward and beneficial to pion/webrtc, do you think it’s a good idea to have this behavior to be enabled when user is using pion/mediadevices? Although the spec doesn’t describe anything about setting RTX by default, this can make the default behavior closer to libwebrtc.

3DRX:
> and it appears to make assumptions about how the other side implements RTX and NACK

I think this shouldn’t be a problem, since the rtx is added before sdp negotiations, and even if a rtx added when remote doesn’t support rtx and nack, the rtx track will always be empty(since no nack packets generated and received, no rtx packets will be send).

Sean-Der:
@3DRX I am going to start an issue. I think we can default this to on! We have multiple issues and I want to make sure we get things right (not break something and regret it)

* A RTX doesn't stop Pion from sending NACKs
* RTX isn't enabled by default (users registering codecs, and not dedicated RTX)
* Congestion Controllers are hard to associate with a a PeerConnection
* More? We have so many Interceptor related issues right now I am losing track :)

My goal
* Fix these issues
* Not break existing users
* Have things on by default, otherwise only power users benefit
* Don't make a decision we regret. I have had a few designs in Pion that I regret because I went too fast.

3DRX:
Thanks for reviewing this @JoeTurki :)

3DRX:
@Sean-Der I’d like to help on fixing the current issues! It seems to only affect pion as a receiver, this PR is all about pion as a sender. I think the issue (A RTX doesn't stop Pion from sending NACKs) won’t block this PR’s work.

3DRX:
There’ve been quite a lot going on here: https://github.com/pion/webrtc/issues/3063. @arjunshajitech did a refactor to fix the issue, but it breaks current interceptor interface. Don’t know if it’s possible to change it at this point.

jech:
> when remote doesn’t support rtx and nack,

What happens if the remote supports NACK but not RTX?  That's the case of most Pion applications.

jech:
> More? We have so many Interceptor related issues right now I am losing track :)

To be fair, we're trying to achieve something difficult: a library that is both flexible enough to be used for low-level stuff (such as highly efficient SFUs) and easy to use for typical applications.  Since we're exposing both low- and high-level functionality, we need to get the layering just right.

There's also https://github.com/pion/webrtc/issues/2994.

Issue #2898, opened 2024-09-05
Advertising / Reading ICE trickle support


delthas:
TL;DR, this is the issue for tracking the TODO at https://github.com/pion/sdp/blob/master/jsep.go#L91

ICE trickling helps speed up the connection establishment. For it to work though, the receiving side should support it. If we send ICE candidates after the SDP payload for clients who do not support ICE trickling, these candidates would be ignored; we could even send an offer without any candidates and the call would fail entirely.

To know whether the other side supports ICE trickling:
- The support can be advertised through the signaling channel (which will work without this ticket)
- The support can be advertised through an SDP ICE option, `a=ice-options:trickle`

https://datatracker.ietf.org/doc/html/rfc8840#section-4 specifically mentions:

> Trickle ICE Agents MUST indicate support for Trickle ICE by including the ice-option "trickle" within all SDP Offers and Answers in accordance to [[RFC8838](https://datatracker.ietf.org/doc/html/rfc8838)].

---

To do this, we probably need to add an "ICETricklingSupported" bool option to webrtc.OfferAnswerOptions that the user can set. If set, the `a=ice-options:trickle` attribute is added to the generated SDP payload.

In addition, we would need to be able to know if the remote side sent the attribute, corresponding to https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/canTrickleIceCandidates in the Web spec. This would probably be a method on SessionDescription?

Sean-Der:
Hey @delthas, Sorry for the delay.

It would be trivial adding `a=ice-options:trickle` to the Offer/Answer (from Pion). Are you interested in doing that? We shouldn't just uncomment the line in `jsep.go`, but instead have pion/webrtc pass it in somehow. 

We don't care if the remote trickles or not! I don't believe any logic needs to be added as a receiver. 

If you aren't interested in contributing tell me and I can have this fixed in 30 minutes :)

ValorZard:
Looking at the latest RFC, why is it called ice2? 

![Image](https://github.com/user-attachments/assets/ed32184b-7151-4eb8-8e75-078d1939c5f5)

JoeTurki:
@ValorZard This is the ice version.

>  While this specification formally relies on [RFC8445], at the time of
   its publication, the majority of WebRTC implementations support the
   version of ICE described in [RFC5245].  The "ice2" attribute defined
   in [RFC8445] can be used to detect the version in use by a remote
   endpoint and to provide a smooth transition from the older
[...]

PR #3097, opened 2025-04-13
Add Advertising / Reading ICE trickle support

ValorZard:
#### Description
Figured I'd try my hand at dealing with this. I'll be honest and say I don't FULLY understand what this is for, since pion already supports trickle ICE, but since this IS part of the spec, might as well.

This is nowhere near done (I will also have to do a pull request to the SDP repo to finish the feature), but I do have some questions

- [ ] the "a=ice-options:tickle" seem to be related to ice2? Whatever that is (see https://datatracker.ietf.org/doc/html/rfc9725#section-4.3.3) Does that mean I have to tack on ice2 to the attribute?
- [ ] how exactly would i add the attribute? I guess I just add a constant here and be done with it https://github.com/pion/sdp/blob/c395deb80b693c4a4f5b28d1ce435f27fa2bc40e/jsep.go#L14

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/2898

### DIFF
--- a/offeransweroptions.go
+++ b/offeransweroptions.go
@@ -9,6 +9,10 @@ type OfferAnswerOptions struct {
 	// VoiceActivityDetection allows the application to provide information
 	// about whether it wishes voice detection feature to be enabled or disabled.
 	VoiceActivityDetection bool
+	// ICETricklingSupported indicates whether the ICE agent should use trickle ICE
+	// If set, the "a=ice-options:trickle ice2" attribute is added to the generated SDP payload.
+	// (See https://datatracker.ietf.org/doc/html/rfc9725#section-4.3.3)
+	ICETricklingSupported bool
 }
 
 // AnswerOptions structure describes the options used to control the answer


ValorZard:
I think this is ready to be reviewed, pretty sure this is the api changes needed

offeransweroptions.go
```diff
@@ -9,6 +9,10 @@ type OfferAnswerOptions struct {
 	// VoiceActivityDetection allows the application to provide information
 	// about whether it wishes voice detection feature to be enabled or disabled.
 	VoiceActivityDetection bool
+	// ICETricklingSupported indicates whether the ICE agent should use trickle ICE
+	// If set, the "a=ice-options:trickle ice2" attribute is added to the generated SDP payload.
```
nils-ohlmeier, line 13: I think the general idea is right here. But do NOT put `ice2` into the `ice-options` here. Pion doesn't support ice2 yet. That is whole different big project.
For this issue and PR we should only add `a=ice-options:trickle`.

ValorZard: removed! thanks so much

JoTurk:
@ValorZard do you want to merge this for v5.2 or should it wait for the next release?

ValorZard:
Uh... im down either way

JoTurk:
@ValorZard Okay, Cool, thank you, I'll update this and get it ready for 5.2.

JoeTurki (approved):
Thank you, I added a test, and fixed two small issues. merging it after the CI runs.

PR #3010, opened 2025-01-16
feat: add H265 payloader and bump RTP dependency

strombergdev:
#### Add H265 payloader support to mediaengine using Pion RTP v1.8.11.

PR #3102, opened 2025-04-22
feat(mediaengine): add h265 to default codecs

andresmanelli:
#### Description

H265 is not in the default codecs list. I propose adding it since the payloader was introduced in #3010.

I wasn't sure if `SDPFmtpLine` is necessary, since I see the defaults are negotiated without it.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -173,6 +173,14 @@ func (m *MediaEngine) RegisterDefaultCodecs() error {
 			PayloadType: 39,
 		},
 		{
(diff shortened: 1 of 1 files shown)


Sean-Der:
Let's do it @andresmanelli! Thank you

Sean-Der:
Moved to #3103 I had to add RTX, thank you for this PR @andresmanelli

andresmanelli:
> Moved to #3103 I had to add RTX, thank you for this PR @andresmanelli 

Thank you !

PR #3103, opened 2025-04-22
Add H265 to RegisterDefaultCodecs

Sean-Der:
H265 will be available by default now


cedricve:
Trying this out as Chrome now release (136), H265 by default. As H265 is also added to the default codecs, I was wondering this now officially works through browser?
https://chromestatus.com/feature/5153479456456704

PR #3113, opened 2025-04-27
Reject candidates from old generation

### LINKED ISSUE #2511: AddICECandidate might not verify ufrag
Quoting https://w3c.github.io/webrtc-pc/#dom-peerconnection-addicecandidate:

> 5. If candidate.usernameFragment is not null, and is not equal to any username fragment present in the corresponding media description of an applied remote description, return a promise rejected with a newly created OperationError.

However, I cannot find the place in the Pion code where we check that that ufrag matches the SDP, so I suspect that we're wrongly accepting candidates from the wrong generation.

### LINKED ISSUE #2993: Pion ignores the ice-ufrag of candidates
I've had this on my todo list for a while, but now that @joeturki seems to be cleaning up the code, it's a good time to write it up.

At https://w3c.github.io/webrtc-pc/#dom-peerconnection-addicecandidate, there's the following language:

> 5. If candidate.usernameFragment is not null, and is not equal to any username fragment present in the corresponding media description of an applied remote description, return a promise rejected with a newly created OperationError.

This requirement ensures that a remote candidate that was sent before renegotiation but arrives after a renegotiation is not erroneously applied to the wrong session.  An example of a protocol that can cause such reordering is WHIP.

As far as I can tell, Pion does not obey this requirement, the ufrag is simply ignored.

JoeTurki:
#### Description

Return an error if a candidate with a username fragment that does not match the username fragment in the remote description is added. This usually indicates that the candidate was generated before the renegotiation.

Last time I tested Chrome silently drops such candidates, firefox returns an error.

Maybe we should also handle "generation" extension.

#### Reference issue
Fixes #2993 
Fixes #2511

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1962,33 +1962,65 @@ func (pc *PeerConnection) RemoteDescription() *SessionDescription {
 // AddICECandidate accepts an ICE candidate string and adds it
 // to the existing set of candidates.
 func (pc *PeerConnection) AddICECandidate(candidate ICECandidateInit) error {
-	if pc.RemoteDescription() == nil {
+	remoteDesc := pc.RemoteDescription()
+	if remoteDesc == nil {
 		return &rtcerr.InvalidStateError{Err: ErrNoRemoteDescription}
 	}
 
 	candidateValue := strings.TrimPrefix(candidate.Candidate, "candidate:")
 
-	var iceCandidate *ICECandidate
-	if candidateValue != "" {
-		candidate, err := ice.UnmarshalCandidate(candidateValue)
-		if err != nil {
-			if errors.Is(err, ice.ErrUnknownCandidateTyp) || errors.Is(err, ice.ErrDetermineNetworkType) {
-				pc.log.Warnf("Discarding remote candidate: %s", err)
-
-				return nil
-			}
-
-			return err
-		}
-
-		c, err := newICECandidateFromICE(candidate, "", 0)
-		if err != nil {
-			return err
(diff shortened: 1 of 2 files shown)


Sean-Der:
I bet some users have code that tears down the PeerConnection entirely on error. This change could trigger that :/

I agree with rejecting candidates and doing a `logger.Error` I just don't want to negatively impact anyone who is just auto upgrading.

Or is it doing wrong by users to not return an error?

JoeTurki:
> I agree with rejecting candidates and doing a logger.Error I just don't want to negatively impact anyone who is just auto upgrading.
> Or is it doing wrong by users to not return an error?

This makes sense, firefox implementation around generations seems extreme, and returning errors already broke few of the tests :)

JoeTurki:
Okay the error is gone, we still test it, I had to do a refactor to make the linter happy. However I don't know how I feel about this change overall, It aligns with other clients, and it shouldn't break any other client but I'm still not confident.

Sean-Der (approved):

JoeTurki:
> Does accepting candidates before the first setRemoteDescription still have the same behavior. I have seen people on that behavior.

- First `setRemoteDescription` call:  The behavior remains the same: we return  `&rtcerr.InvalidStateError{Err: ErrNoRemoteDescription}`.

- Second `setRemoteDescription` call:  Here I think the behavior changes, but in a way that should prevent more problems than it  might cause. we avoid accidentally applying late ICE candidates, like with @jech issue

PR #3116, opened 2025-05-01
Solve data race in ReadSimulcast

cnderrauber:
TrackEndcodings could change when reading

#### Reference issue
Fixes #...

### DIFF
--- a/rtpsender.go
+++ b/rtpsender.go
@@ -437,11 +437,16 @@ func (r *RTPSender) ReadRTCP() ([]rtcp.Packet, interceptor.Attributes, error) {
 func (r *RTPSender) ReadSimulcast(b []byte, rid string) (n int, a interceptor.Attributes, err error) {
 	select {
 	case <-r.sendCalled:
+		r.mu.Lock()
 		for _, t := range r.trackEncodings {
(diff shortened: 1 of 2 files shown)


rtpsender.go
```diff
@@ -437,11 +437,16 @@ func (r *RTPSender) ReadRTCP() ([]rtcp.Packet, interceptor.Attributes, error) {
 func (r *RTPSender) ReadSimulcast(b []byte, rid string) (n int, a interceptor.Attributes, err error) {
 	select {
 	case <-r.sendCalled:
+		r.mu.Lock()
 		for _, t := range r.trackEncodings {
 			if t.track != nil && t.track.RID() == rid {
-				return t.rtcpInterceptor.Read(b, a)
+				reader := t.rtcpInterceptor
```
boks1971, line 443: does `Read` also need it? Maybe for consistency?

boks1971 (approved):

rtpsender.go
cnderrauber: Read will block so can't put it under the lock. It is ok since the same RID encoding has very low chance to change and read a previous rtcp packet does not matter even if it is changed

PR #3118, opened 2025-05-04
make certificate.go partially available on wasm

talentlessguy:
#### Reference issue

(partially) Fixes https://github.com/libp2p/go-libp2p/issues/3277

Go WASM interop supports x509, PEM and other cryptography functions (but not all), so there is no reason not to include those in the wasm build.

This will also fix compilation errors for when something is using webrtc and targets both wasm and server

The PR was tested locally (via a simple web folder with wasm_exec.js and index.html) and I confirmed being able to generate a X509 certificate

### DIFF
--- a/certificate.go
+++ b/certificate.go
@@ -1,9 +1,6 @@
 // SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
 // SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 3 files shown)


Sean-Der (approved):

Sean-Der:
Thank you so much for adding the tests!

PR #3125, opened 2025-05-19
Fix trackDetailsFromSDP not handling fec ssrc

### LINKED ISSUE #3109: trackDetailsFromSDP should also black list FEC ssrc
## The Problem

Unified plan SDP can contain FEC and RTX tracks in the same m= with video track, https://datatracker.ietf.org/doc/html/draft-roach-mmusic-unified-plan-00#section-4.5 .
But currently, pion's implementation thinks it's plan B sdp if FEC track is included.

## The Cause

It seems like the problem is in function `trackDetailsFromSDP` of sdp.go, it only black lists rtx ssrc, but not fec ssrc.

## The Fix

We should also black list fec ssrc in `trackDetailsFromSDP`. And also test `descriptionIsPlanB` against examples in https://datatracker.ietf.org/doc/html/draft-roach-mmusic-unified-plan-00#section-4 .

3DRX:
fix #3109 

TODO
- [x] Add test
- [x] Use consts defined in https://github.com/pion/sdp/pull/210, instead of using string literals

### DIFF
--- a/sdp_test.go
+++ b/sdp_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pion/sdp/v3"
 	"github.com/pion/transport/v3/test"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestExtractFingerprint(t *testing.T) {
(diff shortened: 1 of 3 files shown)


3DRX:
I started to work on FEC in pion as a receiver, for now this PR doesn't provide any additional capabilities (and hopefully doesn't break anything too), but imo removes one of the road blocks toward the goal.

sdp_test.go
```diff
@@ -454,6 +454,44 @@ func TestTrackDetailsFromSDP(t *testing.T) {
...
+		if track.rtxSsrc == nil {
+			assert.Fail(t, "missing RTX ssrc for video track")
+		} else {
+			assert.Equal(t, SSRC(4000), *track.rtxSsrc)
+		}
+		if track.fecSsrc == nil {
+			assert.Fail(t, "missing FEC ssrc for video track")
+		} else {
+			assert.Equal(t, SSRC(5000), *track.fecSsrc)
+		}
```
aalekseevx, line 492: maybe use [require.NotNil](https://pkg.go.dev/github.com/stretchr/testify/require#NotNil)?

aalekseevx (approved):
Looks good to me!

sdp_test.go
aalekseevx: @3DRX, I think there's a panic possible now as `assert` doesn't stop execution. wdyt of using require.NotNil?

3DRX: Oops, my bad, it's fixed now.

PR #3129, opened 2025-05-22
Use space after WMS. Fixes #3128

### LINKED ISSUE #3128: Incorrect MSID semantic in the SDP
### Your environment.
  - Version: v4.1.1
  - Browser: -
  - Other Information: - 

### What did you do?

I'm using a python WebRTC client written using [aiortc/aiortc](https://github.com/aiortc/aiortc) to connect to a pion SFU. I noticed that the [TrackID parsing from the SDP in aiortc](https://github.com/aiortc/aiortc/blob/main/src/aiortc/sdp.py#L562-L571) was failing because of the MSID semantic used in the SDP generated by pion. Specifically, [pion is generating an SDP](https://github.com/pion/webrtc/blob/v4.1.1/peerconnection.go#L2721) with `a=msid-semantic:WMS*` whereas aiortc expects it to be `a=msid-semantic:WMS *` (with a space between `WMS` and `*`).

Here's the [relevant section from the RFC](https://datatracker.ietf.org/doc/html/draft-ietf-mmusic-msid-08#section-3) -
```
The name of the attribute is "msid-semantic".

   The value of the attribute is given by the following ABNF:

     msid-semantic-value = msid-semantic msid-list
     msid-semantic = token ; see [RFC 4566](https://datatracker.ietf.org/doc/html/rfc4566)
     msid-list = *(" " msid-id) / " *"
```

### What did you expect?
Expected `a=msid-semantic:WMS *` in the SDP

### What happened?
Got `a=msid-semantic:WMS*` in the SDP

nils-ohlmeier:
#### Description
This PR adds a space between `WMS` and `*`.

#### Reference issue
Fixes #3128

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2718,7 +2718,7 @@ func (pc *PeerConnection) generateUnmatchedSDP(
 	if err != nil {
 		return nil, err
 	}
-	desc.Attributes = append(desc.Attributes, sdp.Attribute{Key: sdp.AttrKeyMsidSemantic, Value: "WMS*"})
+	desc.Attributes = append(desc.Attributes, sdp.Attribute{Key: sdp.AttrKeyMsidSemantic, Value: "WMS *"})
 
 	iceParams, err := pc.iceGatherer.GetLocalParameters()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Sean-Der (approved):
Thank you @nils-ohlmeier :)

PR #3132, opened 2025-05-23
ConfigureFlexFEC03 helper and play-from-disk-fec example

aalekseevx:
#### Description

1. Add a helper for FlexFEC-03 configuration.
2. Introduce play-from-disk-fec example


interceptor.go
```diff
@@ -160,6 +161,38 @@ func ConfigureSimulcastExtensionHeaders(mediaEngine *MediaEngine) error {
 	)
 }
 
+// ConfigureFlexFEC03 registers flexfec-03 codec with provided payloadType in mediaEngine
+// and adds corresponding interceptor to the registry.
+func ConfigureFlexFEC03(
```
3DRX, line 166: nit pick: should we explicit say here that FEC interceptor should go behind twcc interceptor (`webrtc.RegisterDefaultInterceptors` should be called behind `webrtc.ConfigureFlexFEC03 `)?

3DRX (approved):
LGTM, with a small question about API documentation.

JoeTurki (approved):
Thank you for doing this project.
As a side note, I'm a bit biased but I don't like how the interceptors APIs are top-level (webrtc.ConfigureFlexFEC03), but this is nothing we can change before webrtc 5.

ConfigureNack(...)
ConfigureRTCPReports(...)
ConfigureSimulcastExtensionHeaders(...)
ConfigureTWCCHeaderExtensionSender(...)
ConfigureTWCCSender(...)

aalekseevx:
@Sean-Der, @JoeTurki – I agree that we could have a better API in v5. Do you have any suggestions? My point is that the current helpers work well only for simple cases. When you have multiple interceptors that need to be called in a specific order and share state with user code, it becomes extremely complicated and easy to mess up. For example, it's important to note that sender interceptors are called in reverse order of registration.

I like @jech's approach with extra interceptors added directly to the peer connection, but unfortunately, it's very difficult to arrange them in the correct order when some interceptors are registered via factories while others are added directly to the peer connection.

I suggest adding priority/z-index for the interceptors and sorting them in the chain interceptor. Helpers could then provide meaningful defaults so that everything works out of the box. Users would still have the option to add interceptors manually with their own priorities. This approach requires some refactoring but addresses my specific concerns. @3DRX also reported this issue: https://github.com/pion/interceptor/issues/331

I'm not aware of a quick fix – I suggest creating a separate issue to discuss this before proceeding with implementation and ensuring everyone is satisfied with the API before re-implementing it.

Sean-Der:
@aalekseevx :+1: I thought that was going to be a just putting these functions in another package or making them a method.

IMO this should get merged today! Adding a test to confirm that enabling FlexFEC works is only thing that needs to happen in the short term.

Sorry for the churn/noise :(

aalekseevx:
@JoeTurki , @Sean-Der , thanks for the quick feedback! I'm working on the test

3DRX:
> @3DRX , I'm sorry I should have linked it. I'm talking about this PR series: https://github.com/pion/webrtc/pull/3002/files

No worries, thanks @aalekseevx. I’ve read these PRs before.

aalekseevx:
@JoeTurki, @Sean-Der, I have added a test to verify that FEC is properly configured and present in both parameters and SDP. However, I was unable to write an E2E test with FEC packet exchange between peer connections due to the negotiation issue described in https://github.com/pion/webrtc/issues/3109. I'll be happy to add this test once the fix is merged.

PR #3143, opened 2025-06-05
When looking for simulcast IDs, don't forget to check the packet we already have

kvasilye:
#### Description

The code that looks for Simulcast IDs (mid and rid) currently ignores the first packet.

Fix: check the first packet (that we already have) as well.

Technically it's better when the sending party includes Simulcast IDs on some number of packets after establishing the connection, but this fix helps when the sending party doesn't do that (which might happen).

Since we already have the first packet in `b[:i]`, we just need to call the function that extracts those IDs from the packet - `handleUnknownRTPPacket` - before entering the loop which reads more packets.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1784,7 +1784,20 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream io.Reader, ssrc SSRC) err
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


kvasilye:
I got the commit message wrong again, closing, will submit a new one

PR #2914, opened 2024-10-03
Add Retransmission and FEC to TrackLocal

Sean-Der:
If the MediaEngine contains support for them a SSRC will be generated appropriately

Resolves #1989
Resolves #1675

Issue #1675, opened 2021-02-13
Unable to handle NACK that uses distinct SSRC


ghost:
### Your environment.
  - Version: 3.0.7
  - Browser: Chrome 88.0.4324.150 (Official Build) (x86_64)

### What did you do?
A PeerConnection created with `NewPeerConnection` (not api) logs an error `pc ERROR: 2021/02/13 07:35:09 Incoming unhandled RTP ssrc(3537335204), OnTrack will not be fired. mid RTP Extensions required for Simulcast` after the browser sends first retransmitted packet. Looks like pion/webrtc confuses the retransmitted packet for another stream/track.

After recreating the code from `NewPeerConnection` manually, but registering only some codecs showed that adding the codec below triggers the error.
```
	m.RegisterCodec(webrtc.RTPCodecParameters{
		RTPCodecCapability: webrtc.RTPCodecCapability{"video/rtx", 90000, 0, "apt=96", nil},
		PayloadType:        97,
	}, webrtc.RTPCodecTypeVideo)
```

The rest of the code:

```
func NewPeerConnection() (*webrtc.PeerConnection, error) {
	// pc, _ := webrtc.NewPeerConnection(config)
	m := webrtc.MediaEngine{}

	videoRTCPFeedback := []webrtc.RTCPFeedback{{"goog-remb", ""}, {"ccm", "fir"}, {"nack", ""}, {"nack", "pli"}}

	m.RegisterCodec(webrtc.RTPCodecParameters{
		RTPCodecCapability: webrtc.RTPCodecCapability{"video/VP8", 90000, 0, "", videoRTCPFeedback},
		PayloadType: 96,
	}, webrtc.RTPCodecTypeVideo)

	m.RegisterCodec(webrtc.RTPCodecParameters{
		RTPCodecCapability: webrtc.RTPCodecCapability{"video/rtx", 90000, 0, "apt=96", nil},
		PayloadType:        97,
	}, webrtc.RTPCodecTypeVideo)

	m.RegisterCodec(webrtc.RTPCodecParameters{
		RTPCodecCapability: webrtc.RTPCodecCapability{"audio/opus", 48000, 2, "minptime=10;useinbandfec=1", nil},
		PayloadType:        111,
	}, webrtc.RTPCodecTypeAudio)

	// m.RegisterDefaultCodecs()
	i := &interceptor.Registry{}

	generator, _ := nack.NewGeneratorInterceptor()
	responder, _ := nack.NewResponderInterceptor()
[...]

PR #3157, opened 2025-06-14
Fix BindRemoteStream StreamInfo

3DRX:
#### Provide correct StreamInfo to receiver interceptors

Before this change, interceptor's `BindRemoteStream` gets called twice per track when RTX is used, with only the SSRC being different. Hence, the interceptor can't distinguish between RTX track and video track.

```
ReceiverInterceptor BindRemoteStream, info: &{ map[] 3181198436 0 0 0 0 0 [{http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01 4}] video/AV1 90000 0 level-idx=5;profile=0;tier=0 [{nack } {nack pli} {transport-cc }]}
ReceiverInterceptor BindRemoteStream, info: &{ map[] 1017514322 0 0 0 0 0 [{http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01 4}] video/AV1 90000 0 level-idx=5;profile=0;tier=0 [{nack } {nack pli} {transport-cc }]}
```

After this change, interceptor's `BindRemoteStream` gets called twice, the first time bind to video track, second time bind to RTX track, like this.

```
ReceiverInterceptor BindRemoteStream, info: &{ map[] 559414906 3287567208 0 45 46 0 [{http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01 4}] video/AV1 90000 0 level-idx=5;profile=0;tier=0 [{nack } {nack pli} {transport-cc }]}
ReceiverInterceptor BindRemoteStream, info: &{ map[] 3287567208 0 0 46 0 0 [{http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01 4}] video/AV1 90000 0 level-idx=5;profile=0;tier=0 [{nack } {nack pli} {transport-cc }]}
```

With this change, interceptor can implement internal logic to handle rtx track with video track. This will also make implementing `BindRemoteStream` to FEC track in the future easier.

This shouldn't break anything, since this PR only adds new data (that's originally always 0) to StreamInfo.

### DIFF
--- a/rtpcodec.go
+++ b/rtpcodec.go
@@ -143,16 +143,26 @@ func codecParametersFuzzySearch(
 	return RTPCodecParameters{}, codecMatchNone
 }
 
-// Given a CodecParameters find the RTX CodecParameters if one exists.
+// Given a CodecParameters find the RTX PayloadType.
 func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) PayloadType {
-	aptStr := fmt.Sprintf("apt=%d", needle)
-	for _, c := range haystack {
-		if aptStr == c.SDPFmtpLine {
-			return c.PayloadType
-		}
+	rtxParam := findRTXCodecParameters(needle, haystack)
+	if rtxParam != nil {
+		return rtxParam.PayloadType
 	}
 
 	return PayloadType(0)
+}
+
+// Given a CodecParameters find the RTX CodecParameters if one exists.
+func findRTXCodecParameters(needle PayloadType, haystack []RTPCodecParameters) *RTPCodecParameters {
+	aptStr := fmt.Sprintf("apt=%d", needle)
+	for i, c := range haystack {
+		if aptStr == c.SDPFmtpLine {
+			return &haystack[i]
+		}
+	}
+
+	return nil
 }
 
 // For now, only FlexFEC is supported.
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -19,25 +19,78 @@ import (
 	"github.com/pion/webrtc/v4/internal/util"
 )
 
+type trackStream struct {
+	streamInfo      *interceptor.StreamInfo
+	rtpReadStream   *srtp.ReadStreamSRTP
+	rtpInterceptor  interceptor.RTPReader
+	rtcpReadStream  *srtp.ReadStreamSRTCP
+	rtcpInterceptor interceptor.RTCPReader
+}
+
 // trackStreams maintains a mapping of RTP/RTCP streams to a specific track
 // a RTPReceiver may contain multiple streams if we are dealing with Simulcast.
 type trackStreams struct {
 	track *TrackRemote
 
-	streamInfo, repairStreamInfo *interceptor.StreamInfo
+	mediaStream trackStream
+	rtxStream   trackStream
 
-	rtpReadStream  *srtp.ReadStreamSRTP
-	rtpInterceptor interceptor.RTPReader
-
-	rtcpReadStream  *srtp.ReadStreamSRTCP
-	rtcpInterceptor interceptor.RTCPReader
-
-	repairReadStream    *srtp.ReadStreamSRTP
-	repairInterceptor   interceptor.RTPReader
 	repairStreamChannel chan rtxPacketWithAttributes
+}
 
-	repairRtcpReadStream  *srtp.ReadStreamSRTCP
-	repairRtcpInterceptor interceptor.RTCPReader
+func (s *trackStreams) configureStreams(
+	decodingParams RTPDecodingParameters,
+	codec RTPCodecCapability,
+	globalParams RTPParameters,
+	streamsForSSRC func(SSRC, interceptor.StreamInfo) (*srtp.ReadStreamSRTP, interceptor.RTPReader, *srtp.ReadStreamSRTCP, interceptor.RTCPReader, error),
+) (bool, error) {
+	rtxPayloadType := findRTXPayloadType(globalParams.Codecs[0].PayloadType, globalParams.Codecs)
+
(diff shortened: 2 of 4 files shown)


JoTurk (commented):
I'm outside today, looking at the code from my phone, this is a step in the right direction to fix that part of pion, thank you so much.
Just one thing, can we add a test because it's easy to break this change?

3DRX:
@JoeTurki Thanks for review! Sure, I will look into existing tests to see how to test this kind of things properly.

3DRX:
@JoeTurki I refactored the code a bit for better testability and add test cases for video stream only and video + RTX, this is ready for review again. No rush :)

arjunshajitech:
Hi @3DRX 

In the following line from the Pion code:
https://github.com/pion/webrtc/blob/b8c6f1036cd67f213906c79478fb55d61965c9d2/rtpreceiver.go#L534

Is the call to ```r.rtxPool.Put(b)``` necessary here?

3DRX:
> Is the call to r.rtxPool.Put(b) necessary here?

@arjunshajitech `handleReceivedRTXPackets` in this PR is exactly the same as the old anonymous function in `receiveForRtx`. When I'm doing the refactor I'm just reorganizing code into different functions and variables into different structs, not modifying any logic. Just read through `handleReceivedRTXPackets` and I think you are right, it not necessary.

arjunshajitech:
> > Is the call to r.rtxPool.Put(b) necessary here?
> 
> @arjunshajitech `handleReceivedRTXPackets` in this PR is exactly the same as the old anonymous function in `receiveForRtx`. When I'm doing the refactor I'm just reorganizing code into different functions and variables into different structs, not modifying any logic. Just read through `handleReceivedRTXPackets` and I think you are right, it not necessary.

Just to clarify, I said that `r.rtxPool.Put(b)` is needed in the default case, but it's missing in the code.

3DRX:
Thank you so much @arjunshajitech ! I hope this PR's refactor doesn't change any current behavior and logic, only improves testability so that reviewing is easier. Do you mind opening another PR to make the change?

aalekseevx:
@3DRX, do you think that it would be better to feed RTX/FEC packets to the same stream as the media packets? These streams are not separate in BindLocalStream, so it looks a bit inconsistent from the interceptor's point of view. It's a breaking change, so it could be done using the SettingsEngine I guess. Of course it's not a blocker for this PR, which is definitely a step in the right direction.

3DRX:
> @3DRX, do you think that it would be better to feed RTX/FEC packets to the same stream as the media packets? These streams are not separate in BindLocalStream, so it looks a bit inconsistent from the interceptor's point of view. It's a breaking change, so it could be done using the SettingsEngine I guess. Of course it's not a blocker for this PR, which is definitely a step in the right direction.

Yes, it definately would be better imo. Current way of handling RTX is quite unintuitive to me, with `BindRemoteStream` being called twice. @arjunshajitech @JoeTurki @Sean-Der what do you guys think? To me, It's fine if we introduce a breaking change here, since we don't have 100% correct reciver side RTX and FEC implemented yet, so people probably not using them anyway.

aalekseevx:
@3DRX, I'm concerned not only about breaking RTX implementation in users' interceptors, but also about breaking the existing interceptors' logic which expects packets of only one SSRC inside the stream. We even have this kind of logic in pion's interceptors, which would require fixes. Some examples: [Send side](https://github.com/pion/interceptor/blob/master/pkg/gcc/leaky_bucket_pacer.go#L144). [Receive side](https://github.com/pion/interceptor/blob/master/pkg/report/receiver_interceptor.go#L143).

3DRX:
@JoeTurki @aalekseevx have a good point here, I guess we can also mark this as a reasonable breaking change in future major version? This won’t block this PR though, it’s ready to be merged on my side. Thanks guys :)

JoTurk:
Maybe it's time to start making a v5 milestone? and aim for jan 2026?

3DRX:
> Maybe it's time to start making a v5 milestone? and aim for jan 2026?

@JoeTurki So sorry I missed this, sure we can make the milestone, there's a couple of things I think worth a breaking change.

JoTurk:
@jech Can you please look at this? Will it break any behavior for Galene?

Sean-Der:
@3DRX I would love to hear all your ideas on how to make Pion better! In the past I used the Wiki, but w/e you prefer works for me https://github.com/pion/webrtc/wiki/Big-Ideas

My hope/goal is to try and squeeze peoples ideas in without an API break :)

3DRX:
@Sean-Der Hi, I've updated https://github.com/pion/webrtc/wiki/Big-Ideas, please check it out, thanks :)

jech:
The change is a net positive, it only adds more data to previously unused fields.

On the other hand, I don't think it solves any of the fundamental issues we have, which make the RTX/FEC tracks unusable in practice.

It's a difficult problem.  Pion aims at being suitable both for high-level applications (streaming a video as simply as possible), and at low-level software (high-performance SFUs).  The difficultly lies in designing an API that is transparent to high-level software (create a track and get RTX/FEC handling for free) and low-level software (full control on RTX/FEC, let the SFU decide whether to forward FEC or terminate FEC and regenerate it).

I don't currently see a good solution.  I'd like people to join my suffering by having a look at

 * https://github.com/pion/webrtc/issues/1675
 * https://github.com/pion/webrtc/pull/2914
 * https://github.com/pion/webrtc/issues/2994
 * https://github.com/pion/webrtc/issues/2996

as well as the following patch series:

 * https://github.com/pion/interceptor/pull/297
 * https://github.com/pion/webrtc/pull/3002
 * https://github.com/pion/interceptor/pull/298
 * https://github.com/pion/webrtc/pull/3003

PR #3165, opened 2025-06-20
Add ice-proxy example

amanakin:
#### Description
This PR introduces an example demonstrating how to configure and use a proxy with ICE.

### DIFF
--- a/examples/ice-proxy/README.md
+++ b/examples/ice-proxy/README.md
@@ -1,0 +1,26 @@
+# ICE Proxy
+`ice-proxy` demonstrates Pion WebRTC's capabilities for utilizing a proxy in WebRTC connections.
+
+This proxy functionality is particularly useful when direct peer-to-peer communication is restricted, such as in environments with strict firewalls. It primarily leverages TURN (Traversal Using Relays around NAT) with TCP connections to enable communication with the outside world.
+
+## Instructions
+
+### Download ice-proxy
+The example is self-contained and requires no input.
+
+```bash
+go install github.com/pion/webrtc/v4/examples/ice-proxy@latest
+```
+
+### Run ice-proxy
+```bash
+ice-proxy
+```
+
+Upon execution, four distinct entities will be launched:
+* `TURN Server`: This server facilitates relaying media traffic when direct communication between agents is not possible, simulating a scenario where peers are behind restrictive NATs.
+* `Proxy HTTP Server`: A straightforward HTTP proxy designed to forward all TCP traffic to a specified target.
+* `Offering Agent`: In a typical WebRTC setup, this would be a web browser. In this example, it's a simplified Pion client that initiates the WebRTC connection. This agent attempts direct communication with the answering agent.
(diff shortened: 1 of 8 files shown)


Sean-Der:
Great work @amanakin I would love to merge this!

What is the purpose of TURN in the example? I don't understand the relationship of HTTP Proxying + TURN.

Is it possible to do proxying using ice-tcp instead? Is TCP via TURN a requirement?

amanakin:
> Great work @amanakin I would love to merge this!
> 
> What is the purpose of TURN in the example? I don't understand the relationship of HTTP Proxying + TURN.
> 
> Is it possible to do proxying using ice-tcp instead? Is TCP via TURN a requirement?

Hey, yes, proxying is currently possible (and it makes sense) only via turn
https://github.com/pion/ice/blame/master/gather.go#L706

amanakin:
@Sean-Der 
I also want to update this example by using just two Pion clients, because the Chrome browser doesn't allow TURN requests to localhost. Maybe this is because of: https://localmess.github.io/ (By the way, in Firefox it works.)

Sean-Der:
@amanakin that would be great! Will make it significantly simpler.

also love the in-server TURN example. I think it will be real useful :)

amanakin:
@Sean-Der, I updated the example and added some clarifications to the readme.

amanakin:
Hey @Sean-Der, just pinging

JoTurk:
@amanakin Hello, Is this ready for merge?

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/pion/srtp/v3 v3.0.5
 	github.com/pion/stun/v3 v3.0.0
 	github.com/pion/transport/v3 v3.0.7
+	github.com/pion/turn/v4 v4.0.0
```
JoTurk, line 19: Would love if we could avoid making turn a direct dependency

JoeTurki (approved):
Thank you :)

go.mod
amanakin: I see @JoeTurki , added a separate go module for ice-proxy for that reason.

amanakin:
> @amanakin Hello, Is this ready for merge?

Hey @JoeTurki , yes, it's ready.

JoTurk:
@Sean-Der what do you think we should do? another mod file for this example, or change turn to be a direct dependency? there are trade offs for both options 🤔

examples/ice-proxy/README.md
```diff
@@ -0,0 +1,26 @@
...
+### Download ice-proxy
+
+```bash
+git clone https://github.com/pion/webrtc.git
+cd webrtc/examples/ice-proxy
+```
+
+### Run ice-proxy
+```bash
+go run *.go
```
JoTurk: If we're going to add this example as a module, can you update this run script? also the install script, and we'll merge it right away!

amanakin: Updated!

JoTurk:
@amanakin I undid the package, I think this is better, Also i fixed the wasm test.

PR #3168, opened 2025-06-25
Replace custom atomicBool with sync/atomic.Bool

cppcoffee:
- Remove custom atomicBool implementation
- Replace all atomicBool usages with standard library sync/atomic.Bool


JoeTurki (approved):
Cool :)

PR #3175, opened 2025-07-08
Add PayloadToMimeType map to StreamInfo

5ur3:
https://github.com/pion/interceptor/pull/346

### DIFF
--- a/interceptor.go
+++ b/interceptor.go
@@ -222,6 +222,7 @@ func createStreamInfo(
 	payloadType, payloadTypeRTX, payloadTypeFEC PayloadType,
 	codec RTPCodecCapability,
 	webrtcHeaderExtensions []RTPHeaderExtensionParameter,
+	codecs []RTPCodecParameters,
 ) *interceptor.StreamInfo {
 	headerExtensions := make([]interceptor.RTPHeaderExtension, 0, len(webrtcHeaderExtensions))
 	for _, h := range webrtcHeaderExtensions {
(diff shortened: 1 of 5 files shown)


JoTurk (commented):
I don't know how I feel about adding this, in one hand i see how this is required, but for people who read the code or use pion it might feels tricky to have a member with a generic name and redundant information. I meant for this PR and https://github.com/pion/interceptor/pull/346

5ur3:
Thanks for reviewing it, is it possible to change the member in StreamInfo so it would fit? For example:
- Add a member that would contain only payload type and mime type (minimum amount of redundant information needed for RED FEC decoder interceptor implementation)
- Store negotiated codecs information in private field and make a getter method: `StreamInfo.GetCodecForPayloadType(uint8) (RTPCodecParameters, error)`

Or maybe it can be merged if I provide an example of using this new StreamInfo API by contributing RED FEC decoder interceptor to pion/interceptor/pkg/?

JoTurk:
> Add a member that would contain only payload type and mime type (minimum amount of redundant information needed for RED FEC decoder interceptor implementation)

This sound like a good call, we should only add the information that we need, We might need to refactor streaminfo by the next major version anyways so it contains a slice of codec values.

5ur3:
@JoeTurki Is this better? I have replaced the Codecs field with the PayloadToMimeType map

JoTurk:
> @JoeTurki Is this better? I have replaced the Codecs field with the PayloadToMimeType map

I like this more, but maybe we need to explain what it's doing in the code or maybe rename it to something more readable, not sure what will work best, but as of right now for someone reading the code they might have hard time understanding what it's actually doing!

PR #3182, opened 2025-07-24
Improve RTCP Read

arjunshajitech:
Notify to use ReadSimulcast with rid when multiple tracks are present in RTPReceiver

#### Description

Note: The end-user is using `receiver.ReadRTCP()` instead of `receiver.ReadSimulcastRTCP(track.RID())` for handling simulcast.

The simulcast is configured with three encodings:
```
{ rid: '360', scaleResolutionDownBy: 3.0, maxBitrate: 150_000 },
{ rid: '640', scaleResolutionDownBy: 2.0, maxBitrate: 400_000 },
{ rid: '1024', scaleResolutionDownBy: 1.0, maxBitrate: 1_000_000 }
 ```
 
When the ontrack event is fired in the expected order — for example: 360 -> 640 -> 1024 
Everything seems working fine. No panic occurs, and RTCP packets are read successfully for all simulcast tracks.

Screenshot from 2025-07-24 18-36-57

RTCP is read three times, but internally, all reads point to track[0] (RID 360). 
So while it seems like RTCP is being read for all simulcast layers, it’s actually reading only from the first track each time
This gives the false impression that everything is working, but in reality, RTCP for RID 640 and 1024 is not being read at all.

Trigger a panic only when the ontrack events fire in an unexpected (random) order for example: 640 -> 360 -> 1024

Screenshot from 2025-07-24 18-38-40 

In this case, when trying to read RTCP for RID 640, it incorrectly maps to track[0] (which corresponds to RID 360). Since the RTCP interceptor for that track hasn't been set up yet, a panic occurs and is not properly handled.

#### Reference issue
Fixes #...

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/pion/interceptor"
+	"github.com/pion/logging"
 	"github.com/pion/rtcp"
 	"github.com/pion/srtp/v3"
 	"github.com/pion/webrtc/v4/internal/util"
@@ -70,6 +71,8 @@ type RTPReceiver struct {
 	api *API
 
 	rtxPool sync.Pool
+
+	log logging.LeveledLogger
 }
 
 // NewRTPReceiver constructs a new RTPReceiver.
@@ -78,7 +81,7 @@ func (api *API) NewRTPReceiver(kind RTPCodecType, transport *DTLSTransport) (*RT
 		return nil, errRTPReceiverDTLSTransportNil
 	}
 
-	r := &RTPReceiver{
+	rtpReceiver := &RTPReceiver{
 		kind:      kind,
 		transport: transport,
 		api:       api,
@@ -88,9 +91,10 @@ func (api *API) NewRTPReceiver(kind RTPCodecType, transport *DTLSTransport) (*RT
 		rtxPool: sync.Pool{New: func() any {
 			return make([]byte, api.settingEngine.getReceiveMTU())
 		}},
+		log: api.settingEngine.LoggerFactory.NewLogger("RTPReceiver"),
 	}
 
-	return r, nil
+	return rtpReceiver, nil
 }
 
 func (r *RTPReceiver) setRTPTransceiver(tr *RTPTransceiver) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: rtpreceiver.go [hunk 1]
```diff
@@ -272,7 +272,12 @@ func (r *RTPReceiver) Receive(parameters RTPReceiveParameters) error {
 func (r *RTPReceiver) Read(b []byte) (n int, a interceptor.Attributes, err error) {
 	select {
 	case <-r.received:
-		return r.tracks[0].rtcpInterceptor.Read(b, a)
+		// In non-simulcast mode, RTPReceiver is guaranteed to have exactly one track.
+		// Read directly from that track's RTCP interceptor.
+		if len(r.tracks) == 1 {
+			return r.tracks[0].rtcpInterceptor.Read(b, a)
+		}
+		return 0, nil, errSimulcastRead //nolint:nlreturn
```
JoTurk: Even if this is correct, This will be a breaking change, maybe we can log an error here instead? and in the next major version we can make ReadRTCP return all the tracks.

also you can just add a new line before return instead of `nolint:nlreturn`.

arjunshajitech: I originally agreed with logging a warning instead of changing the behavior.

But after thinking more carefully, I can’t actually see how this would break anything.

In fact, if someone is using this with simulcast, the current behavior might be silently incorrect, like missing RTCP packets and they might wrongly assume everything is working.

arjunshajitech: > also you can just add a new line before return instead of `nolint:nlreturn`.

Got it :)

JoTurk: The logic now returns an error where it previously succeeded, which makes this a breaking change. Even if returning the first track's RTCP for simulcast was incorrect, existing code would continue working. Now, their pipeline will fail - loops that previously processed RTCP will exit on error. and since the error isn't exposed, developers will have to spend time debugging why it's failing, which is something we try to avoid without a major version update.

arjunshajitech: I’ll switch to logging the issue instead of returning an error.

thank you.

JoTurk: Thanks <3

Issue #880, opened 2019-10-17
PeerConnection OnTrack callback never invoked


tylerlong:
### Your environment.
  - Version: v2.1.8
  - Browser: None
  - Other Information - I am trying to create a softphone using SIP and this library. The remote SIP & RTC service have no problem, because I have done a softphone using Node.js: https://github.com/ringcentral/ringcentral-softphone-js . I am trying to make a GoLang version, but I am stuck.

### What did you do?

I have written the following code: https://github.com/ringcentral/ringcentral-softphone-go/blob/master/softphone.go#L137-L220

Just for your reference, here is the **Node.js** version of the GoLang code: https://github.com/ringcentral/ringcentral-softphone-js/blob/ce833c613d5e4bb959235e37addefd10cc9cb81e/index.js#L140-L177 .   The Node.js code works like a charm.

### What did you expect?

I expect that the `peerConnection.OnTrack` callback function will be invoked so that I can get the input audio

### What happened?

`peerConnection.OnTrack` never invoked and I am stuck.

## Some questions:

Could it because I specify the wrong codec? If so, I expect some error message like: no codec to handle the remote track. remote SDP sample:

```
v=0
o=- 5305405169010475891 908790132696250821 IN IP4 104.245.57.182
s=SmcSip
c=IN IP4 104.245.57.182
t=0 0
m=audio 20040 RTP/SAVPF 109 111 18 0 8 9 96 101
a=rtpmap:109 OPUS/16000
a=fmtp:109 useinbandfec=1
a=rtcp-fb:109 ccm tmmbr
a=rtpmap:111 OPUS/48000/2
a=fmtp:111 useinbandfec=1
a=rtcp-fb:111 ccm tmmbr
a=rtpmap:18 g729/8000
a=fmtp:18 annexb=no
a=rtpmap:0 pcmu/8000
a=rtpmap:8 pcma/8000
a=rtpmap:9 g722/8000
a=rtpmap:96 ilbc/8000
a=fmtp:96 mode=20
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-15
a=sendrecv
a=rtcp:20041
a=rtcp-mux
a=setup:actpass
a=fingerprint:sha-1 E2:F8:90:C6:37:7E:82:17:3C:63:CF:55:68:4E:1D:16:17:2C:D9:4D
a=ice-ufrag:atwbXWRm
a=ice-pwd:32SCWdSoU7BgkXxet5g5mlId5E
[...]

PR #3197, opened 2025-08-21
Do not create receiver for ealy media in offerer

boks1971:
#### Description

For offerer, if the remote side sends early media
before the remote description (answer) is received,
the undeclared SSRC processor can create a receiver
and that receiver could be left dangling as
transceiver `mid` is not updated from remote
description answer.

Still leaving the simulcast probe path and only
avoiding creating a receiver for non-simulcast path.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1719,10 +1719,12 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream io.Reader, ssrc SSRC) err
 	// (urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id and urn:ietf:params:rtp-hdrext:sdes:mid)
 	// and even if the RTP stream contains an incorrect MID or RID.
 	// while this can be incorrect, this is done to maintain compatibility with older behavior.
-	if len(remoteDescription.parsed.MediaDescriptions) == 1 {
-		mediaSection := remoteDescription.parsed.MediaDescriptions[0]
-		if handled, err := pc.handleUndeclaredSSRC(ssrc, mediaSection); handled || err != nil {
-			return err
+	if remoteDescription.Type != SDPTypeAnswer || pc.api.settingEngine.handleUndeclaredSSRCWithoutAnswer {
+		if len(remoteDescription.parsed.MediaDescriptions) == 1 {
+			mediaSection := remoteDescription.parsed.MediaDescriptions[0]
+			if handled, err := pc.handleUndeclaredSSRC(ssrc, mediaSection); handled || err != nil {
+				return err
+			}
 		}
 	}
 
@@ -1748,6 +1750,11 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream io.Reader, ssrc SSRC) err
 		RTPHeaderExtensionCapability{sdp.SDESMidURI},
 	)
 	if !audioSupported && !videoSupported {
+		if remoteDescription.Type == SDPTypeAnswer && !pc.api.settingEngine.handleUndeclaredSSRCWithoutAnswer {
+			// if we are offerer, wait for answer with media setion to process this SSRC
+			return errPeerConnEarlyMediaWithoutAnswer
+		}
+
 		// try to find media section by payload type as a last resort for legacy clients.
 		mediaSection, ok := pc.findMediaSectionByPayloadType(payloadType, remoteDescription)
 		if ok {
--- a/errors.go
+++ b/errors.go
@@ -231,6 +231,10 @@ var (
 	errPeerConnSetIdentityProviderNotImplemented = errors.New("TODO SetIdentityProvider")
 	errPeerConnWriteRTCPOpenWriteStream          = errors.New("WriteRTCP failed to open WriteStream")
 	errPeerConnTranscieverMidNil                 = errors.New("cannot find transceiver with mid")
+	errPeerConnEarlyMediaWithoutAnswer           = errors.New(
+		"cannot process early media without SDP answer," +
+			"use SettingEngine.SetHandleUndeclaredSSRCWithoutAnswer(true) to process without answer",
+	)
 
 	errRTPReceiverDTLSTransportNil            = errors.New("DTLSTransport must not be nil")
 	errRTPReceiverReceiveAlreadyCalled        = errors.New("Receive has already been called")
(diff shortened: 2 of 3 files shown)


boks1971:
@Sean-Der @JoTurk @nils-ohlmeier @cnderrauber @davidzhao Found this issue while working on peer connections using bi-directional media. If the answer side sends early media, the undeclared processor could end up creating a dangling receiver.

Please let me know if you have some ideas for writing a test for this. Also, any gotchas I may have overlooked as there are some legacy checks in this path.

JoTurk:
I documented this behavior while working on https://github.com/pion/webrtc/pull/3066 which handles undeclared SSRC it for multiple media sections and handles it like chrome and firefox, I didn't want to remove it at the time because I was worried it would break some older clients, historic reasons for why it was [added](https://github.com/pion/webrtc/commit/c6ad8c78ead92ab0226b7ddbbd5357af0e5f1771#diff-c94e5058864f96b58c5e039ac0485b3a9640613f32a9e31bba905dc3f8438e7aR1138-R1164) in 2019 https://github.com/pion/webrtc/issues/880 

I think it's time to change this behavior if the other side supports `(urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id and urn:ietf:params:rtp-hdrext:sdes:mid)` This way we can keep compatibility with older clients and modern clients, what do you think?

peerconnection.go
```diff
@@ -1719,10 +1718,12 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream io.Reader, ssrc SSRC) err
 	// (urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id and urn:ietf:params:rtp-hdrext:sdes:mid)
 	// and even if the RTP stream contains an incorrect MID or RID.
 	// while this can be incorrect, this is done to maintain compatibility with older behavior.
-	if len(remoteDescription.parsed.MediaDescriptions) == 1 {
-		mediaSection := remoteDescription.parsed.MediaDescriptions[0]
-		if handled, err := pc.handleUndeclaredSSRC(ssrc, mediaSection); handled || err != nil {
-			return err
+	if remoteDescription.Type != SDPTypeAnswer {
```
JoTurk: Other reason for why I didn't change it at the time of the [refactor](https://github.com/pion/webrtc/pull/3066) because this was introduced 6 years ago in: https://github.com/pion/webrtc/commit/c6ad8c78ead92ab0226b7ddbbd5357af0e5f1771#diff-c94e5058864f96b58c5e039ac0485b3a9640613f32a9e31bba905dc3f8438e7aR1138-R1164 And I didn't find any issues reported, changing it would break the client mentioned in the [original issue](https://github.com/pion/webrtc/issues/880). I don't think handling it as an answer is the right fix. Some old/basic SIP or WebRTC clients generate static SDP, so the proper solution is:

1. hide it behind a flag like how Nils suggested.
2. make this path only possibe if (urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id and urn:ietf:params:rtp-hdrext:sdes:mid) aren't used by the other side, like how the handling for multiple undeclared SSRC currently work after https://github.com/pion/webrtc/pull/3066

JoTurk: Maybe `if remoteDescription.Type != SDPTypeAnswer {` as a new default behavior, and make it possible to disable it via a flag?

boks1971: Thank you @JoTurk for the feedback.

A flag definitely makes sense. Will add it to settings engine.

I was hesitant to do based on `answer` type also, but I could not think of a better option. I will read the links you have posted ☝️ and think if there is a better way. Will reach back out after reading it.

boks1971: @JoTurk After reading through the links, I agree that it is better to
1. Add a settings engine flag for this which defaults to existing behaviour. We can later change the default. Naming: maybe `SetNoReceiverOnEarlyMedia`? Any other suggestions?
2. Rather than use answer type, agree with you that this can check for `mid` and `rid` like you mentioned.

Let me make an attempt at this and report back.

JoTurk: @boks1971 Sorry for the confusion, I was referring to the first path (single media section). Adding a fallback flag would break the same behavior Chrome has with legacy clients - and unfortunately, it would also break Pion for many users. Someone even emailed me about this issue with undeclared SSRC with more than one media section: and if they upgrade Pion, things will stop working for them.

This is how chrome handles it by default without a flag https://chromium.googlesource.com/external/webrtc/+/refs/heads/master/call/rtp_demuxer.cc#372 

I think a better fix would be to pass down whether we've received an answer to the undeclaredMediaProcessor, and only handle this once the answer has been received.

boks1971: Got it Joe.

Unfortunately, basing `undeclaredMediaProcessor` on `answer` is not working. That is how it is currently. `startRTP` is called in `SetRemoteDescription` which is what starts `undeclaredMediaProcessor`.

But, the `findMediaSectionByPayloadType` succeeds because the `answer` has media section of the kind and matching payload type. Once that is handed to `handleUndeclaredSSRC`, it fails the checks (no `ssrc` or `rid`) and ends up creating a new receiver.

In my test, a client is publishing a track which creates an `offer` and remote sends `answer`. That kicks off the undeclared media processor. OOB signalling indicates that remote wants to send a track. So, client sends `offer` again. The `answer` corresponding to that `offer` would have the SSRC info. But, early media coming in before that `answer` triggers this path.

The `answer` check was the simplest I could think of. This is definitely tricky with all the legacy stuff out there.

JoTurk: Okay, We can add a flag `handleUndeclaredSSRCWithAnswer` (or just `handleUndeclaredSSRC` or anything you think is correct) keep your answer check only if the flag isn't set. and if someone points out that it breaks for them we point them to the spec and the flag. Until we're able to fix undeclaredMediaProcessor systematically, maybe even re assign the receivers when we receive SSRC? - we're already parting ways with Plan B, So i think it's time to drop it.
Please add it a message in the error at the handler to enable the flag - if used with legacy client.

cnderrauber (approved):
looks good! agree with the flag to make it compatible with old behavior

peerconnection.go
boks1971: @JoTurk Can you please have another look?

Added a flag and made this (no receiver without answer) the default behaviour. Should we flip the sense of the flag so that default works like it is today and new behaviour needs explicit setting?

Also, please suggest changes to the error message/language.

JoTurk (approved):
Perfect, thank you so much and sorry for going back and forth on this.

boks1971:
> Perfect, thank you so much and sorry for going back and forth on this.

Thank you so much Joe. Please do not say sorry. It is great to get such detailed feedback 🙏

PR #2412, opened 2023-02-09
Make Transceiver Negotation more consistent

Sean-Der:
Today a remote `recvonly` transceiver can be satisfied by a `sendonly` or `sendrecv` transceiver. However a remote `sendonly` transceiver can only be satisfied by a `recvonly` transceiver.

This PR updates the answer handling code to be more consistent.

PR #3198, opened 2025-08-21
Disallow incompatible transceiver directions

boks1971:
#### Description

This effectively reverts part of https://github.com/pion/webrtc/pull/2412.

The answer was incorrect under the following conditions
- Answer side added a track, it created a new transceicer with `sendrecv` direction.
- When the remote offer comes in later, SetRemoteDescription tries to find an existing transceiver by type and direction. It ends up picking what was added above even if the remote side is using `sendonly`.
- That results in answer marking the section as `sendrecv` which is incompatible with the offer.

### DIFF
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -8,6 +8,7 @@ package webrtc
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -57,6 +58,19 @@ func (t *RTPTransceiver) SetCodecPreferences(codecs []RTPCodecParameters) error 
 			codec, t.api.mediaEngine.getCodecsByKind(t.kind),
 		); matchType == codecMatchNone {
 			return fmt.Errorf("%w %s", errRTPTransceiverCodecUnsupported, codec.MimeType)
+		}
+	}
+
+	// remove RTX codecs if there is no corresponding primary codec
+	for i := len(codecs) - 1; i >= 0; i-- {
+		c := codecs[i]
+		if !strings.EqualFold(c.MimeType, MimeTypeRTX) {
+			continue
+		}
+
+		if isRTX, primaryExists := primaryPayloadTypeForRTXExists(c, codecs); isRTX && !primaryExists {
+			// no primary for RTX, remove the RTX
+			codecs = append(codecs[:i], codecs[i+1:]...)
 		}
 	}
 
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1160,16 +1160,25 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 				// if transceiver is create by remote sdp, set prefer codec same as remote peer
 				if codecs, err := codecsFromMediaDescription(media); err == nil {
 					filteredCodecs := []RTPCodecParameters{}
+					filteredCodecsPartial := []RTPCodecParameters{}
 					for _, codec := range codecs {
-						if c, matchType := codecParametersFuzzySearch(
+						c, matchType := codecParametersFuzzySearch(
 							codec,
 							pc.api.mediaEngine.getCodecsByKind(kind),
-						); matchType == codecMatchExact {
+						)
+						switch matchType {
+						case codecMatchExact:
 							// if codec match exact, use payloadtype register to mediaengine
 							codec.PayloadType = c.PayloadType
 							filteredCodecs = append(filteredCodecs, codec)
+						case codecMatchPartial:
+							codec.PayloadType = c.PayloadType
+							filteredCodecsPartial = append(filteredCodecsPartial, codec)
+
+						default:
 						}
 					}
+					filteredCodecs = append(filteredCodecs, filteredCodecsPartial...)
 					_ = transceiver.SetCodecPreferences(filteredCodecs)
 				}
 
@@ -2096,7 +2105,8 @@ func (pc *PeerConnection) AddTrack(track TrackLocal) (*RTPSender, error) {
 		// But that will cause sdp inflate. So we only check currentDirection's current value,
 		// that's worked for all browsers.
 		if transceiver.kind == track.Kind() && transceiver.Sender() == nil &&
-			currentDirection != RTPTransceiverDirectionSendrecv && currentDirection != RTPTransceiverDirectionSendonly {
+			currentDirection != RTPTransceiverDirectionSendrecv && currentDirection != RTPTransceiverDirectionSendonly &&
+			(!pc.api.settingEngine.disableTransceiverReuseInRecvonly || currentDirection != RTPTransceiverDirectionRecvonly) {
 			sender, err := pc.api.NewRTPSender(track, pc.dtlsTransport)
 			if err == nil {
 				err = transceiver.SetSender(sender, track)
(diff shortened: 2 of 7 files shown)


boks1971:
Sorry, this is not ready yet. Tests are failing. Reading the spec more to understand better.

boks1971:
This should be ready for review. Sorry, the `AddTrack` changes I had in some intermediate commit were not needed. I got myself tangled with `direction` and `currentDirection` and thought the changes were needed. But, they are not needed.

boks1971:
arrgh no, the `AddTrack` changes are needed I think. Will look at it in a different PR.

The reason is that remote is setting `sendonly`. And `AddTrack` does not know remote is `sendonly`. It just checks local current direction and that is `recvonly` and it ends up adding sender to that transceiver which is incorrect.

cnderrauber:
we can change the `sendrecv`transceiver's direction to `recvonly` instead of filtering it out. 
refer to the rfc 9429 https://www.rfc-editor.org/rfc/rfc9429#section-5.3.1-21.2, 

> A direction attribute, determined by applying the rules regarding the offered direction specified in [[RFC3264](https://www.rfc-editor.org/rfc/rfc9429#RFC3264)], [Section 6.1](https://rfc-editor.org/rfc/rfc3264#section-6.1), and then intersecting with the direction of the associated RtpTransceiver. For example, in the case where an "m=" section is offered as "sendonly" and the local transceiver is set to "sendrecv", the result in the answer is a "recvonly" direction.

boks1971:
> we can change the `sendrecv`transceiver's direction to `recvonly` instead of filtering it out. refer to the rfc 9429 https://www.rfc-editor.org/rfc/rfc9429#section-5.3.1-21.2,
> 
> > A direction attribute, determined by applying the rules regarding the offered direction specified in [[RFC3264](https://www.rfc-editor.org/rfc/rfc9429#RFC3264)], [Section 6.1](https://rfc-editor.org/rfc/rfc3264#section-6.1), and then intersecting with the direction of the associated RtpTransceiver. For example, in the case where an "m=" section is offered as "sendonly" and the local transceiver is set to "sendrecv", the result in the answer is a "recvonly" direction.

It is not filtered out @cnderrauber .

It used to be
```
			return []RTPTransceiverDirection{
				RTPTransceiverDirectionRecvonly,
				RTPTransceiverDirectionSendrecv,
				RTPTransceiverDirectionSendonly,
			}
```
and this changes it back to
```
			return []RTPTransceiverDirection{
				RTPTransceiverDirectionRecvonly,
				RTPTransceiverDirectionSendrecv,
			}
```
So, `recvonly` is one of the possible directions. Just  disabled `sendonly` for that case.

rtptransceiver.go
```diff
@@ -260,10 +260,9 @@ func satisfyTypeAndDirection(
 			return []RTPTransceiverDirection{
 				RTPTransceiverDirectionRecvonly,
 				RTPTransceiverDirectionSendrecv,
-				RTPTransceiverDirectionSendonly,
 			}
 		case RTPTransceiverDirectionSendonly:
-			return []RTPTransceiverDirection{RTPTransceiverDirectionRecvonly, RTPTransceiverDirectionSendrecv}
+			return []RTPTransceiverDirection{RTPTransceiverDirectionRecvonly}
```
cnderrauber, line 280: sorry for the confused comment, I mean this line, does it remove `sendrecv` from mathing list of remote `sendonly`? @boks1971

rtptransceiver.go
```diff
@@ -260,10 +260,9 @@ func satisfyTypeAndDirection(
 			return []RTPTransceiverDirection{
 				RTPTransceiverDirectionRecvonly,
 				RTPTransceiverDirectionSendrecv,
-				RTPTransceiverDirectionSendonly,
```
cnderrauber, line 277:
```
 If an offered media stream is
   listed as sendrecv (or if there is no direction attribute at the
   media or session level, in which case the stream is sendrecv by
   default), the corresponding stream in the answer MAY be marked as
   sendonly, recvonly, sendrecv, or inactive. 
```

According to 3264 https://www.rfc-editor.org/rfc/rfc3264#section-6.1 , `sendonly` is qualified.

rtptransceiver.go
boks1971: ah got it, sorry about the confusion. Yes, it does remove it. If we leave `sendrecv` in there, it breaks like I mentioned in the PR comment ☝️ , i. e.
- Answering side first does `AddTrack`, that creates a transceiver in `sendrecv`.
- When remote offer comes, the above transceiver can be matched against a `sendonly` media section in the offer and change direction to `sendrecv` because this transceiver is trying to send also.

rtptransceiver.go
boks1971: Good catch. Let me include that back.

rtptransceiver.go
cnderrauber: will it work if we change the matched transceiver's direction to correct value? like change the line https://github.com/pion/webrtc/blob/8efd17e592a989bf4e69295aa4e5921d9bf119dd/peerconnection.go#L1189 to

```
				if transceiver.Direction() == RTPTransceiverDirectionInactive || SendRecv {
					transceiver.setDirection(RTPTransceiverDirectionRecvonly)
				}

```

boks1971: I will try that.

The remote has put `sendonly` in the `offer`. This section says, if the local is `sendrecv`, it has to be changed to `recvonly`. Maybe, it can set that and still send media for that transceiver?
https://www.rfc-editor.org/rfc/rfc9429#section-5.3.1-21.2

boks1971: That did not work :-(. Changed it to something like
```
             case direction == RTPTransceiverDirectionSendonly:
                 if transceiver.Direction() == RTPTransceiverDirectionInactive || transceiver.Direction() == RTP  TransceiverDirectionSendrecv {
                      transceiver.setDirection(RTPTransceiverDirectionRecvonly)
                  }
```

And verified that `answer` is setting it to `recvonly` in chrome://webrtc-internals, but the receive side is not rendering media. Will run Chrome in debug and check if there are any interesting logs.

rtptransceiver.go
boks1971: Pushed a change to allow matching `sendrecv` to `sendonly`.

Sean-Der:
@boks1971 I worry this will get accidentally reverted again. Would you mind adding a test or something else that would make things less brittle?

boks1971:
> @boks1971 I worry this will get accidentally reverted again. Would you mind adding a test or something else that would make things less brittle?

There are already tests there @Sean-Der . But, I will look to add more.

Sean-Der:
👍if you think it’s good I’m ok with that.

I broke it and I worry I would break it again. If it’s as simple as flipping some constant values.


My hope is that a beginner could open a PR against Pion and be easy. Want to make it harder to break things is all

boks1971:
> 👍if you think it’s good I’m ok with that.
> 
> I broke it and I worry I would break it again. If it’s as simple as flipping some constant values.
> 
> My hope is that a beginner could open a PR against Pion and be easy. Want to make it harder to break things is all

good point Sean. I will have a look at adding more tests.

PR #3199, opened 2025-08-22
Remove RTX codec if no primary

boks1971:
#### Description

While adding transceivers from SetRemoteDescription, the filtered codecs could filter out the primart codec and leave the RTX codec in. Generating an answer with that fails `SetRemoteDescription` on remote peer due to an unrecognisable codec. Fix it by filtering out RTX is primary is not there.

### DIFF
--- a/rtpcodec.go
+++ b/rtpcodec.go
@@ -5,6 +5,7 @@ package webrtc
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/pion/webrtc/v4/internal/fmtp"
@@ -153,6 +154,38 @@ func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) Paylo
 	}
 
 	return PayloadType(0)
+}
+
+// Given needle CodecParameters, returns if needle is RTX and
+// if primary codec corresponding to that needle is in the haystack of codecs.
+func primaryPayloadTypeForRTXExists(needle RTPCodecParameters, haystack []RTPCodecParameters) (
+	isRTX bool, primaryExists bool,
+) {
+	if !strings.EqualFold(needle.MimeType, MimeTypeRTX) {
+		return
+	}
+
+	isRTX = true
+	parsed := fmtp.Parse(needle.MimeType, needle.ClockRate, needle.Channels, needle.SDPFmtpLine)
+	aptPayload, ok := parsed.Parameter("apt")
+	if !ok {
+		return
+	}
+
+	primaryPayloadType, err := strconv.Atoi(aptPayload)
+	if err != nil || primaryPayloadType < 0 || primaryPayloadType > 255 {
+		return
+	}
+
+	for _, c := range haystack {
+		if c.PayloadType == PayloadType(primaryPayloadType) {
+			primaryExists = true
+
+			return
+		}
+	}
+
+	return
 }
 
 // For now, only FlexFEC is supported.
--- a/rtpcodec_test.go
+++ b/rtpcodec_test.go
@@ -9,6 +9,199 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestFindPrimaryPayloadTypeForRTX(t *testing.T) {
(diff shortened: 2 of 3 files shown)


cnderrauber (approved):
Looks good!!

boks1971:
> Looks good!!

@cnderrauber Thank you for the review. I moved the function to filter out unattached RTX into `RTPTransceiver`'s `SetCodecPreferences` method so that it is always applied whenever codec preferences are set.

rtpcodec.go
```diff
@@ -155,6 +156,27 @@ func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) Paylo
 	return PayloadType(0)
 }
 
+// Given RTX CodecParameters find the primary CodecParameters if one exists.
+func findPrimaryPayloadTypeForRTX(needle RTPCodecParameters, haystack []RTPCodecParameters) PayloadType {
+	parts := strings.Split(needle.SDPFmtpLine, "=")
```
nils-ohlmeier: It might be worth adding an additional sanity check here that the mime type for `needle` is actually the RTX mime type - just in case someone calls this with a non RTX codec.

nils-ohlmeier (approved):
Looks good.
One suggestion for making it more bullet proof, but I'll leave it up to you if want to add that or not.

rtpcodec.go
boks1971: I was thinking about it too @nils-ohlmeier . Thank you.

I am doing it at call site. Because at call site, if the returned payload type is 0 (BTW, that is a valid payload type for PCMU which is a different pain 🙂 , but does not apply in this code), it means that there is no primary for the given RTX and that RTX is removed from list of codecs.

If this function is called with a non-RTX and if this code returns 0, it could remove the codec.

Maybe, this should return two values (PayloadType, error) and return an error if given `needle` is not RTX? WDYT?

boks1971: Or maybe `(PayloadType, bool)` with the second return value indicating if needle is `rtx`?

rtpcodec.go
```diff
@@ -155,6 +156,27 @@ func findRTXPayloadType(needle PayloadType, haystack []RTPCodecParameters) Paylo
...
+	if len(parts) != 2 || parts[0] != "apt" {
+		return PayloadType(0)
+	}
+
+	primaryPayloadType, err := strconv.Atoi(parts[1])
```
JoTurk: Wouldn't this break if there is a `rtx-time`? e.g `apt=90;rtx-time=2000` maybe we can use the new internal fmtp parser https://github.com/pion/webrtc/blob/master/internal/fmtp/fmtp.go

boks1971: ooh, I did not know about the fmtp parser. Will check it out.

rtpcodec.go
nils-ohlmeier: Yes PCMU is a problem. Maybe the right thing to do here is to rename you function to something like `primaryPayloadTypeForRTXExists` and it just returns a bool?
I think that would avoid all (?) of these problems, or?

nils-ohlmeier: Oh and BTW very cool PR. Thanks for fixing this!

boks1971: Sounds good. I think it will still need two return values. I will change it to
```
primaryPayloadTypeForRTXExists(needle, haystack) (isRTX bool, primaryExists bool)
```
Calls site can remove unattached RTX  `if isRTX, exists := primary...(); isRTX && !primaryExists {}`. Can then do the RTX check inside too.

boks1971: Update PR with the change.

rtpcodec.go
boks1971: Changed to use fmtp parser.

I think that would be a nice thing to move it out of internal too.

JoTurk: Cool thank you, yeah a good idea, We can do it in the next minor release :)

JoeTurki (approved):
Thank you :)

PR #3201, opened 2025-08-24
Add partialMatch codecs to transceiver from remote

cnderrauber:
When a recvonly transceiver is created from remote offer, skip partial matched codec will cause the
receiver can't be reused to send the codec.

#### Reference issue
Fixes #...


boks1971 (approved):
Looks good!

PR #3204, opened 2025-08-25
Tests to ensure proper direction in SDP

boks1971:
#### Description

Per @Sean-Der 's  comments in this PR
(https://github.com/pion/webrtc/pull/3198) ([comment](https://github.com/pion/webrtc/pull/3198#issuecomment-3218234554)), adding a couple of more tests to ensure that direction in SDP is correct when remote offer is processed and transceivers are matched up against existing ones using type and direction.

Test flow
- AddTrack first on answer side. This should create a transceiver with `sendrecv`
- Receive remote description offer
- First time, it will try to match using type and direction. o If the remote offer had a media section with `sendonly`, the above should fail to find a match. In this case, a new transceiver will be created. So, there should be 2 transceivers (one from this + one from AddTrack in Step 1). In this case, the answer will have only one media section as the offer had only one slot. And the directions should be `remote = sendonly` and `local = recvonly`. o If the remote offer is `sendrecv`, it should find a match. So, there should be only one transceiver (the one created with AddTrack gets used for receive).  In this case, the answer will have only one media section as the offer had only one slot. And the directions should be `remote = sendrecv` and `local = sendrecv`.
- Then the remote track is removed from offer. Processing that remote decription will find a transceiver by `mid`, but directions are checked in the SDP to ensure they have correct values.

### DIFF
--- a/peerconnection_go_test.go
+++ b/peerconnection_go_test.go
@@ -1285,6 +1285,146 @@ func TestPeerConnection_TransceiverDirection(t *testing.T) {
 	}
 }
 
+func TestPeerConnection_MediaDirectionInSDP(t *testing.T) {
+	lim := test.TimeOut(time.Second * 30)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	createTransceiver := func(pc *PeerConnection, dir RTPTransceiverDirection) (*RTPSender, error) {
+		// AddTransceiverFromKind() can't be used with sendonly
+		if dir == RTPTransceiverDirectionSendonly {
+			codecs := pc.api.mediaEngine.getCodecsByKind(RTPCodecTypeVideo)
+
+			track, err := NewTrackLocalStaticSample(codecs[0].RTPCodecCapability, util.MathRandAlpha(16), util.MathRandAlpha(16))
+			if err != nil {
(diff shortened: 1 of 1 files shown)


boks1971:
Have not seen this one before

 ```
Annotations
1 error and 2 warnings
[lint / Go](https://github.com/pion/webrtc/actions/runs/17210516066/job/48820854178#step:5:166)
issues found
[lint / Metadata](https://github.com/pion/webrtc/actions/runs/17210516066/job/48820854164#step:2:24)
Restore cache failed: Dependencies file is not found in /home/runner/work/webrtc/webrtc. Supported file pattern: go.sum
[lint / Go](https://github.com/pion/webrtc/actions/runs/17210516066/job/48820854178#step:2:16)
Restore cache failed: Dependencies file is not found in /home/runner/work/webrtc/webrtc. Supported file pattern: go.sum
```

boks1971:
@Sean-Der Please let me know if this looks good. This checks the SDP to ensure that it has the proper direction.

Confirmed that these tests fail if I revert https://github.com/pion/webrtc/pull/3198

Sean-Der:
nice, I think that makes a big different @boks1971 

This part has been so fragile, thank you for doing this.

PR #3208, opened 2025-08-27
Match codec order of remote peer

boks1971:
#### Description

Done when creating a transceiver from remote description to respect codec order preference of remote peer.

There was a recent change to include partial matches which overwrote same codecs and also rtx was getting magled.

Change it by removing codecs from search space as matches are found so that a codec match is applied only once.

Also, move RTX matching to separate block to ensure proper RTXes ar matched.

### DIFF
--- a/mediaengine.go
+++ b/mediaengine.go
@@ -630,6 +630,23 @@ func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) e
 			return err
 		}
 
+		addIfNew := func(existingCodecs []RTPCodecParameters, codec RTPCodecParameters) []RTPCodecParameters {
+			found := false
+			for _, existingCodec := range existingCodecs {
+				if existingCodec.PayloadType == codec.PayloadType {
+					found = true
+
+					break
+				}
+			}
+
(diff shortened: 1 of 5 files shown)


mediaengine.go
```diff
@@ -642,9 +658,24 @@ func (m *MediaEngine) updateFromRemoteDescription(desc sdp.SessionDescription) e
 			remoteCodec.RTCPFeedback = rtcpFeedbackIntersection(localCodec.RTCPFeedback, remoteCodec.RTCPFeedback)
 
 			if matchType == codecMatchExact {
-				exactMatches = append(exactMatches, remoteCodec)
+				exactMatches = addIfNew(exactMatches, remoteCodec)
+			} else if matchType == codecMatchPartial {
+				partialMatches = addIfNew(partialMatches, remoteCodec)
+			}
+		}
+		// second pass in case there were missed RTX codecs
```
boks1971, line 667: don't know if this can happen spec wise, but if rtx codec is first in the list, it does not get added, so making a second pass to catch those.

cnderrauber (approved):

Issue #3134, opened 2025-05-30
Add support for emitting inbound-rtp and media-playout stats from PeerConnection.GetStats


Carsinalys:
### What version of Pion are you using?
`v4.1.1`

### Current behaviour
`StatsTypeInboundRTP` ("inbound-rtp") and `StatsTypeMediaPlayout` ("media-playout") are defined in `stats.go`, but they are never instantiated or added to the `StatsReport` returned by `PeerConnection.GetStats()`.

Calling `pc.GetStats()` on an actively receiving connection returns:
- ICE / DTLS / SCTP transport stats
- Candidate pair & certificate stats
- Codec stats
- A synthetic peer-connection node

...but no `inbound-rtp` or `media-playout` objects, so applications cannot obtain basic per-SSRC metrics (bytes/packets/jitter, etc.) or audio-playout metrics.

### Expected behaviour (per WebRTC-Stats spec)
`GetStats()` should include:
1. At least one `InboundRTPStreamStats` entry for every primary SSRC the RTPReceiver is consuming
2. Optionally an `AudioPlayoutStats` entry (or equivalent) describing local audio playout for that track
3. Key fields such as `packetsReceived`, `packetsLost`, `jitter`, `bytesReceived`, `framesDecoded`, etc. should be populated

### Root cause analysis
`PeerConnection.GetStats()` only collects from:
- ICEGatherer / ICETransport
- DTLSTransport / SCTPTransport
- DataChannels
- Certificates
- MediaEngine
- A synthetic PeerConnectionStats

No `collectStats` method exists for:
- RTPReceiver
- RTPSender
- TrackRemote
- Any audio-playout component

Therefore the structs defined in `stats.go` are never instantiated.

## Minimal Repro
```go
package main

import (
    "fmt"
    "github.com/pion/webrtc/v4"
)

func main() {
    pc, _ := webrtc.NewPeerConnection(webrtc.Configuration{})
    // Set up remote SDP, receive an audio/video track...
    stats := pc.GetStats()
    for id, s := range stats {
        fmt.Printf("%s → %s\n", id, s.Type())
    }
    // No item prints "inbound-rtp" or "media-playout"
}
```

### Proposed fix
1. Implement `collectStats()` on media objects
   - `RTPReceiver.collectStats()` → build an `InboundRTPStreamStats` per SSRC
   - (Complementary `RTPSender.collectStats()` for `OutboundRTPStreamStats` would also be useful)
[...]

PR #3221, opened 2025-09-11
Add collection for inbound-rtp stats

drshrey:
### Summary

Implement receiver-side stats for each primary SSRC: emits InboundRTPStreamStats with key fields (ssrc, kind, mid, transportId, codecId) and prepares for populating counters (packetsReceived, packetsLost, jitter, bytesReceived, framesDecoded, etc.). 

Out of scope: 
- This is not getting audio playout stats (media-playout). This will be implemented in a separate PR.

### Issue

Resolves pion/webrtc#3134

### How To Test

1. Run `TestPeerConnection_GetStats` in `stats_go_test.go`.

or 

1. Run a simple receive example and inspect stats:
2. Start any example that receives media (e.g., examples/stats/main.go).
3. Establish a call and let media flow for a few seconds.
4. Call pc.GetStats() and verify:
- At least one object with type == "inbound-rtp" per primary SSRC.
- Fields ssrc, kind, mid, transportId ("iceTransport"), and codecId are present.
- Key counters (packets/bytes/jitter/frames) show expected, increasing values as traffic flows.

### DIFF
--- a/interceptor.go
+++ b/interceptor.go
@@ -7,6 +7,7 @@
 package webrtc
 
 import (
+	"sync"
 	"sync/atomic"
 
 	"github.com/pion/interceptor"
@@ -14,6 +15,7 @@ import (
 	"github.com/pion/interceptor/pkg/nack"
 	"github.com/pion/interceptor/pkg/report"
 	"github.com/pion/interceptor/pkg/rfc8888"
+	"github.com/pion/interceptor/pkg/stats"
 	"github.com/pion/interceptor/pkg/twcc"
 	"github.com/pion/rtp"
 	"github.com/pion/sdp/v3"
@@ -35,8 +37,40 @@ func RegisterDefaultInterceptors(mediaEngine *MediaEngine, interceptorRegistry *
 		return err
 	}
 
+	if err := ConfigureStatsInterceptor(interceptorRegistry); err != nil {
+		return err
+	}
+
 	return ConfigureTWCCSender(mediaEngine, interceptorRegistry)
 }
+
+// ConfigureStatsInterceptor will setup everything necessary for generating RTP stream statistics.
+func ConfigureStatsInterceptor(interceptorRegistry *interceptor.Registry) error {
+	statsInterceptor, err := stats.NewInterceptor()
+	if err != nil {
+		return err
+	}
+	statsInterceptor.OnNewPeerConnection(func(id string, stats stats.Getter) {
+		statsGetter.Store(id, stats)
+	})
+	interceptorRegistry.Add(statsInterceptor)
+
+	return nil
+}
+
+// lookupStats returns the stats getter for a given peerconnection.statsId.
+func lookupStats(id string) (stats.Getter, bool) {
+	if value, exists := statsGetter.Load(id); exists {
+		if getter, ok := value.(stats.Getter); ok {
+			return getter, true
+		}
+	}
+
+	return nil, false
+}
+
+// key: string (peerconnection.statsId), value: stats.Getter
+var statsGetter sync.Map // nolint:gochecknoglobals
 
 // ConfigureRTCPReports will setup everything necessary for generating Sender and Receiver Reports.
 func ConfigureRTCPReports(interceptorRegistry *interceptor.Registry) error {
--- a/interceptor_test.go
+++ b/interceptor_test.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"reflect"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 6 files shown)


stats.go
```diff
@@ -429,11 +429,11 @@ type InboundRTPStreamStats struct {
 	TotalSquaredInterFrameDelay float64 `json:"totalSquaredInterFrameDelay"`
 
 	// PacketsReceived is the total number of RTP packets received for this SSRC.
-	PacketsReceived uint32 `json:"packetsReceived"`
+	PacketsReceived uint64 `json:"packetsReceived"`
 
 	// PacketsLost is the total number of RTP packets lost for this SSRC. Note that
 	// because of how this is estimated, it can be negative if more packets are received than sent.
-	PacketsLost int32 `json:"packetsLost"`
+	PacketsLost int64 `json:"packetsLost"`
```
drshrey: I figured we should conform to the proper signed integer type that we get from stats interceptor's InboundRTPStreamStats in this case since we want to encapsulate the full range of possible values (u)int64 would return (which 32 wouldn't be able to).

drshrey: Nevermind, this is a major change since this is an exported type so it's a breaking change. Will undo this!

rtpreceiver.go
```diff
@@ -401,6 +403,76 @@ func (r *RTPReceiver) Stop() error { //nolint:cyclop
...
+
+			pl := stats.InboundRTPStreamStats.PacketsLost
+			if pl > math.MaxInt32 {
+				inboundStats.PacketsLost = math.MaxInt32
+			} else if pl < math.MinInt32 {
+				// RFC 3550: PacketsLost = expected - received, can be negative for duplicates/reordering
+				inboundStats.PacketsLost = math.MinInt32
+			} else {
+				inboundStats.PacketsLost = int32(pl)
+			}
```
drshrey: @Sean-Der @JoeTurki Since it doesn't seem like we can change the types from int32 to int64 in stats.go (or rather it would be a breaking change since it's an exported type according to the linter failures I was seeing), is this the best way to handle casting of the int64 value we get from inboundStats to int32?

JoTurk: You can just use `//nolint:[linter_name]` this lint rule is just for you to double check that everything is correct and intended, and safe, but it's not smart enough to know that \:)
we have many nolints :)

JoTurk: If you want to be very detailed about it, You can also log a warning when it wraps around :) and just simplify the casting logic here.

interceptor.go
```diff
@@ -35,9 +37,41 @@ func RegisterDefaultInterceptors(mediaEngine *MediaEngine, interceptorRegistry *
...
+// getStatsGetter returns the stats getter for a given peerconnection.statsId.
+func getStatsGetter(id string) (stats.Getter, bool) {
+	if value, exists := statsGetter.Load(id); exists {
+		if getter, ok := value.(stats.Getter); ok {
+			return getter, true
+		}
+	}
+
+	return nil, false
+}
```
philipch07: A small thing, but perhaps we can rename this to something like `lookupStats()` to avoid having "get" twice. Or maybe it's worth renaming the variable from `statsGetter` to something else.

interceptor_test.go
```diff
@@ -353,6 +353,25 @@ func Test_Interceptor_ZeroSSRC(t *testing.T) {
...
+	defer report()
+
+	pc, err := NewPeerConnection(Configuration{})
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, pc.Close())
+	}()
+
+	assert.NotNil(t, pc.statsGetter, "statsGetter should be non-nil with NewPeerConnection")
+}
```
philipch07: I think it would be good to assert whatever the default value of `pc.statsGetter` is as well since asserting that something isn't nil doesn't guarantee that it's the correct value.

interceptor.go
```diff
@@ -35,9 +37,41 @@ func RegisterDefaultInterceptors(mediaEngine *MediaEngine, interceptorRegistry *
...
+	statsInterceptor.OnNewPeerConnection(func(id string, stats stats.Getter) {
+		statsGetter.Store(id, stats)
+	})
+	interceptorRegistry.Add(statsInterceptor)
+
+	return nil
+}
+
+// lookupStatsGetter returns the stats getter for a given peerconnection.statsId.
+func lookupStats(id string) (stats.Getter, bool) {
```
philipch07, line 62: One last tiny thing, I think the comment should have the same function name as the function.

drshrey: Ah yes, thanks for catching this!

philipch07 (approved):
Great job!! Don't forget to squash before merging.

PR #3227, opened 2025-09-14
Apply go modernize suggestions

philipch07:
#### Description
Applies go modernization suggestions, see https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize for more details.

@JoeTurki and I are looking into a linting option for this, but in the meantime applying these changes should be trivial as it only makes the codebase cleaner.

#### Reference issue
N/A

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


JoeTurki (approved):

JoTurk:
Thank you.

> @JoeTurki and I are looking into a linting option for this, but in the meantime applying these changes should be trivial as it only makes the codebase cleaner.

makes sense.

PR #3229, opened 2025-09-19
Cleanup statsGetter after peer is closed

JoeTurki:
#### Description

Cleanup statsGetter after peer is closed to prevent memory leaks.


Sean-Der (approved):
thanks for fixing :)

PR #3230, opened 2025-09-19
Prefer makezero with a cap

JoeTurki:
#### Description

remove `//nolint:makezero` and prefer make 0.


JoTurk:
This also fixes a subtle bug in `handlePendingPackets` where the result packets wouldn't match `m.pendingPackets` if we drop some of them.

Issue #3226, opened 2025-09-12
Add proper statistics to Pion


philipch07:
## Summary

Pion is currently missing some statistics related to various functions. This issue is more like an epic to manage then partition the problem into different parts. Ideally, completing this feature request should introduce no breaking changes.

Currently there are two existing issues related to this:

- #3134 (emitting stats from `PeerConnection.GetStats`) would ideally be resolved by @drshrey's pr: #3221 
- https://github.com/pion/webrtc/issues/2779 (a step towards completing ICECandidatePairStats) would be resolved by my pr: https://github.com/pion/ice/pull/809

Here are additional subtasks from @JoeTurki: 

- make the RTCP sender and receiver more easy to use and more automatic [...maybe we should have an easy mode like the javascript api]
  - @Sean-Der suggested spinning up an ReadRTCP internally, until we detect a user has done it themselves.
    - @at-wat Would you happen to have a better idea for this? There is some discussion in the #dev channel in the Pion discord server if you're interested in joining.
- make stats part of pion/interceptor more integrated and easy to use
- easier and cleaner way for people to send media stats when Pion is the receiver
- test pion RTCP more with other clients
- work on the getStats() api, maybe revive and finish #610 

@JoeTurki also mentioned that @jech had additional comments related to handling stats in Pion.

## Additional context

There is some more discussion that has taken place in the Pion discord server, but this is the gist of the conversation so far.

PR #3234, opened 2025-09-26
Add AudioPlayoutStatsProvider interface for getting media-playout stats

drshrey:
#### Description

- Refactored AudioPlayoutStatsProvider to provider-centric API with AddTrack/RemoveTrack methods, enabling multi-device and multi-track scenarios

- Provider represents a playout device (playout-id); tracks register to providers and stats are automatically collected via pc.GetStats()

#### Sample Usage

```go
// Create provider for a playback device
provider := webrtc.NewAudioPlayoutStatsProvider("speaker-device-1")

// Register tracks with the provider
provider.AddTrack(remoteTrack)

// Inside playback callback
provider.Accumulate(frames*channels, sampleRate, deviceDelay, synthesized)

// Switch devices 
provider.RemoveTrack(remoteTrack)
headphoneProvider.AddTrack(remoteTrack)

// Stats collection via GetStats
for id, stat := range pc.GetStats() {
    if playout, ok := stat.(webrtc.AudioPlayoutStats); ok {
        // consume playout stats
    }
}
```

#### Reference issue

Resolves https://github.com/pion/webrtc/issues/3134 (part 2/2)

Also related to resolving https://github.com/pion/webrtc/issues/3226

### DIFF
--- a/track_remote.go
+++ b/track_remote.go
@@ -7,6 +7,7 @@
 package webrtc
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -32,6 +33,8 @@ type TrackRemote struct {
 	receiver         *RTPReceiver
 	peeked           []byte
 	peekedAttributes interceptor.Attributes
+
+	audioPlayoutStatsProviders []AudioPlayoutStatsProvider
 }
 
 func newTrackRemote(kind RTPCodecType, ssrc, rtxSsrc SSRC, rid string, receiver *RTPReceiver) *TrackRemote {
@@ -235,6 +238,70 @@ func (t *TrackRemote) HasRTX() bool {
 	defer t.mu.RUnlock()
 
 	return t.rtxSsrc != 0
+}
+
+func (t *TrackRemote) addProvider(provider AudioPlayoutStatsProvider) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	for _, p := range t.audioPlayoutStatsProviders {
+		if p == provider {
+			return
+		}
+	}
+
+	t.audioPlayoutStatsProviders = append(t.audioPlayoutStatsProviders, provider)
+}
+
+func (t *TrackRemote) removeProvider(provider AudioPlayoutStatsProvider) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	for i, p := range t.audioPlayoutStatsProviders {
+		if p == provider {
+			t.audioPlayoutStatsProviders = append(t.audioPlayoutStatsProviders[:i], t.audioPlayoutStatsProviders[i+1:]...)
+
+			return
+		}
+	}
+}
+
+func (t *TrackRemote) pullAudioPlayoutStats(now time.Time) []AudioPlayoutStats {
+	t.mu.RLock()
+	providers := t.audioPlayoutStatsProviders
+	t.mu.RUnlock()
+
+	if len(providers) == 0 {
+		return nil
+	}
+
+	var allStats []AudioPlayoutStats
+	for _, provider := range providers {
+		stats, ok := provider.Snapshot(now)
+		if !ok {
+			continue
+		}
+
+		if stats.ID == "" {
+			stats.ID = fmt.Sprintf("media-playout-%d", uint32(t.SSRC()))
+		}
+
+		if stats.Type == "" {
+			stats.Type = StatsTypeMediaPlayout
+		}
+
+		if stats.Kind == "" {
+			stats.Kind = string(MediaKindAudio)
+		}
+
+		if stats.Timestamp == 0 {
+			stats.Timestamp = statsTimestampFrom(now)
+		}
+
+		allStats = append(allStats, stats)
+	}
+
+	return allStats
 }
 
 func (t *TrackRemote) setRtxSSRC(ssrc SSRC) {
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -413,6 +413,7 @@ func (r *RTPReceiver) collectStats(collector *statsReportCollector, statsGetter 
 		mid = r.tr.Mid()
 	}
 	now := statsTimestampNow()
+	nowTime := now.Time()
 	for trackIndex := range r.tracks {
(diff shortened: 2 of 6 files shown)


JoTurk:
@drshrey Sorry for taking time, I read half of it, And i'll post my review tomorrow

track_remote.go
```diff
@@ -237,6 +240,47 @@ func (t *TrackRemote) HasRTX() bool {
 	return t.rtxSsrc != 0
 }
 
+// SetAudioPlayoutStatsProvider registers the audio playout stats provider for this track.
+func (t *TrackRemote) SetAudioPlayoutStatsProvider(provider AudioPlayoutStatsProvider) {
```
JoTurk: One problem with this approach is that playout stats is tied to `playout-id` and not a track, playout-id can be shared between many tracks, we can for example swap audio tracks but keep playout stats, we'll eventually have to link all the play-out ids across all the stats, because naturally it maps to a single playback device in most of the cases.

I think we should think about standardizing playout-id now and where to add / hook stats, what do you think?

track_remote.go
```diff
@@ -237,6 +240,47 @@ func (t *TrackRemote) HasRTX() bool {
...
+func (t *TrackRemote) pullAudioPlayoutStats(now time.Time) (AudioPlayoutStats, bool) {
+	t.mu.RLock()
+	provider := t.audioPlayoutStatsProvider
+	defer t.mu.RUnlock()
+
+	if provider == nil {
+		return AudioPlayoutStats{}, false
+	}
+
+	stats, ok := provider.Snapshot(now, t)
```
JoTurk: Hmm, I think we should unlock the mutex before calling user provided API.

JoTurk:
@at-wat @Sean-Der can you please take a look, and if you have a suggestion about the API design and where to hook stats, playout-ids etc.

audio_playout_stats_provider.go
```diff
@@ -0,0 +1,87 @@
...
+
+// AudioPlayoutStatsProvider is an interface for getting audio playout metrics.
+// The user can provide their own AudioPlayoutStatsProvider implementations, or use
+// the default implementation provided by Pion.
+type AudioPlayoutStatsProvider interface {
+	Snapshot(now time.Time, track *TrackRemote) (AudioPlayoutStats, bool)
+}
+
+// DefaultAudioPlayoutStatsProvider accumulates audio playout stats on behalf of the application.
+type DefaultAudioPlayoutStatsProvider struct {
```
JoTurk: Btw does this struct need to be public?

audio_playout_stats_provider.go
```diff
@@ -0,0 +1,87 @@
...
+	mu sync.Mutex
+
+	stats           AudioPlayoutStats
+	lastSynthesized bool
+}
+
+// NewDefaultAudioPlayoutStatsProvider constructs a default provider with the supplied stats ID.
+func NewDefaultAudioPlayoutStatsProvider(id string) *DefaultAudioPlayoutStatsProvider {
```
JoTurk: Also i feel this name is too long for a public API, what do you think?

audio_playout_stats_provider.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"sync"
+	"time"
+)
+
+// AudioPlayoutStatsProvider is an interface for getting audio playout metrics.
+// The user can provide their own AudioPlayoutStatsProvider implementations, or use
+// the default implementation provided by Pion.
+type AudioPlayoutStatsProvider interface {
+	Snapshot(now time.Time, track *TrackRemote) (AudioPlayoutStats, bool)
```
at-wat: In the sample usage, the stats provider instance is bound to the remote track on creation.
So, I'm not sure passing track on `Snapshot()` make sense.
I think it should be passed via `New...()` if the provider implementation uses the remote track pointer.

> ```go
> provider := webrtc.NewDefaultAudioPlayoutStatsProvider(fmt.Sprintf("media-playout-%d", uint32(remoteTrack.SSRC())))
> remoteTrack.SetAudioPlayoutStatsProvider(provider)
> ```

audio_playout_stats_provider.go
drshrey: Agreed. Perhaps `NewAudioPlayoutStatsProvider`? The default seems unnecessary/ redundant although I do want to keep audio since we may want to have a separate provider for video.

audio_playout_stats_provider.go
drshrey: Ah it does not, will make private, thank you!

track_remote.go
drshrey: Oh thank you good call, especially if Snapshot is calling any TrackRemote method (which would need the same mutex), this could result in deadlock.

audio_playout_stats_provider.go
drshrey: 100% agreed, that is way cleaner and makes more sense given we're already tying it to a specific track to your point. Thank you, will address this!

drshrey: In this case, our default implementation does not even use the TrackRemote object, so we do not need to even pass it into `New...()` 😅

track_remote.go
drshrey: ~~Actually in regards to https://github.com/pion/webrtc/pull/3234/files#r2386496043, we do not need to even pass `t` into Snapshot so hopefully this shouldn't be relevant anymore!~~

Oh actually, it may still be relevant in case others have an implementation where they do pass the TrackRemote object so this issue would then still hold.

track_remote.go
drshrey: (chatted with Joe and we came to the following new model):

A provider will add and remove tracks vs the track setting a provider. This keeps the model flexible to use cases like the track being associated with multiple devices (speaker (provider), headphones (provider)). It also makes for a cleaner and more intuitive abstraction since it's provider-centric: "speaker device is playing track1, headphone provider is playing track1" vs. the old version saying "track1 is being played by speaker device, headphone device". 

This also prevents us from having to add a public method to the TrackRemote API.

JoeTurki (approved):
Thank you a lot, sorry for requesting too much :)

PR #3249, opened 2025-10-19
Expose stats ID for use in interceptor factories

mengelbart:
#### Description

We are already using the ID to access interceptors. If we expose it, we can store data per peerconnection in interceptor factories and give users an API to interceptors of a specific peerconnection.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -35,8 +35,8 @@ import (
 // peer-to-peer communications with another PeerConnection instance in a
 // browser, or to another endpoint implementing the required protocols.
 type PeerConnection struct {
-	statsID string
-	mu      sync.RWMutex
+	id string
+	mu sync.RWMutex
 
 	sdpOrigin sdp.Origin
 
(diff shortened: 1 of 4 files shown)


peerconnection.go
```diff
@@ -597,11 +597,11 @@ func (pc *PeerConnection) GetConfiguration() Configuration {
 	return pc.configuration
 }
 
-func (pc *PeerConnection) getStatsID() string {
+func (pc *PeerConnection) ID() string {
```
drshrey, line 600: @JoeTurki just checking, does this method need to exist in peerconnection_js.go in order for consumers compiling to wasm to be able to access this method (and do we care)? I suspect the wasm build tests are passing because this is a new method, but checking whether it's ok to introduce a new public method here but not in the js/wasm equivalent file.

JoTurk: Good question, we currently don't support stats API in wasm, it would be cool to provide a fallback for stats and pion's interceptor but it will be a lot of work, and we can't support everything.
Thank you.

JoTurk:
@mengelbart sorry for the ping, is this ready for merge?

mengelbart:
I think it is.

PR #3250, opened 2025-10-20
Preferr RRID for retransmission over RTX

nils-ohlmeier:
#### Description

When receiving an offer with simulcast enabled from Chrome the offer contains RRID and RTX both offered to be used for retransmitting missing packets. Without this PR Pion's answer SDP would contain both mechanisms. That results in Chrome actually sending the retransmission twice, once over RRID and a second time over RTX.

This PR results in Pion preferring RRID and omitting RTX from the m-section payloads to avoid this double retransmission.

### DIFF
--- a/rtptransceiver.go
+++ b/rtptransceiver.go
@@ -101,6 +101,18 @@ func (t *RTPTransceiver) setCodecPreferencesFromRemoteDescription(media *sdp.Med
 		return
 	}
 
+	simulcast := false
+	if len(getRids(media)) > 0 {
+		extensions, err := rtpExtensionsFromMediaDescription(media)
+		if err == nil {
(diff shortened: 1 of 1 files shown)


JoTurk (approved):
Was really tricky to test manually but it works, thank you.
I wonder if we can add a unit test 🤔

nils-ohlmeier:
I think to verify only the signaling part should not be too hard.
If you want to verify the retransmission behavior on the wire as well, that would get a lot more tricky.
I'll start looking into a signaling/SDP test.

philipch07:
@nils-ohlmeier I think this PR is good to merge if you can add a test unless there's something else that you maybe wanted to update!

PR #3252, opened 2025-10-21
Add simple datachannel example with demo.html

Sundenis:
This PR adds a new simple -datachannel example to help newcomers understand how to create and test a basic WebRTC DataChannel using Go and a static HTML page (demo.html).

- Includes main.go for signaling and WebRTC setup.
- Includes demo.html to test sending/receiving messages.
- Tested locally and works with the provided signaling server.

### DIFF
--- a/examples/data-channels-simple/README.md
+++ b/examples/data-channels-simple/README.md
@@ -1,0 +1,23 @@
+# WebRTC DataChannel Example in Go
+
+This is a minimal example of a **WebRTC DataChannel** using **Go (Pion)** as the signaling server.
+
+## Features
+
+- Go server for signaling
+- Browser-based DataChannel
+- ICE candidate exchange
+- Real-time messaging between browser and Go server
+
+## Usage
+
+1. Run the server:
+
+```
+go run main.go
+```
+
+2. Open browser at http://localhost:8080
+
+3. Send messages via DataChannel and see them in terminal & browser logs.
+
--- a/examples/data-channels-simple/demo.html
+++ b/examples/data-channels-simple/demo.html
@@ -1,0 +1,88 @@
+<!DOCTYPE html>
+<html>
+  <!--
+		SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+		SPDX-License-Identifier: MIT
+	-->
+<head>
+  <meta charset="utf-8">
+  <title>DataChannel Test</title>
+</head>
+<body>
+  <h2>📡 WebRTC DataChannel Test</h2>
+  <input id="msg" placeholder="Message">
+  <button id="sendBtn" disabled onclick="sendMsg()">Send</button>
+  <pre id="log"></pre>
+
+  <script>
+    const pc = new RTCPeerConnection({iceServers:[{urls:"stun:stun.l.google.com:19302"}]});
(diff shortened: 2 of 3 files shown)


ValorZard:
Hm, what's the difference between this and https://github.com/pion/webrtc/tree/master/examples/data-channels ?

philipch07:
> Hm, what's the difference between this and https://github.com/pion/webrtc/tree/master/examples/data-channels ?

The proposed simple-datachannels automatically performs the handshake which hides the handshake which the user would otherwise have to manually perform. I think I updated that example as this one was being written, but I'm not opposed to this being added to show how the handshake can be automated, although maybe this one can have more comments in the code to distinguish itself from the original datachannels example. Just a suggestion, but unsure if there's more to this PR atm.

ValorZard:
@philipch07 I think that we need some sort of either example or tutorial showing how to setup a signaling server. It's probably the one things I've struggled with when it comes to WebRTC. Granted, we now have the WHIP/WHEP standard, but that doesn't really apply to datachannel applications.

It's why in the https://github.com/pion/example-webrtc-applications/tree/master/ebiten-game example, I included a sample signaling server I wrote for the game (though I don't think the code in it is very good)

JoTurk:
@philipch07 @ValorZard I feel like this example solves another problem of providing an easy to use example for datachannels for new users.  WHIP and WHEP would be cool but datachannels aren't standard too.  I think we can add as much examples as we can to make things easier.
For me I think we should merge this. 
What do you think?

JoeTurki (approved):

philipch07:
> @philipch07 I think that we need some sort of either example or tutorial showing how to setup a signaling server. It's probably the one things I've struggled with when it comes to WebRTC.

I'll see if I can take a look into learning about the signaling server if an example is needed for it.

> I think we can add as much examples as we can to make things easier.

That's fair, it is a bit of a simpler example for datachannels which the name implies.

I did have an issue with the example: I followed the instructions as I checked out the branch and ran it in VSC connected to WSL. When I opened the page in Firefox it wasn't connecting to the channel, but when I used the "simple browser" inside VSC it was fine. I'm what was causing that issue, since every time I opened the page in Firefox it was showing the ICE candidates in my terminal but it would never connect.

JoTurk:
@philipch07  It just needs to gather and send the ICE candidates, which is what i was going to do before merging it and fix the lint, I think the talk here is about the idea of adding a simple / just work. I think we should merge @Sundenis example is a good idea. 

What do you think?

philipch07:
Oops yeah, sorry I was thinking a little out of scope. This pr sounds good to me!

JoTurk:
@ValorZard Are you okay with merging this after some fixes?

ValorZard:
@JoeTurki yeah sure, I'm just more so thinking about like, maybe adding some sort of example for signaling servers specifically in a later PR maybe

JoTurk:
@ValorZard That will be cool btw no pressure but if you can move your whep-like example from being a draft we should merge it.

JoTurk:
@Sundenis hello I fixed the lint issues, and the ICE issue. Merging this now.

JoTurk:
Also i renamed the example directory to data-channels-simple so it's sortable with other examples.

PR #3263, opened 2025-11-13
ECN support

penhauer:
`mux` and `endpoint` structs either implemented `net.Conn` or `net.PacketConn`. Now they implement either `transport.PacketConnSocket` or `transport.NetConnSocket` which include the ReadWithAttributes method. 
`srtp.NewSessionSRTPWithNewSocket` is used for creating SRTP session with the desired functionality.

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pion/logging"
 	"github.com/pion/rtcp"
 	"github.com/pion/srtp/v3"
+	"github.com/pion/transport/v3"
 	"github.com/pion/webrtc/v4/internal/mux"
 	"github.com/pion/webrtc/v4/internal/util"
 	"github.com/pion/webrtc/v4/pkg/rtcerr"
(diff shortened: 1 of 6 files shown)


sirzooro:
[RFC 6679 section 6.4](https://www.rfc-editor.org/rfc/rfc6679.html#section-6.4) says that this should be negotiated using `a=ice-options: rtp+ecn` attribute in SDP. This is not implemented here.

JoTurk:
I'll finish this and do the SDP integration once https://github.com/pion/webrtc/pull/3513 is merged, so we don't have to pass the attributes through the mux layer.

JoTurk:
> [RFC 6679 section 6.4](https://www.rfc-editor.org/rfc/rfc6679.html#section-6.4) says that this should be negotiated using a=ice-options: rtp+ecn attribute in SDP. This is not implemented here.

not used by libwebrtc, I think we'll skip that

JoTurk:
ECN support is merged to ICE, this pr will get obsolete by the DTLS v4 change because mux is getting removed, so we have to close this for now, I'll make sure you're added as coauthor when we add ECN to webrtc after dtls@v4

PR #3270, opened 2025-11-18
Add deterministic NACK/RTX reproduction test for issue #3063

lars-sto:
#### Description
This PR adds a minimal end-to-end test that reproduces the NACK/RTX loop.

The test intentionally drops the second outgoing RTP packet and verifies that:
- exactly one NACK is sent for the missing packet,
- exactly one RTX retransmission is received,
- no additional NACK/RTX packets appear after a short grace period.

The test is currently skipped (t.Skip()) so it doesn't break CI.

#### Reference issue
Minimal Test for issue #3063

### DIFF
--- a/interceptor_test.go
+++ b/interceptor_test.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


interceptor_test.go
```diff
@@ -561,3 +563,164 @@ func testInterceptorNack(t *testing.T, requestNack bool) { //nolint:cyclop
...
+		h := &rtp.Header{}
+		if _, err := h.Unmarshal(c.UserData()); err != nil {
+			return true
+		}
+
+		if h.PayloadType == 96 {
+			mediaPacketCount++
+
+			if mediaPacketCount == 2 {
+				state.lostSeq = h.SequenceNumber
```
JoTurk, line 614: would you mind adding a lock around state so we don't have a race in the test?

JoeTurki (commented):
Thank you a lot for making this, once you address the race condition we can merge this, thank you again.

lars-sto:
Thanks a lot for the quick review, Joe! I should have fixed all the race conditions now. Let me know if you spot anything else!

PR #3274, opened 2025-11-21
Added option to configure DTLS Cipher Suites

sirzooro:
Added new option to SettingEngine to configure DTLS Cipher Suites


sirzooro:
> This is cool, I wanted to add this for the SCTP benchmark.
> 
> Nit: Would you mind adding a simple comment to explain the difference between CipherSuites and CustomCipherSuites because I'm worried that we might confuse some users.

Done, I have improved comments for both options.

PR #3287, opened 2025-12-05
ivfwriter: Add WithTimebase option for precise RTP timestamp handling

### LINKED ISSUE #3286: IVFWriter timestamp precision loss due to millisecond conversion
## Your environment
  - pion/webrtc version: v4.x
  - Codec: VP8/VP9/AV1

  ## What did you do?
  Recorded a 15fps video (225 frames) using IVFWriter.WriteRTP()

  ## What did you expect?
  Video duration: 15 seconds

  ## What happened?
  Video duration: ~16 seconds (accumulated rounding errors from integer division)

  ## Root Cause
  WriteRTP converts RTP timestamps to milliseconds, losing precision:
  relativeTstampMs := 1000 * uint64(...) / i.clockRate  // truncates fractional ms

miya-masa:
#### Description
  Add `WithTimebase(numerator, denominator)` option that enables direct use of RTP timestamps as PTS values, preserving full
  timestamp precision.

  The current implementation converts RTP timestamps to milliseconds before calculating PTS, which causes precision loss. For a
  15fps video, this results in ~16 seconds duration instead of the expected 15 seconds.

  With `WithTimebase(1, 90000)`, RTP timestamps are written directly as PTS values without millisecond conversion.

  #### Reference issue
  Fixes #3286

### DIFF
--- a/pkg/media/ivfwriter/ivfwriter.go
+++ b/pkg/media/ivfwriter/ivfwriter.go
@@ -382,23 +382,18 @@ func WithFrameRate(numerator, denominator uint32) Option {
 	}
 }
 
-// WithTimebase sets the IVF timebase (numerator/denominator) and enables
-// direct use of RTP timestamps as PTS values.
+// WithDirectPTS enables direct use of RTP timestamps as PTS values
+// without millisecond conversion.
 //
-// The timebase defines the time unit for PTS values in the IVF container.
-// For example, WithTimebase(1, 90000) means each PTS unit represents 1/90000 seconds.
-//
-// When this option is used, RTP timestamps are written directly as PTS values
-// without millisecond conversion, preserving full timestamp precision.
-// This is ideal for accurate timing when converting RTP streams to IVF.
+// When this option is used, RTP timestamps are written directly as PTS values,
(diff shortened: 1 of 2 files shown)


pkg/media/ivfwriter/ivfwriter.go
```diff
@@ -361,3 +381,26 @@ func WithFrameRate(numerator, denominator uint32) Option {
...
+//	WithTimebase(1, 90000)
+func WithTimebase(numerator, denominator uint32) Option {
+	return func(i *IVFWriter) error {
+		i.timebaseNumerator = numerator
+		i.timebaseDenominator = denominator
+		i.directPTS = true
+
+		return nil
+	}
+}
```
JoTurk: since we already have `WithFrameRate(numerator, denominator uint32)` we should simplify this to `WithDirectPTS` that does one thing.

miya-masa: Thanks for the review! You're right - I've simplified it to WithDirectPTS() which only enables direct PTS mode. Users can use the existing WithFrameRate(1, 90000) to set the timebase.

PR #3289, opened 2025-12-05
Option to ignore rid pause in a=simulcast:recv

boks1971:
#### Description

Before this change, if the SDP offer has paused rids, SDP answer echoed it back. That causes issues where enabling/disabling of layers is signalled via the signalling channel. The following scenario breaks
1. PeerB signals PeerA to disable all simulcast layers.
2. PeerA sends an SDP offer with RIDs disabled. a. NOTE: A renegotiation is not needed, but happens for a different reason.
3. PeerB signals again to PeerA to re-enable layers.
4. PeerB sends answer for the offer in Step 2 with paused rids.
5. PeerA keeps layers paused.

Adding a setting engine option to not change existing behaviour. Pause state of a rid will be ignored if setting is enabled.


cnderrauber (approved):

Sean-Der:
@boks1971 I am not familiar with RID pausing.

Why would we not want this to just be the default behavior? Why would someone want to enable/disable this?

Is the existing behavior wrong/buggy?

boks1971:
> @boks1971 I am not familiar with RID pausing.
> 
> Why would we not want this to just be the default behavior? Why would someone want to enable/disable this?
> 
> Is the existing behavior wrong/buggy?

Don't think it is buggy @Sean-Der , but not very well defined. According to RFC (https://datatracker.ietf.org/doc/html/rfc8853), using `~` means pause and it looks like either side can do it. But, the actual usage makes it quite complex, i. e. managing layer pauses/resume using SDP. Did not make it the default as I was wary of changing existing behaviour.

nils-ohlmeier:
Sorry I haven't looked at the code, but did re-read the RFC parts to freshen up my memory. So the RFC says that the INITIAL SDP offerer can pause a given RID by using ~ to advertise to the SDP answerer, that this particular RID is not going to be transmitted initially.

To me the wording is pretty clear that:
- neither can the SDP answerer pause a RID, which the offerer hasn't paused
- nor can either side pause any RID after it has been made active (not paused any more)

@boks1971 is that what your code does, or should we adapt it to the RFC?

BTW I'm pretty sure that none of the browsers actually supports the pausing of a RID. I would assume the SDP parser can handle it, but I doubt the browsers do anything useful with that information.

boks1971:
Thank you for the comments @nils-ohlmeier .  I missed the part about the initial pause in the spec. I thought it could be paused at any time. Yeah, spec mentions initial pause explicitly, so must be that case only.

In practice though, it happens during regular operation too. The sequence is
1. Peer A is set up with `setParameters` to do three layers
2. Peer A ends all three layers initials
3. Peer A is requested to stop all layers via a signalling channel message.
4. Peer A does that using `setParameters` and layers are stopped. There is no need for a re-negotiation.
5. But, if some other condition happens on Peer A side needing a re-negotiation, it sends an SDP offer with paused RIDs.
6. Then it is possible to race between signalling and SDP answer (signalling says re-enable layers, but SDP answer says pause it). Seems like browser (headless Chromium via Playwright) is acting on it. Peer A logs showed that it is activating layers using `setParameters`, but there were no frames emitted.

In Step 5 above, it is not initial. It had sent frames on all layers prior, but was later de-activated.

What I did was to add a setting engine option to ignore the paused state on the simulcast receiver side. As I did not want to change the default behaviour, added it as a setting engine option.

PR #3290, opened 2025-12-05
Check for closed receiver before setting up rid

boks1971:


rtpreceiver.go
```diff
@@ -552,6 +552,12 @@ func (r *RTPReceiver) receiveForRid(
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
+	select {
```
Sean-Der: Can you use `haveClosed` instead?

rtpreceiver.go
```diff
@@ -587,6 +593,15 @@ func (r *RTPReceiver) receiveForRtx(
 	rtcpReadStream *srtp.ReadStreamSRTCP,
 	rtcpInterceptor interceptor.RTCPReader,
 ) error {
+	r.mu.Lock()
```
Sean-Der, line 592: Can you use `haveClosed` instead?

peerconnection.go
```diff
@@ -1875,9 +1875,6 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream *srtp.ReadStreamSRTP, ssr
 			}
 
 			if rsid != "" {
-				receiver.mu.Lock()
```
Sean-Der, line 1884: If you use the atomic do you still need locking changes?

Sean-Der:
@boks1971 is it possible to get a test/reproduce for this behavior?

rtpreceiver.go
boks1971: Will change.

peerconnection.go
boks1971: Yeah, I just moved the lock inside. Yeah, there are more internal state like `tracks` in `RTPReceiver` which needs locking. Not sure why the lock was not internal and called from peerconnection.go. Seemed like a good thing to move it inside.

boks1971:
> @boks1971 is it possible to get a test/reproduce for this behavior?

Unfortunately, I do not have a specific repro. Happens in production in a larg(ish) call (40 - 50 participants) and they are subscribing to 60 - 80 tracks each and they are connecting from possibly poorer networks.

I have been trying to get a definitive proof, but also looking all around for possible places. This one seemed like one possible option as a lot of objects created by buffer factory are leaking.

boks1971:
I have been running this in a cluster for a couple of weeks and have not seen issues. Unfortunately, I cannot definitively prove that there were any leaks in this path and this change helped with those leaks. Also harder to write a good test. Can only say it did not cause issues. Seemed like a good change to make to not potentially extend lifecycle of a closed object.

I am okay to close this if this does not look like the right thing to do.

JoTurk:
@boks1971 hello, I pushed a basic scuffed test, feel free to edit it to get the right behavior, but it fails on master. and another test to just assert that we stop, but the TestPeerConnection_Simulcast_LateRIDRSIDAfterReceiverClosed should capture the behavior. if i understand correctly.

boks1971:
> @boks1971 hello, I pushed a basic scuffed test, feel free to edit it to get the right behavior, but it fails on master. and another test to just assert that we stop, but the TestPeerConnection_Simulcast_LateRIDRSIDAfterReceiverClosed should capture the behavior. if i understand correctly.

Thank you so much @JoeTurki. Really appreciate this.

Yeah, I think these tests cover the scenario.

If you are okay approving, will get this merged.

boks1971:
@Sean-Der Will merge this based on @JoeTurki 's review and tests. 

Have also been running this branch in prod for a couple of weeks without any issues.

Without this, I am seeing heaps with a lot of memory attributed to `ReadStreamSRTP.write` (https://github.com/pion/srtp/blob/95d5e685df1d68ab8158b9ae9eab3eac0d5a2c41/stream_srtp.go#L68-L77) which is quite puzzling as there is no allocation there. I am guessing part of it could be due to addressed with this. Going to keep digging after merging this and deploying.

If you have concerns, please let me know and I will address.

Screenshot 2025-12-24 at 17 01 35

PR #3296, opened 2025-12-09
Option to check for fingerprint in DTLS handshake

boks1971:
According to [this]( https://www.potaroo.net/ietf/all-ids/draft-ietf-mmusic-dtls-sdp-08.html#:~:text=The%20certificate%20received%20during%20the,may%20have%20undesirable%20latency%20effects.), DTLS session should be terminated immediately if the fingerprint check fails.

There is also this from DTLS spec which says bad records should be ignored - https://github.com/pion/dtls/pull/759#issuecomment-3609030208, but I think that might apply to application data.

~~This PR adds a setting engine to option to check for certificate fingerprint during handshake and uses the DTLS config `VerifyPeerCertificate` hook to check.~~

~~Added a test which is exactly same as the `TestInvalidFingerprintCausesFailed` with the setting mentioned above enabled. There is no direct proof that the early detection path is taken. I confirmed it using some debug logging, but it would be good to do it in a testable way. Open to ideas on how to do that.~~

~~This PR also includes changes from @JoeTurki in this [PR](https://github.com/pion/webrtc/pull/3295) as I was struggling to get the unit test to work and Joe found the source of the issue.~~

Use `VerifyPeerCertificate` hook of DTLS config to check fingerprint during handshake (unless fingerprint check is disabled).


Sean-Der:
@boks1971 Should we just make this the default behavior?

The current behavior of 'wait until handshake is complete' is just wrong. This just moves the check earlier.

Maybe I am missing some nuance, but I am all for
* Making this the default behavior
* Updating/fixing the existing test

JoTurk:
@Sean-Der @boks1971  the only problem with as the default is that `VerifyPeerCertificate` currently sends a DTLS alert which is as far as i can tell isn't recommended for such errors in DTLS, according to the spec we must tear down the media session, nothing about sending DTLS alerts and I think it's not recommended to be chatty at the DTLS layer with such errors. 
I'm a bit worried but I think it's fine as a default, for now, also because it's a fatal alert. and because our settings are getting too big.
Maybe later we can make the alerts optin for this and other dtls problems.

Sean-Der:
`VerifyPeerCertificate` should we go fix this? Bug in `pion/dtls` that also impacts non-WebRTC users!

100% agree on SettingEngine. I only want to add more flags if users REALLY need them. In this case I can't think of a reason why someone would want to enable/disable this flag. If the current behavior is wrong we shouldn't require users to put more effort in to get non-buggy behavior.

I might be misunderstanding though!

JoTurk:
> VerifyPeerCertificate should we go fix this? Bug in pion/dtls that also impacts non-WebRTC users!

Yes it's kinda related to the other fix in https://github.com/pion/dtls/issues/758 we should make alerts optin, and also optin to send alerts that we don't send right now.

> 100% agree on SettingEngine. I only want to add more flags if users REALLY need them. In this case I can't think of a reason why someone would want to enable/disable this flag. If the current behavior is wrong we shouldn't require users to put more effort in to get non-buggy behavior.

Yeah i agree.

boks1971:
I can make this the default. Will update the PR shortly. There is no test change needed I think. I just need to remove the new test I added.

Also, I think it is fine to send an alert too. IIUC, DTLS inherits a bunch of behaviour from TLS and error handling in TLS is quite simple (as they say it :-) ) - https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.2, i. e. the side detecting the error should send the alert and any fatal alert should close things on either side.

JoTurk:
> Also, I think it is fine to send an alert too. IIUC, DTLS inherits a bunch of behaviour from TLS and error handling in TLS is quite simple (as they say it :-) ) - https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.2, i. e. the side detecting the error should send the alert and any fatal alert should close things on either side.

This is one of the parts where DTLS disagrees with TLS, I think it's fine for our case here, because we have other form of integrity, but sending alerts in dtls is more nuanced.

1. https://datatracker.ietf.org/doc/html/rfc6347#section-4.1.2.7
2. https://datatracker.ietf.org/doc/html/rfc6347#section-4.1.2.1
3. https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.7

boks1971:
> > Also, I think it is fine to send an alert too. IIUC, DTLS inherits a bunch of behaviour from TLS and error handling in TLS is quite simple (as they say it :-) ) - https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.2, i. e. the side detecting the error should send the alert and any fatal alert should close things on either side.
> 
> This is one of the parts where DTLS disagrees with TLS, I think it's fine for our case here, because we have other form of integrity, but sending alerts in dtls is more nuanced.
> 
> 1. https://datatracker.ietf.org/doc/html/rfc6347#section-4.1.2.7
> 2. https://datatracker.ietf.org/doc/html/rfc6347#section-4.1.2.1
> 3. https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.7

Thank you @JoeTurki . I need to read more :-) 

Updated the PR to make the check during handshake the default behaviour.

dtlstransport.go
```diff
@@ -371,6 +371,26 @@ func (t *DTLSTransport) Start(remoteParameters DTLSParameters) error { //nolint:
...
+	dtlsConfig.VerifyPeerCertificate = func(rawCerts [][]byte, _verifiedChains [][]*x509.Certificate) error {
+		if len(rawCerts) == 0 {
+			return errNoRemoteCertificate
+		}
+
+		t.lock.Lock()
+		defer t.lock.Unlock()
+		t.remoteCertificate = rawCerts[0]
+
+		if !t.api.settingEngine.disableCertificateFingerprintVerification {
```
Sean-Der: sorry to be so pedantic, can we do early return? Easier to reason about I did a double take 

```
if t.api.settingEngine.disableCertificateFingerprintVerification {
  return nil
}

			parsedRemoteCert, parseErr := x509.ParseCertificate(t.remoteCertificate)
			if parseErr != nil {
				return parseErr
			}

			return t.validateFingerPrint(parsedRemoteCert)
			```

Sean-Der (approved):

Sean-Der:
nice change! Less LoC is something that makes me soooo happy

dtlstransport.go
boks1971: Will do. Thank you Sean. I prefer early return too.

PR #3299, opened 2025-12-12
Parse opus mappings

JoTurk:
#### Description
This is the first pr of 3. Adds support for more channel mapping families other than 0 (we add support for 1, 2 and 255). while it doesn't make sense for RTP opus as most clients only support mono or stereo audio. Sending audio over SCTP is a thing :)

### DIFF
--- a/pkg/media/oggreader/oggreader.go
+++ b/pkg/media/oggreader/oggreader.go
@@ -7,6 +7,7 @@ package oggreader
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/media/oggreader/oggreader.go
```diff
@@ -97,31 +103,85 @@ func (o *OggReader) readHeaders() (*OggHeader, error) {
...
+	if err := validatePageHeader(pageHeader, payload); err != nil {
+		return nil, err
+	}
+
+	header := parseBasicHeaderFields(payload)
+	if err := parseChannelMapping(header, payload); err != nil {
+		return nil, err
+	}
+
+	return header, nil
```
philipch07, line 115: According to https://datatracker.ietf.org/doc/html/rfc7845#section-5.1.1 after figure 3:

> For channel mapping family 0, this value defaults to 1, and not coded.

But I didn't see it covered here, I'm not sure if it's supposed to be here or somewhere else.

JoTurk: This is true, and also `CoupledCount` but it's intended to be left absent because the values are implied by the spec, and this isn't a decoder, it's a reader / parser, it doesn't conflict with the spec, because we don't have any real audio logic, it's a reader. 
so tldr, i think this is the right approach for separation of concerns. if we add full spec logic and correctness, we'll need to add other things too, and hide many stuff from the user.

philipch07: Oh that makes more sense, I had the idea of readers and decoders mixed up for this.

philipch07 (approved):

PR #3300, opened 2025-12-12
Rework ogg reader so it supports multi tracks

JoTurk:
#### Description
after https://github.com/pion/webrtc/pull/3299 
Introduces a new constructor for ogg reader, expose header parsing APIs, so handling / reading the header is controlled by the user, and not hardcoded to only the first header, we also expose serial id, so users can parser and handle multi track ogg files.


philipch07:
Really cool ^^

PR #3301, opened 2025-12-12
Read opus tags

JoTurk:
#### Description
Add a vorbis tags parser to the ogg parser, so we can parse ARTIST and TITLE and cool shit like that, based on https://github.com/pion/webrtc/pull/3300

### DIFF
--- a/pkg/media/oggreader/oggreader.go
+++ b/pkg/media/oggreader/oggreader.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


JoTurk:
I added more unit tests so OggReader is 100% covered.

pkg/media/oggreader/oggreader.go
```diff
@@ -354,3 +357,124 @@ func generateChecksumTable() *[256]uint32 {
...
+// UserComment is a key-value pair of a vorbis comment.
+type UserComment struct {
+	Comment string
+	Value   string
+}
+
+// ParseOpusTags parses an OpusTags from the page payload.
+func ParseOpusTags(payload []byte) (*OpusTags, error) {
+	const (
+		headerMagicLen = 8
```
philipch07, line 378: I think a comment about where this is from in the spec (preferably w/ a quote so it can be ctrl+f in the future will be good)

JoTurk: thanks. https://www.xiph.org/vorbis/doc/v-comment.html is already linked in OpusTags struct :)

philipch07: I thought it was from https://datatracker.ietf.org/doc/html/rfc7845#section-5.2?

JoTurk: good point added this link too.

PR #3303, opened 2025-12-12
Use ICE options API

JoTurk:
#### Description

Move to the new ICE with options API so we can use the new stuff.

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -8,6 +8,7 @@ package webrtc
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -70,7 +71,7 @@ func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
 	}, nil
 }
 
-func (g *ICEGatherer) createAgent() error { //nolint:cyclop
+func (g *ICEGatherer) createAgent() error {
 	g.lock.Lock()
 	defer g.lock.Unlock()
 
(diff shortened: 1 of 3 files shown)


JoTurk:
I'm going to add tests for all the missing options, I found that we don't actually test many of them.

JoTurk:
~~I want to add a test for relay's timeout but ice doesn't honor WithNet for the turn client. Making a PR to ice ...~~

JoTurk:
Hmm, it actually does, something in the test was just wrong.

icegatherer.go
```diff
@@ -70,87 +71,211 @@ func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
...
+		opts = append(opts, ice.WithDisconnectedTimeout(*g.api.settingEngine.timeout.ICEDisconnectedTimeout))
+	}
+	if g.api.settingEngine.timeout.ICEFailedTimeout != nil {
+		opts = append(opts, ice.WithFailedTimeout(*g.api.settingEngine.timeout.ICEFailedTimeout))
+	}
+	if g.api.settingEngine.timeout.ICEKeepaliveInterval != nil {
+		opts = append(
+			opts, ice.WithKeepaliveInterval(*g.api.settingEngine.timeout.ICEKeepaliveInterval),
+		)
+	}
```
philipch07, line 208:
```suggestion
	if g.api.settingEngine.timeout.ICEKeepaliveInterval != nil {
		opts = append(opts, ice.WithKeepaliveInterval(*g.api.settingEngine.timeout.ICEKeepaliveInterval))
	}
```
small formatting

icegatherer_test.go
```diff
@@ -84,6 +91,241 @@ func TestICEGather_mDNSCandidateGathering(t *testing.T) {
...
+
+	parseCreds := func(sdp string) (string, string) {
+		var ufrag, pwd string
+		for _, l := range strings.Split(sdp, "\n") {
+			l = strings.TrimSpace(l)
+			if strings.HasPrefix(l, "a=ice-ufrag:") {
+				ufrag = strings.TrimPrefix(l, "a=ice-ufrag:")
+			} else if strings.HasPrefix(l, "a=ice-pwd:") {
+				pwd = strings.TrimPrefix(l, "a=ice-pwd:")
+			}
```
philipch07:
```suggestion
			if after, ok := strings.CutPrefix(l, "a=ice-ufrag:"); ok  {
				ufrag = after
			} else if after0, ok0 := strings.CutPrefix(l, "a=ice-pwd:"); ok0  {
				pwd = after0
			}
```
The go extension suggested this, I don't mind the current one or this, it's just style at this point.

icegatherer.go
JoTurk: good eye, It was formatted this way because of LLL before the refactor.

icegatherer_test.go
JoTurk: nice. ty.

Sean-Der (approved):

Sean-Der:
Fantastic!

PR #3305, opened 2025-12-14
Add nohost gather policy

### LINKED ISSUE #3176: SCTP fails when using SetInterfaceFilter and pion is an answerer
### Summary
I want to disable all host candidates and use only server-reflexive (srflx) ones.
To achieve that, I used SettingEngine.SetInterfaceFilter like this:
```
settingEngine.SetInterfaceFilter(func(name string) bool {
    return false
})
```
This works correctly when Pion is an offerer: the ICE connection establishes, and SCTP works as expected.

However, when Pion acts as an answerer, ICE succeeds but SCTP fails in ~80% of cases. The failure happens during SCTP handshake: the incoming COOKIE-ACK never arrives, even though ICE reports "connected".

In about 20% of runs, the full connection including SCTP does succeed — so the behavior is stochastic.

### Investigation
Using strace, I observed that:

In normal operation, Pion creates two distinct UDP sockets (separate file descriptors).

When using SetInterfaceFilter (excluding all interfaces), both srflx and host candidates share the same underlying socket (same FD).

I suspect this causes an internal conflict or race condition when routing incoming packets (especially during SCTP setup), particularly for answerer mode.

### Environment
Pion version: v4.0.10

Remote peer: Google Chrome 138.0.7204.49

OS: Linux (but should be reproducible cross-platform)

### Repro
Use SetInterfaceFilter to exclude all interfaces.

Set up a peer connection where Pion is an answerer.

Wait for Chrome to initiate.

Observe that ICE connects but SCTP often fails.

### Expected behavior
Even with a single srflx candidate and one UDP socket, SCTP should succeed reliably when Pion is an answerer.

### Additional notes
I suspect a socket muxing issue in ICE or DTLS layers when both host and srflx candidates share a socket (fd).

JoTurk:
#### Description
Currently, we don't have a way to limit which candidates we gather beyond lite-mode (host candidates only) and relay mode (relay only). As a result, users end up relying on hacky workarounds like https://github.com/pion/webrtc/issues/3176 to limit to srflx only, which aren't reliable due to how ICE actually works.
Added many tests to make sure that it actually works.
Edit: this pr was changed from SetCandidateTypes to just nohost gather policy which did exist in ORTC https://github.com/w3c/ortc/issues/742
this resolves https://github.com/pion/webrtc/issues/3176

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -170,8 +170,13 @@ func (g *ICEGatherer) resolveCandidateTypes() []ice.CandidateType {
 	if g.api.settingEngine.candidates.ICELite {
 		return []ice.CandidateType{ice.CandidateTypeHost}
 	}
-	if g.gatherPolicy == ICETransportPolicyRelay {
+
+	switch g.gatherPolicy {
+	case ICETransportPolicyRelay:
 		return []ice.CandidateType{ice.CandidateTypeRelay}
+	case ICETransportPolicyNoHost:
+		return []ice.CandidateType{ice.CandidateTypeServerReflexive, ice.CandidateTypeRelay}
+	default:
 	}
 
 	return nil
--- a/icegatherer_test.go
+++ b/icegatherer_test.go
@@ -336,6 +336,155 @@ func gatherCandidatesWithSettingEngine(t *testing.T, se SettingEngine, opts ICEG
 	return candidates
 }
 
+func TestICEGatherer_NoHostPolicyVNet(t *testing.T) { //nolint:cyclop
+	lim := test.TimeOut(time.Second * 20)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	const (
+		stunIP     = "1.2.3.4"
+		stunPort   = 3478
+		externalIP = "1.2.3.10"
+		localIP    = "10.0.0.1"
+		realm      = "pion.ly"
+		timeout    = 3 * time.Second
+	)
+
+	loggerFactory := logging.NewDefaultLoggerFactory()
+
+	wan, err := vnet.NewRouter(&vnet.RouterConfig{
+		CIDR:          "1.2.3.0/24",
+		LoggerFactory: loggerFactory,
+	})
+	assert.NoError(t, err)
+
+	stunNet, err := vnet.NewNet(&vnet.NetConfig{
+		StaticIP: stunIP,
+	})
+	assert.NoError(t, err)
+	assert.NoError(t, wan.AddNet(stunNet))
+
+	clientLAN, err := vnet.NewRouter(&vnet.RouterConfig{
+		StaticIPs: []string{fmt.Sprintf("%s/%s", externalIP, localIP)},
+		CIDR:      "10.0.0.0/24",
+		NATType: &vnet.NATType{
+			Mode: vnet.NATModeNAT1To1,
+		},
+		LoggerFactory: loggerFactory,
+	})
+	assert.NoError(t, err)
+
+	clientNet, err := vnet.NewNet(&vnet.NetConfig{
+		StaticIP: localIP,
+	})
+	assert.NoError(t, err)
+	assert.NoError(t, clientLAN.AddNet(clientNet))
+	assert.NoError(t, wan.AddRouter(clientLAN))
+	assert.NoError(t, wan.Start())
+	defer func() {
+		assert.NoError(t, wan.Stop())
+	}()
+
+	stunListener, err := stunNet.ListenPacket("udp4", net.JoinHostPort(stunIP, fmt.Sprintf("%d", stunPort)))
+	assert.NoError(t, err)
+
+	turnServer, err := turn.NewServer(turn.ServerConfig{
+		Realm:         realm,
+		LoggerFactory: loggerFactory,
+		PacketConnConfigs: []turn.PacketConnConfig{
+			{
+				PacketConn: stunListener,
+				RelayAddressGenerator: &turn.RelayAddressGeneratorStatic{
+					RelayAddress: net.ParseIP(stunIP),
+					Address:      "0.0.0.0",
+					Net:          stunNet,
+				},
+			},
+		},
+	})
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, turnServer.Close())
+	}()
+
+	iceServer := ICEServer{
+		URLs: []string{fmt.Sprintf("stun:%s:%d", stunIP, stunPort)},
+	}
+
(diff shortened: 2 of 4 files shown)


icegatherer.go
```diff
@@ -70,89 +71,226 @@ func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
...
 	}
 
-	g.agent = agent
+	if g.api.settingEngine.candidates.ICELite {
+		return []ice.CandidateType{ice.CandidateTypeHost}
+	}
+
+	if len(candidateTypes) > 0 {
+		return candidateTypes
+	}
```
JoTurk: if relay mode and ice lite are enabled at the same time, we  return  `[]ice.CandidateType{ice.CandidateTypeRelay}` which isn't spec compliant, but safer so we don't leak IPs if the user didn't intended to do that, we should log a warning tho.

philipch07: Is it worth commenting it in the function so it's written in the code that we intentionally "break" the spec?

philipch07 (approved):

icegatherer.go
JoTurk: We don't break the spec, If a user turns on relay-only-mode and ice-lite at the same time, they are breaking the ice-lite spec, We should log an error when that happens, It's not a new behavior, it's how it works currently if you set both.

philipch07: I see, I'm in favor of logging if the user accidentally breaks the spec by setting both at the same time.

Sean-Der:
I am torn on adding `srflx` as a ICEGatherPolicy?

* It's not part of the standard (that is bad to just extend with `srflx`)
* But adding more to the grab bag that is SettingEngine is also not great for users
* If users search 'limit WebRTC to relay only' it will lead them to the same place everywhere.
* Unlikely to be accepted, but maybe we try and make a PR/request that `srflx` be added to w3c spec?

We are improving the spec and making it possible for other WebRTC implementations to get that goodness (even thoughts that don't have the SettingEngine pattern)


I support w/e you think is best.

Sean-Der (approved):

JoTurk:
@Sean-Der I did some digging and i found that ORTC had `nohost` policy, It doesn't give the same control as `SetCandidateTypes` in the original PR but it solves the issue. so I changed the PR to add it.

Sadly it got removed https://github.com/w3c/ortc/issues/742 also found this https://github.com/w3c/webrtc-pc/issues/1644#issuecomment-356485231 so I don't think we have a chance to add it back to the spec.

@nils-ohlmeier @fippo I see you guys in these discussions any thoughts?

nils-ohlmeier:
I see the advantage for the sake of flexibility for Pion, which isn't a browser. TBH the original motivation in #3176 looks pretty slim to me.
As @fippo mentioned a "real" advantage would come from such an option if you could instruct the ICE stack to skip host candidates all together. But it is not really a big advantage, because the ICE stack still needs to create sockets on the interfaces to be able to gather server reflexive candidates anyway. So the only "advantage" would be that a browser would not have to bubble up host candidates into JS land. And that it could not add the host candidate/socket to it's internal ICE check list.
But then modifying the ICE stack with that optimization is a lot of effort and risk for what I assume is a very small use case.

Long story short: I haven't seen any new argument here which would convince the browser developers to add specific support for this limited use case.

JoTurk:
@nils-ohlmeier Thank you so much. Yeah makes sense.

fippo:
The other advantage is that this avoids LNA (https://groups.google.com/a/chromium.org/g/blink-dev/c/CDy8LAs-DoA/m/H5pNeS3WAAAJ has some details)

@alvestrand @guidou since [libWebRTC offers nohost too](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/peer_connection_interface.h;l=282?q=webrtc::rtcconfiguration) WDYT about exposing that to JS too? LNA has changed things since https://github.com/w3c/webrtc-pc/issues/1644

JoTurk:
Sorry i just remembered this today. and sorry for the ping @fippo @alvestrand @guidou is there any update about exposing this to the JS API too?

fippo:
[The change itself is trivial](https://chromium-review.googlesource.com/c/chromium/src/+/7810456) but one needs a good definition. And web-platform tests even though the interesting behavior such as not exposing private IPs as rel-addr on srflx candidates is not testable due to infrastructure requirements.

The libWebRTC had me scratching my head why I got an IPv6 host candidate.  Turns out this was `2000::/3 ` GUA address (ask your favorite minion "what type of IPv6 address is  starting with 2003:") which is allowed by [this](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/p2p/client/basic_port_allocator.cc;l=151;drc=8c0301b5c76ca4d7e7ae01daa9bfe3bc5fe697a4;bpv=1;bpt=1).

The rationale of "if a STUN server would likely return the same address expose it as host rather than srflx" makes sense.

This address did get returned by the STUN server too so is ok to expose.

What does not seem ok is to expose a mdns host candidate in that case as this reveals that the browser has a GUA. Already an existing problem though.

I assume "noPrivateHost" is a better name given the behavior.

JoTurk:
@fippo wouldn't "noPrivate" make more sense in this case, not just for host, but also for srflx, mdns and prflx.
A friend just showed me that he can connect to other users in his isp's private network through a direct private path (prflx learned) and I don't think that there is a direct way to prevent this.

fippo:
That should already be covered by `relay`?

(but yes, ISPs are weird and at times let you connect with private IP addresses even though your public ip is different ...)

JoTurk:
I mean nohost would make sense for most p2p applications that aren't supposed to connect over private host so no reason to leak network details, think of Omegle-like websites 😅 but if that to be extended to noprivatehost I think it would be a good opertunity to do noPrivate.
Relay doesn't make sense for these applications, I chatted with few people that can provide use case for their applications.

fippo:
Yeah but that is what "relay" is for. In particular it prevents the local side from using non-relay candidates (i.e. local ports) for probing and discovery of addresses (ICE is good at that which is why we have prflx)

The argument for no(Private)Host then becomes "we do not need local network access"i. .e. won't *send* to anything LNA considers local (https://wicg.github.io/local-network-access/#ip-address-space-section has a long list).

You'll need to bring it to the W3C WG but I wouldn't expect much progress.

JoTurk:
Relay mode is overkill for simple p2p applications that barely afford to host a static frontend. and want the bare minimum privacy and less fingerprinting for their users (nohost or noprivatehost), and possibly with NoPrivate or srflx-only, not private prflx.
I'll try to get some app owners for to submit a concrete use cases for w3c wg but yeah i suspect it will go nowhere, thank you.

PR #3306, opened 2025-12-14
Test and wire automatic renomination

JoTurk:
#### Description

This is based on https://github.com/pion/webrtc/pull/3303 

Wires renomination in the settings engine, via extendable options, so we can add more options in the future, currently this only wires automatic renomination, as we need to design an APi for selecting peers, and doing manual controllable renomination (will be tricky to get right). 

Added vnet-based tests to make sure it's working as intended.

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/pion/ice/v4"
 	"github.com/pion/logging"
@@ -105,6 +106,7 @@ func (g *ICEGatherer) buildAgentOptions() []ice.AgentOption {
 	options = append(options, g.natRewriteOptions(nat1To1CandiTyp)...)
 	options = append(options, g.timeoutOptions()...)
 	options = append(options, g.miscOptions()...)
+	options = append(options, g.renominationOptions()...)
 
 	requestedNetworkTypes := g.api.settingEngine.candidates.ICENetworkTypes
 	if len(requestedNetworkTypes) == 0 {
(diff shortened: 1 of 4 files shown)


icegatherer.go
```diff
@@ -247,6 +249,31 @@ func (g *ICEGatherer) miscOptions() []ice.AgentOption {
 	return opts
 }
 
+func (g *ICEGatherer) renominationOptions() []ice.AgentOption {
+	renom := g.api.settingEngine.renomination
+	if !renom.enabled && !renom.automatic {
```
wrangelvid, line 254: Shouldn't this be just `!renom.enabled` to allow for renomination with automatic set to false?

settingengine.go
```diff
@@ -114,6 +116,67 @@ type SettingEngine struct {
...
+		return errInvalidRenominationInterval
+	}
+
+	if cfg.generator == nil {
+		cfg.generator = ice.DefaultNominationValueGenerator()
+	}
+
+	e.renomination.enabled = true
+	e.renomination.generator = cfg.generator
+	e.renomination.automatic = true
```
wrangelvid, line 175: Shouldn't this be derived from the cfg? The docstring suggests the user can change this.

wrangelvid (commented):
I understand from your PR description that only automatic renomination is enabled. Some of the function docstrings and the actual code seem to diverge on this. Might make sense to align them in either case if automatic is hardwired or not.

Looking forward to testing this in production!

icegatherer.go
JoTurk: No, because automatic can't be false with this API. we don't export an API to set it, and we don't export an API to renominate.

settingengine.go
JoTurk: The user can't change `automatic`. until we export an option to set it, and an API to use it. we're holding this until we think of an API that's extendable and let's the user select and influence candidates selection. 
I didn't update the godoc after i removed the manual `renomination` and the renominate API, I'll update the comment before merge. Thank you.

settingengine_test.go
```diff
@@ -55,6 +55,57 @@ func TestSetConnectionTimeout(t *testing.T) {
...
+
+	t.Run("AutomaticRenominationEnablesGenerator", func(t *testing.T) {
+		s := SettingEngine{}
+		assert.NoError(t, s.SetICERenomination())
+
+		assert.True(t, s.renomination.enabled)
+		assert.True(t, s.renomination.automatic)
+		assert.Nil(t, s.renomination.automaticInterval)
+		assert.NotNil(t, s.renomination.generator)
+	})
```
philipch07, line 100: This can probably get combined with `EnableWithDefaultGenerator`, if you bring over these asserts:
```go
assert.True(t, s.renomination.automatic)
assert.Nil(t, s.renomination.automaticInterval)
```

settingengine.go
wrangelvid: Sounds good ◡̈

JoTurk: updated the godoc thank you :)

settingengine_test.go
JoTurk: You're right now, but we'll need this test when we add manual renomination so it makes sense as a standalone test.

PR #3309, opened 2025-12-14
Wire address rewrite API

JoTurk:
#### Description
based on https://github.com/pion/webrtc/pull/3303 
Adds the new AddressRewrite API which replaces NAT1To1.
Added many vnet tests.

### DIFF
--- a/settingengine.go
+++ b/settingengine.go
@@ -54,6 +54,7 @@ type SettingEngine struct {
 		IPFilter                 func(net.IP) (keep bool)
 		NAT1To1IPs               []string
 		NAT1To1IPCandidateType   ICECandidateType
+		addressRewriteRules      []ice.AddressRewriteRule
 		MulticastDNSMode         ice.MulticastDNSMode
 		MulticastDNSHostName     string
 		UsernameFragment         string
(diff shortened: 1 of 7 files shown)


settingengine.go
```diff
@@ -270,6 +271,31 @@ func (e *SettingEngine) SetNAT1To1IPs(ips []string, candidateType ICECandidateTy
 	e.candidates.NAT1To1IPCandidateType = candidateType
 }
 
+// SetICEAddressRewriteRules configures address rewrite rules for candidate publication.
+// These rules provide fine-grained control over which local addresses are replaced or
+// supplemented with external IPs.
+// This replaces the legacy NAT1To1 settings, which will be deprecated in the future.
+func (e *SettingEngine) SetICEAddressRewriteRules(rules ...ICEAddressRewriteRule) error {
```
philipch07, line 349: I think we should put a warning in the doc comment for NAT1To1 settings that it will be deprecated and that `addressRewriteRules` should be used instead

Sean-Der: 100% agree.

We should also make an effort to find everywhere that `NAT1To1` is used (even libraries that use Pion) and update the code. So it stops getting recommended.

And some nice migration docs!

Sean-Der (approved):

settingengine.go
JoTurk: Good call we deprecated it in ICE, we should do it in webrtc too, I'll include migration docs in the release and wiki :)

JoTurk: Done, also added two test-examples

PR #3312, opened 2025-12-16
Fix RTP timestamp drift in WriteSample

JoTurk:
#### Description
Because of `uint32(sample.Duration.Seconds() * clockRate)` we used to truncate about 1.2s for 30fps@90khz video every one hour. This caused some a/v dsync over long periods, I had to do some crazy libfaketime + docker stuff to track this issue, and it turned out to be super basic issue lol
fix: Compute skipped ticks using the same float+remainder accumulator as normal sample.
Sadly had to add a lock.

### DIFF
--- a/track_local_static.go
+++ b/track_local_static.go
@@ -226,10 +226,12 @@ func (s *TrackLocalStaticRTP) Write(b []byte) (n int, err error) {
 // TrackLocalStaticSample is a TrackLocal that has a pre-set codec and accepts Samples.
 // If you wish to send a RTP Packet use TrackLocalStaticRTP.
 type TrackLocalStaticSample struct {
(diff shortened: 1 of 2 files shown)


track_local_static.go
```diff
@@ -226,10 +226,12 @@ func (s *TrackLocalStaticRTP) Write(b []byte) (n int, err error) {
 // TrackLocalStaticSample is a TrackLocal that has a pre-set codec and accepts Samples.
 // If you wish to send a RTP Packet use TrackLocalStaticRTP.
 type TrackLocalStaticSample struct {
+	mu         sync.Mutex
```
philipch07: I think it would be nice to leave a comment saying that the mutex is only used in `WriteSample`.

PR #3314, opened 2025-12-18
Implement write deadline

JoTurk:
#### Description
Make write deadline actually work.
Endpoint.SetDeadline(t):
1. Now directly calls e.mux.nextConn.SetDeadline(t).
2. This is a connection-wide operation. it affects every endpoint attached to the mux.
3. This makes SetDeadline the API you call when you want to control the underlying transport's read/write deadlines.

Endpoint.SetReadDeadline(t):
1. Still calls e.buffer.SetReadDeadline(t).
2. The new comment explicitly documents that this sets the read deadline on the internal packet buffer, not on the shared net.Conn.
3. Only reads from that specific Endpoint are affected. other endpoints and the mux's Read loop are unaffected.

### DIFF
--- a/internal/mux/endpoint.go
+++ b/internal/mux/endpoint.go
@@ -80,21 +80,24 @@ func (e *Endpoint) RemoteAddr() net.Addr {
 	return e.mux.nextConn.RemoteAddr()
 }
 
-// SetDeadline sets the read deadline for this Endpoint.
-// Write deadlines are not supported because writes go directly to the shared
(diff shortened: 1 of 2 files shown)


boks1971 (approved):

JoTurk:
Okay so:
1. SetDeadline and SetWriteDeadline directly calls the underlying conn deadline functions.
2. SetWriteDeadline set the write deadline in our internal buffer, which still unblocks, but it's per endpoint and softer than conn deadlines, I documented the differences.

PR #3323, opened 2025-12-25
Add continual gathering policy to SettingEngine

### LINKED ISSUE #3321: Support Continual Gathering
## Motivation

Now that automatic nomination is supported, it would be even stronger if we could keep gathering new ice candidates. Continual gathering has been supported in [pion/ice](https://github.com/pion/ice/blob/master/agent_options.go#L710), and the question is how we can expose it on the WebRTC api.

wrangelvid:
#### Description
Expose pion/ice's ContinualGatheringPolicy via the SettingEngine.

See #3321
I thought setting the gathering policy through the settings engine would be the best place for this as I am not aware of any w3c standard api.

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -157,6 +157,7 @@ func (g *ICEGatherer) buildAgentOptions() ([]ice.AgentOption, error) {
 	options = append(options, g.timeoutOptions()...)
 	options = append(options, g.miscOptions()...)
 	options = append(options, g.renominationOptions()...)
+	options = append(options, g.continualGatheringOptions()...)
 
 	requestedNetworkTypes := g.api.settingEngine.candidates.ICENetworkTypes
 	if len(requestedNetworkTypes) == 0 {
@@ -332,6 +333,15 @@ func (g *ICEGatherer) renominationOptions() []ice.AgentOption {
 	}
 
 	return opts
+}
+
+func (g *ICEGatherer) continualGatheringOptions() []ice.AgentOption {
+	policy := g.api.settingEngine.iceContinualGatheringPolicy
+	if policy == 0 {
+		return nil
+	}
+
+	return []ice.AgentOption{ice.WithContinualGatheringPolicy(policy)}
 }
 
 func legacyNAT1To1AddressRewriteRules(ips []string, candidateType ice.CandidateType) []ice.AddressRewriteRule {
--- a/icegatherer_test.go
+++ b/icegatherer_test.go
@@ -2109,6 +2109,271 @@ func TestICEGatherer_RenominationOptions(t *testing.T) {
 	assert.NotNil(t, se.renomination.generator)
 }
 
+func TestICEGatherer_ContinualGatheringOptions(t *testing.T) {
+	se := SettingEngine{}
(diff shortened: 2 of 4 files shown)


wrangelvid:
@JoeTurki I know we said we would discuss the API in the issue, but this seemed so straightforward that I made a quick PR. I have no issues with completely rewriting this, but I thought it would aid the discussion to have a rough draft.

JoTurk (changes requested):
The problem is that this  changes  fundamental  ICE behavior without defining the observable user contract. We agreed to add this to 4.3 but we need to define behavior.

If we just merge this we'll introduces undefined behavior (UB) around:

1. OnICECandidate(nil)
Today, many apps treat nil as terminal. Continual gathering breaks that assumption unless behavior is redefined. will we send nil after every iteration? will we not send nil at all? we need to define and test this.

2. ICEGatheringState transitions becomes incoherent If the state reaches complete, but candidates can still appear later, the state machine lies. we don't define how the state will be in the peer. and we don't test it either.

so we have two assumptions either:

1. the state must be allowed to transition back to gathering, or

2. complete must be redefined under continual mode.

The PR does assert neither.

we will also need to add vnet tests around this similar to how we did with renomination.

wrangelvid:
That was quick, thank you Joe!
Semantically, I believe it makes to not send `OnICECandidate(nil)` and never change to `complete` since we never stop gathering. Since a user would have to explicitly opt-in via the settings engine, I believe this "breaking" change is fine as long as we document it well (which we don't currently).

Though, lets think about an alternative! We could also treat continual gathering as lots of mini ice-restarts. In that case it would make sense to do a bunch of state transitions and sending nil, but it needs change in pion/ice.

I am tending towards the first approach as it seems semantically more aligned.

Happy to add vnet tests!

wrangelvid:
@JoTurk Added the vnet test. Ones `AddIpToNic` is merged, we can uncomment the test in this PR. Let me know what you think.

JoTurk:
I commented on the transport PR :pray:

wrangelvid:
@JoTurk vnet test enabled and ready for review.

wrangelvid:
@JoTurk Closing this for now. We ended up moving to str0m for various reasons and implemented continual gathering there. Thank you for the initial support!

JoTurk:
Sounds good, good luck.

> and implemented continual gathering there. 

But you didn't actually need to contribute it to str0m since your gather lives on your application logic, right?, you can do the same thing with Pion if you use the ORTC path and `ICEGatherer.OnLocalCandidate` and `ICETransport.AddRemoteCandidate`.
The blocker here is that we wanted to have a defined I/O interactions especially around cycles and gather end and state since the RTCPeerConnection is I/O heavy and we have to follow W3C to some degree, and if we do something custom we have to have to define it because we can't just change the behavior later if people start to use it.

wrangelvid:
> But you didn't actually need to contribute it to str0m since your gather lives on your application logic, right?

You are absolutely right, and this is **not** the reason we switched! We just needed more control over the networking layer,  lower level kernel tasks, and had to cross the cgo boundary so much that rust with the sans-io implementation made more sense. It definitely took much more work to get parity with pion, so I wouldn't recommend our path to others.

PR #3325, opened 2025-12-28
Implement support for remote offers that are not actpass

fippo:
as described in RFC 8842 section 5.3. This is not just for backward compat but allows controlling the DTLS role through the SDP which is useful at times.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -1835,19 +1835,16 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream *srtp.ReadStreamSRTP, ssr
 	rtcpInterceptor := result.rtcpInterceptor
 
 	// try to read simulcast IDs from the packet we already have
-	var mid, rid, rsid string
-	if _, err = handleUnknownRTPPacket(
+	mid, rid, rsid, _, err := handleUnknownRTPPacket(
 		b[:i], uint8(midExtensionID), //nolint:gosec // G115
 		uint8(streamIDExtensionID),       //nolint:gosec // G115
 		uint8(repairStreamIDExtensionID), //nolint:gosec // G115
-		&mid,
-		&rid,
-		&rsid,
-	); err != nil {
+	)
+	if err != nil {
 		return err
 	}
 
-	peekedPackets := []*peekedPacket{}
+	peekedPackets := []*peekedPacket(nil)
 
 	// if the first packet didn't contain simuilcast IDs, then probe more packets
 	var paddingOnly bool
@@ -1868,14 +1865,12 @@ func (pc *PeerConnection) handleIncomingSSRC(rtpStream *srtp.ReadStreamSRTP, ssr
 				attributes: attributes,
 			})
 
-			if paddingOnly, err = handleUnknownRTPPacket(
+			mid, rid, rsid, paddingOnly, err = handleUnknownRTPPacket(
 				b[:i], uint8(midExtensionID), //nolint:gosec // G115
 				uint8(streamIDExtensionID),       //nolint:gosec // G115
 				uint8(repairStreamIDExtensionID), //nolint:gosec // G115
-				&mid,
-				&rid,
-				&rsid,
-			); err != nil {
+			)
+			if err != nil {
 				return err
 			}
 
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -644,10 +644,12 @@ func (r *RTPReceiver) receiveForRtxInternal(
 	track.repairRtcpInterceptor = rtcpInterceptor
 	track.repairStreamChannel = make(chan rtxPacketWithAttributes, 50)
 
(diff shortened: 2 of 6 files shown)


peerconnection.go
```diff
@@ -869,7 +869,14 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
 
 	connectionRole := connectionRoleFromDtlsRole(pc.api.settingEngine.answeringDTLSRole)
 	if connectionRole == sdp.ConnectionRole(0) {
-		connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
+        dtlsRole := dtlsRoleFromRemoteSDP(remoteDesc.parsed)
```
fippo: The naming `dtlsRoleFromRemoteSDP` threw me off. If it was from the remote SDP I would expect it be reverse of the role given in the SDP which it isn't because in ORTC the DTLSTransport's `start` is called with the remote role while the newly added code actually determines the answering *local* role from the remote role given in the SDP.

Probably easiest to rename `dtlsRoleFromRemoteSDP` to `dtlsRoleFromSDP`

peerconnection.go
```diff
@@ -869,7 +869,14 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
...
 	if connectionRole == sdp.ConnectionRole(0) {
-		connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
+        dtlsRole := dtlsRoleFromRemoteSDP(remoteDesc.parsed) 
+        if dtlsRole == DTLSRoleAuto {
+            connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
+        } else if dtlsRole == DTLSRoleClient {
+            connectionRole = connectionRoleFromDtlsRole(DTLSRoleServer)
+        } else if dtlsRole == DTLSRoleServer {
+            connectionRole = connectionRoleFromDtlsRole(DTLSRoleClient)
+        }
```
JoeTurki: nit: we currently "could use tagged switch on enum"  QF1003  lint because it cases conflicts with other linters, but I think it's cleaner to use a switch here
```suggestion
		switch dtlsRole {
		case DTLSRoleAuto:
			connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
		case DTLSRoleClient:
			connectionRole = connectionRoleFromDtlsRole(DTLSRoleClient)
		case DTLSRoleServer:
			connectionRole = connectionRoleFromDtlsRole(DTLSRoleServer)
		default:
		}
```

peerconnection.go
```diff
@@ -869,7 +869,17 @@ func (pc *PeerConnection) CreateAnswer(options *AnswerOptions) (SessionDescripti
...
+		switch dtlsRole {
+		case DTLSRoleAuto:
+			connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
+		case DTLSRoleClient:
+			connectionRole = connectionRoleFromDtlsRole(DTLSRoleServer)
+		case DTLSRoleServer:
+			connectionRole = connectionRoleFromDtlsRole(DTLSRoleClient)
+		case DTLSRoleUnknown:
+			connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
+		}
```
JoTurk, line 880: nit:
```suggestion
		switch dtlsRole {
		case DTLSRoleClient:
			connectionRole = connectionRoleFromDtlsRole(DTLSRoleServer)
		case DTLSRoleServer:
			connectionRole = connectionRoleFromDtlsRole(DTLSRoleClient)
		default:
			connectionRole = connectionRoleFromDtlsRole(defaultDtlsRoleAnswer)
		}
```

JoeTurki (commented):
tip: peerconnection_test.go tests both peerconnection.go and peerconnection_js.go. there is peerconnection_go_test.go (Go only) and peerconnection_js_test.go (JS wasm only).

fippo:
> there is peerconnection_go_test.go

That did the trick (even though the browser should support the same and it seemed to fail on asserting Close does not return an error)!

JoTurk:
> That did the trick (even though the browser should support the same and it seemed to fail on asserting Close does not return an error)!

This test doesn't actually test the browser (We'll add enhanced browsers tests soon).
It tests peerconnection_js.go with some mocks. to make it work for wasm we'll have to update createAnswer and createOffer in peerconnection_js.go, Prob not worth it, and Out of the scope for this PR.

JoTurk:
@fippo If you're going to tag this, please pump the patch version, and not the minor version, we have a different system for pion/webrtc releases \:)  so v4.2.2.

fippo:
(can ride-along with the next release)

JoTurk:
(we do many patch and pre-releases between each minor release, it will prob get released soon with a hot fix or something before 4.3).

PR #3326, opened 2025-12-28
WARP demo

fippo:
for testing WARP / SPED / SNAP - will be referenced by future PRs, based on #3325


fippo:
@juberti when checking the "Act as DTLS client" box before connecting one can see the hello verify request from the pion server:
image
with a cookie which is put into the subsequent client hello from the browser.

Sean-Der (approved):
LGTM! Have fun on this adventure :)

PR #3327, opened 2025-12-28
sctp: implement SNAP / sctp-init

fippo:
implements [draft-hancke-tsvwg-snap.html](https://datatracker.ietf.org/doc/draft-hancke-tsvwg-snap/), i.e. moving the sctp handshake to out-of-band negotiation as `a=sctp-init` in the SDP which saves two network round-trips.

(based on #3326, sctp changes in https://github.com/pion/sctp/pull/449)

### DIFF
--- a/settingengine.go
+++ b/settingengine.go
@@ -124,6 +124,7 @@ type renominationSettings struct {
 	generator         ice.NominationValueGenerator
 	automatic         bool
 	automaticInterval *time.Duration
+	attributeType     *uint16
 }
 
 // NominationValueGenerator generates nomination values for ICE renomination.
@@ -152,9 +153,18 @@ func WithRenominationInterval(interval time.Duration) RenominationOption {
 	}
 }
 
+// WithRenominationNominationAttribute overrides the STUN attribute type used for ICE renomination.
+// If unset, the underlying ICE agent default is used.
+func WithRenominationNominationAttribute(attrType uint16) RenominationOption {
+	return func(cfg *renominationSettings) {
+		a := attrType
+		cfg.attributeType = &a
+	}
+}
+
 var errInvalidRenominationInterval = errors.New("renomination interval must be greater than zero")
 
-// SetICERenomination configures ICE renomination using options for generator and scheduling.
+// SetICERenomination configures ICE renomination using options for generator, scheduling, and attribute type.
 // Manual control is not exposed yet. This always enables automatic renomination with the default
 // generator unless a custom one is provided.
 func (e *SettingEngine) SetICERenomination(options ...RenominationOption) error {
@@ -177,6 +187,7 @@ func (e *SettingEngine) SetICERenomination(options ...RenominationOption) error 
 	e.renomination.generator = cfg.generator
 	e.renomination.automatic = true
 	e.renomination.automaticInterval = cfg.automaticInterval
+	e.renomination.attributeType = cfg.attributeType
 
 	return nil
 }
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -363,6 +363,9 @@ func (g *ICEGatherer) renominationOptions() []ice.AgentOption {
 			return generator()
 		}),
 	}
+	if renom.attributeType != nil {
+		opts = append(opts, ice.WithNominationAttribute(*renom.attributeType))
+	}
 
 	if renom.automatic {
 		interval := time.Duration(0)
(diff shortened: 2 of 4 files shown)


sctpcapabilities.go
```diff
@@ -6,4 +6,5 @@ package webrtc
 // SCTPCapabilities indicates the capabilities of the SCTPTransport.
 type SCTPCapabilities struct {
 	MaxMessageSize uint32 `json:"maxMessageSize"`
+	SctpInit       []byte `json:"sctpInit"`
```
fippo: this wil have to be a string to be comparable? It would be possible to define this as the base64 encoded variant and do the base64 decoding later.

fippo: string it is, using base64 was even more meh

JoTurk: Is it fixed size in snap? if yes we can keep it comparable

JoTurk: but string is fine yeah.

fippo: size depends on the options used so is variable

fippo:
Close enough, what do I need to do to disable the WASM test?

JoTurk:
@fippo peerconnection_test.go is shared, peerconnection_js_test.go is wasm only, and peerconnection_go_test.go is "go native" only.

fippo:
(updated accordingly)

settingengine.go
```diff
@@ -598,6 +599,11 @@ func (e *SettingEngine) EnableSCTPZeroChecksum(isEnabled bool) {
 	e.sctp.enableZeroChecksum = isEnabled
 }
 
+// EnableSctpSnap enables the use of the SCTP SNAP connect optimization.
+func (e *SettingEngine) EnableSctpSnap(isEnabled bool) {
+	e.sctp.enableSnap = isEnabled
+}
+
```
JoTurk, line 617: Is it safe to enable this by default insstead?

fippo: Not until it is further along in the IETF process imo.

peerconnection.go
```diff
@@ -2823,11 +2824,12 @@ func (pc *PeerConnection) startRTP(
 
 	pc.startRTPReceivers(remoteDesc, currentTransceivers)
 	if d := haveDataChannel(remoteDesc); d != nil && d.MediaName.Port.Value != 0 {
-		pc.startSCTP(getMaxMessageSize(d))
+		remoteSctpInit, _ := getSctpInit(d)
```
JoTurk, line 2827: Maybe we should at least log this error?

sctptransport.go
```diff
@@ -497,3 +509,16 @@ func (r *SCTPTransport) BufferedAmount() int {
 
 	return r.sctpAssociation.BufferedAmount()
 }
+
+// GetSctpInit returns the current sctp-init attribute and caches the last created.
+// The caller should hold the lock.
+func (r *SCTPTransport) GetSctpInit() []byte {
+	if len(r.localSctpInit) == 0 {
+		r.localSctpInit, _ = sctp.GenerateOutOfBandToken(sctp.Config{
```
JoTurk: same as the other error.

sctpcapabilities.go
```diff
@@ -6,4 +6,6 @@ package webrtc
 // SCTPCapabilities indicates the capabilities of the SCTPTransport.
 type SCTPCapabilities struct {
 	MaxMessageSize uint32 `json:"maxMessageSize"`
+	// Note: this is the binary sctp-init, not the base64 encoded version.
+	SctpInit string `json:"sctpInit"`
 }
```
JoTurk, line 11: I keep thinking about this, maybe it's actually better to use base64 encoded string here, since the field is json-tagged?

peerconnection.go
fippo: error handling and logging for this needs to be done in SetRemoteDescription, this is in createAnswer so I am not sure the error (base64 decode) can happen?

sctpcapabilities.go
fippo: Same but the responsibility is that the sdp parser validates that the field is base64 encoded

sctptransport.go
fippo: Logging a warning if this fails (and does not create/negotiate) seems appropriate

peerconnection.go
fippo: I think it can happen but should be silently ignored because the offer doesn't have a *valid* sctp-init:
https://www.ietf.org/archive/id/draft-hancke-tsvwg-snap-00.html#name-generating-the-sdp-answer

JoTurk (approved):
thank you :)

sctpcapabilities.go
fippo: (made private which gives us time to decide)

fippo:
Thank you for the feedback!

Issue #3149, opened 2025-06-09
Issue with switching among tracks of same kind, codec & clock rate with ReplaceTrack of RTPSender


selvan:
### Your environment.
  - Version: github.com/pion/webrtc/v4 v4.1.1

### What did you do?

Creating a local track from a TrackRemote (remoteX),
```
localX, _ := webrtc.NewTrackLocalStaticRTP(remoteX.Codec().RTPCodecCapability, remoteX.ID(), remoteX.StreamID(), webrtc.WithRTPStreamID(remoteX.RID()))
```

Creating a another local track from a another TrackRemote (remoteY),
```
localY, _ := webrtc.NewTrackLocalStaticRTP(remoteY.Codec().RTPCodecCapability, remoteY.ID(), remoteY.StreamID(), webrtc.WithRTPStreamID(remoteY.RID()))
```

remoteX and remoteY are of same kind (Audio), codec type (VP8) and clock rate (90000)

Trying to replace localX with localY on the RTPSender with ReplaceTrack method.

### What did you expect?
The remote peer attached to the RTPSender should receive switched video tracks when localX is replaced with localY & vice versa.

### What happened?
The remote peer's video just freeze, when trying to replace localX with localY on the RTPSender.

arjunshajitech:
> The remote peer attached to the RTPSender should receive switched video tracks when localX is replaced with localY & vice versa.

May be you need to request a key frame via PLI.

```
_ = pc.WriteRTCP([]rtcp.Packet{
	&rtcp.PictureLossIndication{
		MediaSSRC: uint32(receiver.Track().SSRC()),
	},
})
```

selvan:
@arjunshajitech Have created sample code to demonstracte the issue,

Running the same code:
https://github.com/selvan/webrtc/blob/issue-3149/examples/swap-tracks-via-replace/README.md

PLI before ReplaceTrack: https://github.com/selvan/webrtc/blob/d9734013bb06fa7476c8e1ad5b87858a42ce9b76/examples/swap-tracks-via-replace/main.go#L180

arjunshajitech:
@selvan 

1) The code doesn't read RTP packets from the remote track and write them to the local track created in the `ontrack` event. This needs to be updated first.

Update the code in ontrack after the `trackCount++`

```
go func() {
			rtpBuf := make([]byte, 1400)
			for {
				n, _, err := track.Read(rtpBuf)
				if err != nil {
					if err == io.EOF {
[...]

PR #3337, opened 2026-01-01
Add WithRTPSequenceNumber

JoTurk:
#### Description
We have WithRTPTimestamp but we don't have a way to set the initial sequence number useful for tests and for manually replacing tracks.

Doesn't solve https://github.com/pion/webrtc/issues/3149 but makes it easier to work around it.

### DIFF
--- a/track_local_static.go
+++ b/track_local_static.go
@@ -33,7 +33,8 @@ type TrackLocalStaticRTP struct {
 	codec             RTPCodecCapability
 	payloader         func(RTPCodecCapability) (rtp.Payloader, error)
 	id, rid, streamID string
-	rtpTimestamp      *uint32
+	initalTimestamp   *uint32
+	initialSeqNumber  *uint16
 }
 
 // NewTrackLocalStaticRTP returns a TrackLocalStaticRTP.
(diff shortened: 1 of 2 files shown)


track_local_static.go
```diff
@@ -296,12 +304,18 @@ func (s *TrackLocalStaticSample) Bind(t TrackLocalContext) (RTPCodecParameters,
...
+		options = append(options, rtp.WithTimestamp(*s.rtpTrack.initalTimestamp))
+	}
+
+	if s.rtpTrack.initialSeqNumber != nil {
+		s.sequencer = rtp.NewFixedSequencer(*s.rtpTrack.initialSeqNumber)
+	}
+
+	if s.sequencer == nil {
+		s.sequencer = rtp.NewRandomSequencer()
 	}
```
philipch07, line 319: Tiny nit but would it be more clear to have this be an if + else-if because `rtp.NewFixedSequencer()` guarantees that `s.sequencer` is non-nil

JoTurk: I wrote it this way because i felt that they aren't coupled, and would be anti idiomatic go. the first block tries to configure the sequencer from initialSeqNumber, the second if is a default / fall back.

After this PR I'm planning to improve replacing track so that:

1. If the track is being replaced with another one it inherit the original 
2. callers can still override the sequencer and/or set a custom initial sequence number.
3. if no sequencer is provided, we reliably fall back to creating one (random).


So under these conditions, i think the code will be easier to read and maintain, what do you think? WIll change it if you think otherwise, just sharing my thoughts :)

philipch07: That makes sense, I'm alright with it then.

PR #3339, opened 2026-01-06
SetConfiguration now updates ICEGatherer's servers

wrangelvid:
#### Description
After using `SetConfiguration` I found that the new ice servers have not been used. Turns out after creating a peer connection the ice gatherer is created with the current configuration, but it is never updated with new servers when the configuration changes

This pr should fix the issue and implement the spec properly. We just need to wait on https://github.com/pion/ice/pull/866 to land.

### DIFF
--- a/icegatherer.go
+++ b/icegatherer.go
@@ -114,6 +114,40 @@ func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
 	}, nil
 }
 
+// updateServers updates the ICE servers and gather policy.
+// If called before gathering starts, the new servers will be used for initial gathering.
+// If called after gathering has started, the new servers will be used on the next ICE restart.
+func (g *ICEGatherer) updateServers(servers []ICEServer, policy ICETransportPolicy) error {
+	g.lock.Lock()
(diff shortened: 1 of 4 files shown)


wrangelvid:
@JoTurk thanks for getting the pion/ice PR in so quickly. This is ready for review.

icegatherer.go
```diff
@@ -114,6 +114,40 @@ func (api *API) NewICEGatherer(opts ICEGatherOptions) (*ICEGatherer, error) {
 	}, nil
 }
 
+// UpdateServers updates the ICE servers and gather policy.
+// If called before gathering starts, the new servers will be used for initial gathering.
+// If called after gathering has started, the new servers will be used on the next ICE restart.
+func (g *ICEGatherer) UpdateServers(servers []ICEServer, policy ICETransportPolicy) error {
```
JoTurk: I think this should be private.

JoTurk (commented):
Changes look good to me overall, and makes us closer to the browser behavior. thank you.

icegatherer.go
wrangelvid: Done.

PR #3340, opened 2026-01-07
Fix typo in custom-logger example

boushley:
#### Description
Fix a typo in the custom-logger example application where `Infof` was calling `Trace` instead of `Info` breaking the pattern.


Sean-Der:
@boushley your back! thanks for the fix :)

PR #3343, opened 2026-01-08
Extract a `DTLSConn` interface to wrap `dtls.Conn`

juberti-oai:
#### Description
This PR allows for runtime replacement of the `dtls.Conn` implementation from pion/dtls with a test or injected implementation. It defines a new `DTLSConn` interface, along with a factory function for creating a `DTLSConn` which can be specified in `SettingEngine`.

By default, a `DTLSConn` based on `dtls.Conn` is created, and for the most part, the `DTLSConn` methods are straight thunks to `dtls.Conn`. The one exception is the key extractor method, which calls `ConnectionState()` and then returns a `KeyingMaterialExporter`, which is the only data that pion/webrtc needs from `ConnectionState()`.

lint passes, tests pass

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -68,7 +68,6 @@ type DTLSConn interface {
 	net.Conn
 	ConnectionState() (dtls.State, bool)
 	SelectedSRTPProtectionProfile() (dtls.SRTPProtectionProfile, bool)
-	RemoteSRTPMasterKeyIdentifier() ([]byte, bool)
 	HandshakeContext(ctx context.Context) error
 	Handshake() error
 }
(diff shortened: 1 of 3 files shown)


Sean-Der:
LGTM! What do you think about not exporting this? Do you want this for tests/prototyping WARP in this repo or outside also?

juberti-oai:
I need it to be exported so I can use it from TC, but happy to mark it as experimental if you're worried about committing to the specific interface shape

Sean-Der:
I can’t make it experimental unfortunately. If we API break (even if no one uses code) some automation downstream forces me to bump or revert. 

I’m just trying to save myself heart ache! If you need this API public works for me. I thought this was just going to be for tests/internal use.

Would a go.mod replace work for your use case? If you want an alternative dtls implementation this is what I did for project that had to use OS crypto

juberti-oai:
Hmm. The dtls.Conn class has a wider interface, so that's not ideal. I could refactor that class slightly, but would that be a breaking change on its own? Or maybe I could just add a method and that would be OK (as I'd just replace the methods pion/webrtc needs?)

juberti-oai:
FWIW, I think this sort of indirection is nice in general, it gives the developer a lot of flexibility, so I still think the interface is the best path forward.

It's tricky though if you have to get the interface right on the first try...

Sean-Der:
You just need to implement the methods pion/webrtc needs. You can have your own pion/dtls stub that uses openssl and just a relative path in mod replace.

If it’s not possible totally in support of merging. It’s just this has to stay forever, and no one is going to use it after DTLS 1.3 is done.

juberti-oai:
OK. Let me do a tiny PR against dtls that adds a new method for key extraction, and then I can try the go.mod swap.

PR #3355, opened 2026-01-09
Use LoggerFactory when interceptors are registered

sirzooro:
Updated interceptor configuration functions to support options. Added options to configure logger factory and interceptor-specific options. Updated NewAPI to pass LoggerFactory when default interceptors are registered.

### DIFF
--- a/interceptor_option.go
+++ b/interceptor_option.go
@@ -1,0 +1,79 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+//go:build !js
+// +build !js
+
+package webrtc
+
+import (
+	"github.com/pion/interceptor/pkg/nack"
+	"github.com/pion/interceptor/pkg/report"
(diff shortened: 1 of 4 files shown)


interceptor_option.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"github.com/pion/interceptor/pkg/nack"
+	"github.com/pion/interceptor/pkg/report"
+	"github.com/pion/interceptor/pkg/stats"
+	"github.com/pion/interceptor/pkg/twcc"
+	"github.com/pion/logging"
+)
+
+// InterceptorOptions contains options for configuring interceptors.
+type InterceptorOptions struct {
```
JoTurk: Does this need to be public?

sirzooro: No, changed

sirzooro:
> Wish we could do this without introducing this many public APIs in webrtc, but ig we have to.

I do not like it too, unfortunately there was no other easy way to inject LoggerFactory. This is another thing to cleanup in the future when webrtc v5 will be released.

PR #3357, opened 2026-01-12
Added support for Cryptex (RFC 9335)

sirzooro:
This is final part of Cryptex ([RFC 9335](https://www.rfc-editor.org/rfc/rfc9335)) implementation for pion/webrtc. Public API is based on [WebRTC Extensions specification](https://w3c.github.io/webrtc-extensions/#rtp-header-extension-encryption) (W3C Editor's Draft from 19 December 2025).

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -42,6 +42,7 @@ type DTLSTransport struct {
 	remoteCertificate     []byte
 	state                 DTLSTransportState
 	srtpProtectionProfile srtp.ProtectionProfile
+	cryptexMode           srtp.CryptexMode
 
 	onStateChangeHandler   func(DTLSTransportState)
 	internalOnCloseHandler func()
@@ -191,12 +192,20 @@ func (t *DTLSTransport) GetRemoteCertificate() []byte {
 	return t.remoteCertificate
 }
 
-func (t *DTLSTransport) startSRTP() error {
+// startSRTP requires the caller holds the lock.
+func (t *DTLSTransport) startSRTP() error { //nolint:cyclop
 	srtpConfig := &srtp.Config{
 		Profile:       t.srtpProtectionProfile,
 		BufferFactory: t.api.settingEngine.BufferFactory,
 		LoggerFactory: t.api.settingEngine.LoggerFactory,
 	}
+
+	if t.cryptexMode == srtp.CryptexModeEnabled || t.cryptexMode == srtp.CryptexModeRequired {
+		opt := srtp.Cryptex(t.cryptexMode)
+		srtpConfig.LocalOptions = append(srtpConfig.LocalOptions, opt)
+		srtpConfig.RemoteOptions = append(srtpConfig.RemoteOptions, opt)
+	}
+
 	if t.api.settingEngine.replayProtection.SRTP != nil {
 		srtpConfig.RemoteOptions = append(
 			srtpConfig.RemoteOptions,
@@ -714,4 +723,27 @@ func (t *DTLSTransport) streamsForSSRC(
 		rtcpReadStream:  rtcpReadStream,
 		rtcpInterceptor: rtcpInterceptor,
 	}, nil
+}
+
+func (t *DTLSTransport) getCryptexMode() srtp.CryptexMode {
+	t.lock.RLock()
+	defer t.lock.RUnlock()
+
+	return t.cryptexMode
+}
+
+func (t *DTLSTransport) setCryptexMode(mode srtp.CryptexMode) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	t.cryptexMode = mode
+}
+
+// rtpHeaderEncryptionNegotiated reports if RFC 9335 RTP Header Extension Encryption ("Cryptex")
+// has been negotiated and is enabled for this transceiver.
+func (t *DTLSTransport) rtpHeaderEncryptionNegotiated() bool {
+	t.lock.RLock()
+	defer t.lock.RUnlock()
+
+	return t.cryptexMode == srtp.CryptexModeEnabled || t.cryptexMode == srtp.CryptexModeRequired
 }
--- a/errors.go
+++ b/errors.go
@@ -53,6 +53,10 @@ var (
 	// RTCPMuxPolicy was made after PeerConnection has been initialized.
 	ErrModifyingRTCPMuxPolicy = errors.New("rtcp mux policy cannot be modified")
 
+	// errModifyingRTPHeaderEncryptionPolicy indicates that an attempt to modify
+	// RTPHeaderEncryptionPolicy was made after PeerConnection has been initialized.
+	errModifyingRTPHeaderEncryptionPolicy = errors.New("rtp header encryption policy cannot be modified")
+
 	// ErrModifyingICECandidatePoolSize indicates that an attempt to modify
 	// ICECandidatePoolSize was made after PeerConnection has been initialized.
 	ErrModifyingICECandidatePoolSize = errors.New("ice candidate pool size cannot be modified")
@@ -134,6 +138,21 @@ var (
 
 	// ErrNoSRTPProtectionProfile indicates that the DTLS handshake completed and no SRTP Protection Profile was chosen.
 	ErrNoSRTPProtectionProfile = errors.New("DTLS Handshake completed and no SRTP Protection Profile was chosen")
+
+	// ErrRTPHeaderEncryptionRequired indicates that the local endpoint requires
+	// RFC 9335 Cryptex negotiation, but the remote SDP did not negotiate it.
+	ErrRTPHeaderEncryptionRequired = errors.New("rtp header extension encryption required")
+
+	// ErrRTPHeaderEncryptionModeChangeNotSupported indicates that the remote SDP attempted
+	// to change the negotiated RTP header extension encryption mode after SRTP has started.
+	// Pion's SRTP sessions cannot currently switch Cryptex mode without a full restart.
+	ErrRTPHeaderEncryptionModeChangeNotSupported = errors.New(
+		"rtp header extension encryption mode change not supported")
+
+	// ErrRTPHeaderEncryptionMixedModeNotSupported indicates that the remote SDP negotiated
+	// RTP header extension encryption for only some RTP media sections. Pion currently
+	// configures Cryptex as a connection-wide SRTP behavior and cannot support mixed mode.
+	ErrRTPHeaderEncryptionMixedModeNotSupported = errors.New("rtp header extension encryption mixed mode not supported")
 
 	// ErrFailedToGenerateCertificateFingerprint indicates that we failed to generate the fingerprint
 	// used for comparing certificates.
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -120,12 +120,13 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
 	pc := &PeerConnection{
 		id: fmt.Sprintf("PeerConnection-%d", time.Now().UnixNano()),
 		configuration: Configuration{
-			ICEServers:           []ICEServer{},
-			ICETransportPolicy:   ICETransportPolicyAll,
-			BundlePolicy:         BundlePolicyBalanced,
-			RTCPMuxPolicy:        RTCPMuxPolicyRequire,
-			Certificates:         []Certificate{},
-			ICECandidatePoolSize: 0,
+			ICEServers:                []ICEServer{},
+			ICETransportPolicy:        ICETransportPolicyAll,
+			BundlePolicy:              BundlePolicyBalanced,
+			RTCPMuxPolicy:             RTCPMuxPolicyRequire,
+			Certificates:              []Certificate{},
+			ICECandidatePoolSize:      0,
+			RTPHeaderEncryptionPolicy: RTPHeaderEncryptionPolicyNegotiate,
 		},
 		isClosed:                                &atomic.Bool{},
 		isCloseDone:                             make(chan struct{}),
@@ -271,6 +272,10 @@ func (pc *PeerConnection) initConfiguration(configuration Configuration) error {
 			}
 		}
 		pc.configuration.ICEServers = sanitizedICEServers
+	}
+
+	if configuration.RTPHeaderEncryptionPolicy != RTPHeaderEncryptionPolicyUnknown {
+		pc.configuration.RTPHeaderEncryptionPolicy = configuration.RTPHeaderEncryptionPolicy
 	}
 
 	return nil
@@ -642,6 +647,13 @@ func (pc *PeerConnection) SetConfiguration(configuration Configuration) error { 
 	}
 
 	pc.configuration.ICEServers = configuration.ICEServers
+
+	if configuration.RTPHeaderEncryptionPolicy != RTPHeaderEncryptionPolicyUnknown {
+		if configuration.RTPHeaderEncryptionPolicy != pc.configuration.RTPHeaderEncryptionPolicy {
+			return &rtcerr.InvalidModificationError{Err: errModifyingRTPHeaderEncryptionPolicy}
+		}
+		pc.configuration.RTPHeaderEncryptionPolicy = configuration.RTPHeaderEncryptionPolicy
+	}
 
 	return nil
 }
@@ -1208,7 +1220,22 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 	default:
 		pc.canTrickleICECandidates = ICETrickleCapabilityUnknown
 	}
+
+	cryptexMode, cryptexErr := pc.getCryptexModeFromRemoteSDP(desc.parsed)
 	pc.mu.Unlock()
+	if cryptexErr != nil {
+		return cryptexErr
+	}
+
+	// If SRTP is already established, we can't safely switch Cryptex mode without
+	// recreating SRTP sessions (which is not supported by pion/srtp today).
+	if isRenegotiation {
+		if srtpSession, err := pc.dtlsTransport.getSRTPSession(); err == nil && srtpSession != nil {
+			if cryptexMode != pc.dtlsTransport.getCryptexMode() {
+				return &rtcerr.InvalidModificationError{Err: ErrRTPHeaderEncryptionModeChangeNotSupported}
+			}
+		}
+	}
 
 	// Disable RTX/FEC on RTPSenders if the remote didn't support it
 	for _, sender := range pc.GetSenders() {
@@ -1376,6 +1403,7 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 			iceDetails.Password,
 			fingerprint,
 			fingerprintHash,
+			cryptexMode,
 		)
 		if weOffer {
 			pc.startRTP(false, &desc, currentTransceivers)
@@ -1383,6 +1411,100 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 	})
 
 	return nil
+}
+
+func cryptexNegotiatedInSDP(desc *sdp.SessionDescription) (negotiatedForAnyMedia, negotiatedForAllMedia bool) {
+	negotiatedForAnyMedia = false
+	negotiatedForAllMedia = true
+	haveAnyMedia := false
+	haveRTPMedia := false
+
+	if _, hasCryptex := desc.Attribute(sdp.AttrKeyCryptex); hasCryptex {
+		return true, true
+	}
+
+	for _, media := range desc.MediaDescriptions {
+		if media.MediaName.Port.Value == 0 {
+			continue
+		}
+
+		haveAnyMedia = true
+		if media.MediaName.Media == mediaSectionApplication {
+			continue
+		}
+		haveRTPMedia = true
+
+		hasCryptex := isCryptexSet(media)
+		negotiatedForAnyMedia = negotiatedForAnyMedia || hasCryptex
+		negotiatedForAllMedia = negotiatedForAllMedia && hasCryptex
+	}
+
+	if !haveAnyMedia {
+		return false, false
+	}
+	if !haveRTPMedia {
+		// If the SDP contains only non-RTP media sections (for example, m=application
+		// for data channels), Cryptex is not applicable. Treat "negotiated for all RTP
+		// media" as true (vacuously) so RTPHeaderEncryptionPolicyRequire doesn't reject
+		// data-only SDP.
+		return false, true
+	}
+
+	return negotiatedForAnyMedia, negotiatedForAllMedia
+}
+
+// getCryptexModeFromRemoteSDP requires the caller holds pc.mu.
+//
+// nolint:cyclop
+func (pc *PeerConnection) getCryptexModeFromRemoteSDP(parsed *sdp.SessionDescription) (srtp.CryptexMode, error) {
+	negotiatedForAnyMedia, negotiatedForAllMedia := cryptexNegotiatedInSDP(parsed)
+	var mode srtp.CryptexMode
+
+	// When the SDP contains no active RTP media (for example, an initial datachannel-only
+	// negotiation), we still need to select a stable Cryptex mode for the DTLS transport.
+	// Otherwise, when RTP m-lines are added later via renegotiation, we would attempt to
+	// change Cryptex mode after SRTP has already been created.
+	noActiveRTPMedia := !negotiatedForAnyMedia && negotiatedForAllMedia
+	if noActiveRTPMedia {
+		switch pc.configuration.RTPHeaderEncryptionPolicy {
+		case RTPHeaderEncryptionPolicyDisable:
+			return srtp.CryptexModeDisabled, nil
+		case RTPHeaderEncryptionPolicyRequire:
+			return srtp.CryptexModeRequired, nil
+		case RTPHeaderEncryptionPolicyUnknown, RTPHeaderEncryptionPolicyNegotiate:
+			return srtp.CryptexModeEnabled, nil
+		}
+	}
+
+	// We can't support mixed Cryptex negotiation across RTP m-lines.
+	// Accept only: (a) session-level Cryptex, or (b) media-level Cryptex on all RTP m-lines.
+	if pc.configuration.RTPHeaderEncryptionPolicy != RTPHeaderEncryptionPolicyDisable &&
+		negotiatedForAnyMedia && !negotiatedForAllMedia {
+		return srtp.CryptexModeDisabled, &rtcerr.InvalidAccessError{Err: ErrRTPHeaderEncryptionMixedModeNotSupported}
+	}
+
+	switch pc.configuration.RTPHeaderEncryptionPolicy {
+	case RTPHeaderEncryptionPolicyDisable:
+		mode = srtp.CryptexModeDisabled
+	case RTPHeaderEncryptionPolicyUnknown, RTPHeaderEncryptionPolicyNegotiate:
+		if negotiatedForAnyMedia {
+			mode = srtp.CryptexModeEnabled
+		} else {
+			mode = srtp.CryptexModeDisabled
+		}
+	case RTPHeaderEncryptionPolicyRequire:
+		switch {
+		case !negotiatedForAllMedia:
+			return srtp.CryptexModeDisabled, &rtcerr.InvalidAccessError{Err: ErrRTPHeaderEncryptionRequired}
+		case negotiatedForAnyMedia:
+			mode = srtp.CryptexModeRequired
+		default:
+			// This is case for data-only SDP with no session-level Cryptex attribute.
+			mode = srtp.CryptexModeDisabled
+		}
+	}
+
+	return mode, nil
 }
 
 func (pc *PeerConnection) configureReceiver(incoming trackDetails, receiver *RTPReceiver) {
@@ -2787,6 +2909,7 @@ func (pc *PeerConnection) startTransports(
 	iceRole ICERole,
 	dtlsRole DTLSRole,
 	remoteUfrag, remotePwd, fingerprint, fingerprintHash string,
+	cryptexMode srtp.CryptexMode,
 ) {
 	// Start the ice transport
 	err := pc.iceTransport.Start(
@@ -2816,6 +2939,8 @@ func (pc *PeerConnection) startTransports(
 			}
 		}()
 	}
+
+	pc.dtlsTransport.setCryptexMode(cryptexMode)
 
 	// Start the dtls transport
 	err = pc.dtlsTransport.Start(DTLSParameters{
@@ -2872,8 +2997,9 @@ func (pc *PeerConnection) generateUnmatchedSDP(
 	}
 
 	isPlanB := pc.configuration.SDPSemantics == SDPSemanticsPlanB
-	mediaSections := []mediaSection{}
+	cryptexEnabledByPolicy := pc.configuration.RTPHeaderEncryptionPolicy != RTPHeaderEncryptionPolicyDisable
 
+	mediaSections := []mediaSection{}
 	// Needed for pc.sctpTransport.dataChannelsRequested
 	pc.sctpTransport.lock.Lock()
 
@@ -2888,10 +3014,13 @@ func (pc *PeerConnection) generateUnmatchedSDP(
 		audio := make([]*RTPTransceiver, 0)
 
 		for _, t := range transceivers {
-			if t.kind == RTPCodecTypeVideo {
+			switch t.kind {
+			case RTPCodecTypeVideo:
 				video = append(video, t)
-			} else if t.kind == RTPCodecTypeAudio {
+			case RTPCodecTypeAudio:
 				audio = append(audio, t)
+			case RTPCodecTypeUnknown:
+				// nothing to do
 			}
 			if sender := t.Sender(); sender != nil {
 				sender.setNegotiated()
@@ -2899,21 +3028,36 @@ func (pc *PeerConnection) generateUnmatchedSDP(
 		}
 
 		if len(video) > 0 {
-			mediaSections = append(mediaSections, mediaSection{id: "video", transceivers: video})
+			mediaSections = append(mediaSections, mediaSection{
+				id:           "video",
+				transceivers: video,
+				cryptex:      false,
+			})
 		}
 		if len(audio) > 0 {
-			mediaSections = append(mediaSections, mediaSection{id: "audio", transceivers: audio})
+			mediaSections = append(mediaSections, mediaSection{
+				id:           "audio",
+				transceivers: audio,
+				cryptex:      false,
+			})
 		}
 
 		if pc.configuration.AlwaysNegotiateDataChannels || pc.sctpTransport.dataChannelsRequested != 0 {
-			mediaSections = append(mediaSections, mediaSection{id: "data", data: true})
+			mediaSections = append(mediaSections, mediaSection{
+				id:   "data",
+				data: true,
+			})
 		}
 	} else {
 		for _, t := range transceivers {
 			if sender := t.Sender(); sender != nil {
 				sender.setNegotiated()
 			}
-			mediaSections = append(mediaSections, mediaSection{id: t.Mid(), transceivers: []*RTPTransceiver{t}})
+			mediaSections = append(mediaSections, mediaSection{
+				id:           t.Mid(),
+				transceivers: []*RTPTransceiver{t},
+				cryptex:      false,
+			})
 		}
 
 		if pc.configuration.AlwaysNegotiateDataChannels || pc.sctpTransport.dataChannelsRequested != 0 {
@@ -2946,6 +3090,7 @@ func (pc *PeerConnection) generateUnmatchedSDP(
 		nil,
 		pc.api.settingEngine.getSCTPMaxMessageSize(),
 		false,
+		cryptexEnabledByPolicy,
 	)
 }
 
@@ -2983,6 +3128,22 @@ func (pc *PeerConnection) generateMatchedSDP(
 	isExtmapAllowMixed := isExtMapAllowMixedSet(remoteDescription.parsed)
 	localTransceivers := append([]*RTPTransceiver{}, transceivers...)
 
+	cryptexEnabledByPolicy := pc.configuration.RTPHeaderEncryptionPolicy != RTPHeaderEncryptionPolicyDisable
+	cryptexAtSessionLevel := false
+	cryptexAtMediaLevel := false
+	if cryptexEnabledByPolicy {
+		if includeUnmatched {
+			// Offers are generated with session-level Cryptex to avoid mixed per-media negotiation.
+			cryptexAtSessionLevel = true
+		} else {
+			// Answers must mirror the attribute level (session vs media) used by the remote offer.
+			negotiatedForAnyMedia, _ := cryptexNegotiatedInSDP(remoteDescription.parsed)
+			_, offerHasSessionCryptex := remoteDescription.parsed.Attribute(sdp.AttrKeyCryptex)
+			cryptexAtSessionLevel = negotiatedForAnyMedia && offerHasSessionCryptex
+			cryptexAtMediaLevel = negotiatedForAnyMedia && !offerHasSessionCryptex
+		}
+	}
+
 	detectedPlanB := descriptionIsPlanB(remoteDescription, pc.log)
 	if pc.configuration.SDPSemantics != SDPSemanticsUnifiedPlan {
 		detectedPlanB = descriptionPossiblyPlanB(remoteDescription)
@@ -3005,7 +3166,11 @@ func (pc *PeerConnection) generateMatchedSDP(
 				pc.sctpTransport.lock.Unlock()
 			}
 
-			mediaSections = append(mediaSections, mediaSection{id: midValue, data: true, sctpInit: localSctpInit})
+			mediaSections = append(mediaSections, mediaSection{
+				id:       midValue,
+				data:     true,
+				sctpInit: localSctpInit,
+			})
 			alreadyHaveApplicationMediaSection = true
 
 			continue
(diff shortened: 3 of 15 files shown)


dtlstransport.go
```diff
@@ -43,6 +43,7 @@ type DTLSTransport struct {
 	remoteCertificate     []byte
 	state                 DTLSTransportState
 	srtpProtectionProfile srtp.ProtectionProfile
+	cryptexMode           CryptexMode
```
boks1971: is this field used?

peerconnection.go
```diff
@@ -118,15 +118,21 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
 	// Some variables defined explicitly despite their implicit zero values to
 	// allow better readability to understand what is happening.
 
+	defaultRTPHeaderEncryptionPolicy := RTPHeaderEncryptionPolicyNegotiate
```
boks1971: :nit: rename to `rtpHeaderEncryptionPolicy`? `default` indicates it is the default value, but it is changed in code here.

peerconnection.go
```diff
@@ -118,15 +118,21 @@ func (api *API) NewPeerConnection(configuration Configuration) (*PeerConnection,
...
+	if api.settingEngine.disableRTPHeaderEncryption {
+		defaultRTPHeaderEncryptionPolicy = RTPHeaderEncryptionPolicyUnknown
```
boks1971: :nit: maybe call this enum `Disabled` rather than `Unknown`?

rtptransceiver.go
```diff
@@ -253,6 +253,28 @@ func (t *RTPTransceiver) Direction() RTPTransceiverDirection {
...
+	if dtlsTransport == nil {
+		if receiver := t.Receiver(); receiver != nil {
+			dtlsTransport = receiver.Transport()
+		}
+	}
+	if dtlsTransport == nil {
+		return false
+	}
+
+	dtlsTransport.lock.RLock()
```
boks1971: would be good to add a method to `dtlsTransport` instead of holding lock of some other module from outside.

peerconnection.go
```diff
@@ -596,6 +609,14 @@ func (pc *PeerConnection) SetConfiguration(configuration Configuration) error {
 
 	pc.configuration.ICEServers = configuration.ICEServers
 
+	if configuration.RTPHeaderEncryptionPolicy != RTPHeaderEncryptionPolicyUnknown {
+		if configuration.RTPHeaderEncryptionPolicy != pc.configuration.RTPHeaderEncryptionPolicy ||
```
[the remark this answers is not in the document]
boks1971: That suggestion potentially looks right? This is a bit hard to read. Looks like any modification is invalid. Is that correct?

Then it might be easier to check using something like, but I do not know the exact requirement here.
```
if config.policy != pc.config.policy && !pc.config.disabled {
     invalid
}
```

JoTurk:
I'm going to review this soon, I think we should try to avoid adding new files to pion/webrtc, i think all the new files can be moved to already existing files :)

dtlstransport.go
sirzooro: Yes, in `DTLSTransport.startSRTP`, `DTLSTransport.Start` and `RTPTransceiver.RTPHeaderEncryptionNegotiated`

peerconnection.go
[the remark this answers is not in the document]
sirzooro: I thought about this more and decided to replace SettingsEngine option with new value in RTPHeaderEncryptionPolicy. It is more readable and not so confusing.

peerconnection.go
sirzooro: Removed this part, now code always passes RTPHeaderEncryptionPolicyNegotiate

peerconnection_js.go
```diff
@@ -51,6 +51,7 @@ func (api *API) NewPeerConnection(configuration Configuration) (_ *PeerConnectio
 			err = recoveryToError(e)
 		}
 	}()
+
```
JoTurk: can we remove this noise?

peerconnection_test.go
```diff
@@ -280,6 +280,10 @@ func TestPeerConnection_GetConfiguration(t *testing.T) {
 	assert.Equal(t, expected.ICETransportPolicy, actual.ICETransportPolicy)
 	assert.Equal(t, expected.BundlePolicy, actual.BundlePolicy)
 	assert.Equal(t, expected.RTCPMuxPolicy, actual.RTCPMuxPolicy)
+
+	// This field is not tested because it is not supported in the JavaScript/Wasm bindings.
+	// assert.Equal(t, expected.RTPHeaderEncryptionPolicy, actual.RTPHeaderEncryptionPolicy)
```
JoTurk: Is there a reason to add this comment? can we just test this in _go_test?

peerconnection.go
sirzooro: Value Unknown is used for any unrecognized value assigned in JS code. Most likely this extension would be part of JSEP one day. I added Disabled as additional value here.

rtptransceiver.go
sirzooro: Changed

sirzooro:
> I'm going to review this soon, I think we should try to avoid adding new files to pion/webrtc, i think all the new files can be moved to already existing files :)

I have moved contents of these files elsewhere and removed them.

rtpheaderencryptionpolicy.go
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
```
JoTurk: can we please move / merge this file with other files? i don't think we should add any files to this package. it's already so hard to scroll, what do you think?

dtlsparameters.go
```diff
@@ -7,4 +7,18 @@ package webrtc
 type DTLSParameters struct {
 	Role         DTLSRole          `json:"role"`
 	Fingerprints []DTLSFingerprint `json:"fingerprints"`
+
+	// srtpCryptexMode is a local-only knob used by the ORTC-style API to inform
+	// the DTLS transport how SRTP should be configured once the DTLS handshake
+	// completes. It is not part of the DTLS signaling parameters.
+	srtpCryptexMode srtpCryptexMode `json:"-"`
```
JoTurk: I don't feel adding this to `DTLSParameters` is a right idea, it breaks the assumption that it's a signaling struct, it's fine if we have to.

errors.go
```diff
@@ -53,6 +53,10 @@ var (
 	// RTCPMuxPolicy was made after PeerConnection has been initialized.
 	ErrModifyingRTCPMuxPolicy = errors.New("rtcp mux policy cannot be modified")
 
+	// ErrModifyingRTPHeaderEncryptionPolicy indicates that an attempt to modify
+	// RTPHeaderEncryptionPolicy was made after PeerConnection has been initialized.
+	ErrModifyingRTPHeaderEncryptionPolicy = errors.New("rtp header encryption policy cannot be modified")
```
JoTurk: do we need to export this error?

JoTurk (commented):
sorry this took time, it's mostly fine, it would be cooler if we could clean the negations / knob area so we don't carry multiple states and types, the mapping between srtpCryptexMode and RTPHeaderEncryptionPolicy could get cleaner, but it's not a blocker IMO. 

it would be nice to add a clear test to guarantee that "negotiate" only turns on when actually negotiated.

I'll continue with the test part of the review tomorrow.

thank you. Sorry this took a while.

peerconnection_js.go
sirzooro: removed

peerconnection_test.go
sirzooro: No, it is no longer needed. Removed.

rtpheaderencryptionpolicy.go
sirzooro: Other `RTPsomething` types have their own files, so I added new one for this one too. Maybe we should move all of them to one common file?

dtlsparameters.go
sirzooro: Removed this field and `srtpCryptexMode` type, it was no longer needed.

errors.go
sirzooro: Other similar errors were public. Changed to private.

rtpheaderencryptionpolicy.go
JoTurk: I agree, I plan to refactor them later. Let's merge this as its.

JoTurk:
@sirzooro I just noticed now, why we add and test for `a=cryptex` in applications media sections?

sirzooro:
> @sirzooro I just noticed now, why we add and test for `a=cryptex` in applications media sections?

`a=cryptex` has to be added to all sections which are part of `BUNDLE`, and application section is in one bundle with video and audio sections. Beside this there is no reason to add it there.

JoTurk:
@sirzooro this doesn't sound right. per spec this rule only apply to rtp sections not applications.

> If BUNDLE is in use as per [[RFC9143](https://www.rfc-editor.org/rfc/rfc9335.html#RFC9143)] and the "a=cryptex" attribute is present for a media line, it MUST be present for all RTP-based "m=" sections belonging to the same bundle group. This ensures that the encrypted Media Identifier (MID) header extensions can be processed, allowing RTP streams to be associated with the correct "m=" section in each BUNDLE group as specified in [Section 9.2](https://www.rfc-editor.org/rfc/rfc9143#section-9.2) of [[RFC9143](https://www.rfc-editor.org/rfc/rfc9335.html#RFC9143)]. When used with BUNDLE, this attribute is assigned to the TRANSPORT category [[RFC8859](https://www.rfc-editor.org/rfc/rfc9335.html#RFC8859)].

https://www.rfc-editor.org/rfc/rfc9335.html#section-4

Maybe i missed other spec?

sirzooro:
> @sirzooro this doesn't sound right. per spec this rule only apply to rtp sections not applications.
> 
> > If BUNDLE is in use as per [[RFC9143](https://www.rfc-editor.org/rfc/rfc9335.html#RFC9143)] and the "a=cryptex" attribute is present for a media line, it MUST be present for all RTP-based "m=" sections belonging to the same bundle group. This ensures that the encrypted Media Identifier (MID) header extensions can be processed, allowing RTP streams to be associated with the correct "m=" section in each BUNDLE group as specified in [Section 9.2](https://www.rfc-editor.org/rfc/rfc9143#section-9.2) of [[RFC9143](https://www.rfc-editor.org/rfc/rfc9335.html#RFC9143)]. When used with BUNDLE, this attribute is assigned to the TRANSPORT category [[RFC8859](https://www.rfc-editor.org/rfc/rfc9335.html#RFC8859)].
> 
> https://www.rfc-editor.org/rfc/rfc9335.html#section-4
> 
> Maybe i missed other spec?

You are right, I interpreted this incorrectly. I will fix this.

sirzooro:
> You are right, I interpreted this incorrectly. I will fix this.

Fixed, `a=cryptex` is no longer added to application section.

JoTurk:
im sry i found few issues, i'll double check tomorrow morning before i comment.

fippo:
https://github.com/w3c/webrtc-extensions/issues/47#issuecomment-3878343882 -- currently pushing for some simplification of the API. I don't see why anyone would want the extra complication of not requiring cryptex but dealing with the result at a transceiver level.

peerconnection.go
```diff
@@ -1368,6 +1386,57 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+		}
+
+		haveMedia = true
+		hasCryptex := isCryptexSet(media)
+		negotiatedForAnyMedia = negotiatedForAnyMedia || hasCryptex
+		negotiatedForAllMedia = negotiatedForAllMedia && hasCryptex
+	}
+
+	if !haveMedia {
+		return false, false
```
JoTurk, line 1563: Won't this make us throw ErrRTPHeaderEncryptionRequired if we only have datachannels in the required mode?

peerconnection.go
```diff
@@ -1368,6 +1386,57 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+		if media.MediaName.Port.Value == 0 || media.MediaName.Media == mediaSectionApplication {
+			continue
+		}
+
+		haveMedia = true
+		hasCryptex := isCryptexSet(media)
+		negotiatedForAnyMedia = negotiatedForAnyMedia || hasCryptex
+		negotiatedForAllMedia = negotiatedForAllMedia && hasCryptex
+	}
+
```
JoTurk, line 1561: I think we're going with always negotiating at session-level, if yes, can we consider enforcing/normalizing to session-level only, and simplifying this helper?

peerconnection.go
```diff
@@ -3010,7 +3112,11 @@ func (pc *PeerConnection) generateMatchedSDP(
 				}
 				mediaTransceivers = append(mediaTransceivers, transceiver)
 			}
-			mediaSections = append(mediaSections, mediaSection{id: midValue, transceivers: mediaTransceivers})
+			mediaSections = append(mediaSections, mediaSection{
+				id:           midValue,
+				transceivers: mediaTransceivers,
+				cryptex:      cryptex,
+			})
 		case sdpSemantics == SDPSemanticsUnifiedPlan || sdpSemantics == SDPSemanticsUnifiedPlanWithFallback:
```
JoTurk, line 3383: Does this mean we will never offer cryptex again on later renegotiations if the first negotiation doesn't echo cryptex?
It's fine if we don't support cryptex for renegotiations, but i feel this should be explicit somewhere.

peerconnection.go
```diff
@@ -2799,6 +2869,8 @@ func (pc *PeerConnection) startTransports(
 		}()
 	}
 
+	pc.dtlsTransport.cryptexMode = cryptexMode
```
JoTurk: is this racy?

sdp.go
```diff
@@ -364,6 +364,7 @@ func addDataMediaSection(
 	shouldAddCandidates bool,
 	dtlsFingerprints []DTLSFingerprint,
 	midValue string,
+	cryptex bool,
```
JoTurk: can we remove this?

JoTurk (commented):
sorry for being late, more comments.

peerconnection.go
sirzooro: Yes, fixed

peerconnection.go
sirzooro: I changed this to make session-level preferred. Media-level may be needed as a fallback, so I kept support for it and updated code to respond with attribute at proper level (session or media).

peerconnection.go
sirzooro: Yes, I fixed this. Unfortunately now it is not possible to update SRTP context on the fly or recreate it, so renegotiation actually cannot change cryptex mode. I also added extra check for this.

peerconnection.go
sirzooro: Possibly, fixed.

sdp.go
sirzooro: Removed

peerconnection.go
fippo: Interesting point! The RFC only says
> Generating subsequent SDP offers and answers MUST use the same procedures for including the "a=cryptex" attribute as the ones on the initial offer and answer.

but this is generation, not handling of subsequent answers. The problem is if you do this:
```
O: cryptex
A: no
(create SRTP sessions)
O: cryptex
A: cryptex
(create more SRTP sessions)
```
you end up with a weird mix of cryptex and non-cryptex. Technically this is valid and distinguishable and... you are asking for trouble. And the `require` policy should prevent this.

sirzooro: When cryptex attribute is at session level, after renegotiation like in this example old SRTP sessions should start using cryptex too. I am not sure why someone would want to do this. Plus this is a potential attack vector, to trick peer to downgrade its security a bit for existing sessions by disabling header encryption (`require` policy prevents this). So even for `negotiate` policy it makes sense to prevent renegotiation of cryptex.

fippo: Good point - @juberti since I see anything in the RFC is it worth pointing it out as an errata?

fippo: hrm... the RFC allows modification:
> Both endpoints can change the Cryptex support status by modifying the session as specified in Section 8 of [RFC3264]

but does not specify what happens to existing SRTP streams?

sirzooro: > hrm... the RFC allows modification:
> 
> > Both endpoints can change the Cryptex support status by modifying the session as specified in Section 8 of [RFC3264]
> 
> but does not specify what happens to existing SRTP streams?

[RFC 3264 section 8.3.4](https://www.rfc-editor.org/rfc/rfc3264#section-8.3.4) clarifies this:

> 8.3.4 Changing Attributes
> Any other attributes in a media description MAY be updated in an
> offer or answer.  Generally, an agent MUST send media (if the
> directionality of the stream allows) using the new parameters once
> the SDP with the change is received.

lgrahl: I don't see a reason why one would ever want to downgrade (or even upgrade) an existing media track. I'm not even sure how one would implement such a transition, since the change on the SRTP level is inherently racing with the O/A exchange.

Perhaps this really is something to clarify as an errata to the RFC.

sirzooro: > I'm not even sure how one would implement such a transition, since the change on the SRTP level is inherently racing with the O/A exchange.

Cryptex packets use special values for RTP Extended Header Profile field, so other peer can tell if Cryptex was used or not and decode RTP packet properly.

lgrahl: @sirzooro Ah, thanks! It still begs the question if there is any need for such a transition, especially if the resulting implementation is complex. Hopefully that question is in good hands with @fippo and @juberti. :slightly_smiling_face:

JoTurk: @lgrahl  My original question was mainly about avoiding behavioral surprises. If we don't support cryptex renegotiation, that behavior should be explicit, e.g. we should not offer cryptex again after the initial negotiation and should not renegotiate it if the peer starts offering it later.

I think it would be good to go with far as defining this in tests as well, until cryptex renegotiation is actually supported (we'll need to add live reconfigure to SCTP like how we did with ICE recently). that way we avoid silent behavior changes between versions when related code changes (which has happened to us many times before, and it makes reviews a lot harder).

fippo: I don't think it is too hard to enable/disable cryptex on the fly at srtp level but I don't really see a practical use-case for it.
Throwing an error until that is supported (if ever) seems right.

sirzooro: @JoTurk I can change code to not offer cryptex again when it was not negotiated in initial offer/answer negotiation, but this will address this issue only one scenario, when other peer would like to enable cryptex later in renegotiation. Opposite scenario is when cryptex is negotiated, and another peer tries to disable it during renegotiation. This opposite scenario is even worse from security perspective. I propose to keep it as-it, and add support for cryptex renegotiation later if/when someone would need it.

murillo128: Not sure if I understood the whole issue, but note that `cryptex` attribute only indicate the capability to receive and process cryptex packets. Sending side behavior  is not controlled by SDP signaling but by app APIs (`negotiate`, `require`).

It could be possible for a forwarding proxy/gateway to renegotiate the SDP to modify the upstream destination server, and change the cryptex flag. It makes not sense for a terminating RTP endpoint to modify the cryptex flag between negotiation.

Also note that cryptex may not be bidirectional, if offerer (A) doesn't add the cryptex attribute in the SDP offer and answerer (B) doesn't add the attribute in the answer SDP, B could still send cryptex packets to A. 

Even further, even if both A and B support cryptex in the SDP O/A, they could still send each other plain rtp packet.

It is up to local policy (i.e. APIs) to decide what to do when the other side does not support receiving cryptex or what to do with non-cryptex rtp packet is received.

fippo: @murillo128 thank you, I missed this is declarative (or did not want to see?)

The W3C API controls negotiation but does that mean it is also missing a way to control and determine whether encryption is used? I would strongly prefer not to because this will be a headache.

murillo128: I would neither like an API for wether encryption is used on the sender side if the received supports it. I think sender should always use it in that case.

fippo: @murillo128 made a great point in https://mailarchive.ietf.org/arch/msg/avt/pGnete_6IqExxTERNl44_9GqVRU/ which probably requires W3C API changes.

The most pragmatic thing here is to ship cryptex with the two policy options *disabled* and *negotiate* for the time being and add "require" with a new definition of "mandatory to negotiate **and use**" later which shouldn't require many changes to the PR.

juberti: are these the right definitions?
- `disabled`: don't offer a=cryptex, ignore incoming cryptex, and don't send cryptex even if the remote side supports it in SDP
- `negotiate`: offer a=cryptex, handle incoming cryptex (or not), and send cryptex if the remote side supports it in SDP (as usual)
- `require`: offer a=cryptex, fail on incoming non-cryptex, and fail if the remote side doesn't support cryptex in SDP

JoTurk:
@sirzooro is this ready for another round of review, sorry about the conflicts btw.

JoTurk:
Since this is merged https://webrtc-review.googlesource.com/c/src/+/451800 we can use libwebrtc to test it.

peerconnection.go
JoTurk: @sirzooro based on my understanding of the libwebrtc impl, we'll have to always declare `a=cryptex` if enabled on our side, and fail in the `require` mode if the remote didn't send cryptex. I think we can make changes to srtp so it has updateOptions/updateSession similar to the API we added to ICE recently, I think it should be a lite change.
/cc @fippo

sirzooro:
> Since this is merged https://webrtc-review.googlesource.com/c/src/+/451800 we can use libwebrtc to test it.

@JoTurk  Thanks for the info. I will update this PR, address new comments and do some tests with libwebrtc. I will let you know when this PR will be ready for next round of review.

fippo:
one can now test in Chrome 151+ with `--enable-features=RtcRtpHeaderEncryptionPolicy`

zshang-oai:
@sirzooro just checking how's going with this PR

peerconnection.go
```diff
@@ -1385,6 +1413,100 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+	// change Cryptex mode after SRTP has already been created.
+	noActiveRTPMedia := !negotiatedForAnyMedia && negotiatedForAllMedia
+	if noActiveRTPMedia {
+		switch pc.configuration.RTPHeaderEncryptionPolicy {
+		case RTPHeaderEncryptionPolicyDisable:
+			return srtp.CryptexModeDisabled, nil
+		case RTPHeaderEncryptionPolicyRequire:
+			return srtp.CryptexModeRequired, nil
+		case RTPHeaderEncryptionPolicyUnknown, RTPHeaderEncryptionPolicyNegotiate:
+			return srtp.CryptexModeEnabled, nil
```
zshang-oai, line 1475: **[P2] Leave RTP encryption undecided until the first RTP negotiation**

Selecting `CryptexModeEnabled` here for a data-only session commits the default-policy peer to a mode before any RTP capability has been exchanged. Once DTLS/SRTP is initialized, the mode-change guard in `SetRemoteDescription` prevents the first real media negotiation from selecting the mode the peer supports.

I reproduced this in both offerer/answerer roles:

1. Connect a default/`negotiate` peer to a `disable` peer, initially with only a data channel.
2. Wait for both SRTP sessions to be initialized.
3. Add the first audio transceiver and renegotiate.

The second negotiation fails with `InvalidModificationError: rtp header extension encryption mode change not supported`. Neither initial SDP contained RTP media or advertised Cryptex, so there was no previous RTP encryption agreement to change.

[RFC 3264 §8.1](https://datatracker.ietf.org/doc/html/rfc3264#section-8.1) describes adding media through a subsequent offer, and [RFC 9335 §4](https://www.rfc-editor.org/rfc/rfc9335.html#section-4) defines the SDP indication of Cryptex receive support. This is a default-policy interoperability regression in that normal data-first workflow, rather than an argument that every added stream must be accepted.

Could we distinguish an initialized data/DTLS transport from a committed RTP mode, and defer the latter or allow SRTP configuration when RTP is first negotiated? Simply defaulting this branch to `Disabled` would move the same problem to Cryptex-capable peers. A regression test should wait for transport establishment before adding media, with both supporting and non-supporting peers in both roles.

peerconnection.go
```diff
@@ -1208,7 +1220,22 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
 	default:
 		pc.canTrickleICECandidates = ICETrickleCapabilityUnknown
 	}
+
+	cryptexMode, cryptexErr := pc.getCryptexModeFromRemoteSDP(desc.parsed)
 	pc.mu.Unlock()
+	if cryptexErr != nil {
+		return cryptexErr
```
zshang-oai, line 1227: **[P2] Reject invalid Cryptex SDP before committing description/signaling state**

These new error paths run after `pc.setDescription` and the media-engine update. Returning the expected Cryptex error therefore does not leave the PeerConnection in its pre-call state.

For an initial audio offer from a `require` peer, applying an answer without `a=cryptex` returns `ErrRTPHeaderEncryptionRequired`, but the observed state is:

```text
signalingState = stable           // was have-local-offer
currentRemoteDescription != nil  // contains the rejected answer
pendingLocalDescription == nil   // the outstanding offer was consumed
```

Retrying the answer with the missing attribute added then fails with `InvalidModificationError: invalid proposed signaling state transition: stable->SetRemote(answer)->stable`. I also reproduced the offer-side case: a rejected offer leaves `have-remote-offer` and a pending remote description, so a corrected offer cannot be retried.

[JSEP, RFC 9429 §5.10](https://www.rfc-editor.org/rfc/rfc9429.html#section-5.10) explicitly requires restoring the pre-operation session state when applying a remote description returns an error. The observable state here violates that requirement and makes the policy rejection difficult for an application to recover from.

Please perform the Cryptex policy/mixed-mode/mode-change checks before committing affected state, or roll back all affected state on failure. The mode-change error immediately below has the same ordering issue. Regression tests should assert preserved current/pending descriptions and signaling state, plus a successful corrected-description retry; checking only `ErrorIs` misses the corruption.

dtlstransport.go
```diff
@@ -191,12 +192,20 @@ func (t *DTLSTransport) GetRemoteCertificate() []byte {
...
 	srtpConfig := &srtp.Config{
 		Profile:       t.srtpProtectionProfile,
 		BufferFactory: t.api.settingEngine.BufferFactory,
 		LoggerFactory: t.api.settingEngine.LoggerFactory,
 	}
+
+	if t.cryptexMode == srtp.CryptexModeEnabled || t.cryptexMode == srtp.CryptexModeRequired {
+		opt := srtp.Cryptex(t.cryptexMode)
+		srtpConfig.LocalOptions = append(srtpConfig.LocalOptions, opt)
+		srtpConfig.RemoteOptions = append(srtpConfig.RemoteOptions, opt)
```
zshang-oai, line 218: **[P2] Keep advertised receive capability separate from outbound encryption mode**

The same `cryptexMode` controls both SRTP directions, but `getCryptexModeFromRemoteSDP` derives it from the remote SDP. With a default-policy offer containing `a=cryptex` and an answer omitting it, Pion initializes both directions as `Disabled`: it stops accepting Cryptex even though its own offer advertised receive support.

I verified the packet-level consequence using PeerConnection-created sessions over ICE/DTLS. After that asymmetric offer/answer exchange, I encrypted a `0xC0DE` RTP packet with the answerer's actual DTLS-exported SRTP keys and sent it into the established transport. The offerer's read timed out; an ordinary SRTP packet with the same keys and transport was received successfully. This isolates the loss to the receive transform rather than connectivity or unauthenticated input.

[RFC 9335 §4](https://www.rfc-editor.org/rfc/rfc9335.html#section-4) defines `a=cryptex` as the advertising endpoint's receive capability. Under the declaration-based interpretation already discussed [in this PR](https://github.com/pion/webrtc/pull/3357#discussion_r2894578578), the peer's receive declaration should determine our outgoing encryption, not withdraw our advertised receive support. The RFC also conditions unilateral sending on mutual verification of support, so I am not claiming an unconditional MUST permitting every asymmetric exchange; this is the concrete implementation consequence that needs resolving alongside that discussion.

Could we configure receive acceptance from the local policy/capability and send encryption from the peer's receive support, with packet-level tests for the asymmetric case? `negotiate` receive handling should retain support for both conventional SRTP and Cryptex under that model.

peerconnection.go
```diff
@@ -1385,6 +1413,100 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+	haveAnyMedia := false
+	haveRTPMedia := false
+
+	if _, hasCryptex := desc.Attribute(sdp.AttrKeyCryptex); hasCryptex {
+		return true, true
+	}
+
+	for _, media := range desc.MediaDescriptions {
+		if media.MediaName.Port.Value == 0 {
+			continue
```
zshang-oai, line 1428: **[P2] Include zero-port bundle-only RTP in the Cryptex policy check**

A zero port does not always mean rejected media in an offer. [RFC 9143 §6](https://www.rfc-editor.org/rfc/rfc9143.html#section-6) and [§7.2](https://www.rfc-editor.org/rfc/rfc9143.html#section-7.2) define the `a=bundle-only` exception: that media section requests acceptance within its BUNDLE group.

I reproduced this with an initial offer whose data section is the BUNDLE tag and whose audio section has port zero plus `a=bundle-only`, with **no `a=cryptex` anywhere**. The relevant shape is below; unrelated SDP fields are omitted:

```sdp
a=group:BUNDLE 1 0
m=application 9 UDP/DTLS/SCTP webrtc-datachannel
a=mid:1
m=audio 0 UDP/TLS/RTP/SAVPF 111
a=mid:0
a=bundle-only
```

This loop skips the audio section, classifies the offer as data-only, and reaches the vacuous-success branch. With `require`, `SetRemoteDescription` incorrectly succeeds and selects `CryptexModeRequired`; `CreateAnswer` accepts audio at port 9 without Cryptex. With `negotiate`, the same offer incorrectly selects `Enabled` instead of an unencrypted outgoing mode.

The reproduction places shared transport/IDENTICAL attributes on the BUNDLE tag, following [§7.1.3](https://www.rfc-editor.org/rfc/rfc9143.html#section-7.1.3), so this is not relying on duplicate attributes in the bundle-only section.

Please account for BUNDLE membership and `bundle-only` before treating a zero-port RTP section as rejected. The effective Cryptex capability also needs to respect BUNDLE attribute inheritance, rather than requiring a literal copy on every section. Add a data-tag/bundle-only-audio regression for both policies; `require` should not accept this RTP offer without the required peer capability.

rtptransceiver.go
```diff
@@ -252,6 +252,25 @@ func (t *RTPTransceiver) Direction() RTPTransceiverDirection {
...
+	if dtlsTransport == nil {
+		if receiver := t.Receiver(); receiver != nil {
+			dtlsTransport = receiver.Transport()
+		}
+	}
+	if dtlsTransport == nil {
+		return false
+	}
+
+	return dtlsTransport.rtpHeaderEncryptionNegotiated()
```
zshang-oai, line 271: **[P2] Report this transceiver's negotiation state, not the shared transport's mode**

This returns a shared DTLS-transport flag, so a transceiver can report negotiated encryption before it has participated in any offer/answer exchange.

Using only the public API, I connected two default-policy peers with audio, waited for the connection, then added a video transceiver without creating another offer. The result was:

```text
connected audio:                 mid="0" encrypted=true
new video, no offer created:     mid=""  encrypted=true
```

The [December 19 WebRTC Extensions draft cited by this PR](https://github.com/w3c/webrtc-extensions/blob/38b4cc557eec29a330ea9f603d21fae9db3a3bcf/index.html#L1403-L1408) explicitly initializes each new transceiver's `[[RtpHeaderEncryptionNegotiated]]` slot to `false` and requires the getter to return that slot. Sharing a transport with an already-negotiated transceiver does not satisfy that per-transceiver initialization contract.

Please track the negotiated result on the transceiver and update it with the applicable successful negotiation, rather than treating the transport mode as the public API state. The minimal regression is an established encrypted audio connection followed by `AddTransceiverFromKind(RTPCodecTypeVideo)`: the new transceiver must still return false before negotiation. The draft does not fully specify later update timing, so the creation case is the clearest mismatch.

Related API-completeness issue: this method is only added to the native implementation, not `rtptransceiver_js.go`. A program referencing `(*webrtc.RTPTransceiver).RTPHeaderEncryptionNegotiated` compiles natively but fails with an undefined-method error for `GOOS=js GOARCH=wasm`. Please expose the corresponding getter in the JS wrapper as well.

zshang-oai (commented):
Reviewed at `b10b8d169b98f780c4b0e2ee79bdec8760fdda0c`.

The core packet-encryption path looks close to the stated goal. The focused Cryptex tests passed, as did a local 24-case round-trip matrix covering AES-128-GCM, AES-256-GCM and AES-128-CM/HMAC-SHA1, both `negotiate` and `require`, and one-byte extensions, two-byte extensions, CSRC-only and bare RTP packets. Those checks used PeerConnection-created SRTP sessions over ICE/DTLS; they are not browser interoperability or full RTPSender/OnTrack coverage.

I reproduced the five integration issues below at this head. The main remaining work is the negotiation lifecycle, preserving state when SDP is rejected, keeping receive capability separate from outbound encryption, handling BUNDLE's zero-port exception, and reporting per-transceiver API state accurately. I would address those before calling this the final integration. The asymmetric-capability comment explicitly distinguishes the declaration-based interpretation discussed in this PR from a blanket RFC MUST.

Verification limits: I have not established a complete native-suite pass or Chrome/libwebrtc interoperability. The added reproduction tests were local review checks, not changes pushed to this PR.

sirzooro:
> @sirzooro just checking how's going with this PR

Sorry for late response. I still remember about this PR. Some IRL things needed my attention in recent months and I did not have much free time. Hopefully now I will be able to finish this PR.

Thank you for your review. Looks that most of the issues needs ability to update Cryptex status in created RTSP session, what needs `pion/srtp` change. This is another thing that I started and have to finish.

I will push rebased version of my code, with some fixes for other discovered issues.

zshang-oai (commented):
Reviewed at `4e60d59d0041b02d65199a6f5a79c2daf1fc78d9`.

I reproduced three remaining issues in Cryptex negotiation: the answerer can advertise receive support without enabling it, a rejected description can change the active encryption mode, and a valid BUNDLE answer can be rejected.

The focused Cryptex tests and WASM compilation passed. The broader native test run stopped in `TestMulticastDNSCandidates` after multicast bind errors and a timeout. Browser interoperability was not tested.

peerconnection.go
```diff
@@ -1386,6 +1416,183 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+		localDesc = pc.pendingLocalDescription
+	}
+	if localDesc == nil || localDesc.parsed == nil {
+		// We haven't generated our own SDP yet (for example, when answering the very first
+		// offer of a negotiation). Our upcoming answer mirrors the remote's declared Cryptex
+		// support, so localCryptexMode already reflects what we will declare for receiving.
+		return localCryptexMode, localCryptexMode, nil
+	}
+
+	remoteCryptexMode, err = pc.cryptexModeFromSDP(localDesc.parsed)
```
zshang-oai, line 1574: **[P2] Update the receive mode when applying a new local answer**

During renegotiation, `localDesc` here is the answerer’s previous local description. If that answer omitted Cryptex and the new remote offer includes it, the receive mode stays disabled. `CreateAnswer` then advertises Cryptex, but applying that answer never updates the receiver.

I reproduced this with the same two PeerConnections: signaling removed Cryptex from the first offer after `SetLocalDescription`, then delivered the next offer unchanged. [JSEP §5.4](https://www.rfc-editor.org/rfc/rfc9429.html#section-5.4) permits that capability filtering, and [RFC 9335 §4](https://www.rfc-editor.org/rfc/rfc9335.html#section-4) permits changing the advertisement during renegotiation. This is an application-controlled signaling case; no endpoint replacement or configuration change is involved.

After the second exchange succeeded, the sender transmitted Cryptex packets but the answerer delivered none, compared with 10/10 before renegotiation.

Please update the inbound SRTP mode when applying the local description that advertises the new receive capability, and cover this transition with an RTP delivery test.

peerconnection.go
```diff
@@ -1188,6 +1200,22 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+	pc.mu.Lock()
+	localCryptexMode, remoteCryptexMode, cryptexErr := pc.getCryptexModes(desc.parsed)
+	pc.mu.Unlock()
+	if cryptexErr != nil {
+		return cryptexErr
+	}
+
+	// Applies the negotiated Cryptex modes to the already-started SRTP session, if any.
+	if isRenegotiation {
+		if err := pc.dtlsTransport.updateCryptexModes(localCryptexMode, remoteCryptexMode); err != nil {
```
zshang-oai, line 1239: **[P2] Preserve active SRTP modes when rejecting a description**

This updates the running SRTP session before `setDescription` validates the signaling transition.

On an established default-policy Cryptex connection, I submitted another answer with Cryptex removed while signaling was already stable. `SetRemoteDescription` returned an invalid-transition error and preserved the current remote SDP, but subsequent RTP packets switched from encrypted headers (`0xc0de`) to clear header extensions (`0xbede`). The media payload remained SRTP-encrypted.

A caller handling the rejected answer therefore still gets a change to its established session. Please validate before mutating the active transport, and ensure any subsequent failure restores the previous modes. A regression should check that a rejected description leaves both the stored SDP and packet encryption unchanged.

peerconnection.go
```diff
@@ -1386,6 +1416,183 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+	if media.MediaName.Port.Value == 0 && !bundleOnly {
+		return false, false, false
+	}
+
+	if media.MediaName.Media == mediaSectionApplication {
+		return true, false, false
+	}
+
+	hasCryptex = isCryptexSet(media)
+	if !hasCryptex && bundleOnly && bundleTagMedia != nil {
```
zshang-oai: **[P2] Resolve Cryptex inheritance for ordinary bundled sections**

The `bundleOnly` condition limits inheritance too narrowly. Cryptex belongs to the TRANSPORT category. In answers and subsequent offers, [RFC 9143 §7.1.3](https://www.rfc-editor.org/rfc/rfc9143.html#section-7.1.3) places those attributes on the tagged section and applies them to the other members of the BUNDLE group, including ordinary nonzero-port sections without `a=bundle-only`.

I reproduced this with an audio/video answer whose media ports were both `9`. Session-level Cryptex was accepted, and copies on both media sections were accepted. Moving Cryptex solely to the audio BUNDLE tag caused `SetRemoteDescription` to reject it with `rtp header extension encryption mixed mode not supported`.

Please determine the effective capability using BUNDLE membership and offer/answer context. Add an answer with tag-only Cryptex to the negotiation tests.

JoTurk:
@sirzooro I'm going to test this PR against libwebrtc this weekend.

sirzooro:
All issues reported so far should be fixed. Please take a look on this PR again.

zshang-oai (approved):

fippo:
commit message says RFC 9334

sirzooro:
> commit message says RFC 9334

good catch, changed

peerconnection.go
```diff
@@ -1386,6 +1437,206 @@ func (pc *PeerConnection) SetRemoteDescription(desc SessionDescription) error {
...
+	// change Cryptex mode after SRTP has already been created.
+	noActiveRTPMedia := !negotiatedForAnyMedia && negotiatedForAllMedia
+	if noActiveRTPMedia {
+		switch pc.configuration.RTPHeaderEncryptionPolicy {
+		case RTPHeaderEncryptionPolicyDisable:
+			return srtp.CryptexModeDisabled, nil
+		case RTPHeaderEncryptionPolicyRequire:
+			return srtp.CryptexModeRequired, nil
+		case RTPHeaderEncryptionPolicyUnknown, RTPHeaderEncryptionPolicyNegotiate:
+			return srtp.CryptexModeEnabled, nil
```
JoTurk, line 1605: This cases data-only peers to incorrectly enable Cryptex without peer support, I confirmed this against chrome, please remove the `noActiveRTPMedia` based early return.

peerconnection.go
```diff
@@ -2829,6 +3088,9 @@ func (pc *PeerConnection) startTransports(
 		}()
 	}
 
+	pc.dtlsTransport.setLocalCryptexMode(localCryptexMode)
+	pc.dtlsTransport.setRemoteCryptexMode(remoteCryptexMode)
+
```
JoTurk, line 3093: Nit: while i was testing `renegotiation` with toggling Cryptex on / off, this causes a race if ICE didn't connect before the renegotiation and we can switch to earlier values.

peerconnection.go
JoTurk: nit: ^^^ I confirmed that this can happen if Pion rejects the renegotiation SDP, and it can prematurely toggle cryptex even if we reject `SetRemoteDescription`

peerconnection.go
JoTurk: ^ i confirmed that this and the renegotiation race can break later media tracks if the original call starts with a datachannel only, since is this is a common setup for SFUs we should fix this. This happens even if the user doesn't use cryptex at all.

JoTurk (changes requested):
Hello i tested against Chrome, everything seems to work as intended. There are just 2 related bugs that can break normal calls, we should fix them before this is merged

peerconnection.go
fippo: yep, datachannel-only is mean for cryptex. It also means that if you want to check cryptex-consistency of a bundle whose first/primary m-line is the datachannel one you need to special-case that...

https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/pc/sdp_offer_answer_unittest.cc;l=3440?q=SdpOfferAnswerTest.CryptexRequiredDatachannelOnly is the C++ test I had for this.

sirzooro:
@JoTurk thank you for testing this. I have fixed reported issues.

JoTurk:
thank you @sirzooro i'll test again tonight.

JoTurk (approved):
I tested again yesterday and it works as intended, thank you

Issue #3335, opened 2025-12-31
WARP tracking issue


fippo:
## Summary
WARP is an acronym (coined by @juberti) that stands for "WebRTC Abridged Roundtrip Protocol" which cuts down the amount of round-trip involved in establishing a WebRTC connection from 6 to 2.

The pion implementation spans multiple repos and PRs which are tracked here. One part is SNAP which skips two round trips in opening the datachannel:
* SCTP SNAP implementation https://github.com/pion/sctp/pull/449
* WebRTC SNAP implementation https://github.com/pion/webrtc/pull/3327

The second (and more complicated) part is SPED, embedding the DTLS handshake in the STUN/ICE connection establishment. This is going to span multiple PRs in several repos:
* STUN changes https://github.com/pion/stun/pull/260
* ICE changes to read and write the attributes https://github.com/pion/ice/pull/861
* ICE changes to insert and process the attributes with the correct semantics (PR TBD, might be more than one)
* ICE changes to make the API nonblocking so that DTLS can start at the same time as DTLS https://github.com/pion/ice/pull/862
* DTLS changes to intercept/get notified about inbound and outbound DTLS packets and inject them from non-network sources https://github.com/pion/dtls/pull/766
* WebRTC changes to glue everything together, mainly in the ICE and DTLS transport and how they are started (PR TBD)
* WebRTC changes that allow handling non-actpass offers so that the client (i.e. a browser) can send the DTLS client hello with the first outgoing STUN packets https://github.com/pion/webrtc/pull/3325

Last but not least
* A demo for testing all this https://github.com/pion/webrtc/pull/3326

🐻 with me, this is going to be a bit of a mess

PR #3362, opened 2026-01-19
Implement SPED (STUN Protocol for Embedding DTLS)

fippo:
part of #3335

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -13,6 +13,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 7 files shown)


JoTurk:
Hello @fippo is this just blocked by https://github.com/pion/dtls/pull/766 ? Maybe we need a better way to provide a socket for dtls, we had a related issue recently. If this was the case I can propose an API.

fippo:
Yeah, that one still has different points of view on the approach

cnderrauber:
> Hello @fippo is this just blocked by [pion/dtls#766](https://github.com/pion/dtls/pull/766) ? Maybe we need a better way to provide a socket for dtls, we had a related issue recently. If this was the case I can propose an API.

This is an amazing feature for all WebRTC connections!! Is there a progess about the new API? I'd be happy to participate if there's anything I can do.

ValorZard:
@cnderrauber the most central place you can find out about all of this stuff is probably: https://github.com/fippo/warp-snap-sped

might be nice to have an issue on there on which WebRTC implementations are adding this.

Off the top of my head, i know str0m, pion, and libwebrtc are currently implementing this.

cnderrauber:
@ValorZard I have tested the open PRs against Chrome WARP flags, and it works well. Looks like the dtls pr is blocked by the API discussion, so I'm curious about the new API proposal.

fippo:
Last I tested recent changes to libWebRTC (past M146) broke interop. We're still trying to figure out the way forward for that.

JoTurk:
Since we're breaking the DTLS API for dtls 1.3 we can look into exposing a better API or sane I/O style API and let the user handle the socket themself, Will this work for you?

PR #3371, opened 2026-02-10
Add StartContext to ICE and DTLS Transport

HashimTheArab:
## Summary
- Adds `StartContext` method to `ICETransport` that accepts a `context.Context` parameter, allowing callers to control cancellation and timeouts for ICE connectivity checks
- Refactors `Start` to delegate to `StartContext` with `context.Background()`, maintaining full backwards compatibility


Sean-Der:
Seems useful @HashimTheArab! I’m curious what’s the bug/use case your solving for?

ShlomoCodes:
We're using this for Minecraft Nethernet connections (https://github.com/lactyy/go-nethernet) and the transport methods can hang indefinitely if the remote peer never responds. Without this option we have to use this fragile hack:
```go
func withContextCancel(ctx context.Context, f func() error, cancel func()) error {
	errCh := make(chan error, 1)
	go func() {
		errCh <- f()
	}()
	select {
	case <-ctx.Done():
		if cancel != nil {
			cancel()
		}
		go func() { <-errCh }()
		return ctx.Err()
	case err := <-errCh:
		return err
	}
}
```

Passing a context will allow us to simplify and improve all of the call sites:
```go
  // Before
  if err := withContextCancel(ctx, func() error {
      return conn.ice.Start(nil, d.ice, &iceRole)
  }, func() {
      _ = conn.ice.Stop()
  }); err != nil {
      return fmt.Errorf("start ICE: %w", err)
  }

  // After
  if err := conn.ice.StartContext(ctx, nil, d.ice, &iceRole); err != nil {
      return fmt.Errorf("start ICE: %w", err)
  }
  ```


(I am not the author of the PR)

**Comment by JoTurk** [comment]
@RealSexMC this is interesting I'm curious why you don't use pion/ice directly? seems more fit for your use, but I might be missing something. We're currently working on improving pion/ice and making it more suitable for non-webrtc use.

**Comment by didntpot** [comment]
@RealSexMC that is very interesting, couldn't you

**Comment by HashimTheArab** [comment]
NetherNet is a full WebRTC stack, it uses ICE + DTLS + SCTP with data channels (ReliableDataChannel and UnreliableDataChannel) negotiated via SDP. We use pion's ORTC API to set up each transport individually since the signaling is custom (not standard WebSocket/HTTP), but the transports themselves are standard WebRTC.

**Comment by HashimTheArab** [comment]
Anything else required to get this merged?

**Comment by JoTurk** [comment]
Can you please fix the lint?

**Comment by HashimTheArab** [comment]
Should be good now

**Comment by HashimTheArab** [comment]
I've updated this to also add a StartContext method to DTLS transport

**Comment by HashimTheArab** [comment]
Hey is anything else needed to merge this?

**Review by JoTurk — APPROVED** [review]
lgtm.

**Review by didntpot — APPROVED** [review]

**Comment by JoTurk** [comment]
@HashimTheArab Can you please update this branch and fix the tests before we can merge this. thank you.

**Comment by HashimTheArab** [comment]
Fixed the issues. The test-i386 failure is from the master branch. Do I need to add additional testing for that codecov/patch?

**Comment by HashimTheArab** [comment]
Hey @JoTurk please let me know what's needed to get this merged, the test coverage is failing due to 66% but I don't see any tests to meaningfully add?

File: icetransport.go [hunk 1]
```diff
@@ -134,7 +147,7 @@ func (t *ICETransport) Start(gatherer *ICEGatherer, params ICEParameters, role *
 	}
 	t.role = *role
 
-	ctx, ctxCancel := context.WithCancel(context.Background())
+	ctx, ctxCancel := context.WithCancel(ctx)
 	t.ctxCancel = ctxCancel
```
JoTurk, line 153: This is a problem in the current ICE design, but it affects this PR, currently https://github.com/pion/ice/blob/main/transport.go ICE just returns an error if the context is cancelled after the connection checks started, and it doesn't call agent's close by itself (we should change in the next major version). But for this PR we should also call close when the context get cancelled.

JoTurk (commented):
Sorry for ghosting this PR. I lost track.

icetransport.go
HashimTheArab: Should be good now

icetransport.go
```diff
@@ -161,6 +174,12 @@ func (t *ICETransport) Start(gatherer *ICEGatherer, params ICEParameters, role *
...
 	t.lock.Lock()
 	if err != nil {
+		if ctx.Err() != nil {
+			t.lock.Unlock()
+			_ = t.Stop()
+			t.lock.Lock()
+		}
+
 		return err
 	}
```
JoTurk, line 190: Sorry for all the back and forth on this, just before we merge this please consider returning the context cancellation error instead of the ICE error, when the caller cancels early.

```suggestion
	if err != nil {
		if ctxErr := ctx.Err(); ctxErr != nil {
			t.lock.Unlock()
			_ = t.Stop()
			t.lock.Lock()

			return ctxErr
		}

		ctxCancel()
		t.ctxCancel = nil

		return err
	}
```

HashimTheArab: Thanks for the feedback, check now please

icetransport.go
```diff
@@ -161,6 +174,17 @@ func (t *ICETransport) Start(gatherer *ICEGatherer, params ICEParameters, role *
 	// Reacquire the lock to set the connection/mux
 	t.lock.Lock()
 	if err != nil {
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			t.lock.Unlock()
+			_ = t.Stop()
+			t.lock.Lock()
+
+			return ctxErr
+		}
```
sirzooro: Use `context.Cause(ctx)` instead of `ctx.Err()`

JoTurk: @sirzooro I don't think we would want a Pion api to return aribitary erros espically if this is used with third party created contexts, this will be kinda unpredicatable. and also sounds like a policy or API change across all of pion in a new major version and not just a single API and keep the rest.

dtlstransport_test.go
```diff
@@ -216,7 +216,7 @@ func (c *failingPacketConn) SetReadDeadline(time.Time) error  { return nil }
 func (c *failingPacketConn) SetWriteDeadline(time.Time) error { return nil }
 
 func TestDTLSTransport_Start_ErrICEConnectionNotStarted(t *testing.T) {
-	transport := &DTLSTransport{state: DTLSTransportStateNew}
+	transport := &DTLSTransport{api: NewAPI(), state: DTLSTransportStateNew}
```
JoTurk: I'm sorry can we revert this and check if t.api is null? calls with `t.api` null shouldn't panic.

icetransport.go
sirzooro: This would be useful when used with complex context trees. Cancellation cause set few levels up the tree is propagated down when context is cancelled. This allows to improve logs, it is immediately obvious that connection was terminated due to something what happened elsewhere. Generic error "context cancelled" does not carry this information.

JoTurk: I see the point, but it's outside of the scope of this change, we shouldn't introduce a new pattern or error handling randomly in a single new API and keep the rest. Because having unpredictable APIs and maintaining different behaviors is way worse than losing some context, and If we would do it, we should do it across all of pion and we should use a linter to force it, and it should be in a new release.

JoTurk: @sirzooro I made an issue https://github.com/pion/webrtc/issues/3446 does this look good for you?

sirzooro: Thanks, I am ok with this.

dtlstransport_test.go
HashimTheArab: Sorry for the delay this is fixed

didntpot (approved):

HashimTheArab:
Hey @JoTurk could you approve the workflows and let me know if we need anything else to get this merged? Thanks

JoTurk:
Hello, I tested this more and i found 3 issues, I pushed commits to fix them, I'll test again and if this is ready to merge I'll merge it, Thank you.

JoTurk (approved):
Thank you, and sorry for all the back and forth.

HashimTheArab:
Hey no worries thank you! I could have designed this better from the start, I appreciate your help

HashimTheArab:
When do you think itll be merged? Thanks again

JoTurk:
I'll merge it once the CI passes :)

PR #3391, opened 2026-03-05
Support passing RTP extensions when writing samples.

fancycode:
This is needed to set the `sdp.SDESMidURI` and `sdp.SDESRTPStreamIDURI` RTP header extensions when sending samples to the different layers of simulcast tracks. At least I didn't find another way to do this.

I considered reusing `rtp.Extension` but that doesn't expose the internal data, so additional accessors and a `NewExtension(id uint8, payload []byte) *Extension` function would have been required in `rtp`.

### DIFF
--- a/track_local_static.go
+++ b/track_local_static.go
@@ -230,6 +230,15 @@ func (s *TrackLocalStaticRTP) Write(b []byte) (n int, err error) {
 	return len(b), s.writeRTP(packet)
 }
 
+// SampleRTPExtension is a RTP extension that can be added to packets generated
+// when writing Samples to a TrackLocalStaticSample.
+type SampleRTPExtension struct {
+	// ID is the negotiated extension id.
+	ID uint8
+	// Payload is the payload of the extension to add to the packet.
+	Payload []byte
+}
+
 // TrackLocalStaticSample is a TrackLocal that has a pre-set codec and accepts Samples.
 // If you wish to send a RTP Packet use TrackLocalStaticRTP.
 type TrackLocalStaticSample struct {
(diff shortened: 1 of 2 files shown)


fancycode:
Improved test coverage and rebased to current master.

Not sure about the API test failure, the new parameters to `WriteSample` are optional, so existing callers will still work as before.

JoTurk:
> Not sure about the API test failure, the new parameters to WriteSample are optional, so existing callers will still work as before.

it's still a breaking change, please refer to [Adding to a function](https://go.dev/blog/module-compatibility#adding-to-a-function)


does this https://github.com/pion/webrtc/discussions/1841 solve your issue?

fancycode:
> it's still a breaking change, please refer to [Adding to a function](https://go.dev/blog/module-compatibility#adding-to-a-function)

Right, I didn't think about this - introduced a new function.

> does this #1841 solve your issue?

Yes, copying most of the existing `TrackLocalStaticSample` code into my project obviously works (and this is what I'm doing right now), but I assumed this is something others might also need to send simulcast streams from files, which is the reason for my PR.

JoTurk:
Hello @fancycode  I think we can get this API or similar API targeted for v4.3, I added it to the milestone.
Thank you.

JoTurk:
@Sean-Der thoughts on this?

JoTurk (approved):

JoTurk:
the other option is to make this generic, so we can add other metadata in the future, it can be called WriteSamplesWithOptions for example

JoTurk:
@fancycode thank you and sorry it took a while, new APIs usually take a lot of time, sorry again

track_local_static.go
```diff
@@ -230,6 +230,15 @@ func (s *TrackLocalStaticRTP) Write(b []byte) (n int, err error) {
...
 }
 
+// SampleRTPExtension is a RTP extension that can be added to packets generated
+// when writing Samples to a TrackLocalStaticSample.
+type SampleRTPExtension struct {
+	// ID is the negotiated extension id.
+	ID uint8
+	// Payload is the payload of the extension to add to the packet.
+	Payload []byte
+}
```
JoTurk, line 240: Thinking about it before merge, this design breaks multiple PCs and one track writes, and per packet extension, i think interceptors work better here

PR #3410, opened 2026-04-11
Add multi-track ogg writing API

JoTurk:
#### Description
First of 2 or 3 PRs to rework the OGG writer, like how we did with the reader to support multiple track writing and metadata.

### DIFF
--- a/pkg/media/oggwriter/oggwriter.go
+++ b/pkg/media/oggwriter/oggwriter.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"sync"
 
 	"github.com/pion/rtp"
 	"github.com/pion/rtp/codecs"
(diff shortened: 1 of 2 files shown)


JoTurk:
@FrantaBOT can you take a look here.

pkg/media/oggwriter/oggwriter.go
```diff
@@ -17,251 +18,869 @@ import (
...
 	pageHeaderSignature                = "OggS"
+	pageHeaderSize                     = 27
+	maxOggPageSegments                 = 255
+	noGranulePosition                  = ^uint64(0)
 )
 
 var (
-	errFileNotOpened    = errors.New("file not opened")
-	errInvalidNilPacket = errors.New("invalid nil packet")
+	errFileNotOpened        = errors.New("file not opened")
```
FrantaBOT, line 38: Wouldn't it better to call it something else then file? The output doesn't need to be file

JoTurk: yeah good point, but this error was always here from the old file writer we didn't remove it in this PR, I'll add a new error for the stream writer.

JoTurk: done also fixed an edge case where we didn't call close.

FrantaBOT (approved):

PR #3412, opened 2026-04-16
Parse rtp extensions from padding only empty packet

arjunshajitech:
#### Description

While testing simulcast with chrome (Version 147.0.7727.101 (Official Build) (64-bit)) strange behavior was noticed.
Chrome was sending first packet as empty with extension properly populated,
The original implementation discarded empty packets without parsing the extensions.
This pr fixes the issue and also adds relevant test cases.


JoTurk (approved):
Good catch, thank you Arjun.

PR #3413, opened 2026-04-17
Update (current|pending)RemoteDescription SDP after addIceCandidate

fippo:
as described in
  https://w3c.github.io/webrtc-pc/#dom-peerconnection-addicecandidate

(noticed while pondering https://github.com/pion/ice/issues/912#issuecomment-4267673815)

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2070,8 +2070,15 @@ func (pc *PeerConnection) RemoteDescription() *SessionDescription {
 
 // AddICECandidate accepts an ICE candidate string and adds it
 // to the existing set of candidates.
-func (pc *PeerConnection) AddICECandidate(candidate ICECandidateInit) error {
-	remoteDesc := pc.RemoteDescription()
+func (pc *PeerConnection) AddICECandidate(candidate ICECandidateInit) error { //nolint:cyclop
+	pc.mu.Lock()
+	defer pc.mu.Unlock()
+
+	remoteDesc := pc.currentRemoteDescription
+	if pc.pendingRemoteDescription != nil {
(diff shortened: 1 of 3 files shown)


fippo:
@JoTurk 🤔 any idea what causes the test failures? They are either totally uncorrelated or I need to move the test to the go layer...

JoTurk:
@fippo there is a race in attribute.

peerconnection.go
```diff
@@ -2112,7 +2119,23 @@ func (pc *PeerConnection) AddICECandidate(candidate ICECandidateInit) error {
...
+
+	if pc.pendingRemoteDescription != nil {
+		if err := pc.pendingRemoteDescription.UpdateWithCandidate(candidate); err != nil {
+			return err
+		}
+	}
+	if pc.currentRemoteDescription != nil {
+		if err := pc.currentRemoteDescription.UpdateWithCandidate(candidate); err != nil {
+			return err
+		}
```
JoTurk, line 2135: This updates the candidates' attributes while `startRTP` is reading them. which cases the test to fail with race detected.

fippo: but why does `startRtp` look at the candidates in the SDP instead of getting them from the ICE agent? Also why does anything RTP not consider ICE a virtual socket? 🤔

JoTurk: Sorry i missed this so this is partly done for historic reasons. during early calls (we'll change this with ice v5, where we'll move more ownership to ice), also check the mux in this repo, especially at the close handling, currently we have multiple semi ownership for sockets. 
There is a plan to move most of the ownership and write to one place, also so we can do low level socket optimizations and similar stuff.

PR #3424, opened 2026-05-17
Expose SCTP assoc metadata in getStats

JoTurk:
#### Description
Expose SCTP metadata in our stats, so users can get negotiated SCTP features. currently this exposes If Interleaving is enabled, partial reliability mode, zero-checksum in send and receive.

this is based on change in SCTP https://github.com/pion/sctp/pull/467


JoTurk:
I updated the interface to match my suggested w3c/webrtc-stats interface.

RingoMar (approved):

JoTurk:
Some of the tests are failing because there is a race in tests, found when we changed to interleaving by default!

PR #3425, opened 2026-05-17
Add a setting to toggle interleaving

JoTurk:
#### Description
A setting to switch if I-DATA is enabled in SCTP, we're going to enable it by default only after 4.3, and even after this some users might want to disable it for whatever reasons.

based on https://github.com/pion/sctp/pull/466 and requires https://github.com/pion/sctp/pull/443 to get merged


JoTurk:
https://github.com/pion/sctp/pull/466#issuecomment-4503414727

PR #3426, opened 2026-05-21
Fix DC OnClose race, add default DC open handler

sirzooro:
## Summary

This PR addresses two related reliability issues with `DataChannel` close handling and unhandled incoming data channels.

---

### `DataChannel`: fire `OnClose` immediately if already closed + prevent duplicate calls

**Problem:** If `OnClose` was registered after the channel had already transitioned to `DataChannelStateClosed`, the handler would never be called.

The late-registration race for `OnClose` is rarely observed in production or even in localhost-based unit tests, due to network and scheduling latency. However, when it does occur in tests it causes flaky failures — the `OnClose` handler silently never fires — which was the direct motivation for this fix. The scenario where a remote peer closes a data channel immediately after opening it should be handled reliably regardless of timing.

**Fix:**

- Added a `closeHandlerOnce sync.Once` field to `DataChannel`.
- `OnClose()` now resets the once-guard when a new handler is registered. If the channel is already closed at registration time, the handler is fired immediately in a goroutine.
- `onClose()` now wraps the handler call in `closeHandlerOnce.Do(...)`, ensuring the handler is called at most once per registration.

### `PeerConnection`: default `OnDataChannel` handler closes undeclared channels

**Problem:** Some applications are designed so that only one side opens data channels and the other side only accepts them. Such an app typically registers `OnDataChannel` to handle expected channels, but may not consider the case where the remote peer opens an unexpected channel. Before this change, any application that never called `OnDataChannel` (or reset it to `nil`) would silently accept every incoming channel and leave it open — leaking resources.

**Fix:**

- `NewPeerConnection` now initialises `onDataChannelHandler` to a new `defaultOnDataChannelHandler`.
- `defaultOnDataChannelHandler` calls `d.Close()` on any incoming channel, logging a warning on error.
- `OnDataChannel(nil)` restores the default handler instead of setting the field to `nil`.

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -53,6 +53,7 @@ type DataChannel struct {
 	onOpenHandler       func()
 	dialHandlerOnce     sync.Once
 	onDialHandler       func()
+	closeHandlerOnce    sync.Once
 	onCloseHandler      func()
 	onBufferedAmountLow func()
 	onErrorHandler      func(error)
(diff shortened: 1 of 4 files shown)


JoTurk (approved):
I feel like we fixed this in multiple locations few times.

PR #3427, opened 2026-05-22
Fix a race in benchmark

JoTurk:
#### Description
This race happens more after we moved to interleaving, ig because the path is faster.


JoTurk:
the macos upgrade broke the macos test.

PR #3431, opened 2026-05-25
Make RID parsing more robust

sirzooro:
`getRids` panics on empty simulcast token, verify token is non-empty.


JoTurk (approved):
good catch, thank you.

PR #3432, opened 2026-05-26
Fix wasm null JS event handlers before release

alinz:
I ran into a frustrating bug when using pion/webrtc in a WASM app:
after closing a peer connection, every subsequent operation would
blow up with syscall/js: call to released function in the browser
console. Took a while to track down.

### DIFF
--- a/datachannel_js.go
+++ b/datachannel_js.go
@@ -181,18 +181,22 @@ func (d *DataChannel) Close() (err error) {
 
 	// Release any handlers as required by the syscall/js API.
 	if d.onOpenHandler != nil {
+		d.underlying.Set("onopen", js.Null())
 		d.onOpenHandler.Release()
 	}
 	if d.onCloseHandler != nil {
 		d.onCloseHandler.Release()
 	}
 	if d.onClosingHandler != nil {
+		d.underlying.Set("onclosing", js.Null())
 		d.onClosingHandler.Release()
 	}
 	if d.onMessageHandler != nil {
+		d.underlying.Set("onmessage", js.Null())
 		d.onMessageHandler.Release()
 	}
 	if d.onBufferedAmountLow != nil {
+		d.underlying.Set("onbufferedamountlow", js.Null())
 		d.onBufferedAmountLow.Release()
 	}
 	if d.onErrorHandler != nil {
--- a/datachannel_js.go
+++ b/datachannel_js.go
@@ -180,32 +180,22 @@ func (d *DataChannel) Close() (err error) {
 	d.underlying.Call("close")
 
 	// Release any handlers as required by the syscall/js API.
-	// Null out each JS property before Release() to prevent "call to released
-	// function": close() queues browser events (e.g. close, closing) that fire
-	// asynchronously; if Release() runs first, those callbacks call a Go FuncOf
-	// that is no longer in funcs[].
 	if d.onOpenHandler != nil {
-		d.underlying.Set("onopen", js.Null())
 		d.onOpenHandler.Release()
 	}
 	if d.onCloseHandler != nil {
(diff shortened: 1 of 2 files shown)


datachannel_js.go
```diff
@@ -180,22 +180,32 @@ func (d *DataChannel) Close() (err error) {
...
 		d.onOpenHandler.Release()
 	}
 	if d.onCloseHandler != nil {
+		d.underlying.Set("onclose", js.Null())
 		d.onCloseHandler.Release()
 	}
 	if d.onClosingHandler != nil {
+		d.underlying.Set("onclosing", js.Null())
 		d.onClosingHandler.Release()
 	}
```
JoTurk: Hello, clearing onClose and onError this early will prevent us from receiving the close event later. We'll need to fix this before this is merged, thank you.

JoTurk (changes requested):

JoTurk:
Seems like the CI is stuck for some reason, That's weird.

JoTurk:
It's also stuck for other PRs https://github.com/pion/mdns/pull/275 seems like a github issue...

datachannel_js.go
```diff
@@ -181,18 +181,22 @@ func (d *DataChannel) Close() (err error) {
 
 	// Release any handlers as required by the syscall/js API.
 	if d.onOpenHandler != nil {
+		d.underlying.Set("onopen", js.Null())
 		d.onOpenHandler.Release()
 	}
 	if d.onCloseHandler != nil {
 		d.onCloseHandler.Release()
 	}
```
JoTurk: I think we should defer nullifying and releasing onclose/onerror after close is triggered,.

alinz: Thanks for the feedback, It got a little bit more complicated than I thought, but please have a look at my implementation [here](https://github.com/pion/webrtc/pull/3432/commits/ffaad8c148a3465c771a944eaba0fd2d1f148852)

datachannel_js.go
```diff
@@ -177,27 +182,59 @@ func (d *DataChannel) Close() (err error) {
...
+		}
+		return js.Undefined()
+	})
+	d.underlying.Set("onclose", wrapperHandler)
+	if oldCloseHandler != nil {
+		oldCloseHandler.Release()
 	}
+	d.onCloseHandler = nil
+	d.onCloseFunc = nil
+	d.onErrorHandler = nil
```
JoTurk: I think before we merge this we just need to make sure we don't override the previous close if the user calls close twice or more for whatever reason.

Thank you.

datachannel_js.go
```diff
@@ -177,27 +186,63 @@ func (d *DataChannel) Close() (err error) {
...
 	}
-	if d.onErrorHandler != nil {
-		d.onErrorHandler.Release()
+
+	// Defer nullifying onclose and onerror until the channel has fully closed.
+	// An error event may still fire during the close handshake (between calling
+	// close() and the onclose event), and the user-provided OnClose callback
+	// must still be invoked. Install the wrapper only once so repeated Close()
+	// calls don't override the original wrapper before it fires.
+	if !d.closeWrapperInstalled {
```
JoTurk, line 216: Nit: we need to avoid installing the wrapper if already closed., but we can do this later. I think.

JoTurk (approved):
Thank you and sorry for the back and forth.

alinz:
My apologies @JoTurk , but the Lint Meta data keeps failing, and I don't have any clue how to solve this.

JoTurk:
> My apologies @JoTurk , but the Lint Meta data keeps failing, and I don't have any clue how to solve this.

Thank you for your fixes. the CI is failing because we have lint rules for commit messages, but it doesn't matter I'm going to squash merge it.

Thank you again.

PR #3438, opened 2026-06-01
fix: default audio channels to 1 when not specified in SDP

### LINKED ISSUE #3127: Default audio channels number
### Your environment.
  - Version: v4.1.1 (also a same behaviour for latest master)

### What did you do?
After updating Pion to version 4.1.1, we encountered an issue when using the G722 codec.
In PR #3021, logic was changed that codec matching is no longer based only on mimeType and fmtp parameters. Now it also checks the sample rate and number of channels.

In our case (and it's default for Chrome), sender in SDP announces `a=rtpmap:9 G722/8000`. According to [RFC 4556](https://datatracker.ietf.org/doc/html/rfc4566#section-6) it indicates 1 channel:
> For audio streams, <encoding parameters> indicates the number of audio channels.  This parameter is OPTIONAL and may be omitted if the number of channels is one, provided that no additional parameters are needed.

However, current implementations of [codecsFromMediaDescription](https://github.com/pion/webrtc/blob/v4.1.1/sdp.go#L1039) use 0 channels as default. Should we align its logic with RFC?

### What did you expect?
1 channel number after decoding SDP description for `a=rtpmap:9 G722/8000`

toller892:
## Problem

When an SDP  line omits the channel count (e.g., ), the  function defaults  to . Per RFC 4556 Section 6:

> For audio streams, \<encoding parameters\> indicates the number of audio channels. This parameter is OPTIONAL and may be omitted if the number of channels is one.

This means the default should be  for audio, not .

## Impact

After PR #3021 added channel count verification to codec matching, this causes G722 (and potentially other audio codecs that omit the channel count) to fail matching when the remote SDP uses a different payload type (e.g., Chrome's dynamic PT 9 for G722).

While  in  has a workaround that defaults 0 to 1 during comparison, the stored codec value is still incorrect and could affect other code paths that use  directly (e.g., stats reporting, interceptor configuration).

## Fix

In , after parsing the encoding parameters, default  to  for audio media types when the value is . This aligns the parsed value with the RFC specification.

Also adds a test case verifying G722 with no explicit channel count gets .

Fixes #3127

### DIFF
--- a/commit_msg.txt
+++ b/commit_msg.txt
@@ -1,0 +1,12 @@
+fix: default audio channels to 1 when not specified in SDP
+
+RFC 4556 Section 6 states that for audio streams, the encoding
+parameters indicate the number of channels, and this parameter
+is OPTIONAL and may be omitted if the number of channels is one.
+
+Previously, codecsFromMediaDescription defaulted channels to 0
+when not specified. This caused codec matching issues after PR #3021
+added channel count verification, breaking G722 and other audio
+codecs that omit the channel count from their rtpmap line.
+
+Fixes #3127
--- a/sdp.go
+++ b/sdp.go
@@ -1124,6 +1124,13 @@ func codecsFromMediaDescription(mediaDescr *sdp.MediaDescription) (out []RTPCode
 			channels = uint16(val)
 		}
 
+		// RFC 4556 Section 6: For audio streams, the encoding parameters
+		// indicate the number of channels. This parameter is OPTIONAL and
+		// may be omitted if the number of channels is one.
+		if channels == 0 && strings.EqualFold(mediaDescr.MediaName.Media, "audio") {
+			channels = 1
+		}
+
 		feedback := []RTCPFeedback{}
 		for _, raw := range codec.RTCPFeedback {
 			split := strings.Split(raw, " ")
--- a/sdp_test.go
+++ b/sdp_test.go
@@ -1426,6 +1426,23 @@ func TestCodecsFromMediaDescription(t *testing.T) {
 		})
 		assert.NoError(t, err)
 	})
+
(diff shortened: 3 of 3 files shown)


commit_msg.txt
```diff
@@ -0,0 +1,12 @@
...
+RFC 4556 Section 6 states that for audio streams, the encoding
+parameters indicate the number of channels, and this parameter
+is OPTIONAL and may be omitted if the number of channels is one.
+
+Previously, codecsFromMediaDescription defaulted channels to 0
+when not specified. This caused codec matching issues after PR #3021
+added channel count verification, breaking G722 and other audio
+codecs that omit the channel count from their rtpmap line.
+
+Fixes #3127
```
JoTurk, line 12: Can you please omit this file.

sdp.go
```diff
@@ -1124,6 +1124,13 @@ func codecsFromMediaDescription(mediaDescr *sdp.MediaDescription) (out []RTPCode
 			channels = uint16(val)
 		}
 
+		// RFC 4556 Section 6: For audio streams, the encoding parameters
+		// indicate the number of channels. This parameter is OPTIONAL and
+		// may be omitted if the number of channels is one.
+		if channels == 0 && strings.EqualFold(mediaDescr.MediaName.Media, "audio") {
+			channels = 1
+		}
```
JoTurk, line 1132: this shouldn't apply to opus i think:

> The media subtype ("opus") goes in SDP "a=rtpmap" as the encoding
      name.  The RTP clock rate in "a=rtpmap" MUST be 48000, and the
      number of channels MUST be 2.
> https://www.rfc-editor.org/rfc/rfc7587.html

JoTurk (changes requested):

PR #3442, opened 2026-06-03
rm redundant lock

solos:
rm redundant lock

#### Description

The d.ReadyState uses atomic, so the RLock is not necessary.

#### Reference issue
None


Sean-Der:
Nice catch @solos !

JoTurk (approved):
Looking at blame this wasn't updated after moving to `ReadyState()` :)

PR #3445, opened 2026-06-05
Optimize SampleBuilder and reset cross-sample state on Flush

solos:
#### Description
Improve SampleBuilder performance and make Flush() a clean boundary for a new RTP stream.
Performance:
- Replace lastSampleTimestamp *uint32 with uint32 + hasLastSampleTimestamp to avoid a heap allocation on every emitted sample
- Pre-size sample payload buffers from the total RTP payload length
- Compute Sample.Duration with integer arithmetic instead of float64 Flush semantics:
- After purging buffered packets, Flush() now clears lastSampleTimestamp, droppedPackets, and paddingPackets
- Prevents post-flush packets from being misclassified as padding from a prior stream when timestamps collide
- Prevents PrevDroppedPackets on the next sample from including drops that occurred during the flush of leftover buffer data Samples built during Flush() still capture PrevDroppedPackets at build time; only state for packets received after Flush() is reset. Tests:
- Add TestSampleBuilder_FlushResetsState
- Add TestSampleBuilder_ValidSampleSameTimestampAfterFlush

#### Reference issue
None

### DIFF
--- a/pkg/media/samplebuilder/samplebuilder.go
+++ b/pkg/media/samplebuilder/samplebuilder.go
@@ -318,9 +318,16 @@ func (s *SampleBuilder) buildSample(purgingBuffers bool) *media.Sample {
 	}
 	samples := afterTimestamp - sampleTimestamp
 
+	var duration time.Duration
+	if s.sampleRate == 0 {
+		duration = 0
+	} else {
+		duration = time.Duration(samples) * time.Second / time.Duration(s.sampleRate)
+	}
+
 	sample := &media.Sample{
 		Data:               data,
-		Duration:           time.Duration(samples) * time.Second / time.Duration(s.sampleRate),
+		Duration:           duration,
 		PacketTimestamp:    sampleTimestamp,
 		PrevDroppedPackets: s.droppedPackets,
 		Metadata:           metadata,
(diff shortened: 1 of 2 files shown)


pkg/media/samplebuilder/samplebuilder.go
```diff
@@ -310,7 +320,7 @@ func (s *SampleBuilder) buildSample(purgingBuffers bool) *media.Sample {
 
 	sample := &media.Sample{
 		Data:               data,
-		Duration:           time.Duration((float64(samples)/float64(s.sampleRate))*secondToNanoseconds) * time.Nanosecond,
+		Duration:           time.Duration(samples) * time.Second / time.Duration(s.sampleRate),
```
JoTurk: The old behavior returned invalid value (+Inf) if the sample-rate is 0.00f, this will just panic, both aren't ideal, but if there is a setup where users can pass sample rate to applications, this can be used to trigger panic/DoS.

Maybe we can check and return an error early if the user created a samplebuilder with 0 sample rate?

JoTurk (commented):
Change looks good to me, but we'll have to manually test it because samplebuilder has many moving parts.

pkg/media/samplebuilder/samplebuilder.go
solos: Yes, you are right. If we return an error,  the caller will need to change too. What about set the Duration to 0.

JoTurk:
Hello, hope you didn't close this pr because it took long to merge, we just need to make sure it doesn't bring any regressions because sample builder has a lot of moving parts I'll try to test it manualy tonight.
Thank you.

PR #3447, opened 2026-06-10
Fix bundle-only data channel startup

zshang-oai:
Pion currently treats every zero-port application `m=` section as rejected. This violates [RFC 8843, Section 6](https://www.rfc-editor.org/rfc/rfc8843.html#section-6), which permits a bundled `m=` section to use port zero with `a=bundle-only`.

Start SCTP when the application section has either a nonzero port or `a=bundle-only`. Zero-port sections without `a=bundle-only` remain rejected.

### DIFF
--- a/datachannel_test.go
+++ b/datachannel_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/pion/transport/v4/test"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // expectedLabel represents the label of the data channel we are trying to test.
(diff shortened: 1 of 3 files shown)


datachannel_test.go
```diff
@@ -277,6 +278,53 @@ func TestDataChannel_Open(t *testing.T) {
...
+	require.NoError(t, err)
+	answerGatheringComplete := GatheringCompletePromise(answerPC)
+	require.NoError(t, answerPC.SetLocalDescription(answer))
+	<-answerGatheringComplete
+	require.NoError(t, offerPC.SetRemoteDescription(*answerPC.LocalDescription()))
+
+	require.Eventually(t, func() bool {
+		return dataChannel.ReadyState() == DataChannelStateOpen
+	}, 5*time.Second, 10*time.Millisecond)
+}
```
JoTurk: datachannel_test.go is shared between go and wasm/js tests (this is why the test breaks) should move this to datachannel_go_test.go.

JoTurk (approved):
Thank you so much.

PR #3453, opened 2026-06-19
Parse pt= parameter in RID attributes

asnyatkov:
Parse the optional pt= parameter from a=rid lines as defined in RFC 8851 Section 4, storing the allowed payload types in the simulcastRid struct.

RFC 8851 defines the RID attribute syntax as:

  a=rid:<id> <direction> [pt=<fmt list>;<restrictions>]

The pt= parameter lists payload types allowed in the RTP stream. For example:

  a=rid:1 send pt=96,97;max-width=1280;max-height=720
  a=rid:2 send pt=96,97;max-width=640;max-height=360

Changes:
- Parse pt= from RID attributes in getRids()
- Add payloadTypes field to simulcastRid struct
- Write pt= back in SDP answer in addTransceiverSDP(), filtering to only include payload types that match available codecs
- Comprehensive test coverage for parsing and round-trip

### DIFF
--- a/sdp.go
+++ b/sdp.go
@@ -282,13 +282,64 @@ func trackDetailsToRTPReceiveParameters(trackDetails *trackDetails) RTPReceivePa
 	return RTPReceiveParameters{Encodings: encodings}
 }
 
+// parseRidPayloadTypes parses the pt= parameter from a=rid parameters string
+// per RFC 8851 Section 4. The params string is the space-separated portion after
+// "<id> <direction>", e.g. "pt=96,97;max-width=1280".
+func parseRidPayloadTypes(params string) []uint8 {
+	var payloadTypes []uint8
+	for param := range strings.SplitSeq(params, ";") {
+		ptStr, ok := strings.CutPrefix(param, "pt=")
+		if !ok {
+			continue
+		}
+		for pt := range strings.SplitSeq(ptStr, ",") {
+			if ptVal, err := strconv.ParseUint(pt, 10, 8); err == nil {
+				payloadTypes = append(payloadTypes, uint8(ptVal)) //nolint:gosec
+			}
+		}
+	}
+
+	return payloadTypes
+}
+
+// buildRidPTParam maps offer payload types to answer payload types via semantic
+// codec matching per RFC 8851 Section 6.3, and returns the " pt=<types>" string
+// for the answer's a=rid line. Returns empty string if no types match.
+func buildRidPTParam(offerPTs []uint8, offerCodecs, answerCodecs []RTPCodecParameters) string {
+	if len(offerPTs) == 0 {
+		return ""
+	}
+
+	validPTs := []string{}
+	for _, pt := range offerPTs {
+		offerCodec := findCodecByPayload(offerCodecs, PayloadType(pt))
+		if offerCodec == nil {
+			continue
+		}
+		answerCodec, matchType := codecParametersFuzzySearch(*offerCodec, answerCodecs)
+		if matchType == codecMatchNone {
+			continue
+		}
+		validPTs = append(validPTs, strconv.FormatUint(uint64(answerCodec.PayloadType), 10))
+	}
+	if len(validPTs) == 0 {
+		return ""
+	}
+
+	return " pt=" + strings.Join(validPTs, ",")
+}
+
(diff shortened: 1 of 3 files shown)


sdp.go
```diff
@@ -282,13 +282,56 @@ func trackDetailsToRTPReceiveParameters(trackDetails *trackDetails) RTPReceivePa
...
+func buildRidPTParam(payloadTypes []uint8, availablePTs map[uint64]bool) string {
+	if len(payloadTypes) == 0 {
+		return ""
+	}
+	validPTs := []string{}
+	for _, pt := range payloadTypes {
+		if availablePTs[uint64(pt)] {
+			validPTs = append(validPTs, strconv.FormatUint(uint64(pt), 10))
+		}
+	}
```
JoTurk: `availablePTs` comes from the answer and payloadTypes here comes from the offer, right? if yes we should match PT values and return the values from the answer. and not compare PT values between the answer and the offer. We have some helpers for this, because we do it in multiple places.

>    Note: In the case that the answerer uses different PT values to
   represent a codec than the offerer did, the "a=rid" values in the
   answer use the PT values that are present in its answer.
> https://datatracker.ietf.org/doc/rfc8851/

JoTurk (changes requested):

sdp.go
asnyatkov: The axiom of WebRTC development: no matter how many RFCs you read, there's always another one, essentially creating infinite set of RFCs.

I'll fix the logic, but there's very confusing item 6.4 in that RFC ("intersections") which is not covered by any existing tests. Can we address that later?

sdp.go
```diff
@@ -282,13 +282,64 @@ func trackDetailsToRTPReceiveParameters(trackDetails *trackDetails) RTPReceivePa
...
+	for _, pt := range offerPTs {
+		offerCodec := findCodecByPayload(offerCodecs, PayloadType(pt))
+		if offerCodec == nil {
+			continue
+		}
+		answerCodec, matchType := codecParametersFuzzySearch(*offerCodec, answerCodecs)
+		if matchType == codecMatchNone {
+			continue
+		}
+		validPTs = append(validPTs, strconv.FormatUint(uint64(answerCodec.PayloadType), 10))
```
JoTurk, line 323: this should just match the payload-type mapped from the offer, please check setCodecPreferencesFromRemoteDescription, and we don't have to implement full 6.4.

sdp_test.go
```diff
@@ -1650,3 +1650,287 @@ func TestSctpInit(t *testing.T) {
...
+		// PT 200 should be filtered out
+		require.Equal(t, "high", parsedRids[0].id)
+		require.ElementsMatch(t, []uint8{96, 98}, parsedRids[0].payloadTypes)
+
+		require.Equal(t, "medium", parsedRids[1].id)
+		require.ElementsMatch(t, []uint8{96, 98}, parsedRids[1].payloadTypes)
+
+		// PT 200 not available, so no PT in output
+		require.Equal(t, "low", parsedRids[2].id)
+		require.Empty(t, parsedRids[2].payloadTypes)
```
JoTurk, line 1924: I think this test asserts wrong behavior. low should be removed here?

> If the "a=rid" line contains a "pt=", the answerer is allowed to discard one or more media formats from a given "a=rid" line. If the answerer chooses to discard all the media formats from an "a=rid" line, the answerer MUST discard the entire "a=rid" line. If the offer did not contain a "pt=" for a given "a=rid" line, then the answer MUST NOT contain a "pt=" in the corresponding line. 

RID without pt is not equivalent.

> The optional "pt=<fmt-list>" lists one or more PT values that can be used in the associated RTP stream. If the "a=rid" attribute contains no "pt", then any of the PT values specified in the corresponding "m=" line may be used. 

I think we should remove the RID line and its references in the answer, both quotes are from rfc8851.

JoTurk (commented):
I'm so sorry @asnyatkov I lost track of this and I got busy with the DTLS 1.3 project, please ping me if you have any questions or want another review.

sdp.go
JoTurk: > 6.4 in that RFC ("intersections") which is not covered by any existing tests. Can we address that later?

yeah sure, this should be a followup.

PR #3461, opened 2026-06-28
Add support for non-zero channel mapping

JoTurk:
#### Description
This PR adds support for writing non-zero channel mapping. While this isn't super useful for WebRTC media track Opus streams, but useful for some Opus over sctp usage, This completes the feature set supported by the reader.


JoTurk:
After this is merged, I'll add a simple example that records an entire room into a single Ogg file, with each participant stored as a separate track and identified using OpusTags :)

FrantaBOT (approved):

PR #3476, opened 2026-07-14
Set the default MTU to 1191

JoTurk:
#### Description
follow up on https://github.com/pion/sctp/pull/476 since we overwrite SCTP mtu here too, not sure if this is safe or correct for RTP.


fippo (approved):

JoTurk:
I think i'll add a new constant for SCTP MTU and won't touch RTP since it's not affected by this issue since it's not affected by the DTLS buffer.

PR #3482, opened 2026-07-23
Defer ICE restart until SetLocalDescription

maum-meliora:
`CreateOffer(&OfferOptions{ICERestart: true})` restarts the ICE transport immediately, before the offer is ever applied. The restart synchronously drops the selected candidate pair and moves the connection back to checking, so merely creating a restart offer tears down an active connection whether or not the offer is ever used.

The iceRestart definition states "Applying the generated description will restart ICE" ([webrtc-pc §4.2.6](https://www.w3.org/TR/webrtc/#dom-rtcofferoptions-icerestart)), so creating the offer should not.

## Fix

Generate the new ICE credentials and hold them pending, applying them to the transport only when `SetLocalDescription` commits the offer. While pending, the offer advertises the new credentials but the live transport keeps the old ones, so an unused restart offer no longer drops the connection.

Deferring the restart also defers its gathering cycle to `SetLocalDescription`. A `GatheringCompletePromise` created in between would otherwise resolve against the previous cycle's Complete state; it now waits for the new cycle, so non-trickle callers need no change.

## Behavior note

The restart now takes effect at `SetLocalDescription` rather than at `CreateOffer`. Standard `CreateOffer` → `SetLocalDescription` flows and `GatheringCompletePromise` are unaffected. The narrow observable change: code that created a restart offer and relied on the transport restarting before (or without) applying it now sees the restart only once `SetLocalDescription` commits the offer.

An offerer ICE restart now proposes freshly generated credentials rather than any static ufrag/pwd configured via `SetICECredentials`, so the restart's credentials always differ from the current ones (webrtc-pc §4.2.6). The answerer's restart path is unchanged here.

## Testing

- `CreateOffer(ICERestart)` no longer mutates the live transport, and the proposed credentials apply only once `SetLocalDescription` commits the offer.
- A `GatheringCompletePromise` created before `SetLocalDescription` resolves against the new gathering cycle.
- Concurrent `CreateOffer(ICERestart)` / `CreateOffer(nil)` calls, serialized by `pc.mu`, always propose new credentials.
- `go test -race` across the ICE-restart and signaling suites; `golangci-lint` clean.

### DIFF
--- a/gathering_complete_promise.go
+++ b/gathering_complete_promise.go
@@ -28,8 +28,7 @@ func GatheringCompletePromise(pc *PeerConnection) (gatherComplete <-chan struct{
 	pc.setGatherCompleteHandler(func() { done() })
 	// While a restart is pending, the current Complete belongs to the previous cycle, so let the
 	// handler resolve once SetLocalDescription starts the new one instead of resolving now.
-	pending, _ := pc.pendingICERestartCredentials.Load().(*iceCredentials)
-	if pending == nil && pc.ICEGatheringState() == ICEGatheringStateComplete {
+	if !pc.hasPendingICERestart() && pc.ICEGatheringState() == ICEGatheringStateComplete {
 		done()
 	}
 
--- a/gathering_complete_promise.go
+++ b/gathering_complete_promise.go
@@ -14,6 +14,11 @@ import (
 // It is better to not use this function, and instead trickle candidates.
 // If you use this function you will see longer connection startup times.
 // When the call is connected you will see no impact however.
+//
+// A promise created after CreateOffer(ICERestart) waits for the restart's gathering cycle,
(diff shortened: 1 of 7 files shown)


JoTurk (changes requested):
Thank you for the fix.
The problem is this will introduces a new state machine for ICE and already cause problems with racing for some states. 
ICE generation should be owned by pion/ice, credentials, candidates, gathering state, end-of-candidates, and completion are properties of one ICE generation and need to be done in a single generation (we also have a pending PR in ice to add generation extension). 
I think this all can be done on ICEGatherer with supporting primitives in pion/ice.
PC shouldn't retain or generate ICE generations or conflict with ICE state and role.

maum-meliora:
Thanks for the feedback. I opened pion/ice#945 to move the
generation state there, as you suggested. The webrtc part can be
picked up again once the ice side is ready. Closing this PR.

PR #3483, opened 2026-07-24
Send nil interceptor attributes on RTP writes

kcaffrey:
Currently an interceptor.Attributes map is allocated for each RTP write.  However, interceptors only strictly need to allocate an attributes map when wanting to set a value. None of the default interceptors currently allocate a new map when passed nil on the write path. The gcc.SendSideBWE interceptor, when TWCC is negotiated, does allocate attributes for every outbound packet. Thus, by sending nil instead of allocating a new map, we can save one allocation per `WriteRTP()`
 
 All interceptors in pion/interceptor are already safe with nil attributes on the write path, and pion/interceptor#431 made `GetRTPHeader()` and `GetRTCPPackets()` safe to call on nil attributes.  Users with custom interceptors should ensure that attributes is always checked for nil before calling `Set()`.
 
 The benchmark added in this package shows a reduction of allocations for a write+read round trip from 8 allocs/op to 7 allocs/op.
 
 Running the benchmark with pion/ice#943 (not yet merged) and pion/srtp#378 (merged but not yet released) results in 1 alloc/op. The remaining allocation is in the [replay detector](https://github.com/pion/transport/blob/main/replaydetector/replaydetector.go#L56-L68). I plan to make a parallel API for the replay detector to have `Check(seq)` and `Accept(seq)` to avoid the closure allocation. Combined with this PR, once srtp is updated to include that, there will be zero allocations in the hot path of reading and writing an RTP packet.


JoTurk (changes requested):
We shouldn't add a settingsengine for this, we should just patch attributes methods so they allocate/reconstruct when called with nil instead of panic and after we tag interceptors it should be safe to just change it to nil here with no setting engine.

kcaffrey:
I first looked at that, but unless I'm missing something, that would require changing [`Set()`](https://github.com/pion/interceptor/blob/a01a5831c7ec0a53e24df159ccb87db67f5f5135/attributes.go#L31-L33) to take a pointer receiver instead:

```go
func (a *Attributes) Set(key, val any) {
	if *a == nil {
		*a = make(Attributes, 1)
	}
	(*a)[key] = val
}
```

I _think_ this breaks back-compat guarantees, which is why I went with the `SettingsEngine` approach instead. Admittedly I'm not quite sure, but I think there are some edge cases where changing to a pointer receiver could cause consumer code to no longer compile.

-----
Edit: Here is an example of such a breakage. This would compile today, but not with changing Set to a pointer receiver:

```go
type Foo interface {
	Get(key any) any
	Set(key any, value any)
}

a := interceptor.Attributes{}
var b Foo = a
```
-----

Let me know if you would still like me to go with the approach to change `Set()` to a pointer receiver or whether you think the setting engine approach is fine. Thanks!

kcaffrey:
Also, in theory, I think a consumer can write an interceptor with code like this that compiles today:

```go
foo := interceptor.Attributes{}
foo["bar"] = 7
```

So merely patching `Set()` wouldn't avoid a theoretical back-compat breakage if we switched to writing with `nil` attributes. Perhaps this is unlikely enough that it's worth it anyway?

JoTurk:
@kcaffrey but adding a setting engine doesn't solve any of this and it doesn't make the methods op. it just makes this optimization useful for a few users.

I think we should:
1. make sure that all our interceptors reconstruct if att is nil (i think we already do) https://github.com/pion/interceptor/blob/main/pkg/nack/generator_interceptor.go#L117 
2. for cache methods like GetRTCPPackets / GetRTPHeader we should just let it cache miss (it's better than panic).
3. for pure set i'm not sure if we should keep it panic so the user know that they need to construct attributes if nil.

we should add pointer receiver methods that reconstruct / allocate and plan to change this to a pointer in the next interceptors major release (should be really soon) to avoid having to do all of this.

JoTurk:
> foo := interceptor.Attributes{}
foo["bar"] = 7

I also think this is fine, if all of our built-in interceptors check for nil before write, a custom interceptor can pass nil down the chain. what shouldn't panic is mainly GetRTCPPackets and GetRTPHeader, and maybe our Set.

kcaffrey:
@JoTurk I can make a PR to interceptor to address your points, but would like to clarify a few things first.

> make sure that all our interceptors reconstruct if att is nil

I manually audited all of the interceptors in pion/interceptor and all of the first party interceptors currently allocate when needed (and pass on) a new attributes map when passed nil. This is verified for at least default interceptors by a test in this PR. So on that front, we are fine. One concern I have is enforcing this going forwards. I can add tests for all the current interceptors to verify they don't panic with nil attributes, but it doesn't prevent a new interceptor from being added that does panic on nil. I think this is probably fine, but I want to make sure this is an acceptable risk for you, and check whether you think the tests are worth it.

> for cache methods like GetRTCPPackets / GetRTPHeader we should just let it cache miss (it's better than panic).

This is reasonable; if attributes is nil we can unmarshal and skip the `Set()`.

> for pure set i'm not sure if we should keep it panic so the user know that they need to construct attributes if nil

I think panicking on nil is safest, as there is no way to adhere to the API contract with a nil attributes and the current API.

> we should add pointer receiver methods that reconstruct / allocate

Do you mean add parallel methods? Go doesn't allow the same method name with both a pointer receiver and value receiver. So would you want something like `func (a *Attributes) SetOrInit(key, value any)`? The problem I see with this is that if we want to change `Set()` to a pointer receiver in a later version, `SetOrInit()` lives in an awkward spot. I suppose we could remove it in the next major release of interceptor?

---------

Even with all of the above, though, it seems like changing `WriteRTP()` to pass nil attributes would be a breakage change for users, since `Set()` would still panic when called on nil attributes, and consumers of pion/webrtc could be using custom interceptors that call `Set()` on the write path without checking for nil (which would currently work since we never pass nil).

> adding a setting engine doesn't solve any of this and it doesn't make the methods op. it just makes this optimization useful for a few users.

One small point of disagreement I have here is that adding a setting engine option _does_ avoid back-compat breakages. All existing code will continue to work as-is, and if a user opts into that setting and their code breaks, that is no longer a back-compat breakage. I do agree that it means the optimization is only useful for a few users. I tried to get clever and have the option default to on when the user is using the built in interceptors but I didn't see a clean way to do that.

This last point does actually raise a new idea that is worth considering. What if we added a new optional method for interceptors (perhaps via an optional secondary interface), something like `HandlesNilAttributes() bool`. Then pion/webrtc, instead of having a setting engine option, could look at all registered interceptors and if all declare that they handle nil attributes, then we send `nil` in `WriteRTP()`. This would enable the optimization for the vast majority of users.

So I see a few options going forward:
1. Make `GetRTCPPackets`/`GetRTPHeader`nil-safe, always send `nil` attributes inside of `WriteRTP()`, and accept the back-compat breakage for custom interceptors that call `Set()` or make raw map writes without first checking for nil.
2. Add a setting engine option for whether to send `nil` attributes on write, keeping full backwards compatibility, but making the optimization only useful for a small subset of users.
3. Have an opt in interface for interceptors to declare they are nil-safe, and send `nil` attributes on the write path if-and-only-if the interceptor chain for writes is fully nil-safe. Any user using the default interceptors gets the full optimization, but any user with custom interceptors would not without opting in first.

I now have a slight preference for the third option. How would you like me to proceed? (And thanks for the feedback!)

JoTurk:
@kcaffrey 
> Do you mean add parallel methods? 

We can add private methods and switch our interceptors to them. Until we move to the next major release.

> One small point of disagreement I have here is that adding a setting engine option does avoid back-compat breakages. All existing code will continue to work as-is

very few users create custom interceptors and i think all the users that create custom interceptors are either aware to handle nil attribute or fix it if it panics, making this an op-tin will help very a few users. and just add more noise to the already noisy setting engine.
Most users use default interceptors, custom interceptors are currently in a weird position where normal users don't define them because they aren't super useful and power users skip them and use custom buffer to have more control (what we do at my work).

Built-in interceptors already pass nill down the chain on some errors https://github.com/pion/interceptor/blob/main/pkg/nack/generator_interceptor.go#L114  and we handle nil before read, so it's not really unexpected or a new behavior.

I think it's just a better trade off to do it without a setting engine.

kcaffrey:
pion/interceptor#431 makes the cache methods nil-safe. After that merges and gets released, I'll update this PR to send `nil` attributes on the write path with no setting engine option.

kcaffrey:
@JoTurk I have updated this PR according to our discussion. PTAL, thanks!

JoTurk (approved):
Thank you so much :)

PR #3489, opened 2026-08-05
Fix #2867

### LINKED ISSUE #2867: SCTP transport start and stop should be safe to call concurrently
There's a chance that a SetRemoteDescription can start the SCTP transport while also stopping it which in some cases can leak data channels and some goroutines. Start ought to check the state of itself prior to creating any resources.

mojl:
#### Description
This fixes #2867. `Stop()` now changes the state to closed always, and `Start()` checks the state twice, once at the start and once before allocating an association, and aborting in case `Stop()` was called during the handshake.

#### Testing
The only way I could test this locally is by sleeping inside `Start()`.
Add a hook to `sctptransport.go`:
```Go
var sctpStartDelay = func(*SCTPTransport) {}
```
And call it in line 150:
```Go
sctpStartDelay(r)
```

And test:
```Go
func TestSCTPTransportStartStopRace(t *testing.T) {
	const startDelay = 3 * time.Second

	offerPC, answerPC, err := newPair()
	require.NoError(t, err)

	defer closePairNow(t, offerPC, answerPC)

	answerSCTP := answerPC.SCTP()
	inStart := make(chan struct{})

	sctpStartDelay = func(r *SCTPTransport) {
		if r != answerSCTP {
			return
		}

		close(inStart)
		time.Sleep(startDelay)
	}

	_, err = offerPC.CreateDataChannel("race", nil)
	require.NoError(t, err)

	require.NoError(t, signalPair(offerPC, answerPC))

	<-inStart

	require.NoError(t, answerPC.Close())

	time.Sleep(startDelay)


	// wasn't the main issue, but this passes too now
	// assert.Equal(t, SCTPTransportStateClosed, answerSCTP.State())
	assert.Nil(t, answerSCTP.association())
}
```

### DIFF
--- a/sctptransport.go
+++ b/sctptransport.go
@@ -121,10 +121,14 @@ func (r *SCTPTransport) GetCapabilities() SCTPCapabilities {
 //
 //nolint:cyclop
 func (r *SCTPTransport) Start(capabilities SCTPCapabilities) error {
-	if r.isStarted {
+	r.lock.Lock()
+	if r.isStarted || r.state == SCTPTransportStateClosed {
+		r.lock.Unlock()
+
 		return nil
 	}
 	r.isStarted = true
+	r.lock.Unlock()
 
 	maxMessageSize := capabilities.MaxMessageSize
 	if maxMessageSize == 0 {
@@ -149,6 +153,12 @@ func (r *SCTPTransport) Start(capabilities SCTPCapabilities) error {
 	}
 
 	r.lock.Lock()
+	if r.state == SCTPTransportStateClosed {
+		r.lock.Unlock()
+		sctpAssociation.Abort("")
+
+		return nil
+	}
 	r.sctpAssociation = sctpAssociation
 	r.state = SCTPTransportStateConnected
 	dataChannels := append([]*DataChannel{}, r.dataChannels...)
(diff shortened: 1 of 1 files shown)


sctptransport.go
```diff
@@ -149,6 +153,12 @@ func (r *SCTPTransport) Start(capabilities SCTPCapabilities) error {
 	}
 
 	r.lock.Lock()
+	if r.state == SCTPTransportStateClosed {
+		r.lock.Unlock()
+		sctpAssociation.Abort("")
+
+		return nil
+	}
```
JoTurk, line 161: This won't close a handshake until we merge and switch to https://github.com/pion/sctp/pull/473  sorry for that, I'll try to get it merged.

JoTurk (commented):
lgtm thank you, we just need to merge https://github.com/pion/sctp/pull/473 first, to avoid creating a race.

sctptransport.go
mojl: Oh, I wasn't aware, thanks for the review :)

JoTurk: @mojl https://github.com/pion/sctp/pull/473 got merged, we can update and merge this now, do you want to update it or do you want me to do that?

mojl: I would love to do that, I will be working on it 👍

JoTurk: @mojl thank you

PR #3491, opened 2026-08-07
Reuse DataChannel IDs safely within negotiated SCTP limits

idy:
## What changed

- reuse DataChannel IDs only after SCTP reports reset completion
- keep overlapping and detached DataChannel generations associated with the correct stream ID
- classify locally opened channels even when they are created after SCTP startup
- enforce the negotiated SCTP stream limit for generated and explicit DataChannel IDs
- roll back transport membership and ID reservations when opening a channel fails
- expose a separate SCTP handshake retransmission setting

## Why

DataChannel IDs are SCTP stream IDs. Reusing an ID before both reset directions complete can associate delayed reset or acknowledgement traffic with a new channel generation. Never releasing an ID avoids that race but eventually exhausts the negotiated stream range.

This change tracks local generations through reset completion so IDs can be reused without allowing traffic from an older generation to affect the new channel. It also rejects explicit IDs outside the negotiated range and cleans up failed opens.

## Dependencies and relationship to existing work

This PR depends on pion/sctp#485. The temporary `replace` directive pins the reviewed fork commit containing that SCTP work; it should be replaced with an upstream release after the SCTP change lands.

This is an alternative implementation that covers the goals of #3405 and adds handling for overlapping generations, detached and dynamically created channels, explicit-ID bounds, and failure rollback.

This PR is not intended to invalidate the work in #3405. If maintainers prefer to continue with #3405, please feel free to close this PR and reuse or adapt any tests and fixes that are useful.

## Validation

- `go test ./...`
- affected DataChannel race tests repeated successfully under `go test -race`
- `go vet ./...`
- `go mod tidy -diff`
- `staticcheck -checks 'all,-SA6002,-ST1016' .`

The full race suite also reaches two existing TrackLocal timestamp and sequence-number timing assertions. They fail identically on a clean `upstream/main` checkout and do not produce race-detector reports; the DataChannel-focused race suite passes.

### DIFF
--- a/datachannel.go
+++ b/datachannel.go
@@ -183,10 +183,10 @@ func (d *DataChannel) open(sctpTransport *SCTPTransport) error { //nolint:cyclop
 
 		return err
 	}
-	d.sctpTransport.registerLocalDataChannelStream(stream)
+	generation := d.sctpTransport.registerLocalDataChannelGeneration(*d.id)
 	dc, err := datachannel.Client(stream, cfg)
 	if err != nil {
-		d.sctpTransport.unregisterLocalDataChannelStream(stream)
+		d.sctpTransport.unregisterLocalDataChannelGeneration(*d.id, generation)
 		d.mu.Unlock()
 
 		return err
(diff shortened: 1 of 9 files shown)


idy:
Superseded by #3534, which carries the same change from a dedicated branch rebased on current `main`.

PR #3504, opened 2026-08-21
Skip DTLS HelloVerifyRequest by default

fippo:
The HelloVerifyRequest cookie exchange only validates the remote address, which the ICE connectivity checks have already done in the context of WebRTC so disable this by default.

### DIFF
--- a/dtlstransport.go
+++ b/dtlstransport.go
@@ -517,10 +517,18 @@ func (t *DTLSTransport) toDTLSServerOptions(sharedOpts []dtls.Option) []dtls.Ser
 		clientAuth = *t.api.settingEngine.dtls.clientAuth
 	}
 
+	// The HelloVerifyRequest cookie exchange only validates the remote
+	// address, which the ICE connectivity checks have already done, so it
+	// is skipped unless explicitly requested via the SettingEngine.
+	skipHelloVerify := true
+	if skip := t.api.settingEngine.dtls.insecureSkipHelloVerify; skip != nil {
+		skipHelloVerify = *skip
+	}
+
 	serverOpts = append(serverOpts,
 		dtls.WithClientAuth(clientAuth),
 		dtls.WithClientCAs(t.api.settingEngine.dtls.clientCAs),
-		dtls.WithInsecureSkipVerifyHello(t.api.settingEngine.dtls.insecureSkipHelloVerify),
+		dtls.WithInsecureSkipVerifyHello(skipHelloVerify),
 	)
 
 	if t.api.settingEngine.dtls.serverHelloMessageHook != nil {
--- a/settingengine.go
+++ b/settingengine.go
@@ -67,7 +67,7 @@ type SettingEngine struct {
 		SRTCP *uint
 	}
 	dtls struct {
-		insecureSkipHelloVerify       bool
+		insecureSkipHelloVerify       *bool
 		disableInsecureSkipVerify     bool
 		retransmissionInterval        time.Duration
 		ellipticCurves                []dtlsElliptic.Curve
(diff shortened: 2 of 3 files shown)


JoTurk (commented):
Wouldn't this be dangerous to do for lite agents?

fippo:
pion/webrtc isn't a lite agent? I don't think it should be removed in dtls

JoTurk:
Also I remember from the DTLS 1.3 spec,  but i wasn't convinced that we should skip it by default because of ice-lite agents and because the RFC still recommends performing cookie exchange by default. and it just permits servers to  skip it.
 
> DTLS servers SHOULD perform a cookie exchange whenever a new handshake is being performed. If the server is being operated in an environment where amplification is not a problem, e.g., where ICE [[RFC8445](https://www.rfc-editor.org/info/rfc9147/#RFC8445)] has been used to establish bidirectional connectivity, the server MAY be configured not to perform a cookie exchange. The default SHOULD be that the exchange is performed,

JoTurk:
> pion/webrtc isn't a lite agent? I don't think it should be removed in dtls

I think many of our users use pion for SFUs as a lite-agent. and switching this by default can be bad I think.

fippo:
don't think there is an attack for lite either. It will check the remote binding request validity too.
Now there is a theoretical replay attack but executing it still requires a DTLS handshake that matches the fingerprint. That is a weakness of lite in general.

JoTurk:
One concern i have is that i think this will make it possible to make an ice-lite agent be used as a DoS amplifier against arbitrary addresses with just a spoofed packet? the exchange is like 5x-10x? with Pion's default server flights? it might be dumb I'm not sure.

And I'm not sure I feel comfortable about going against the spec recommendation for the the default. All the power users can already flip this option if they see fit. maybe I'm just being paranoid @Sean-Der  @theodorsm what do you think?

fippo:
RFC 6347 has (unsurprisingly) this:
>  "DTLS servers SHOULD perform a cookie exchange whenever a new handshake is being performed. 
> If the server is being operated in an environment where amplification is not a problem, the server MAY be 
> configured not to perform a cookie exchange. The default SHOULD be that the exchange is performed, however."

which is exactly what pion/webrtc does now. DTLS 1.3 got rid of it. Nobody else I know of (mediasoup/janus) uses it either.

JoTurk:
This changed a bit in DTLS 1.3 i think you missed my earlier comment https://github.com/pion/webrtc/pull/3504#issuecomment-5367036708 I referenced rfc-9147

I think "where ICE [[RFC8445](https://www.rfc-editor.org/info/rfc9147/#RFC8445)] has been used to establish bidirectional connectivity" implies full ice agent, and the keyword is still "may", and pion still uses the recommendation default. 
For ice-lite amplification is a problem, because we'll have roughly 5-10x amplification factor with Pion's default flights.

fippo:
amplification attacks are currently done through TURN usually (nobody has found time to include padding in those binding requests).

The amplification attack requires spoofing and valid ice credentials. And hardly effective, this would be paced at DTLS timing?

JoTurk:
Okay I agree about the practical risk being small, unless I'm missing something. Maybe if we skip cookie exchange only if LITE isn't enabled, then we won't not be going against the spec? but even then we'll be deviating from the spec's recommended behavior without a compelling reason, in a minor / patch release :)

theodorsm:
> Maybe if we skip cookie exchange only if LITE isn't enabled

@JoTurk I second this approach. I am also unsure about the versioning, there you are more on-top of things.

Firefox does not use the cookie exchange for WebRTC using DTLS 1.2 (from what I can observe in my captures).

PR #3507, opened 2026-08-27
Fix panic in RTX repair reader for malformed packets

GerardGao:
## Summary

Fix a panic in the RTX repair reader (`maybeStartRepairStreamReader`) when the repair interceptor returns malformed data:

- A zero-length read (`i == 0`) made the padding length read `b[i-1]` index `b[-1]`, panicking the process.
- A truncated packet with the extension bit set caused the extension length field to be parsed from stale bytes beyond the packet.
- The header length was tracked in a `uint16`, so a large extension length field wrapped around and let malformed packets bypass the probe check and be forwarded as bogus RTX packets.

## Changes

- Skip zero-length reads and packets truncated before the extension header, returning the buffer to the pool.
- Track the header length as an `int` to avoid `uint16` wraparound.

## Verification

- New test `TestRTPReceiverRepairReaderMalformedPacketNoPanic` feeds a zero-length read, a truncated packet and a valid RTX packet through the repair reader and asserts the valid packet is still forwarded correctly. It panics on `main` and passes with this change.
- Full repo test suite with `-race` passes.
- golangci-lint clean for the changed lines.

### DIFF
--- a/rtpreceiver.go
+++ b/rtpreceiver.go
@@ -667,7 +667,7 @@ func (r *RTPReceiver) receiveForRtxInternal(
 }
 
 // maybeStartRepairStreamReader starts repair processing when needed. The caller must hold r.mu.
-func (r *RTPReceiver) maybeStartRepairStreamReader(track *trackStreams) { //nolint:cyclop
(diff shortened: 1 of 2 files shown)


JoTurk (approved):

JoTurk:
@GerardGao Thank you. I tagged in v4.2.19.

PR #3508, opened 2026-08-27
Fix data race in TrackRemote.read

GerardGao:
## Problem

`TrackRemote.read` consumed `t.peekedPackets` while holding only a read lock:

```go
t.mu.RLock()
if len(t.peekedPackets) != 0 {
    peekedPkt = t.peekedPackets[0]
    t.peekedPackets = t.peekedPackets[1:] // write: advances the slice header
}
t.mu.RUnlock()
```

Popping is a write, so a read lock is the wrong boundary: `RLock` holders do
not exclude each other, and two readers can take the same head element while
the header only advances once.

## Scope

This is not a regression, and it is not reachable through the `peek` path.
As pointed out in review, `peek` completes before `OnTrack` fires and does not
overlap user reads, and simulcast does not by itself produce concurrent reads
on a single `TrackRemote`. `Read` is exported API, so the guard still matters
for callers that read one track from more than one goroutine.

## Fix

Take the full lock while consuming the slice. No test is added.

## Verification

- `go build ./...` and `go vet ./...` clean
- `go test -race .` passes
- `golangci-lint` v2.10.1 (the version CI pins): 0 issues
- both changed lines are already covered by the existing suite, so patch
  coverage stays at 100%

### DIFF
--- a/track_remote.go
+++ b/track_remote.go
@@ -130,14 +130,14 @@ func (t *TrackRemote) Read(b []byte) (n int, attributes interceptor.Attributes, 
 }
 
 func (t *TrackRemote) read(b []byte) (n int, attributes interceptor.Attributes, err error) {
-	t.mu.RLock()
+	t.mu.Lock()
 	receiver := t.receiver
 	var peekedPkt *peekedPacket
 	if len(t.peekedPackets) != 0 {
 		peekedPkt = t.peekedPackets[0]
 		t.peekedPackets = t.peekedPackets[1:]
 	}
-	t.mu.RUnlock()
+	t.mu.Unlock()
 
 	if receiver.haveClosed() {
 		return 0, nil, io.EOF


fallenmi (approved):
AI disclosure: I used OpenAI Codex to assist this review; I verified the results locally against exact head `27a876af32a2cddabf523fb0b0eb59c0360fa70b`.

Popping `peekedPackets` mutates the slice header, so the full lock is the right boundary here. In an independent 16-reader / 8,192-packet oracle, exact base `37a88bd2` reports races at `track_remote.go:136-138` and consumes packets more than once, while this head passes 20/20 runs under the race detector. Replaying the submitted regression gives the same base-RED/head-GREEN result, with the head passing 10/10.

I also tested GitHub's synthetic merge `77b98534` over current main `278b7f9c`: the regression passes 10/10, the TrackRemote-focused race suite passes, and `go vet ./...` is clean. All 21 live checks are green.

Looks good to me.

JoTurk (changes requested):
Nice correctness, but I don't think this was a regression. and please remove the test file it seems to be deliberately triggering the race and the test doesn't seem correct (we don't add regression test files anyways). changing the lock from a read to a write lock is a good patch tho (I don't think it should be tested).

Simulcast does not mean multiple concurrent reads/writes on one TrackRemote ,I also don't see a realistic normal application call pattern where the same TrackRemote would be read concurrently by multiple goroutines.

GerardGao:
@JoTurk thanks for the review. Removed the test file as requested, keeping only the lock change (`RLock` to full `Lock`) in `track_remote.go`.

Agreed on the concurrent-read pattern: the realistic concern was the simulcast probe `peek` goroutine racing an application read, but I'm fine with the lock-only patch.

track_remote.go
```diff
@@ -130,14 +130,18 @@ func (t *TrackRemote) Read(b []byte) (n int, attributes interceptor.Attributes,
 }
 
 func (t *TrackRemote) read(b []byte) (n int, attributes interceptor.Attributes, err error) {
-	t.mu.RLock()
+	// A full lock is required here: consuming peekedPackets mutates the
+	// slice header, which is not safe under a shared (read) lock when other
+	// goroutines read or write the slice concurrently (e.g. the simulcast
+	// probing peek or other concurrent readers).
```
JoTurk: Please remove this comment it's not accurate and not needed, peak happened before ontrack and it didn't overlap with user reads. Simulcast didn't translate into multiple reads too.

JoTurk (changes requested):

GerardGao:
@JoTurk removed the comment in d71a1950. The diff is now only the `RLock` → `Lock` change in `track_remote.go`, with no test file and no comment.

```
 func (t *TrackRemote) read(...) {
-	t.mu.RLock()
+	t.mu.Lock()
 	receiver := t.receiver
 	var peekedPkt *peekedPacket
 	if len(t.peekedPackets) != 0 {
 		peekedPkt = t.peekedPackets[0]
 		t.peekedPackets = t.peekedPackets[1:]
 	}
-	t.mu.RUnlock()
+	t.mu.Unlock()
```

Verified on this head: `go build ./...`, `go vet ./...` and `go test -race .` all pass.

GerardGao:
@JoTurk rebased onto latest `main` and squashed into a single commit. The diff is unchanged — only `RLock` → `Lock` in `track_remote.go`, no test file, no comment.

I also rewrote the PR description: it still carried the simulcast/`peek` framing you rejected, plus a "Reproducer" section describing the test that's now removed. You were right on both counts — I walked the call path and `peek` does complete before `OnTrack`, so it doesn't overlap user reads. The description now just states that popping under a read lock is the wrong boundary for exported `Read`, and notes this isn't a regression.

The `lint / Metadata` failure was unrelated to the code — a commit-message body line over 72 characters — and is fixed by the squash. On the new head: `go build ./...`, `go vet ./...` and `go test -race .` all pass, `golangci-lint v2.10.1` (the version CI pins) reports 0 issues, and patch coverage stays at 100% without the removed test.

track_remote.go
GerardGao: Removed — the comment is gone from the current head, and the diff is now just the `RLock` → `Lock` change.

On the substance: you're right. I traced the path and `peek` completes before `OnTrack` fires, so it doesn't overlap user reads, and simulcast alone doesn't produce concurrent reads on a single `TrackRemote`. I've dropped that framing from the PR description as well, and kept the rationale to the narrow point that popping mutates the slice header, so a read lock is the wrong boundary for exported `Read`.

JoTurk (approved):

PR #3510, opened 2026-08-28
Release ICE connection on closed transport

GerardGao:
## Problem

`ICETransport.StartContext` has an asymmetric cleanup path. When the
transport is stopped while `Dial`/`Accept` is in flight and the connection
subsequently succeeds, StartContext returns `errICETransportClosed`
without releasing the resources it just acquired:

```go
if t.State() == ICETransportStateClosed {
    return errICETransportClosed // iceConn leaked, ctxCancel dangling
}
```

The error path above it properly clears `ctxCancel`; this branch does not.
Today `Stop` happens to close the gatherer in most interleavings, but the
cleanup must not depend on that: if the ICE agent's close semantics change,
the connection established by the successful Dial would leak.

## Fix

Mirror the error path: clear the cancel function and close the gatherer
(which releases the ICE agent and its sockets). Closing the gatherer is
idempotent and serialized by its lock, so racing with a concurrent `Stop`
is safe.

## Reproducer

A regression test deterministically drives StartContext into this branch
by holding the transport lock while the ICE connection is established, so
the Dial goroutine can only observe the Closed state:

- Before the fix: fails — `ctxCancel` is left dangling and the connection
  is not released.
- After the fix: passes under `-race` (5 consecutive runs), the cancel
  function is cleared, and the connection's goroutines wind down.

A temporary stress test (20 iterations of racing Stop against
StartContext, repeated 3x) showed no double-close panic and no deadlock.

## Verification

- `go test -race .` — all tests pass
- `golangci-lint run` (v2.10.1, same as CI) — 0 issues
- `go vet` — clean
- Patch coverage: 100% (codecov target is 70%)

### DIFF
--- a/icetransport.go
+++ b/icetransport.go
@@ -191,16 +191,22 @@ func (t *ICETransport) StartContext(
 
 	if t.State() == ICETransportStateClosed {
 		// The transport was stopped while Dial/Accept was in flight.
-		// Release the resources acquired by the successful connection
-		// attempt, mirroring the error path above. Closing the gatherer
-		// releases the underlying ICE agent and its sockets; it is
-		// idempotent and serialized by the gatherer lock, so racing
-		// with a concurrent Stop is safe.
+		// Release the cancel function, mirroring the error path above.
 		ctxCancel()
 		t.ctxCancel = nil
-		if t.gatherer != nil {
-			_ = t.gatherer.Close()
+
+		// Closing the gatherer releases the underlying ICE agent and its
+		// sockets. It must happen outside the transport lock: the close
+		// synchronously invokes the user's OnStateChange callback, which
+		// may re-enter the transport and would deadlock on t.lock. The
+		// close is idempotent and serialized by the gatherer lock, so
+		// racing with a concurrent Stop is safe.
+		gatherer := t.gatherer
+		t.lock.Unlock()
+		if gatherer != nil {
+			_ = gatherer.Close()
 		}
+		t.lock.Lock()
 
 		return errICETransportClosed
 	}
--- a/icetransport.go
+++ b/icetransport.go
@@ -190,6 +190,18 @@ func (t *ICETransport) StartContext(
 	}
 
 	if t.State() == ICETransportStateClosed {
+		// The transport was stopped while Dial/Accept was in flight.
+		// Release the resources acquired by the successful connection
+		// attempt, mirroring the error path above. Closing the gatherer
+		// releases the underlying ICE agent and its sockets; it is
+		// idempotent and serialized by the gatherer lock, so racing
+		// with a concurrent Stop is safe.
+		ctxCancel()
+		t.ctxCancel = nil
+		if t.gatherer != nil {
+			_ = t.gatherer.Close()
+		}
+
 		return errICETransportClosed
 	}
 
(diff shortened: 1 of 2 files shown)


fallenmi (changes requested):
The new cleanup calls `t.gatherer.Close()` while `ICETransport.StartContext` still holds `t.lock`. `ICEGatherer.close()` synchronously calls `setState(ICEGathererStateClosed)`, and `setState` synchronously invokes the public `OnStateChange` handler. If that handler re-enters this transport through a method such as `transport.Start(...)`, it blocks forever on the same `t.lock`, so the outer `StartContext` can never return.

I reproduced this deterministically on exact head `1dc7cc7fd7ad85c03f1e732a719719423b9f47a9` by extending the submitted closed-dial choreography with an `OnStateChange(Closed)` handler that calls `transport.Start`. The exact logical base passes the oracle 3/3 when gatherer cleanup occurs after the outer `StartContext` releases the lock. This head times out with the stack `StartContext:202 -> ICEGatherer.close/setState:570/646 -> OnStateChange -> StartContext:104`, and GitHub's synthetic merge `7c5bd7ec` over current `main` does the same.

Please perform the gatherer cleanup outside `t.lock` (reacquiring it as needed for the existing deferred-unlock structure), or otherwise avoid invoking gatherer/user callbacks under the transport lock, and add a callback-reentrancy regression.

The submitted regression itself passes 20/20 under the race detector, the ICE-focused suite passes five repetitions, the full exact-head `go test -race ./...` suite passes, and all 21 live checks are green; those paths do not exercise this callback re-entry.

AI disclosure: I used OpenAI Codex to help inspect the locking and callback paths, construct and run the exact base/head and synthetic-merge oracles, execute the focused/full test suites, and draft this review. I manually verified the code paths, captured stacks, exact commits, live checks, and conclusion.

GerardGao:
Thanks for the detailed review — confirmed the deadlock: `ICEGatherer.setState` synchronously invokes the `OnStateChange` handler, so closing the gatherer under `t.lock` deadlocks when the handler re-enters the transport.

Fixed by releasing the transport lock around the gatherer close (reacquiring it for the existing deferred unlock), and added a callback-reentrancy regression test: an `OnStateChange(Closed)` handler that calls `transport.Start` now completes, where it previously deadlocked (reproduced on the old head with the stack `StartContext → ICEGatherer.close/setState → OnStateChange → StartContext`).

The original closed-dial choreography is now shared by both tests via a `runClosedDuringDial` helper. Local verification: both tests pass under `-race` (3 runs), the ICE-focused suite passes, full `go test -race ./...` passes, and `golangci-lint` is clean.

fallenmi (approved):
Thanks — this addresses my prior deadlock finding. On exact head `17dcb5b7a45363d943a06b3f98d23922e61214b2`, the gatherer is captured under `t.lock`, the lock is released before `gatherer.Close()`, and it is reacquired for the existing deferred unlock. The added `OnStateChange(Closed)` regression re-enters through `transport.Start`, directly covering the callback path that previously deadlocked.

I rechecked the one-commit old-to-new delta and a source-order oracle: the reviewed head `1dc7cc7f` calls `Close()` while the transport lock is held, while this head and GitHub's current synthetic merge `b144040b24ec8153a5c2aebe568db95213445803` both have the `Unlock -> Close -> Lock` sequence. The two changed-file blobs are identical between the head and that merge. All 21 exact-head check runs across 11 check suites are successful, including all 9 Actions workflows. The PR is mergeable but currently behind `main` by three unrelated commits; GitHub's current merge is clean.

Disclosure: I used OpenAI Codex to compare the exact revisions, trace the lock/callback ordering, verify the regression coverage and current merge blobs, and inspect every live CI layer. I verified the source paths, SHAs, interactions, policy, and conclusion before submission.

PR #2563, opened 2023-09-07
Generate answer to match group bundle in offer

cnderrauber:
#### Description
User report negotiation failed in https://github.com/livekit/livekit/issues/1850.

When the client sends offer to stop a transceiver (inactive), it removes the media section from group bundle, but pion still includes the media section in the answer, making negotiation failed by `Failed to set remote answer sdp: A BUNDLE group in answer contains a MID='3' that was not in the offered group`.

PR #3522, opened 2026-09-04
Accept non-bundled offers in generated answer

winklemad:
`generateMatchedSDP` discards the ok value of `Attribute(sdp.AttrKeyGroup)`, so an offer with no `a=group:BUNDLE` becomes an empty bundle group rather than no group at all. No mid then matches, and every media section in the answer is rejected with port 0. The application gets a well-formed answer and no error, but the session is dead.

Repro is a plain single-m-line audio offer with mid, ice, fingerprint and setup but no `a=group`: the answer comes back as `m=audio 0 UDP/TLS/RTP/SAVPF 111`. With `a=group:BUNDLE 0` added to the same offer it is `m=audio 9 ...`.

This came in with #2563, which needed the group to drop mids the offerer had removed; the absent-group case was not considered. #2950 notes that non-bundled offers are valid input. Browsers always bundle, so it mostly shows up with GStreamer's webrtcbin, whose `bundle-policy` defaults to `none`, and with SIP gateways.

There is one design decision in here that you may want to make differently. When the offer has no BUNDLE group the answer now keeps the first media section and rejects the rest, because only one ICE and DTLS transport exists (candidates are only added to section 0), and it emits no group of its own, since RFC 8843 section 7.3 says an answerer may only include a BUNDLE group the offerer requested. Simply passing nil would accept every section and put a BUNDLE group in the answer that the offer never asked for, which is what happened before #2563.

While in there: the group was read with a cutset `TrimLeft(value, "BUNDLE")` over the first `a=group` attribute only, so an offer carrying `a=group:LS 0 1` produced `a=group:BUNDLE 0 1` in the answer. Groups are now matched on their semantics field, since RFC 5888 allows several of them.

Tests cover the answer for an offer with no group, an offer with an LS group, and populateSDP with two sections and no remote group. All three fail without the change, and the existing bundle tests are untouched.

### DIFF
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -3091,7 +3091,7 @@ func (pc *PeerConnection) generateMatchedSDP(
 	pc.sctpTransport.lock.Lock()
 	defer pc.sctpTransport.lock.Unlock()
 
-	var bundleGroup *string
+	var bundleGroup *remoteBundleGroup
 	// If we are offering also include unmatched local transceivers
 	if includeUnmatched { //nolint:nestif
 		if !detectedPlanB {
@@ -3119,9 +3119,22 @@ func (pc *PeerConnection) generateMatchedSDP(
 			}
 		}
 	} else if remoteDescription != nil {
-		groupValue, _ := remoteDescription.parsed.Attribute(sdp.AttrKeyGroup)
-		groupValue = strings.TrimLeft(groupValue, "BUNDLE")
-		bundleGroup = &groupValue
+		bundleGroup = &remoteBundleGroup{}
+
+		// a description can carry several groups with different semantics
+		// (RFC 5888), so look for the BUNDLE one instead of taking the first.
+		for _, attribute := range remoteDescription.parsed.Attributes {
+			if attribute.Key != sdp.AttrKeyGroup {
+				continue
+			}
+
+			if semantics, mids, _ := strings.Cut(attribute.Value, " "); semantics == "BUNDLE" {
+				bundleGroup.mids = mids
+				bundleGroup.present = true
+
+				break
+			}
+		}
 	}
 
 	if pc.configuration.SDPSemantics == SDPSemanticsUnifiedPlanWithFallback && detectedPlanB {
--- a/peerconnection.go
+++ b/peerconnection.go
@@ -2959,6 +2959,7 @@ func (pc *PeerConnection) generateUnmatchedSDP(
 		mediaSections,
 		pc.ICEGatheringState(),
 		nil,
+		pc.configuration.BundlePolicy,
 		pc.api.settingEngine.getSCTPMaxMessageSize(),
 		false,
 	)
@@ -3160,6 +3161,7 @@ func (pc *PeerConnection) generateMatchedSDP(
 		mediaSections,
 		pc.ICEGatheringState(),
 		bundleGroup,
+		pc.configuration.BundlePolicy,
 		pc.api.settingEngine.getSCTPMaxMessageSize(),
 		ignoreRidPauseForRecv,
 	)
--- a/peerconnection_test.go
+++ b/peerconnection_test.go
@@ -968,3 +968,77 @@ func TestICETrickleCapabilityString(t *testing.T) {
 		assert.Equal(t, tt.expected, tt.value.String())
 	}
 }
+
+func TestPeerConnection_AnswerNonBundledOffer(t *testing.T) {
+	lim := test.TimeOut(time.Second * 10)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	const offerHead = "v=0\r\n" +
+		"o=- 0 0 IN IP4 127.0.0.1\r\n" +
+		"s=-\r\n" +
+		"t=0 0\r\n"
+
+	const audioSection = "m=audio 9 UDP/TLS/RTP/SAVPF 111\r\n" +
+		"c=IN IP4 0.0.0.0\r\n" +
+		"a=ice-ufrag:abcd\r\n" +
+		"a=ice-pwd:abcdefghijklmnopqrstuvwx\r\n" +
+		"a=fingerprint:sha-256 " +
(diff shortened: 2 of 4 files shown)


JoTurk (changes requested):
Can you explain what interoperability are we actually trying to support? Pion doesn't support non-bundle. I assume this is intended to target single m-line non-bundles? if that's the case can we communicate this and / or reject non-bundled multi-m lines SDP?

winklemad:
@JoTurk good questions — let me lay out the target and what the PR actually does.

**What breaks today:** an offer that arrives with no `a=group:BUNDLE` line at all. GStreamer's `webrtcbin` defaults to `bundle-policy=none` and emits exactly that, and some SIP↔WebRTC gateways do too. Since the group-matching added in 2225386c (Sep 2023), `generateMatchedSDP` reads the group with `Attribute(group)` and treats a missing group as an empty-string BUNDLE, so the first section's mid never matches it and **every** m-line comes back with port 0 — valid SDP, but the whole session is silently dead.

**Scope — yes, single transport.** Pion has one bundled transport, so this does not try to support multiple independent transports. When there is no BUNDLE group it keeps only the first non-data section and rejects every later section with port 0 (the `matchBundleGroup != nil && !matchBundleGroup.present` branch in `sdp.go`, which has a comment to that effect). So:

- single-m-line non-bundle offer → that m-line is answered, session works;
- multi-m-line non-bundle offer → first m-line answered, the rest rejected with port 0.

That last case is the "reject the extra m-lines" behaviour you're describing — a multi-m-line non-bundle offer gets a single-transport answer rather than a broken multi-transport one, and the port-0 rejection is the standard way we tell the offerer we only took the first.

If you'd prefer to make the limitation more explicit — e.g. hard-reject a non-bundled multi-m-line offer in `SetRemoteDescription` rather than soft-rejecting the extra lines, or expand the comment — I'm happy to do it whichever way you want. Just let me know.

JoTurk:
Sounds good, this implementation should respect bundle policy https://github.com/pion/webrtc/blob/f06c094949c8efc53b9a2fd4c6df1b26b3f9e940/bundlepolicy.go#L15-L35

winklemad:
Thanks @JoTurk. Digging into it: for a non-bundle-aware remote the policy semantics differ — `max-bundle` keeps one media track, `balanced` keeps one per media type, `max-compat` keeps all — but the latter two are defined as "on separate transports," which pion doesn't have (single ICE/DTLS transport). So on pion the only *working* answer to a non-bundled offer is a single m-line regardless of policy; keeping more than one without a BUNDLE group would just produce another broken session.

Given that, two ways to "respect the policy" cleanly:

1. **Keep the single-m-line behavior under all policies** (what the PR does now) as a graceful degradation — this is exactly `max-bundle`, and it's a strict improvement for `balanced`/`max-compat` too, since today all of them answer a non-bundled offer with every m-line at port 0 (dead session). This also fixes the default (`balanced`).
2. **Only accept the non-bundled offer under `max-bundle`**, and leave `balanced`/`max-compat` rejecting it until pion can gather separate transports.

I'd lean toward (1) since it fixes the common default and never emits a worse answer than today, but it's your call — happy to gate it on `BundlePolicyMaxBundle` (option 2) if you'd rather keep the door open for real per-transport support later. Which do you prefer and I'll push it right away.

JoTurk:
we should do 2, If a user wants a max-bundle behavior, they should opt-in, it will be unexpected behavior for pion to change the default and break the PC spec, and if we break the spec for `balanced` and `max-compat` we will not be able to break this behavior in the future if we actually started supporting other transport options.

winklemad:
Done — went with option 2. The keep-first-m-line handling for a non-bundled offer is now gated on `BundlePolicyMaxBundle`; `balanced` and `max-compat` fall through to the existing path unchanged, so nothing about the default behavior or the spec changes, and a real separate-transport implementation can extend them later without a compat break.

`populateSDP` takes the policy through to that decision, and I added unit coverage for both sides (max-bundle keeps the first section + rejects the rest; balanced leaves the spec behavior intact). Ready for another look.

JoTurk (changes requested):
Other than rejecting the m-lines we need to make sure to stop the transceivers, it should be part of the negotiation state and not just the output SDP. this will not stop pion's startRTPSenders or the internal receiver which can break non-bundled peers anyways.

winklemad:
You're right — the port-0 in the answer is only the output SDP; `startRTPSenders` still fires for the rejected sections and the receiver stays up, so it has to be part of the negotiation state.

My plan: in `generateMatchedSDP`, when the max-bundle / non-bundled path rejects the non-first sections, call `Stop()` on their transceivers before `startRTPSenders`, so the sender and the internal receiver are both torn down and the rejected m-line follows from the stopped transceiver rather than a raw port-0 override in `populateSDP`.

One alignment question before I wire it in: there's already partial stopped/stopping-transceiver handling in `generateMatchedSDP` (some of it commented out). Would you rather I (a) stop the rejected transceivers and let the existing stopped→rejected-m-line machinery emit the port 0 (dropping my explicit `populateSDP` override), or (b) keep the SDP-level rejection and additionally stop the transceivers? I'd lean toward (a) since it keeps a single source of truth, but I'll match whichever fits the model you have in mind.

winklemad:
Still keen on your call between (a) and (b) above, and one useful data point from CI in the meantime: the WASM job (which runs against the browser's native WebRTC) rejects this case outright — `Failed to set remote offer sdp: max-bundle is used but no bundle group found` — so the browser enforces the max-bundle rejection itself. That lines up with your point that the rejection needs to live in negotiation state, not just the output SDP.

I've held off finalizing because the implementation and its test differ between **(a)** stop the rejected transceivers and let the existing stopped→rejected-m-line machinery emit port 0 (dropping the explicit `populateSDP` override), vs **(b)** keep the SDP-level rejection *and* stop the transceivers. I lean (a). Once you confirm the direction I'll wire it up and green CI (the WASM test needs a skip, like the other browser-divergent cases). Thanks!

PR #3524, opened 2026-09-07
Fire OnSignalingStateChange after the description is applied

atirna:
#### Description

`OnSignalingStateChange` handlers now run after the description has been fully applied, instead of in the middle of `SetLocalDescription`/`SetRemoteDescription`.

Before, the event was fired from `setDescription`, right after `signalingState` was updated but before the rest of the description application ran, assigning transceiver mids, updating the media engine, configuring receivers and starting transports. Since the handler is started in its own goroutine, a handler that immediately calls `CreateAnswer` could race the rest of `SetRemoteDescription` and hit:

```
cannot find transceiver with mid: "0"
```

That is the sequence from #3370: register an `OnSignalingStateChange` handler, and on `SignalingStateHaveRemoteOffer` call `CreateAnswer` + `SetLocalDescription` inside it (with `SetRemoteDescription` already applied). The error is intermittent only because the handler goroutine usually loses the race; a `time.Sleep` in the handler hides it.

Per the W3C spec, the signaling state transition is part of applying the description, and `SignalingStateHaveRemoteOffer` means "a remote description, of type offer, has been successfully applied", so firing the event mid-application exposes an inconsistent state.

The fix moves the dispatch to the two public entry points: `setDescription` now returns the next state, and each caller fires the event through `defer` after the description has been fully applied. Delivery semantics are otherwise unchanged (handler still runs in its own goroutine).

#### Reference issue
Fixes #3370


atirna:
shortened the commit subject for the metadata check. source diff unchanged; the signaling-state regression still passes under -race.

PR #3533, opened 2026-09-11
Do not block DataChannel accept on a stuck OPEN

idy:
#### Description

`SCTPTransport.acceptDataChannels` handles incoming streams one at a time: each iteration calls `datachannel.Accept`, which runs `datachannel.Server` and blocks on `stream.ReadSCTP` with no deadline until the DCEP DATA_CHANNEL_OPEN arrives. If one stream's OPEN never becomes readable, the loop stops and no further DataChannel is accepted. ICE and existing channels keep working, so the connection looks healthy but can no longer open channels.

This can happen under a burst of incoming channels:

1. The receive window is `maxReceiveBufferSize` minus the unread bytes in every stream's reassembly queue.
2. The peer opens many channels and sends data right after each OPEN (which RFC 8832 allows).
3. Streams waiting behind the blocked accept have no reader yet, so their data keeps the window full.
4. Once the window is full, `acceptPayloadData` drops incoming DATA. The stream is still created and queued on `acceptCh` before the drop. If the dropped chunk is a stream's OPEN, the accept loop waits on that stream.
5. The retransmitted OPEN can only be accepted once the window drains, and the window only drains when the streams queued behind the stuck accept are read. Neither can happen, so the loop waits forever.

Separately, an `io.EOF` from a single stream's OPEN read (the peer reset that stream) was treated as the association running out of streams, and the accept loop exited.

#### Changes

- `acceptDataChannels` now only calls `AcceptStream` and starts a goroutine per new stream. It exits only when `AcceptStream` fails.
- Each goroutine reads the OPEN with a read deadline, then clears the deadline. On any failure (timeout, `io.EOF`, invalid OPEN) it closes just that stream, so the peer sees a reset.
- The deadline defaults to 10s and can be changed with the new `SettingEngine.SetSCTPDataChannelOpenTimeout`.

Concurrency is what breaks the deadlock: streams behind a stuck one get their DataChannel and reader, the window drains, and the retransmitted OPEN is accepted. The deadline is a backstop for an OPEN that never arrives.

Behaviour changes:
- `OnDataChannel` can fire in a different order from the order the OPENs arrived.
- A malformed OPEN or a failed ACK write now closes only that stream. Before, it fired `OnError`/`OnClose` and stopped accepting.

#### Tests

`sctptransport_accept_test.go` connects two raw SCTP associations over `net.Pipe`. A wrapper on the client's connection strips every non-first DATA/I-DATA fragment on one stream, so the server accepts that stream but never gets a complete OPEN.

- `TestSCTPTransportAcceptDataChannelsStuckOpen`: later channels are accepted while the stuck stream is pending, the stuck stream is reset after the timeout, and accepting still works afterwards.
- `TestSCTPTransportAcceptDataChannelsStreamEOF`: the peer resets the stuck stream, the server closes only that stream, and later channels are still accepted.

Both tests fail on current `main`: in the first, later channels are never accepted; in the second, the server never closes the stuck stream. Both pass with this change under `-race`.


idy:
Superseded by #3535, which carries the same change from a dedicated branch rebased on current `main`.

PR #3550, opened 2026-09-16
Fix data race and leaked goroutines in test waits

maum-meliora:
## Problem

Several tests wait for a callback by parking a goroutine on a `sync.WaitGroup`
and selecting on a timeout. `TestPeerConnection_GetStats` then reassigns
`dcWait` for a second wait while the goroutine parked by the first one is still
on it, which `-race` reports on `main`:

```
WARNING: DATA RACE
Write at 0x... by goroutine ...:
  github.com/pion/webrtc/v4.TestPeerConnection_GetStats()
      stats_go_test.go:1553
Previous write at 0x... by goroutine ...:
  github.com/pion/webrtc/v4.waitWithTimeout.func1()
      stats_go_test.go:1169
```

The shape also leaks. On timeout the parked goroutine stays blocked in
`wg.Wait()` after the test returns, so the next test fails its
`test.CheckRoutines` startup check and one timing failure becomes several
unrelated ones. `TestPlanBMediaExchange` has no timeout at all and hangs until
`test.TimeOut(30s)` takes the run down with it.

## Fix

Count the callbacks down into a channel closed on the last one and select on it
directly, so nothing is parked. `untilConnectionState` is built on the same
helper and returns `<-chan struct{}`, so its call sites become `<-ch` instead
of `.Wait()`; that conversion is mechanical across 12 files. Test-only, no
production change.

## Verification

- Removing a test's signaling call so the awaited callback never arrives: on
  `main` the following test fails `CheckRoutines` and `TestPlanBMediaExchange`
  panics at 30s; with this change each test fails on its own timeout and the
  following one passes.
- `go test ./...` passes and `-race` no longer reproduces the race above.
- `GOARCH=386` and `GOOS=js GOARCH=wasm` build; golangci-lint clean.


JoTurk (approved):
Thank you

PR #3551, opened 2026-09-16
Fix duplicate compatible codec payload selection

zshang-oai:
## Summary

Fix duplicate payload selection when an offer contains multiple compatible codec formats. For example, an H265 offer with PT96 (`level-id=93;profile-id=1`) and PT97 (`level-id=186;profile-id=1`) can currently receive an answer with `96 96` and conflicting format parameters on PT96. The corresponding RTX payload can also be duplicated.

The selection loop introduced in #3208 chooses the first compatible local codec, but removes the last FMTP-compatible codec from the remaining search space. Those can be different entries. The selected codec remains available and can be used twice. The H265 level-compatible matching added in #3506 exposes this; that matching behavior is retained.

## Changes

- Remove the exact selected payload from the remaining codec list.
- Prefer an available compatible same-payload entry during each selection, when the initially selected canonical payload also occurs in the current media section.
- Preserve existing canonical remapping from earlier media sections, exact-match precedence, partial-match fallback, and RTX associations.
- Add literal-SDP H265/RTX regressions and direct selection/remapping coverage in the existing transceiver test file. Existing codec-order tests remain unchanged.

The payload-identity preference is deliberately conditional and per-selection, not a global assignment guarantee. Mixed-overlap offers can still be remapped to distinct compatible canonical payloads. It does not force offer and answer payload numbers to match when the established cross-media mapping requires otherwise. Related work in #3498 and #3453 is complementary; neither currently fixes this selection/removal mismatch.

## Validation

- Regression tests fail against upstream `65a0eef1` before the production change and pass with it.
- New regressions and existing codec-order tests: `go test -race -count=10 -cpu=1,4` with the focused test-name filter.
- Expanded Transceiver/Codec/MediaEngine/ReplaceTrack tests: `go test -race -count=3 -cpu=1,4` with the relevant test-name filter.
- All packages pass `go test -race -shuffle=on -count=1 -timeout=10m -skip '^TestMulticastDNS' ./...` on macOS/arm64 with Go 1.24.0.
- Full tracked-package lint with CI-pinned golangci-lint v2.10.1, all four Pion metadata checks, and `git diff --check` pass.

The unfiltered suite times out in `TestMulticastDNSHostNameConnection` on this host. The same test also times out with the original upstream production code restored through a Go source overlay; mDNS is not claimed as passing coverage.


rtptransceiver_h265_test.go
```diff
@@ -0,0 +1,167 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
```
Sean-Der: @zshang-oai can you put this in existing test file

zshang-oai:
@Awuqing, would you be willing to review this follow-up? It preserves the H265 compatibility matching from #3506 and fixes duplicate payload selection when multiple compatible H265 formats are offered.

rtptransceiver_h265_test.go
JoTurk: Maybe we should add a lint rule for this somehow. This always happen with AI PRs

zshang-oai: Done in 590aed27: moved both regressions into the existing rtptransceiver_test.go, without changing their test bodies.

rtptransceiver.go
```diff
@@ -119,6 +122,24 @@ func (t *RTPTransceiver) setCodecPreferencesFromRemoteDescription(media *sdp.Med
 				remoteCodec,
 				leftCodecs,
 			)
+			// When equivalent payloads are offered together (for example H265
+			// levels), preserve their identities instead of using one twice. If
+			// the canonical payload is absent from this media section, retain
+			// the existing remapping to codecs negotiated by earlier sections.
+			if matchType == codecMatchExact && matchCodec.PayloadType != remoteCodec.PayloadType &&
+				offeredPayloads[matchCodec.PayloadType] {
```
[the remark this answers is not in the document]
zshang-oai: Thanks, I reproduced that mapping: local [96, 97] with offer [96, 112] produces [97, 96]. That shows a limit of the per-selection preference, but both selected payloads remain distinct and compatible, and each retains its offered FMTP; it does not reproduce the duplicate-payload bug.

The existing tests require canonical remapping, including 125 to 106 across media sections, and the focused table also preserves the singleton 97 to 96 case. An unconditional same-payload-first pass would change that behavior. A conditional reservation phase would be a broader assignment-policy change.

I have clarified the PR description that this is an available, per-selection preference rather than a global identity guarantee. I propose keeping the algorithm scoped to removing the selected codec while preserving existing remapping behavior.

JoTurk (approved):
Looks good from my side, but should wait for others I think

rtptransceiver_h265_test.go
sirzooro: > Maybe we should add a lint rule for this somehow. This always happen with AI PRs

We can also add instructions for agents, e.g. for [Gitub Copilot](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions). I also found that there is generic [AGENTS.md](https://agents.md/) standard. This may not resolve this issue fully, but at least reduce it.

rtptransceiver.go
```diff
@@ -182,6 +172,35 @@ func (t *RTPTransceiver) setCodecPreferencesFromRemoteDescription(media *sdp.Med
...
+		return matchCodec, matchType
+	}
+
+	for _, codec := range localCodecs {
+		if codec.PayloadType != remoteCodec.PayloadType {
+			continue
+		}
+		_, identityMatch := codecParametersFuzzySearch(remoteCodec, []RTPCodecParameters{codec})
+		if identityMatch == codecMatchExact {
+			return codec, matchType
```
boks1971: I think the `matchType` here is `codecMatchExact`, but should it explicitly return `codecMatchExact` for readability?

boks1971 (approved):
Thank you for addressing this.

LGTM!, but would be good to get more 👀 as this is a tricky bit and remember fighting it when I worked on #3208

Awuqing:
It seems the co-author is Codex? I think this PR needs human review to ensure the situation is properly handled and avoid unknown issues.

Awuqing:
> It seems the co-author is Codex? I think this PR needs human review to ensure the situation is properly handled and avoid unknown issues.

> @Awuqing, would you be willing to review this follow-up? It preserves the H265 compatibility matching from #3506 and fixes duplicate payload selection when multiple compatible H265 formats are offered.


I think this PR is useful for improving H.265 RTP payload handling.

rtptransceiver.go
Awuqing: Yeah, this should be `codecMatchExact` for better readability.

Awuqing (approved):

PR #3552, opened 2026-09-18
Fix DTLS role during renegotiation

JoTurk:
#### Description

Preserve the negotiated DTLS role in subsequent answers, currently switching offerers could change the advertised role and causes other impls to reject the answer.

Pion offers actpass => browser (active) => browser later offer (actpass) 

1. before the fix: Pion answers with active, browsers reject the answer 
2. after the fix: Pion answers with passive

I hit this bug at my work. We should revisit this once we implement DTLS restart (we should do it when we upgrade to dtls@v4).

#### Reference issue

I googled to see if someone reported it and I didn't see any Pion issues but I found this stackoverflow question, maybe we should ping them :) https://stackoverflow.com/questions/76391362/how-to-fix-set-ssl-role-for-the-transport-exception


JoTurk:
apparently firefox has the same bug, SDP handshake is hard :) https://bugzilla.mozilla.org/show_bug.cgi?id=1329028

fippo:
I think it is a bit more complicated... if you look at Chromium/libWebRTC it still follows the "offers should be actpass" rule. That gets put into the SDP but has basically no effect on the established connection. Which is what this does?

And hey, tagging @docfaraday for FF things and checking if they are still relevant.

JoTurk:
@fippo I think this is actually a spec bug, offer is always `actpass` but then with established DTLS role you have this edge case with old offer changing its role. I think this is pretty dumb IMO, this should never happen unless the new offer is doing a DTLS restart with a new `tls-id`, why touch the roles if the assoc is still in use??

fippo:
It is a grey^Wpitch black area. 8842 kinda solves it by removing the earlier constraint on offers being actpass but how do you know your peer does 8842?

Effectively you ignore changes to these things when the underlying state machine is not receptive for them. In the DTLS case that would be the next offer/answer changing the remote fingerprint.

fippo (approved):
Doing this in createAnswer seems right

JoTurk:
Good context thank you, I'll revisit this with DTLS restart :)

fippo:
probably good to check that the side that sent the offer does not feel like it now needs to change the DTLS role. Usually it would not but :imp: peer...

